PR #10, opened 2017-03-01
Keep trailing newlines in rendered templates

### LINKED ISSUE #9: Final newline stripped from templated files
If a template file ends with a new line voodoo will strip that new line when outputting the rendered template. This causes flake8 to complain about `W292 no newline at end of file` for all templated python files. New lines at the end of a template file should be preserved by voodoo.

timheap:
Jinja2 defaults to stripping trailing newlines, which upsets tools like flake8 when used to validate Python files rendered from a template.

Fixes #9.

Two `.strip()` calls was removed from the tests which were masking this issue. The tests now compare the files completely, including the trailing new line.

### DIFF
--- a/tests/test_render.py
+++ b/tests/test_render.py
@@ -24,7 +24,7 @@ def test_render(dst):
     control = read_content(join(dirname(__file__), 'ref.txt')).strip()
     assert generated == control
 
-    assert_file(dst, u'doc', u'mañana.txt')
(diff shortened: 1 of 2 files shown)


timheap:
I think the tests are failing because of the test that checks the `mañana.txt` file. It was erroring for me locally before I even made any changes. Unfortunately, Travis CI is having some issues itself right now so I can not check the test log.

jpscaletti:
Thank you!

PR #28, opened 2019-06-03
WIP - Allowing specifying extra paths to source templates from.

### LINKED ISSUE #15: Enhancement: Better support for template inheritance
One of `Jinja2`'s prime features is the support of template inheritance.
http://jinja.pocoo.org/docs/2.10/templates/#template-inheritance

The problem I see with `Voodoo`'s implementation is that parent files will also get copied over to the project folder. 

Wouldn't it be better to reserve the `.tmpl` suffix only for those parent files?
Those could still be used as base templates, but will be omitted from copying over.
This inverts the selection process compared to the status quo.

Voodoo could then try to render all non-suffixed text files present under the template path.
Any non-text file (e.g. image files) would simply be copied. (Likely this could be achieved by catching exceptions that arise when trying to open/render non-text files. Hence not requiring an extensive list of file extensions for discerning.)

pykong:
This is a work in progress PR. The feature in question is implemented, yet open tasks remain to make the PR merge ready.

My perfectly working implementation of the `extra_path` feature was based on a prior version of `copier`. In the meantime, `copier` underwent some changes, for example ditching YAML for TOML for configuration. I have now rebased my changes unto the most recent `copier` version - hence I need to gauge whether everything still works fine.

**Tasks:**
- [x] ~~Should we allow to specify multiple paths vs just one?~~ **YES**
- [x] ~~Write tests for the new feature~~
- [x] ~~Investigate: `copier.json` gets copied too~~ #29
- [x] ~~Polish code and make tests pass~~
- [x] ~~Update README.md~~

**Test-driving the feature**
1. Fetch PR: `git fetch origin pull/28/head:extra_path` (to your local copier repo)
2. Clone example templates:  `https://github.com/pykong/templates` (to a different dir)
3. Adapt `copier.json`, especially the tasks so it won't fail due to missing executables
4. Run like so: `copier .sources/web/ dest/ --extra-path .templates/.templates/`
5. Examine the generated files to acknowledge the magic that happened: _e.g. The upgraded template inheritance is very prominently displayed in the rendered `.gitignore` file for example._

closes #15


pykong:
@jpscaletti
Have you tested the `extra_paths` feature out, yet?

The PR is now a merge candidate. When you find the time could you please do a code review?
Do you agree with the tests I wrote?

Issue #30, opened 2019-06-05
Update Readme: Configuration file format


pykong:
@jpscaletti 

Three config file formats (`TOML`, `YAML` and `JSON`) find mention in the readme and the code base. As of the most recent version `copier.yaml` does not seem to work for me any longer, forcing me to switch to `json`.

Could you therefore maybe clarify which configuration file formats we can actually use?
Does the `README.md` require update in this regard?

pykong:
Indeed `CHANGELOG.md` states that: 

> The recommended configuration file is now  copier.toml

https://github.com/jpscaletti/copier/blob/master/CHANGELOG.md#version-23-2019-04

Hence `README.md` need to reflect that change more explicitly.
The screenshot needs to be updated as well.

jpscaletti:
After some reflection, I restored the `copier.yaml` or `copier.yml` as default configuration formats.
The TOML and JSON formats are also still supported and will continue to be so in the future.

pykong:
Great. Personally, I find the array syntax of `yaml` cleaner than that of `json` or `toml`.
This is especially relevant to the tasks list.

jpscaletti:
By the way, I uploaded the `2.4` version to PyPi today (Is now `2.4.1` because I forgot to exclude the yaml files by default).

PR #32, opened 2019-06-08
copier.yml -> copier.toml

pykong:
Some fixes, too.

closes: #30


pykong:
No longer relevant due to switching back to `YAML` as default format:
https://github.com/jpscaletti/copier/issues/30#issuecomment-500141480

Issue #22, opened 2019-05-12
Enhancement: Cleanup on failure


pykong:
To prevent defunct remnants of a failed attempt to create a project an already created folder should be deleted when the creation process fails (e.g. task runner experiences error).

jpscaletti:
I consider this a feature.

But the removal of failed attempts can be achieved right now simply by wrapping the `copier.copy()` call in a try...except clause:

```python
import shutil

try:
    copier.copy(src, dst)
except:
    shutil.rmtree(dst, ignore_errors=True)
```

pykong:
The consequence of not having a cleanup procedure is that, in case of failure to render the template, the user will get stranded with a defunct dir.

This is bad UX IMHO.
Would you consider merging #37 if I made the the cleanup process dependent on a command line flag? Just my suggestion ...

jpscaletti:
🤔

jpscaletti:
🤔

jpscaletti:
ok, you are right

pykong:
Great! You made the right choice! The ability to change one's mind is a tell tale sign of a good developer in my experience.

PR #37, opened 2019-06-09
Implemented cleanup on failure.

pykong:
This is almost ready to merge.

Two points:
- As the test run with `tmp/` as the destination, the test never gets a chance to actually remove. How can we address this?
- `shutil.rmtree` fail when called with a non-existing path, `ignore_errors` does not prevent this

Closes: #22


tests/test_cleanup.py
```diff
@@ -0,0 +1,15 @@
...
+from .. import copier
+
+from .helpers import assert_file, render, PROJECT_TEMPLATE, DATA, filecmp
+
+
+@pytest.mark.slow
+def test_cleanup(dst):
+    with pytest.raises(CalledProcessError):
+        copier.copy("./tests/demo_cleanup", dst, quiet=True)
+    assert not (dst).exists()
```
pykong: The test will always find a non-existing path, as temporary dir gets removed.

copier/main.py
```diff
@@ -137,6 +137,10 @@ def copy(
             skip=skip,
             quiet=quiet,
         )
+    except:  # noqa
+        print("Something went wrong. Removing destination dir.")
+        shutil.rmtree(dst_path, ignore_errors=True)
```
pykong: Will raise when called with non-existent path.

pykong:
It appears the error happens on the teardown of the test fixture.
The cleanup procedure I wrote successfully removes the temporary target, but this makes the teardown upset as it no longer has any dir to remove.

Explicitly ignoring errors resolves this problem:
https://github.com/jpscaletti/copier/blob/302dde9b17a8198e08fe8c8cf034ed226112f752/tests/conftest.py#L17

Changing to:

```py
request.addfinalizer(lambda: shutil.rmtree(dst, ignore_errors=True))
```

Is there any good reason not to do it?

pykong:
@jpscaletti 

The tests are now green, except for `3.5`.
I do not have python 3.5 installed locally, as I am mainly using python 3.7.

Hence could you probably figure out what is going wrong?

jpscaletti:
Is my fault, the tests were failing in master as well because of an extra comma.
However I probably wasn't clear enough at #22 .
I don't want that copier to auto-removes the destination folder if something fails.
I consider that a feature, not a bug.
But if you want to have it, it can be implemented in a python file that calls `copier`, since any  exception propagates up, but I don't think it should be in `copier`.

jpscaletti:
Thanks ;)

PR #44, opened 2019-06-10
Removed cruft from yaml blocks

pykong:
I am not sure what the
```
---

```
was about. I think it was cruft. Removed it.


jpscaletti:
That’s YAML’s optional “beginning of document” marker https://stackoverflow.com/questions/53092786/use-of-in-yaml

I agree is just noise, but we remove it from that yaml, it should be also removed from the other yaml code below.

pykong:
Done.

Issue #42, opened 2019-06-09
Enhancement - Type annotate code base


pykong:
@jpscaletti What do you think of hardening `copier` with endowing all functions with type annotations?

Like the central copy function for example:

```py
def copy(
    src_path: str,
    dst_path: str,
    data: Optional[dict] = None,
    *,
    exclude: Optional[List[str]] = None,
    include: Optional[List[str]] = None,
    tasks: Optional[List[str]] = None,
    envops: Optional[dict] = None,
    extra_paths: Optional[List[str]] = None,
    pretend: bool = False,
    force: bool = False,
    skip: bool = False,
    quiet: bool = False
) -> None:
```

jpscaletti:
Not a big fan because it makes the code uglier, I'm not very convinced of the benefits, and because I'm to lazy to add them... but I would be happy to merge something like this and discover what I was missing that make so many people excited about it.

pykong:
Obviously, we won't be able to discuss pros and cons of type annotations in depth here.
(About the lazy thing: https://github.com/dropbox/pyannotate or Instagram/MonkeyType may do the grunt of the work for you in the future - required you got great test coverage. )

I despise the verbosity that it brings for many signatures.
In my work on `copier` there has been at least one instance where type annotations would have caught me calling a function with the wrong order of arguments earlier.

I am going to try bringing forth some extensive annotations and we can then decide whether it is something we wanna have in `copier`.

pykong:
In my attempt to manually type annotate `copier` I made up my mind. 

At the current complexity, type annotations will not be of a big profit for `copier`.
From my personal experience, there is a break-even point between additional complexity and additional safety for using type hints. The small scale of the current `copier` code base does not yet justify their addition.

At a later point in time, we may resort to it though. Especially once either `pyannotate` or `MonkeyType` becommes usable. Closing this for now.

jpscaletti:
I'm reopening this issue as I would like to try this idea.

pykong:
This is surprising. But all the better.
Feel free to summon me in case you need any assistance.

pykong:
@jpscaletti Have you made progress on this matter?

pykong:
I am working on **PR** right, which fully type annotates the whole code base, excluding tests.
I am hoping to have it ready by next Sunday.

Issue #49, opened 2019-06-17
Enhancement - Simplify creation of configuration object 


pykong:
This issue relates solely to improving the architecture it does not address any bug or suggests any new functionality.

## Creating the configuration object
`copier` needs to take in account options coming from:
1. defaults, Like `DEFAULT_EXCLUDE` and `DEFAULT_INCLUDE`
2. files - the `copier.yml`
3. user input - CLI args, prompt

It also needs do some basic validation and ensure default values for certain options like, `extra_paths` defaulting to `[]` instead of `None`.

The entirety of the configuration options I herein call the configuration object, albeit this is in the current implementation is not strictly correct as we deal with independent variable not brought under the umbrella of a single `dict` yet.

## The problem
The current approach to creating the configuration object as used in `main.copy_local` does not strike me as `DRY` or pythonic. 

Consider for example that we duplicated the same command sequence four times:
https://github.com/jpscaletti/copier/blob/18b2c70c3d582a9d3fc241dec7df34c45dcdd1b3/copier/main.py#L203-L216

Also, default values are not defined in a single place:
https://github.com/jpscaletti/copier/blob/18b2c70c3d582a9d3fc241dec7df34c45dcdd1b3/copier/main.py#L132-L133

## A cleaner way
While working on `SublimeLinter 4` I learned that the problem of creating a single configuration object from different precursor objects is better solved by sequential merger of dicts.

Such a pattern not only allows to define default values in a single place but also easily sets the order of precedence. In this hierarchy of precedence, the default options would reside at the lowest tier, being overridden by options sourced from a `copier.yml` and those provided at the command line.

```py
conf_obj = {**default_opts, **yaml_opts, **cli_opts}
```

As an additional advantage, this approach would also enable a much more straightforward validation of the args.

Related: #45

pykong:
@jpscaletti Can you see this issue I described and would be open to a PR which will give us a simpler more robust code? (Still passing all tests of course ;-) )

pykong:
Memo:

I really like how `pydantic` offers both validation and providing default values at the same time.
We could use this to first merge args provided by command line, configuration files and potentially environment variables and then validate against a defined scheme which will at the same time fill out all None values with defaults. Very powerful. Very elegant.

https://github.com/samuelcolvin/pydantic#a-simple-example

This is all done using the modern type hint syntax. Unfortunately, this means we will not be able to use `pydantic` for copier as we still support py 3.5.

jpscaletti:
[...]

PR #52, opened 2019-07-04
Adding type annotations

pykong:
@jpscaletti You said you wanted to "try this idea". So here it is:

This **PR** introduced a complete type annotation of the entire `copier` module.
*MyPy and all tests run green in my local setup.*

The general Pros and Cons of type annotations apply here as well.

#### Cons - Why not to merge:
The code becomes bulkier even bloated at some places. Especially as we employ many functions with long signatures, those signatures will definitely turn into multi-line monstrosities. At other places, the amount of code grew due to introducing new variables or type conversions.

#### Pros - Why merge:
It enforces stringency and makes refactoring a joy. Working on another **PR**, that for #49, the significant refactoring involved killed half the tests in my first attempt. Now my second attempt is being guided by the type annotations. It makes it easier as I have a clear picture and immediate feedback on what newly introduced functions should return. You might say "everything clicks in place" as lego pieces do.

#### TODO:
- [x]  ~~Are type aliases named idiomatically?~~ _No, idioms found._
- [x]  Check whether types could be made more stringent.
    - [x] Is `Optional` overused?
- [x]  Remove `mypy_extensions` dependency.
- [x]  Make Travis' tests green.
- [x]  Add type checking to the test routine.
- [x] Drop Python 3.5 support.

Closes: #42

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -5,7 +5,7 @@ import re
 import shutil
 import subprocess
 from pathlib import Path
-from typing import Any, Dict, Sequence, Tuple, Optional, Union
+from typing import Any, Dict, Sequence, Tuple, Optional, Union, Callable, List
 
 from . import vcs
 from .tools import (
@@ -28,6 +28,9 @@ __all__ = ("copy", "copy_local")
 StrOrPath = Union[str, Path]
 OptSeqStrOrPath = Optional[Sequence[StrOrPath]]
 OptSeqStr = Optional[Sequence[str]]
+AnyByStr = Dict[str, Any]
+IntOrStr = Union[int, str]
+CheckPathFunc = Callable[[StrOrPath], bool]
 
 # Files of the template to exclude from the final project
 DEFAULT_EXCLUDE: Tuple[str, ...] = (
@@ -59,7 +62,7 @@ def copy(
     include: OptSeqStr = None,
     skip_if_exists: OptSeqStr = None,
     tasks: OptSeqStr = None,
-    envops: Dict[str, Any] = None,
+    envops: AnyByStr = None,
     extra_paths: OptSeqStr = None,
     pretend: bool = False,
     force: bool = False,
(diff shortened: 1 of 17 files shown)


copier/main.py
```diff
@@ -177,82 +179,84 @@ def resolve_source_path(path):
...
-    user_extra_paths = config_data.pop("_extra_paths", None)
+    user_extra_paths: List[str] = config_data.pop("_extra_paths", None)
     if not extra_paths:
         extra_paths = [str(resolve_source_path(p)) for p in user_extra_paths or []]
 
     user_data = config_data if flags["force"] else query_user_data(config_data)
     data.update(user_data)
     data.setdefault("folder_name", dst_path.name)
-    render = get_jinja_renderer(src_path, data, extra_paths, envops)
+    engine = get_jinja_renderer(src_path, data, extra_paths, envops)
```
pykong: Better rename to `rend`?

pykong:
@jpscaletti The tests run now green for` Python 3.6+`.
However, I do not know how to make type annotations work with `Python 3.5`.

Unless we drop `python 3.5` support and start supporting only Python 3.6 or later like so many modern python packages nowadays do and you also do with [sqla-wrapper](https://github.com/jpscaletti/sqla-wrapper), this represents a show stopper.

copier/main.py
```diff
@@ -177,82 +179,84 @@ def resolve_source_path(path):
...
     must_filter, must_skip = get_name_filters(exclude, include, skip_if_exists)
 
     if not flags["quiet"]:
         print("")  # padding space
 
-    for folder, _, files in os.walk(str(src_path)):
-        rel_folder = folder.replace(str(src_path), "", 1).lstrip(os.path.sep)
-        rel_folder = render.string(rel_folder)
-        rel_folder = rel_folder.replace("." + os.path.sep, ".", 1)
+    for _folder, _, files in os.walk(str(src_path)):
```
pykong: Trying out `typing.cast()` here, instead of defining new variables.

copier/tools.py
```diff
@@ -50,13 +63,19 @@ def printf_block(e, action, msg="", style=STYLE_WARNING, indent=0, quiet=False):
 no_value = object()
 
 
-def required(value):
+def required(value: Any, **kwargs: Any) -> Any:
```
pykong: Return type should be identical to the type of parameter `value`.
I have not been able to make it work via `TypeVar`.

pykong:
@jpscaletti 

I consider 162dc56 to be a merge candidate.

Please decide if you want to close this **PR** or drop support for `Python 3.5` in favor of hardening the code base via type hints.

In the latter case, we may also consider adding  [pytest-mypy](https://github.com/dbader/pytest-mypy) to the test sequence.

copier/tools.py
pykong: Got it to work via `TypeVar` now.

copier/main.py
pykong: Prevented the introduction of new variables by moving type declaration before the for loop.

copier/main.py
pykong: Leaving it now with `engine`.

PR #54, opened 2019-07-11
Cleanup and move to Python 3.6

pykong:
With dropping `Python 3.5` support in #52 it is time to take advantage of `Python 3.6` features. 
In the same wake, some needed cleanup on the code base might be done.
This is what this **PR** is about.

Some notable changes:

### Using Python 3.6 features:
- Ditching old `format` for the new `f-strings`.
- Feeding `Path` to `shutil.rmtree`.

### Cleanups
- Dropped support for `voodoo.json`.
- Removed `print` statements from tests.
- Removed redundant commands from `cli`.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -5,7 +5,7 @@ import re
 import shutil
 import subprocess
 from pathlib import Path
-from typing import Any, Dict, Sequence, Tuple, Optional, Union
+from typing import Any, Dict, Sequence, Tuple, Optional, Union, Callable, List
 
 from . import vcs
 from .tools import (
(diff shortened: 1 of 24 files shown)


copier/main.py
```diff
@@ -290,11 +304,19 @@ def render_folder(dst_path, rel_folder, flags):
...
+    engine: Renderer,
+    must_skip: CheckPathFunc,
+    flags: Dict[str, bool],
+) -> None:
     """Process or copy a file of the skeleton.
     """
+    content: Optional[str]
     if source_path.suffix == ".tmpl":
-        content = render(source_path)
+        content = engine(source_path)
```
jpscaletti: I think `render` was a much better name. Without knowing what `engine` is, `engine(something)` is a mistery. I'm going to change it back.

jpscaletti: I think `render` was a much better name. Without knowing what `engine` is, `engine(path)` is a mistery.
`render(path)` and `render.string(string)` were much clearer.

pykong: Agreed `engine` is bad, because it is unclear.
There happened to be some naming collision confusing `mypy`.

As you have by now merged this **PR** I am going to open a separate **PR** to rename this identifier. Might take til next weekend for me to pull it off.

PR #62, opened 2019-09-13
Enable tests for Python 3.8 on CI

Jamim:
Hello,

Python 3.8 is [almost there](https://pythoninsider.blogspot.com/2019/08/python-380b4-is-now-available-for.html), so it would be nice to see tests for it on CI.

These changes also:
 - add more precise classifiers

Best regards!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,14 @@
+dist: bionic
 language: python
+python:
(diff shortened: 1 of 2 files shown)


jpscaletti:
I didn't know they already had any version of python 3.8. Thanks!

PR #70, opened 2019-09-17
Basing config file identification on regex

pykong:
Just a minor fix.

### DIFF
--- a/copier/config/user_data.py
+++ b/copier/config/user_data.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+import re
 
 from ..tools import HLINE, INDENT, printf_block, prompt
 from ..types import AnyByStrDict, StrOrPath, PathSeq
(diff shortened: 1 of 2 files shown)


pykong:

jpscaletti:
Isn't this just slower?
I mean, it would make sense if the extensions were more unpredictable, but they aren't.

pykong:
@jpscaletti 
I believe the solution to be more robust.
Speed isn't the issue - considering a couple of milliseconds more it would take to counter check the files against the regex. Especially, assuming here that the foremost use case is to make project templates once in a while not millions over millions requiring a high-performance solution.

Instead the question is one of design:
> To what extent should we prevent a user from shooting her foot?

#### Two options:

1) We are being pre-emptive:
If a user does not put enough care into properly naming her `copier.yaml` files, should we try to counter this?
2) We are being conservative:
The user is responsible for correctly naming the files. In case files are ill-named we do nothing.

See, also which touches a very similar question #77 

If you think we should be conservative rather than being pre-emptive:
Just close this **PR**.

#### My own take:
**TL/DR**: Merge
Generally, option 2) reverberates most with my essentialist take on things.
However, in this particular case, I believe there is a point for widening the selection of files to include _uppercased file suffixes like `YAML` or `YML`_ which indeed occur in the wild. Hence this **PR** makes sense.

Issue #67, opened 2019-09-16
Sharing identical files between templates


pykong:
I share many identical files between different template folders.
I'd really like to keep my templates DRYer and include those files by reference only.

@jpscaletti
Do you get an idea of how we could achieve this?

_Some ideas:_
- We allow for a multi-level inheritance whereby a template folder inherits templates from a parent template.
- We allow specifying folder whose contents or a list of file paths to be copied over to the destination folder.

Relates to: #58

jpscaletti:
A soft link? Copy & Paste? The inheritance system for templates sounds like over-engineering it to me.

jpscaletti:
I was writing that Jinja's inheritance mechanism doesn't work with files outside the templates folder, but I just remembered the one of Jinja's loader allows you to specify several folders as sources...

So yes, in theory, we could use Jinja's [ChoiceLoader](https://jinja.palletsprojects.com/en/2.10.x/api/#jinja2.ChoiceLoader) or maybe a [PrefixLoader](PrefixLoader) by default and have a list of optional "extra templates folders" in the `copier.yml` file to add to the current working folder (added by default). In that way, you could call a group of templates to use however you want.

How it's sound to you?

pykong:
Thanks for your answer. You appear to a have much more knowledge about Jinja2's engine than me.
Let me try to translate your suggestions into code. Should be ready by the weekend.

My suggestion is that once we merge (or reject) the **PR** resulting from that you may release `copier 3`.

pykong:
I wonder if we couldn't just:
- provide a list of additional source folder paths
- iterate over this list against `copy_local`

We would need to define an order of precedence within that list and versus the original `src_path` setting.

I am still researching whether using Jinja2's ChoiceLoader would offer any benefits.
As far as my current understanding goes ChoiceLoader would only provide additional lookup paths for templates but not instructable to include additional folders.

pykong:
I believe we were talking about different things.

**You**: How to teach Jinja2 to locate additional files...
**Me**: How can we keep our templates DRY...

We can already specify additional lookup paths via the `extra_paths` setting.
Yet, this is only "half-DRY" as we include file content from another location but would still require to do so in an `*.tmpl` file. Hence we may need to include a lot of the same of those files in multiple templates. Not exactly DRY.

Nevermind I have an elegant idea on how to address this.
[...]

PR #79, opened 2019-09-23
WIP - Folder inheritance

pykong:
**WIP** - Do not merge yet.

## What this is about
This **PR** introduces an inheritance mechanism for template folders. Now we can have multiple child templates that inherit files and subfolders from a single parent. Thereby the need to duplicate the same set of files and subfolders among many templates is eliminated resulting in much DRYer templates.

The implementation rests upon the new optional `_extends` setting which defines a path to a parent template folder. If `_extends` is a given then the referenced parent folder is rendered to the destination path before the child folder. 
Conflicts will be either resolved manually via the prompting mechanism or by giving `skip` / `force` flags.

_This is some wickedly powerful feature_. Maybe this is something other packages do not possess (I do not know, but could be a good selling point). I was surprised by how well it turned out working with only minimally invasive modifications to the code. Still, I am testing this further and try to evaluate all corner cases and pitfalls. Eventually, we could release this as an _experimental feature_.

### TODO:
- [ ] Processing `copier.yml` of the parent template folder
  - [x] Preventing same tasks to run on each level
  - [ ] What options should be passed from child to parent?
- [ ] Validate against path collision
- [x] Write more tests
- [ ] Remove `extra_paths`
- [ ] Update docs
- [ ] Add `extends` to cli args
- [ ] Rename `extends` for consistency? (maybe to `parent_path`)

Closes: #67


tests/demo_inheritance/conflicted/child/copier.yml
```diff
@@ -0,0 +1 @@
+_extends: ./tests/demo_inheritance/conflicted/parent
```
Yajo: If this works as it is written here, I'd consider this a bug.

You cannot predict from which folder the user will launch copier. Instead, all paths should always be relative to the very `copier.yml` file that defines them.

So, in this case, this should be:

```suggestion
_extends: ../parent
```

... because from this `copier.yml` file, the parent folder is found there.

Also have you considered the possibility of extending several at once? We could define this as a list:

```suggestion
_extends:
  - ../parent
```

The same applies for other test cases here.

pykong: @Yajo Thx. I will investigate this. Please consider opening a short issue if you think there is indeed a bug or potential improvement here, so we can properly track things.

Yajo: Well, this is WIP, I don't think I need another issue, isn't it?

Yajo:
I'm not sure I understand the feature, maybe if you explain the use case with an example I'll understand better.

If the point is having a set of templates to be able to extend them, what's the difference between `_extends` and `_extra_paths`?

copier/main.py
```diff
@@ -108,7 +116,7 @@ def copy(
         Skip files that already exist, without asking
 
     - quiet (bool):
-        Suppress the status output
+        Suppress the status outputm
```
Yajo: Typo

pykong:
I am no longer pursuing this **PR**.
Closing this in favor of only transcluding from other config files: #117

Issue #78, opened 2019-09-23
Are skip and force mutually exclusive?


pykong:
In my understanding, the `skip` and `force` options represent two mutually exclusive resolution strategies.

If so we should reflect this fact in:
- ~~the cli arg parsing (`argparse` provides ways to specify [mutual exclusivity](https://docs.python.org/3/library/argparse.html#mutual-exclusion))~~
- in the `Flags` config object (`pydantic`s validators are very easy to use)
- the docs of course

**EDIT:**
We should validate in a single place. The best place is the config submodule. It will not make a difference to the user if error messages are properly generated.

jpscaletti:
Hummm.. yes they are. In fact,  the "Overwrite filename (y/n)" prompt is really choosing between
force (overwrite) and skip (leave the file as it is).

pykong:
The prompt does but not the **CLI**. 
But do not bother I have a fix for this in the making... ;-)

PR #80, opened 2019-09-28
Enforcing mutual exclusivity of skip and force.

pykong:
This minor `PR` implements:
- enforcing mutual exclusivity of the flags `skip` and `force` at config level
- a single test for that

Closes: #78


pykong:

PR #82, opened 2019-09-28
Moving reference files to own folder

pykong:
Tidying up the tests directory.
The descriptive Folder name makes the purpose of those files clear also.


pykong:
_(**PR**s are piling up. It would be great if could move forward with those, so we may proceed towards releasing copier 3.)_

Issue #88, opened 2019-10-30
Copy updates as 1st-class citizens


Yajo:
The scaffolding landscape in Python currently is pretty bad, thanks for bringing to life this nice project.

Handling updates is pretty bad still. Imagine:

1. You have a scaffolding. 
1. You copy it and start a new project.
1. Your new project evolves in a separate git repo.
1. You update your scaffolding.
1. You want to apply your scaffolding to your copied project again, but don't want the default answers; you want the ones you used last time.
1. You don't want the updates to touch the files that you already touched, unless they've been touched also in those lines in the source scaffolding.

[`cruft`](https://github.com/timothycrosley/cruft/) handles this pretty well, as it stores all answers in a json file, which also happens to include the git commit of the source when it was copied, and this allows it to extract the diff since that commit to the latest and apply only that diff to the destination.

**How should `copier` *copy* (lol) this behavior?**

First, when copying:

1. When a scaffolding is copied, store all answers in a file. It could be something like `.copied.yml`.
1. Store in `.copied.yml` also the source of the copy (e.g. `_copied_from: gh:my/scaffolding`).
1. When the scaffolding was a git repo, store also the copied commit. It could be a hidden key in `.copied.yml` (e.g.: `_copied_commit: o3ho4ho5hoho2hoh2o`).

Then, support running `copier` without arguments, which:

1. Fails if `.copied.yml` is absent.
1. Fails if `_copied_from` is absent in that file.
1. If `_copied_commit` is absent, it should be the equivalent of `copier $_copied_from .`.
1. If `_copied_commit` is present, it should be the equivalent of `copied --from-commit $_copied_commit --to-commit HEAD $_copied_from .`.

As you probably have guessed, we need to add new flags `--from-commit` and `--to-commit` (which defaults to `HEAD`), which:

1. Fails if source is not a git repo.
1. If present, will generate a patch from `--from-commit` to `--to-commit`.
1. It will only apply that diff.
1. Does nothing if both flags are equal.

**Why all of this?**

1. We need to be able to just execute `copier` in a pre-copied scaffolding and let it get only the diff since it was copied last time.
1. We need to be able to ensure a project is updated with its scaffolding, i.e. to raise an error in CI when that's not the case.

pykong:
@Yajo 
Thanks for your note and all the input provided!
Indeed your input is very welcome as I do not know how others are using it.
Feedback is vital so I can better decide what to improve.

[...]

PR #103, opened 2019-11-21
Add .copier-answers.yml to remember last answers

Yajo:
- If the copy destination exists and has a `.copierlog.yml`, all questions asked in the template `copier.yml` file will get the answer found in `.copierlog.yml` by default.
- For now, there are no `_private` variables in the copierlog, but in the future there might be, so I'm documenting that now.
- Add docs for some of Copier's hidden features.
- Add `to_nice_yaml` filter, to be able to dump beautiful YAML. [This name matches that on Ansible](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#filters-for-formatting-data) just in case somebody wants to share templates among both engines, although do not take that as an official support statement.
- Add a test that copies a folder with default data, then copies with altered data, and then copies again with default data and still the altered values have been used.
- This is a partial improvement towards completion of #88.

@Tecnativa TT20357

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,9 @@ function.
 Since version 3.0, only Python 3.6 or later are supported. Please use the
 2.5.1 version if your project runs on a previous Python version.
 
-## The copier.yml file
+## The `copier.yml`ile
 
-If a `copier.yml`, or `copier.yaml` file is found in the root of the project,
+If a `copier.yml`, or `copier.yaml` file is found in the root of the template,
 it will be read and used for two purposes:
 
 ### Prompt the user for information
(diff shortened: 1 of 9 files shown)


pykong:
Thanks for this **PR**, this signifcantly enhances overall user experience!
I wonder would the entries be retrieved with lower latency if they were written to a `.sqlite` file instead?

Yajo:
I think we should stick with YAML. This file will be git-tracked in most scenarios, and won't be very big. There are almost zero chances that one template asks so many questions that generates a file of more than 1MB here, so performance shouldn't be a bottleneck in this point. Using YAML seems the best option here.

pykong:
What is the strategy for emptying the file? We need to prevent it getting swamped eventually.

Yajo:
The file shouldn't be emptied. It should just get updated with latest answers after a normal update process.

Maybe you're getting confused by the "log" part of the name. It's not really a usage log, but an answers log. Maybe a name like `.copier-answers.yml` would be more intuitive?

pykong:
`.copier-answers.yml` would better reflect the purpose of the file.
If it is supposed to live under the `copier` directory, you may shorten the file name to just
`answers.yml`. Or maybe make it `past-answers.yml`.

Yajo:
> If it is supposed to live under the `copier` directory

Not, it's supposed to live in the `dst_path` root. I prefer a dot-prefixed file in any case (that's the most common practice for files that complement other projects (`.gitlab-ci.yml`, `.travis.yml`, `.editorconfig`, `.gitignore`... all start with dot)) and I'd also prefer some name that indicates specifically that this file comes from copier (just like those examples also). Ideas:

- `.copierlog.yml`, of course
- `.copied.yml`
- `.copier-answers.yml`
- `.copier-past-answers.yml`

Since the purpose is to keep a log of last answers, _and in the future it will not only contain answers_ (that's in an upcoming PR I have in the oven :wink:), and also keeping in mind it's a specific & short name, my vote still goes for `.copierlog.yml` (if that matters :sweat_smile:).

pykong:
Make it `.copier-answers.yml`, then. Otherwise LGTME. 👍

Yajo:
Changed now.

I wonder if you could please merge this. It's the only piece missing for opening the next PR.

Yajo:
Well, it's done. The next iteration is #106. Please merge so we can move forward :pray: :innocent:

Yajo:
Anything else here to be done to merge?

Thanks!

pykong:
Sorry, for the delay. Switching to new workstation right now so I can finally develop again.
Thanks for the PR.

PR #109, opened 2020-01-17
Handle KeyboardInterrupt elegantly

Yajo:
Without this patch, a full traceback is printed when the user hits <kbd>Ctrl</kbd> + <kbd>C</kbd>.

Now it's more elegant:

@Tecnativa TT20357

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -1,26 +1,43 @@
 #!/usr/bin/env python
 import sys
+from textwrap import dedent
 
 from plumbum import cli, colors
 
+from .config.objects import UserMessageError
 from .main import copy
 from .types import AnyByStrDict, OptStr
 from .version import __version__
 
 
+def handle_exceptions(method):
+    def _wrapper(*args, **kwargs):
+        try:
+            try:
+                return method(*args, **kwargs)
+            except KeyboardInterrupt:
+                raise UserMessageError("Execution stopped by user")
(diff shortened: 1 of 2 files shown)


pykong:
@Yajo Ah, this is a small gem of a commit, which would definitely enhance the user experience.
Could you make the tests green, plz?

pykong:
@Yajo I have given you push access, as I learnt to trust in your abilities and commitment to quality code. I believe this could speed-up things. For bigger **PR** please wait for my review though.

Regarding this PR: 
My cleanup policy is to not let the user be stranded with an incomplete target dir.
Hence the nascent target dir should be deleted subsequent to `KeyboardInterrupt`.
This behavior needs to be implemented and covered by tests.

How should we handle pre-existing target dir though?

Yajo:
> I have given you push access, as I learnt to trust in your abilities and commitment to quality code.

Thanks! That's an honor :blush: 




> the nascent target dir should be deleted subsequent to `KeyboardInterrupt`.

This exception guard is at the top level, just for the UX. It doesn't alter the way lower methods work, so this shouldn't be the problem.


> How should we handle pre-existing target dir though?

That's a problem indeed. Maybe just ensure git is clean before doing things, and of course never delete it. I'm adding that check to a PR I'm cooking.

In any case, as new features land, #110 seems more necessary.

PR #111, opened 2020-01-22
Updates respecting evolved git history

### LINKED ISSUE #88: Copy updates as 1st-class citizens

Yajo:
Fix #88 the easiest way possible. Changes summary:

- A new `--vcs-ref` flag that indicates which commit/ref you want to copy, and only applies when the source directory is git-versioned.
- A new `--no-diff` flag to `copier update` skips the smartypants diff behavior.

What it does?

- Checks you are updating a git-versioned destination from a git-versioned source.
- Clones the destination into temp dir.
- Does a `copier copy -f tmp_src tmp_dst` (yes, a temp dir also).
- Gets the git diff between `tmp_dst` and `dst_path`.
- Performs a normal `copier update dst_path`.
- Applies the git diff at the end, to try to respect downstream project evolution.

@Tecnativa TT20357


pykong:
@Yajo Thanks for this **PR*. I am going to review it no later than coming weekend.

pykong:
**LGTM**. Merge as you see fit.

Yajo:
Thanks, I guess this feature can have some bugs, but only time will tell. Merging to have the feature in the next release as asked in #112.

PR #113, opened 2020-01-23
Support custom templates suffix

Yajo:
The `.tmpl` suffix doesn't seem to be a very widely adopted standard on the Jinja world. Others like `.jinja` or `.j2` usually have better IDE integration.

It's not like `.tmpl` is a wrong default, so here I'm leaving it as a default but letting template developers to change it to another one that fits better their environment.

Apart from this, if you really had to produce any `.tmpl` files in your copies, having files that ended with `.tmpl.tmpl` seemed a bit weird, while now you can change that suffix to whatever fits better your project.

@Tecnativa TT20357


pykong:
Generally I prefer erring on the side of constraint rather than configurability.
However, I see the point in allowing custom template suffixes.

**PR**: A test for configuring different suffixes should be implemented.
Otherwise LGTM

Yajo:
> Generally I prefer erring on the side of constraint rather than configurability.
> However, I see the point in allowing custom template suffixes.

Yes, but the fact is that for instance in this vscode extension there's not just a jinja language, but many jinja metalanguages with their corresponding syntax highlighting: https://github.com/samuelcolvin/jinjahtml-vscode

Reconfiguring all of that is PITA. I just prefer to suffix templates with `.jinja` and all works out of the box.



> **PR**: A test for configuring different suffixes should be implemented.

It's there, you probably missed it.

PR #115, opened 2020-01-29
Sort answers file

Yajo:
This way it will be more readable and it will produce less diff if the template copier.yml file is updated.

@Tecnativa TT20357


pykong:
@Yajo Thanks! I see you have an eye for tiny improvement that could improve the user experience! ;-)
This neat little **PR** would definetly be a good addtion to `copier`.

Feel free to merge, once:
- you have implmented a test case
- made tests green

Issue #110, opened 2020-01-21
More class, less dangling methods


Yajo:
The more I hack into copier, the more I see the code is structured weirdly.

It seems like copier is thought as a library which also lets CLI usage, instead of the other way around.

Since #106, we leverage Plumbum for the CLI stuff, and it feels way more pythonic and comfortable to write it like this.

Now, the `CopierApp` class and its friends feature [the `invoke()` method](https://plumbum.readthedocs.io/en/latest/api/cli.html#plumbum.cli.application.Application.invoke), which should be just enough to let downstream to use Copier as a lib in other python codebases.

If we refactor the codebase to make it more focused on that, we will save the pain to keep on passing arguments here and there. We will easily know if we're updating or copying, know if the user forced the execution, and so on.

---

This is my mind map about what the new object model would be and what information could be obtained from each object:

- [ ] `Answer`: what has been responded to a question. It can be obtained from asking the user, from API call, last answers in the `Subproject`, or defaults from the `Question` (which may be templated). It returns the best match for each `Question`.
- [ ] `Question`: contains a question definition. It is obtained from the `Template`. Each question is related to an answer
- [ ] `Questionary`: contains all questions and helpers to ask to the user.
- [ ] `Template`: available versions, the questionary, local clone location, options defined in the template.
- [ ] `Subproject`: is repo? is dirty? last update version, last answers, local path, folder exists?
- [ ] `Settings`: the evolution of `ConfigData`. It won't require passing through `make_config()` (I always thought that's a weird design). It will hold references to all external settings sources (which are the `Template`, and the API call, and after #235 it will also be user config file). It will *not* hold answers data.
- [ ] `Renderer`: takes care of rendering the files to disk.
- [ ] `Worker`: combines all other objects and has 2 main methods: `copy()` and `update()`.

pykong:
> The more I hack into copier, the more I see the code is structured weirdly.
>
> It seems like copier is thought as a library which also lets CLI usage, instead of the other way around.

I am not sure I can concur, still I am committed to maximum code quality and hence I am open for improvements even radical refactoring required it makes sense.

> the pain to keep on passing arguments here and there. 

Carrying the extensive signatures around is indeed a major pain point which I could not tackle via my last big refactoring.

**TL/DR:** I am open to suggestions/PR. Ideally those should be made/merged before I am going to work on #79 again which could happen in the coming weeks.

pykong:
@Yajo What is your the status of this?
[...]

PR #120, opened 2020-02-21
Reduce code complexity

Yajo:
As explained in #110, the dangling parameters for methods keep on complicating development.

A full code refactoring is too much work for now, but I did a little refactoring which consists in:

- Merge `Flags` into `ConfigData`, and remove anything related exclusively to flags.
- Any method that uses arguments from `ConfigData` gets instead a `conf: ConfigData` argument.
- `get_jinja_renderer()` removed; it's not useful as its little work can easily be done by `Renderer` itself.

This way, whenever we need a new config, we can just add it to `ConfigData` and use it more widely.

A full refactoring is still needed IMHO, where we have a main `Copier` class which handles its state and everything it needs (or it could be `CopierApp` instead), but that's more a design decision to be taken by the project leader.

@Tecnativa TT20357

### DIFF
--- a/copier/config/__init__.py
+++ b/copier/config/__init__.py
@@ -1,1 +1,1 @@
-from .factory import Flags, make_config  # noqa
+from .factory import make_config  # noqa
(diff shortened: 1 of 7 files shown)


pykong:
@Yajo Thanks for this **PR**.
I am going to have a look into it next Sunday.

PR #121, opened 2020-02-21
Add migrations support

### LINKED ISSUE #119: Use git tags in updates by default
One thing that floats around my mind is that the update system merged in #111 should use git tags instead of git commits.

Maybe we should implement this before #112.

This way, when running pre and post tasks, we could discriminate by version. We can leverage [one of these methods](https://stackoverflow.com/a/6972866/1468388) to provide template migration scripts:

```yaml
# copier.yml
_migrations:
  v1.0.2:
    pre:
      - git init
    post:
      - pre-commit install
```

When the template gets updated, it knows its old version and the new one. Using both, it can know which tasks to run.

Yajo:
This commit fixes #119. Summary:

- Depend on packaging to get a good implementation of PEP 440.
- Document new behaviors.
- Add support for migrations.
- Checkout by default the most modern git tag available in the template.
- Reuse `run_tasks` to run migrations too.
- Use `git describe --tags --always` to obtain template commit, to get a version parseable by PEP 440.
- Update `test_updatediff` to with new behaviors.

@Tecnativa TT20357


pykong:
LGTME. Merge as you see fit.

PR #126, opened 2020-02-25
Remove .vscode/ and .prettierrc.yml

pykong:
@Yajo Thanks for your continued contribution to this package, you are really pushing the envelope!

I feel we should remove both the configuration for `Visual Studio Code` and `prettier`.
In the case of the former: I want this project to be editor/IDE-agnostic. Hence, I do not want to vouch for any particular editor by including its config file. Feel free to update your `.gitignore` in accordance.

In the case of the latter: While `prettier` is an awesome tool, it belongs to JavaScript land.
For formatting, I want to stick to standard python tools (bools) or at least to something that runs with `pre-commit`

Merge this **PR** once you are sure it does not break your local configuration (just in case...).


pykong:
OK. In case of `prettier` I do not know of any other tool to format yaml files.
At least we should move its minimal configuration to `.pre-commit-config.yaml`

Yajo:
I don't like this PR :sweat_smile: 

Supplying `.vscode` doesn't mean you cannot use any other editor, it just means that using this one can make your life easier out of the box. Just like python, or docker, good tools have "Batteries built-in but swappable".

About prettier, it's simply awesome. Not only for YAML, we're using it here also to format TOML, MD, and JSON. The fact it's not written in python is IMO not enough reason to get rid of it, as long as pre-commit handles that transparently (it does).

Copier 3 should be the best scaffolding tool in the planet. It should serve to scaffold projects in any language, not only python, and it should be a decent competitor to https://yeoman.io/ (from the JS world, also usable for python projects). We should be welcoming to other languages from the core.

pykong:
@Yajo Thanks for your answer.

I am afraid but I do not want configs for specific editors/IDEs in this project out of principle.
Could you, therefore, please remove it? That would be very welcome.

However, we indeed need to keep the `pretttierrc` file, as prettier does not support TOML's tables but expects all its entries at root level => We can not put it inside `pyproject.toml`. (It does not fare well with the other hack I am using to feed flake8 with the `pyproject.toml` file.)

Issue #105, opened 2019-11-24
Ditch Travis CI for GitHub actions


pykong:
With the arrival of GitHub Actions a CI tool native to GitHub is here.
Switching over from Travis CI to GitHub Actions should be considered.

pykong:
Eventually https://github.com/pre-commit/action might be a good choice to run our pre-commits.
See also: https://github.com/pykong/copier/pull/104

PR #127, opened 2020-02-25
Migrate to poetry and GitHub Actions

pykong:
I  love [poetry](https://python-poetry.org/).
It makes everything related to package development a pure joy.

This **PR** migrates from `setup.py` to `pyproject.toml`.
This file can also hold the configuration settings for all tools, like black, flake8, isort and pytest.
Hence it largely tidies up our root folder as a nice bonus.

All tests run green locally, however, I have not yet updated the .travis.yml, as I am ditching Travis CI altogether with another **PR** (kind of hen-egg-problem here ;-))

I have not checked the build system.
The latter I have never used so far anyway and intend to switch to [flit](https://github.com/takluyver/flit) later on.

Closes: #105 

@Yajo Do you have any objections to merging this **PR**?


Yajo:
I love poetry also, go for it, but please do not merge until CI goes :heavy_check_mark:.

Set as WIP until you have ready travis' replacement then please.

Now that we're replacing stuff, you might consider replacing that nasty outdated Makefile in favor of https://www.pyinvoke.org/ :wink:

pykong:
@Yajo Great. Seems we have the same ideas about tooling.
Indeed my next suggestion was to also ditch the Makefile.
See: #129

pykong:
I needed to extend the scope of this **PR** to also migrate to GitHub Actions as otherwise, CI would fail.

Issue #118, opened 2020-02-19
Enhancement: Unify exclude/include


pykong:
The readme states:

```md
# Shell-style patterns files/folders that must not be copied.
_exclude:
  - "*.bar"
  - ".git"
  - ".git/*"

# Shell-style patterns files/folders that *must be* copied, even if
# they are in the exclude list
_include:
  - "foo.bar"
```

Probably it would be better to just have a single setting `exclude` but enhance it with a negating syntax that would allow to specify paths to include nonetheless. Maybe we could take gitignores syntax as an inspiration here. A leading `!` will make a path non-ignored. See: https://git-scm.com/docs/gitignore#_examples

I have found the following libraries which mplement the kind of pattern matching we are looking for:
- https://github.com/cpburnz/python-path-specification
- https://github.com/tsileo/dirtools
- https://github.com/untzag/gitsieve

pykong:
@Yajo I can implement this. 
The package [pathspec](https://github.com/cpburnz/python-path-specification) works like a breeze. :-)
However, it does not yet feature unicode support. Hence files like `ñana.txt` would never get excluded, because of their special char.

I assume for you, as well as other Spanish speaking users such unicode is absolutely vital?

Yajo:
I like the idea of this PR, and also using a good lib that handles this transparently, but yes it should support unicode. Otherwise it's a regression.

Another idea is to use regexps, but that's maybe not as comfortable as a proper glob system. `pre-commit` uses regexs, in case it matters.

PR #128, opened 2020-02-26
gitignore-style pattern matching

pykong:
@Yajo Please have a look.

- gitignore style matching for exclude and skip
- abdicate pattern matching to [pathspec](https://github.com/cpburnz/python-path-specification)
- expands tests
- updates readme
- works with unicode

Closes: #118


pykong:
@Yajo Do not want to be impolite, but I need to merge this now before you could have a look.

Issue #117, opened 2020-02-18
Transclude other YAML files


pykong:
As already described in #116, I like to think up a simpler and more robust way than **PR** #79.

**I like to have a way to reuse settings across projects.**
For example most of my projects run an identical chain of commands _(e.g. start VE, init git, etc. ...)_.

Instead of allowing some complex folder inheritance as proposed by **PR** #79, which potentially opens a can of worms with many unknown side effects, simply allowing to reference other yaml files and their settings would cover most my use cases.

Potential ways to accomplish this are described here:
https://stackoverflow.com/questions/528281/how-can-i-include-a-yaml-file-inside-another

Or we employ Jinja for this purpose as well. That would make sense, as we are already including other files via Jinja.

Yajo:
One solution would be to allow jinja inside `copier.yml` itself, but that'd be a simplified jinja that wouldn't include variables that will be later obtained from `copier.yml` itself.

Another solution would be to add a `_include_manifests:` key, which lets you include other files such as `copier.01.yml`, `copier.02.yml`, etc. All those get merged in order, and the final result is used as a single `copier.yml` file.

We really need #110 fixed, or it will keep complicating everything... :thinking:

PR #136, opened 2020-02-27
Include other config files

pykong:
This **PR** allows reusing settings from other config files.

It works its magic via: [pyyaml-include](https://github.com/tanbro/pyyaml-include)
(See its readme for use instruction)

This also means we need to ditch [ruamel.yam](https://pypi.org/project/ruamel.yaml/)l for [PyYaml](https://github.com/yaml/pyyaml)
In the past the former had a much better reputation than the latter.
But it seems no more, see:
- https://github.com/pre-commit/pre-commit/issues/897

@Yajo Do you see any reason we should not merge this **PR**?

Closes: #117


pykong:
@Yajo Can you have a look plz? (Just in case GitHub did not inform you about being assigned as a reviewer.)

tests/demo_config_transclude/demo/copier.yml
```diff
@@ -0,0 +1 @@
+!include ../include_me.yml
```
Yajo: IMHO this test is invalid.

You have to test the worst case: what happens if you have to include 2 files that define the same keys?

pykong: I am expanding the tests: #146

PR #144, opened 2020-02-28
Safer & faster YAML dump

Yajo:
Leverage pyyaml for a couple of jobs:

- Harden the possibility of dumping unsafe YAML, which was almost impossible already, but it's safer now.
- Do not sort twice; pyyaml already sorts.

@Tecnativa TT20357


pykong:
Great! THX!

PR #146, opened 2020-02-28
Include test for key collision when including more than two files

pykong:


copier.sublime-project
```diff
@@ -0,0 +1,28 @@
...
+			"file_regex": "^[ ]*File \"(...*?)\", line ([0-9]*)",
+			"name": "Anaconda Python Builder",
+			"selector": "source.python",
+			"shell_cmd": "\"/home/ben/Projects/copier/.venv/bin/python\" -u \"$file\""
+		}
+	],
+	"folders":
+	[
+		{
+			"path": "/home/ben/Projects/copier"
```
Yajo: Didn't you say you want no editor-specific files? 🤔

Imagine someone isn't called Ben! 😁😁

Yajo:
I'm not sure these configurations should be invalid. They seem valid to me. Instead, if there are overlapping keys, Copier should take the latest.

copier.sublime-project
pykong: Ouch. Yes that one slipped through.
I killed it once it entered `master`. :-|

pykong:
> I'm not sure these configurations should be invalid. They seem valid to me. Instead, if there are overlapping keys, Copier should take the latest.

I'll have a look into that. THX!

pykong:
I believe the key collision tests are correct, as YAML does not allow duplicate keys:
https://stackoverflow.com/questions/47668308/duplicate-key-in-yaml-configuaration-file

I do not think that overlapping keys should pass silently, therefore.

Yajo:
OK, we can consider that as a future feature for now, but I guess that some heavy user of this will feel that as a need. :man_shrugging:

PR #147, opened 2020-02-28
Use None instead of "None" for str answer defaults

Yajo:
Any other method, if it gets casted from `None` it will return the answer default.

In the case of `str`, `None` would default to `"None"`, which is not desired.

@Tecnativa TT20357


Yajo:
BTW I'm adding cache to builds, this should save a couple of mins per run.

PR #156, opened 2020-03-04
Do not attempt to run latter tasks in previous copies

Yajo:
Due to copier not reloading the `copier.yml` file when doing the previous step to diff updates, it was trying to run tasks from let's say `v2` of the template after copying its `v1` version.

After this change, when copying `v1`, its `copier.yml` is reloaded, loading tasks for that version only (and executing them as usual).

The patch includes a git bundle that is a git repo with a couple of commits required to replicate the bug.

@Tecnativa TT20357


Yajo:
CI is way faster now, for free!

PR #158, opened 2020-03-09
More space among questions

Yajo:
As questions can include now a more detailed help and types, it would be helpful to have them more spaced. It looks better.


pykong:
Memory is quite cheap these days, so I do not mind those additional bytes needed. 😄 

LGTM.

PR #160, opened 2020-03-12
Tasks and migrations env utilities

### LINKED ISSUE #95: Ability to use source template path in tasks section
Sorry for opening so many feature requests, but it's so you have them on your radar. :sweat_smile: 

The tasks section allows you to write scripts in a YAML file. That can be useful for short scripts, but if you need something serious, it becomes uncomfortable.

If there were some var such as `[[ source_dir ]]` then I could use tasks such as:

```yaml
# copier.yml
_exclude:
  - post-clone*
_tasks:
  - pip install "[[ source_dir ]]/post-clone-dependencies.txt"
  - "[[ source_dir ]]/post-clone.py"
```

### LINKED ISSUE #157: Leverage invoke for tasks and migrations
I'm developing a big template: https://github.com/Tecnativa/doodba-copier-template

While thinking about migrations and pre/post tasks, it seems that commands in a yaml file isn't the best option:

1. It must be sh syntax, which is :hankey: 
1. Not all envs have sh (windows).
1. Big migrations will not be much maintainable.

So, I'm thinking that we should refactor that part. YAML is not a programming language, and thus maintaining it would be awful in big templates.

As we discussed in #129 that invoke seems a good alternative, I'm thinking if we should leverage it for migrations. It provides full flexibility while not adding much complexity to the YAML definitions.

I'm adding milestone 3.0.0 as this is an important breaking change.

How should it work?

Well, assuming many projects will want to bootstrap a `tasks.py` file for their copied templates, let's use a default [collection](http://docs.pyinvoke.org/en/stable/concepts/loading.html) `copier_tasks.py` (you know, it could be `copier_tasks/__init__.py` instead if you want).

If copier finds the desired hooks in that file, it will leverage invoke to run them.

As Invoke would run under copier's venv (in case copier is installed in one via, i.e., pipx), those tasks would have access to all of copier dependencies, which include pyyaml and plumbum. Quite helpful ones for migrations.

Yajo:
The rendering context now includes `_copier_conf`, which is a copy of the raw `ConfigData` object, but removing anything not JSON-serializable (a.k.a. `data.now` and `data.make_secret`).

This is useful for a number of purposes:

1. Imagine you have migration scripts but you want to store them on your template and NOT in your copies. Then you can use `_copier_conf.src_path` to find the local clone and call anything from there. This fixes #95 and fixes #157.
2. In the execution environment, it includes `$STAGE` (can be `task`, `after` or `before`), and in the case of migrations also `$VERSION_{FROM,TO,CURRENT}` to let migration scripts know where they are in the moment they're called, so we can reuse migration scripts and unleash their full power.

Now you can write your scripts in any language you want, and use any task execution system you want. Tests include a bash script and a python script, and docs include how to integrate with Invoke.

@Tecnativa TT20357


Yajo:
As a present, I added smarter caching.

1st run: 3m29s https://github.com/pykong/copier/actions/runs/55047625
2nd run: 1m12s https://github.com/pykong/copier/actions/runs/55049232

Issue #173, opened 2020-04-08
Autodoc


Yajo:
Readme's api docs are outdated. It would be best to have some kind of autodoc system. Even more if you think that fixing #110 would change them a lot.

pawamoy:
[mkdocs](https://github.com/mkdocs/mkdocs) + [mkdocstrings](https://github.com/pawamoy/mkdocstrings) + [mkdocs](https://github.com/squidfunk/mkdocs-material) :smile: ?

PR #190, opened 2020-04-23
Add prompt templating

### LINKED ISSUE #90: Prompt self-templating
We need to go from:

```yml
# copier.yml
smtp_host: example.com
smtp_user: postmaster@example.com
```

To: 

```yml
# copier.yml
smtp_host: example.com
smtp_user: postmaster@[[smtp_host]]
```

pawamoy:
fix #90 

TODO:
- [x] docstrings
- [x] tests
- [x] documentation

### DIFF
--- a/copier/config/factory.py
+++ b/copier/config/factory.py
@@ -84,9 +84,12 @@ def make_config(
     config_data, questions_data = filter_config(file_data)
     config_data.update(_metadata)
     del _metadata
-    config_data["data"] = query_user_data(questions_data, answers_data, not force)
     args = {k: v for k, v in locals().items() if v is not None and v != []}
+    env_ops = EnvOps(**config_data.get("envops", {}))
+    config_data["data"] = query_user_data(
+        questions_data, answers_data, not force, env_ops
+    )
     args = {**config_data, **args}
-    args["envops"] = EnvOps(**args.get("envops", {}))
+    args["envops"] = env_ops
     args["data"].update(config_data["data"])
     return ConfigData(**args)
--- a/copier/config/user_data.py
+++ b/copier/config/user_data.py
@@ -1,15 +1,19 @@
 import json
 import re
+from functools import partial
 from pathlib import Path
 from typing import Any, Callable, Dict
 
 import yaml
+from jinja2 import UndefinedError
+from jinja2.sandbox import SandboxedEnvironment
 from plumbum.cli.terminal import ask, choose, prompt
 from plumbum.colors import bold, info, italics
 from yamlinclude import YamlIncludeConstructor
 
-from ..tools import INDENT, printf_exception
-from ..types import AnyByStrDict, OptStrOrPath, PathSeq, StrOrPath
+from ..tools import INDENT, get_jinja_env, printf_exception
+from ..types import AnyByStrDict, Choices, OptStrOrPath, PathSeq, StrOrPath
+from .objects import DEFAULT_DATA, EnvOps, UserMessageError
 
 __all__ = ("load_config_data", "query_user_data")
 
@@ -108,8 +112,43 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
         return answer
 
 
+def render_value(value: Any, env: SandboxedEnvironment, context: AnyByStrDict) -> str:
+    """Render a single templated value using Jinja.
+
+    If the value cannot be used as a template, it will be returned as is.
+    """
+    try:
+        template = env.from_string(value)
+    except TypeError:
+        # value was not a string
+        return value
+    try:
+        return template.render(**context)
+    except UndefinedError as error:
+        raise UserMessageError(str(error)) from error
+
+
+def render_choices(
+    choices: Choices, env: SandboxedEnvironment, context: AnyByStrDict
+) -> Choices:
+    """Render a list or dictionary of templated choices using Jinja."""
+    render = partial(render_value, env=env, context=context)
+    if isinstance(choices, dict):
+        choices = {render(k): render(v) for k, v in choices.items()}
+    elif isinstance(choices, list):
+        for i, choice in enumerate(choices):
+            if isinstance(choice, (tuple, list)) and len(choice) == 2:
+                choices[i] = (render(choice[0]), render(choice[1]))
+            else:
+                choices[i] = render(choice)
+    return choices
+
+
 def query_user_data(
-    questions_data: AnyByStrDict, answers_data: AnyByStrDict, ask_user: bool
+    questions_data: AnyByStrDict,
+    answers_data: AnyByStrDict,
+    ask_user: bool,
+    envops: EnvOps,
 ) -> AnyByStrDict:
     """Query the user for questions given in the config file."""
     type_maps: Dict[str, Callable] = {
@@ -120,12 +159,19 @@ def query_user_data(
         "str": str,
         "yaml": parse_yaml_string,
     }
-    result: AnyByStrDict = {}
+    env = get_jinja_env(envops=envops)
+    result = DEFAULT_DATA.copy()
+
+    _render_value = partial(render_value, env=env, context=result)
+    _render_choices = partial(render_choices, env=env, context=result)
+
     for question, details in questions_data.items():
         # Get default answer
-        answer = default = answers_data.get(question, details.get("default"))
+        answer = default = answers_data.get(
+            question, _render_value(details.get("default"))
+        )
         # Get question type; by default let YAML decide it
-        type_name = details.get("type", "yaml")
+        type_name = _render_value(details.get("type", "yaml"))
         try:
             type_fn = type_maps[type_name]
         except KeyError:
@@ -134,13 +180,19 @@ def query_user_data(
             # Generate message to ask the user
             message = f"\n{INDENT}{bold | question}? Format: {type_name}\n🎤 "
             if details.get("help"):
-                message = f"{info & italics | details['help']}\n{message}"
+                message = (
+                    f"{info & italics | _render_value(details['help'])}\n{message}"
+                )
             # Use the right method to ask
             if type_fn is bool:
                 answer = ask(message, default)
             elif details.get("choices"):
-                answer = choose(message, details["choices"], default)
+                choices = _render_choices(details["choices"])
+                answer = choose(message, choices, default)
             else:
                 answer = prompt(message, type_fn, default)
         result[question] = cast_answer_type(answer, type_fn)
+
+    for key in DEFAULT_DATA:
+        del result[key]
     return result
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -17,8 +17,10 @@ from .config.objects import ConfigData, EnvOps
 from .types import (
     AnyByStrDict,
     CheckPathFunc,
+    Filters,
     IntSeq,
     JSONSerializable,
+    LoaderPaths,
     StrOrPath,
     StrOrPathSeq,
     T,
@@ -103,15 +105,30 @@ def to_nice_yaml(data: Any, **kwargs) -> str:
     return result or ""
 
 
+def get_jinja_env(
+    envops: EnvOps,
+    filters: Optional[Filters] = None,
+    paths: Optional[LoaderPaths] = None,
+    **kwargs: Any,
+) -> SandboxedEnvironment:
+    """Return a pre-configured Jinja environment."""
+    loader = FileSystemLoader(paths) if paths else None
+    # We want to minimize the risk of hidden malware in the templates
+    # so we use the SandboxedEnvironment instead of the regular one.
+    # Of couse we still have the post-copy tasks to worry about, but at least
+    # they are more visible to the final user.
+    env = SandboxedEnvironment(loader=loader, **envops.dict(), **kwargs)
+    default_filters = {"to_nice_yaml": to_nice_yaml}
+    default_filters.update(filters or {})
+    env.filters.update(default_filters)
+    return env
+
+
 class Renderer:
     def __init__(self, conf: ConfigData) -> None:
         envops: EnvOps = conf.envops
         paths = [str(conf.src_path)] + list(map(str, conf.extra_paths or []))
-        # We want to minimize the risk of hidden malware in the templates
-        # so we use the SandboxedEnvironment instead of the regular one.
-        # Of couse we still have the post-copy tasks to worry about, but at least
-        # they are more visible to the final user.
-        self.env = SandboxedEnvironment(loader=FileSystemLoader(paths), **envops.dict())
+        self.env = get_jinja_env(envops=envops, paths=paths)
         self.conf = conf
         answers: AnyByStrDict = {}
         # All internal values must appear first
@@ -133,7 +150,6 @@ class Renderer:
             _copier_answers=answers,
             _copier_conf=conf.copy(deep=True, exclude={"data": {"now", "make_secret"}}),
         )
-        self.env.filters["to_nice_yaml"] = to_nice_yaml
 
     def __call__(self, fullpath: StrOrPath) -> str:
         relpath = (
--- a/copier/types.py
+++ b/copier/types.py
@@ -1,5 +1,15 @@
 from pathlib import Path
-from typing import Any, Callable, Dict, Optional, Sequence, TypeVar, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    TypeVar,
+    Union,
+)
 
 # simple types
 StrOrPath = Union[str, Path]
@@ -21,5 +31,8 @@ OptAnyByStrDict = Optional[AnyByStrDict]
 
 # miscellaneous
 CheckPathFunc = Callable[[StrOrPath], bool]
+Choices = Union[List[Any], Dict[Any, Any]]
 T = TypeVar("T")
 JSONSerializable = (dict, list, str, int, float, bool, type(None))
+Filters = Dict[str, Callable]
+LoaderPaths = Union[str, Iterable[str]]
--- a/tests/test_templated_prompt.py
+++ b/tests/test_templated_prompt.py
@@ -1,0 +1,179 @@
+import io
+from datetime import datetime
+
+import pytest
+
+from copier.config import make_config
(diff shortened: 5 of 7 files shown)


pawamoy:
Hmmm... CONTRIBUTING does not mention Poetry at all, nor mypy. It mentions tox, but tox is not a dependency listed in pyproject.toml. I guess it should be updated :sweat_smile:

Yajo:
You guess right. The docs are not good right now. :man_shrugging: To be done in #173

Yajo:
I took the freedom to convert this PR to draft, given you still have some TODO points pending.

Please ping me when it's ready to review, thanks!

pawamoy:
Should I add docs in the big "Advanced prompt formatting" YAML block, or in second one below?

pawamoy:
@Yajo ready :slightly_smiling_face:

copier/config/factory.py
```diff
@@ -84,9 +84,12 @@ def make_config(
     config_data, questions_data = filter_config(file_data)
     config_data.update(_metadata)
     del _metadata
-    config_data["data"] = query_user_data(questions_data, answers_data, not force)
     args = {k: v for k, v in locals().items() if v is not None and v != []}
+    env_ops = EnvOps(**config_data.get("envops", {}))
+    config_data["data"] = query_user_data(
+        questions_data, answers_data, not force, env_ops.dict()
```
Yajo, line 90: IMHO you better pass directly the `EnvOps` to benefit from typing.

```suggestion
        questions_data, answers_data, not force, env_ops
```

copier/config/user_data.py
```diff
@@ -108,8 +112,36 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
         return answer
 
 
+def render_string(template, env, context):
```
Yajo, line 115: Missing types.

copier/config/user_data.py
```diff
@@ -108,8 +112,36 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
...
+    try:
+        return env.from_string(template).render(**context)
+    except TypeError:
+        # template was not a string
+        return template
+    except UndefinedError as error:
+        raise UserMessageError(str(error)) from error
+
+
+def render_choices(choices, env, context):
```
Yajo, line 126: Missing types.

copier/config/user_data.py
```diff
@@ -108,8 +112,36 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
...
+                choices[i] = render(choice)
+    return choices
+
+
 def query_user_data(
-    questions_data: AnyByStrDict, answers_data: AnyByStrDict, ask_user: bool
+    questions_data: AnyByStrDict,
+    answers_data: AnyByStrDict,
+    ask_user: bool,
+    envops: dict,
```
Yajo, line 144: Following one comment above:

```suggestion
    envops: EnvOps,
```

copier/config/user_data.py
```diff
@@ -120,12 +152,20 @@ def query_user_data(
         "str": str,
         "yaml": parse_yaml_string,
     }
-    result: AnyByStrDict = {}
+    env = SandboxedEnvironment(undefined=StrictUndefined, **envops)
```
Yajo, line 155: Is there no way to extend `copier.tools.Renderer` for this use case instead of having to do all of this manually here?

tests/test_templated_prompt.py
```diff
@@ -0,0 +1,153 @@
+import io
+from datetime import datetime
+
+import pytest
+
+from copier.config.objects import DEFAULT_DATA, EnvOps, UserMessageError
+from copier.config.user_data import query_user_data
+
+answers_data = DEFAULT_DATA.copy()
+envops = EnvOps().dict()
```
Yajo, line 10:
```suggestion
envops = EnvOps()
```

copier/config/user_data.py
pawamoy: Doesn't seem like it. The `Renderer` needs a `ConfigData` object which needs `query_user_data()`. Making `query_user_data()` use a `Renderer` would create a logical cyclic dependency :confused:

copier/config/user_data.py
pawamoy: Done.

copier/config/user_data.py
pawamoy: Done.

tests/test_templated_prompt.py
pawamoy: Done.

copier/config/user_data.py
pawamoy: Done.

copier/config/factory.py
pawamoy: Done.

pawamoy:
Hmmm... it's passing in local. Is it possible to re-run a job in GitHub workflows?

pawamoy:
Could be a discrepancy in Pydantic versions. Maybe wiping the cache would fix it.

pawamoy:
Silly me... forgot to add the new test demo directory... Should be fine now.

copier/types.py
```diff
@@ -21,5 +22,6 @@
 
 # miscellaneous
 CheckPathFunc = Callable[[StrOrPath], bool]
+Choices = Union[Sequence[AnyOrTupleAny], Dict[Any, Any]]
```
Yajo, line 25: You can't specify a list's length by static typing because they're mutable: https://stackoverflow.com/a/44833864/1468388

So the closest you can get is:

```suggestion
Choices = Union[List[Any], Tuple[Any, Any], Dict[Any, Any]]
```

copier/config/user_data.py
```diff
@@ -108,8 +112,36 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
...
+    render = partial(render_string, env=env, context=context)
+    if isinstance(choices, dict):
+        choices = {render(k): render(v) for k, v in choices.items()}
+    elif isinstance(choices, list):
+        for i, choice in enumerate(choices):
+            if isinstance(choice, (tuple, list)) and len(choice) == 2:
+                choices[i] = (render(choice[0]), render(choice[1]))
+            else:
+                choices[i] = render(choice)
+    return choices
```
Yajo, line 137: By using `NativeEnvironment`, all this is very simplified:

```suggestion
    return render(repr(choices))
```

Check this interactive session so you understand what's happening here:

```py
>>> from jinja2.nativetypes import NativeEnvironment                                                                                                                                                              

>>> choices = {"{{ name1 }}": 'value1', '{{ name2 }}':2}                                                                                                                                                          

>>> NativeEnvironment().from_string(repr(choices)).render(name1=0, name2="hey")                                                                                                                                   
{'0': 'value1', 'hey': 2}


>>> choices = ""                                                                                                                                                                                                  

>>> NativeEnvironment().from_string(repr(choices)).render(name1=0, name2="hey")                                                                                                                                   
''


>>> choices = "good"                                                                                                                                                                                              

>>> NativeEnvironment().from_string(repr(choices)).render(name1=0, name2="hey")                                                                                                                                   
'good'
```

Actually it's so much simplified that you could even remove this method.

copier/config/user_data.py
```diff
@@ -108,8 +112,36 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
...
 
+def render_string(template: Any, env: SandboxedEnvironment, context: AnyByStrDict):
+    """Render a templated-string using Jinja."""
+    try:
+        return env.from_string(template).render(**context)
+    except TypeError:
+        # template was not a string
+        return template
+    except UndefinedError as error:
+        raise UserMessageError(str(error)) from error
```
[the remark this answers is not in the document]
pawamoy: > and the returned result would be typed as expected

But what is *expected* here? If the `type` is `str` and the `default` is `"3"`, we expect a string, not an integer. In that case it's still fine because the answer will be cast again as `str`. But if the type is not set, I think the user still expects the default to be a string and not an integer, and the default `yaml` type will leave it as an integer.

copier/config/user_data.py
```diff
@@ -120,12 +152,20 @@ def query_user_data(
...
+
+    _render_string = partial(render_string, env=env, context=result)
+    _render_choices = partial(render_choices, env=env, context=result)
+
     for question, details in questions_data.items():
         # Get default answer
-        answer = default = answers_data.get(question, details.get("default"))
+        answer = default = _render_string(
+            answers_data.get(question, details.get("default"))
+        )
```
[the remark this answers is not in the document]
pawamoy: Oh yes, I was too quick on that one, thanks.

copier/types.py
```diff
@@ -1,9 +1,10 @@
 from pathlib import Path
-from typing import Any, Callable, Dict, Optional, Sequence, TypeVar, Union
+from typing import Any, Callable, Dict, Optional, Sequence, Tuple, TypeVar, Union
 
 # simple types
 StrOrPath = Union[str, Path]
 AnyByStrDict = Dict[str, Any]
+AnyOrTupleAny = Union[Any, Tuple[Any, Any]]
```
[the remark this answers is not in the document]
pawamoy: My typing-fu is not the best yet :sweat_smile: Thanks

copier/types.py
pawamoy: I wanted to represent the possibility for the list to be composed of both strings and tuples of 2 strings.

copier/config/user_data.py
pawamoy: Wow, this is great! Thanks, I'll use that.

pawamoy: In fact using `repr` will not work:

```python
>>> from jinja2.nativetypes import NativeEnvironment

>>> nenv = NativeEnvironment(variable_start_string='[[', variable_end_string=']]')

>>> choices = [['one', 'first'], ['two', 'second'], ['three', 'third']]

>>> nenv.from_string(repr(choices)).render()
TemplateSyntaxError: unexpected ']'
```

copier/config/user_data.py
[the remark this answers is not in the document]
Yajo: Hmm I see, you're right.

Then I guess the best you can do is modify the method name. Calling it `render_string` and expecting things that are not strings is confusing to the reader.

Maybe something like `render_if_string`? :thinking: 

Also, since all the type casting is indeed done later, I guess all of the comments about using `NativeEnvironment` should be ignored. Let me resolve them too.

copier/config/user_data.py
Yajo: hummmm.

Well, I guess this can't be improved then. Thanks!

copier/config/user_data.py
```diff
@@ -120,12 +152,20 @@ def query_user_data(
         "str": str,
         "yaml": parse_yaml_string,
     }
-    result: AnyByStrDict = {}
+    env = SandboxedEnvironment(undefined=StrictUndefined, **envops.dict())
```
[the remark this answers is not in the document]
Yajo: Solved in https://github.com/pykong/copier/pull/190/files#r416481797.

copier/config/user_data.py
```diff
@@ -108,8 +112,36 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
...
+    try:
+        return env.from_string(template).render(**context)
+    except TypeError:
+        # template was not a string
+        return template
+    except UndefinedError as error:
+        raise UserMessageError(str(error)) from error
+
+
+def render_choices(choices: Choices, env: SandboxedEnvironment, context: AnyByStrDict):
```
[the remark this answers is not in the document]
Yajo: Solved in https://github.com/pykong/copier/pull/190/files#r416481797.

copier/types.py
Yajo: Yes, you're right.

Also, after all, YAML will never send us a tuple. It always evaluates to lists. So chances are to get a list or a dict.

copier/config/user_data.py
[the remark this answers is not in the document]
pawamoy: I agree that `render_string` is confusing! Maybe `render_value` then, since we're trying to render the *values* of the YAML keys? Otherwise `render_if_string` is fine as well!

copier/types.py
pawamoy: Done.

copier/config/user_data.py
```diff
@@ -120,12 +154,21 @@ def query_user_data(
         "str": str,
         "yaml": parse_yaml_string,
     }
-    result: AnyByStrDict = {}
+    env = get_env(
+        envops=envops, filters={"to_nice_yaml": to_nice_yaml}, undefined=StrictUndefined
```
pawamoy, line 158: Still passing `undefined=StrictUndefined` in kwargs I didn't want to make this behavior the default. But maybe we should? Not sure.

copier/tools.py
```diff
@@ -103,15 +103,29 @@ def to_nice_yaml(data: Any, **kwargs) -> str:
     return result or ""
 
 
+def get_env(
+    envops: EnvOps,
+    filters: Optional[Dict[str, Callable]] = None,
```
pawamoy, line 108: Should I add `Filters = Dict[str, Callable]` in `copier.types`?

copier/tools.py
```diff
@@ -103,15 +103,29 @@ def to_nice_yaml(data: Any, **kwargs) -> str:
...
+    paths: Optional[Union[str, Iterable[str]]] = None,
```
pawamoy, line 109: Should I add `LoaderPaths = Union[str, Iterable[str]]` in `copier.types`?

copier/tools.py
```diff
@@ -103,15 +103,29 @@ def to_nice_yaml(data: Any, **kwargs) -> str:
...
+    **kwargs: Any,
+):
+    loader = FileSystemLoader(paths) if paths else None
+    # We want to minimize the risk of hidden malware in the templates
+    # so we use the SandboxedEnvironment instead of the regular one.
+    # Of couse we still have the post-copy tasks to worry about, but at least
+    # they are more visible to the final user.
+    env = SandboxedEnvironment(loader=loader, **envops.dict(), **kwargs)
+    env.filters.update(filters or {})
```
pawamoy, line 118: As said in another comment, maybe we should add filters here by default.

copier/tools.py
```diff
@@ -103,15 +103,29 @@ def to_nice_yaml(data: Any, **kwargs) -> str:
...
     def __init__(self, conf: ConfigData) -> None:
         envops: EnvOps = conf.envops
         paths = [str(conf.src_path)] + list(map(str, conf.extra_paths or []))
-        # We want to minimize the risk of hidden malware in the templates
-        # so we use the SandboxedEnvironment instead of the regular one.
-        # Of couse we still have the post-copy tasks to worry about, but at least
-        # they are more visible to the final user.
-        self.env = SandboxedEnvironment(loader=FileSystemLoader(paths), **envops.dict())
+        self.env = get_env(
+            envops=envops, filters={"to_nice_yaml": to_nice_yaml}, paths=paths
```
pawamoy, line 127: We could remove the filters parameter if we set them by default.

copier/config/user_data.py
pawamoy: So I took a shot at de-duplicating code between `Renderer` and `query_user_data`, let me know what you think.

copier/config/user_data.py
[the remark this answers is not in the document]
Yajo: > Maybe `render_value`

:ok_hand:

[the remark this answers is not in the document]
pawamoy: Done :slightly_smiling_face:

copier/config/user_data.py
Yajo: It would be nice to be able to configure undefined behavior in `_envops`, though. But that can be a separate improvement.

I'd like copier to be widely used outside of python projects.

I guess using more relaxed checks makes it easier for templaters. So, unless there's a clear advantage, I'd vote against being strict. Also, having different behaviors depending on the context might be more confusing to the templater.

So I'd just leave the same undefined behavior as we had until now.

About deduplicating the addition of `to_nice_yaml`, go ahead. :ok_hand:

copier/tools.py
Yajo: As you please.

copier/tools.py
Yajo: As you please.

copier/config/user_data.py
```diff
@@ -108,8 +112,38 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
         return answer
 
 
+def render_value(value: Any, env: SandboxedEnvironment, context: AnyByStrDict) -> str:
+    """Render a single templated value using Jinja."""
```
Yajo, line 116: Let's make this even more obvious:

```suggestion
    """Render a single templated value using Jinja.

    If the value cannot be used as a template, it will be returned as is.
    """
```

copier/config/user_data.py
```diff
@@ -108,8 +112,38 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
...
+    try:
+        return env.from_string(value).render(**context)
+    except TypeError:
+        # value was not a string
+        return value
```
Yajo, line 121: Also maybe we should handle errors more explicitly:

```suggestion
    try:
        template = env.from_string(value)
    except TypeError:
        # value was not a string
        return value
    try:
        return template.render(**context)
```

I'm worried about the case that the template itself is badly programmed and raises a `TypeError` when rendered.

copier/tools.py
Yajo: After reading more code, yes, this seems the best place for that.

copier/tools.py
```diff
@@ -103,15 +103,29 @@ def to_nice_yaml(data: Any, **kwargs) -> str:
     return result or ""
 
 
+def get_env(
```
Yajo, line 106: I guess this makes it more explicit. "Env" can mean many things.

```suggestion
def get_jinja_env(
```

copier/config/user_data.py
pawamoy: Continuing discussion in related comments.

copier/config/user_data.py
pawamoy: OK, let's leave it non-strict, and add a strict option to EnvOps in another PR :slightly_smiling_face: I'll have to comment out the test that was checking for undefined variables.

PR #205, opened 2020-05-07
Add minimum copier version verification

### LINKED ISSUE #198: Assert minimal copier version
The template needs a new key:

```yaml
# copier.yml
_min_copier_version: v3.1.0
```

Otherwise the template might fail badly if a template is done for a version that i.e. includes #190 and is executed under a copier that doesn't have it.

@pawamoy You might be interested in this.

pawamoy:
Fix #198 

TODO:
- [x] docs

### DIFF
--- a/copier/config/factory.py
+++ b/copier/config/factory.py
@@ -1,11 +1,12 @@
 from typing import Tuple
 
+import semver
 from plumbum import local
 from plumbum.cmd import git
 
 from .. import vcs
 from ..types import AnyByStrDict, OptAnyByStrDict, OptBool, OptStr, OptStrSeq
-from .objects import DEFAULT_DATA, ConfigData, EnvOps, NoSrcPathError
+from .objects import DEFAULT_DATA, ConfigData, EnvOps, NoSrcPathError, UserMessageError
 from .user_data import load_answersfile_data, load_config_data, query_user_data
 
 __all__ = ("make_config",)
@@ -28,6 +29,20 @@ def filter_config(data: AnyByStrDict) -> Tuple[AnyByStrDict, AnyByStrDict]:
             if v.get("secret"):
                 conf_data["secret_questions"].add(k)
     return conf_data, questions_data
+
+
+def verify_minimum_version(version: str) -> None:
+    """Raise an error if the current Copier version is less than the given version."""
+    # Importing __version__ at the top of the module creates a circular import
+    # ("cannot import name '__version__' from partially initialized module 'copier'"),
+    # so instead we do a lazy import here
+    from .. import __version__
+
+    if semver.compare(__version__, version) == -1:
+        raise UserMessageError(
+            f"This template requires Copier version >= {version}, "
+            f"while your version of Copier is {__version__}."
+        )
 
 
 def make_config(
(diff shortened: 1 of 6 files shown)


copier/config/factory.py
```diff
@@ -1,11 +1,14 @@
 from typing import Tuple
 
+import semver
 from plumbum import local
 from plumbum.cmd import git
 
+import copier
```
Yajo, line 7: Avoid this, do a relative import instead please.

copier/config/factory.py
```diff
@@ -58,6 +70,10 @@ def make_config(
     answers_data = DEFAULT_DATA.copy()
     answers_data.update(load_answersfile_data(dst_path, answers_file))
     answers_data.update(data or {})
+
+    if "_min_copier_version" in answers_data:
+        verify_minimum_version(answers_data["_min_copier_version"])
```
Yajo, line 75: I don't think this parameter should be saved in the answers file.

The one that declares the minimal version should be always the template, not the user IMHO.

Am I right?

copier/config/factory.py
```diff
@@ -81,6 +97,11 @@ def make_config(
                 _metadata["commit"] = git("describe", "--tags", "--always").strip()
     # Obtain config and query data, asking the user if needed
     file_data = load_config_data(src_path, quiet=True)
+
+    if "_min_copier_version" in file_data:
+        min_copier_version = file_data["_min_copier_version"]
+        verify_minimum_version(min_copier_version)
```
Yajo, line 103: [EAFP](https://docs.python.org/3.8/glossary.html#term-eafp):

```suggestion
    try:
        verify_minimum_version(file_data["_min_copier_version"])
    except KeyError:
        pass
```

copier/config/factory.py
pawamoy: Hmm I think you're right. Even when updating, this is covered by the verification of the min version in the template copier.yml, so we don't need to copy it in the answers file.

copier/config/factory.py
pawamoy: Absolute imports are recommended in PEP8, but I guess this is to maintain consistency across the project?

pawamoy: Also, I'm getting this:

```
from .. import __version__, vcs
E   ImportError: cannot import name '__version__' from partially initialized module 'copier' (most likely due to a circular import)
```

Yajo: Oh yes, of course. This is due to how that poetry plugin hacks the `__version__`.

I guess you'll be safer lazy-loading that:

```py
def copier_version() -> str:
    from .. import __version__
    return __version__
```

pawamoy: Do you really want to add this function, in `tools` maybe, or do I simply add `from .. import __version__` in the `verify_minimum_version` function?

copier/config/factory.py
```diff
@@ -30,6 +31,17 @@ def filter_config(data: AnyByStrDict) -> Tuple[AnyByStrDict, AnyByStrDict]:
     return conf_data, questions_data
 
 
+def verify_minimum_version(version: str) -> None:
+    """Raise an error if the current Copier version is less than the given version."""
+    from .. import __version__
```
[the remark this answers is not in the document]
pawamoy: Done :)

Yajo:
I was gonna merge but realized you didn't mention this in the readme. Are you going to do it or are you going to leave that for the other PR?

pawamoy:
Oh yes sorry, I'll add this in the very PR.

PR #209, opened 2020-05-18
Fix path to jinja get_template on Windows

H0R5E:
With a jinja Environment get_template call, the path must always be given with forward slashes (see https://github.com/pallets/jinja/issues/767). On windows, in `Renderer.__call__`, relpath is returning with backslashes, so I added a replace statement to switch them in this case.


Yajo:
Thanks, but this needs a test.

Yajo:
Oh what happened?

pawamoy:
I have the same issue, this should be reopened.

Yajo:
See #213

PR #236, opened 2020-07-28
Docs: Add a warning about "!include" usage with YAML files

achaussier:


README.md
```diff
@@ -376,6 +376,16 @@ common_setting: "1"
 !include other_place/include_me.yml
 ```
 
+> **Warning**: You can't have in the same file a top `!include` tag and classic YAML
```
Yajo, line 379: Please don't use quoting for things that are not quotes :blush:

achaussier: Rebase done and quoting removed.

PR #244, opened 2020-08-03
s/pykong/copier-org

pawamoy:
Also add a "thanks" mention for @pykong 🙂


pawamoy:
I'm not sure if third-party services like codecov will accept the new name (copier-org).

PR #248, opened 2020-08-10
A little polish on docs

Yajo:
- Prepare for a new release.
- Add configuring page, with more explicit docs and examples for each option.
- Remove configuration stuff from creation page.
- Comparisons are ugly, so remove them from the index page.


pawamoy:
Whew, nice 🎉

PR #253, opened 2020-08-13
Add --prereleases support

Yajo:
Before this patch, Copier didn't ignore prereleases when detecting latest template tag.

This is mostly a bug because there's no way to safely upgrade a template to the latest non-prerelease tag automatically.

This a behavioral change that probably didn't hit anybody out there, but enough to make a new big release.


Yajo:
Closed by mistake, continues in #254

Issue #154, opened 2020-02-28
Update CONTRIBUTING.md


pykong:
Once the dev setup has stabilized it will be time to reflect the changes back into the dev docs:
`CONTRIBUTING.md`.

See also: #129 #132

pawamoy:
Seeing how you improved CONTRIBUTING.md in [this commit](https://github.com/copier-org/copier/commit/4f36db4c72004ae4d124aa391db55737b8ecfdf3), maybe this issue could be closed?

Maybe just one thing I noticed: you track VSCode settings and set the path to the interpreter to `.venv/bin/python`, but if contributors do not tell Poetry to install the venv in the project, they will have to change that setting:

```diff
-  "python.pythonPath": ".venv/bin/python"
+  "python.pythonPath": "/home/pawamoy/.cache/pypoetry/virtualenvs/copier-AExR1Oeb-py3.9/bin/python"
```

Not sure how to circumvent that though.

Yajo:
Yes, that's because of https://github.com/microsoft/vscode-python/issues/8372.  For now, the only way to use both tools sanely and share out-of-the-box settings is to tell poetry to save the venv local to the project. 

I should document that.

Issue #129, opened 2020-02-26
Ditch Makefile


pykong:
I like to replace the old Makefile with something more modern.

My first idea was to use `pyproject.toml`'s script section to run with poetry:
[python-scripts](https://medium.com/octopus-wealth/python-scripts-26e3d0bd5277)

@Yajo suggests [pyinvoke](https://www.pyinvoke.org/).

Yajo:
Woah! I didn't know of poetry scripts... Then it's good to have a single tool indeed :+1:

Yajo:
I noticed that poetry's scripts section is to describe executables installed when installing your package. In our case, there'd be just one script called `copier`.

Then, using `poetry run $command` would be the equivalent to running in 2 steps:

```bash
poetry shell
> $command
```

Nothing special here. You could even run an external program, not related with copier.

So it's not the same as invoke, which is basically a make replacement in python. After this, I still consider invoke more suited for our purpose on replacing the makefile, as these are tasks more targeted to development automation, not for end user usage.

Actually invoke can be used even to install poetry and use it to bootstrap the dev env.

pykong:
@Yajo Thanks for the input.
What you think - how many commands do we really need?
(e.g. what should we salvage from the old Makefile?)

I like to strike the right balance here between a dev setup that is a lightweight and unimposing as possible and one that provides reasonable structure to the dev process.

Yajo:
Develop: to set up the development environment.
Test: to run all tests for the CI.

pawamoy:
I personally like to keep a Makefile in my projects, with its rules calling `poetry run`. It's just faster to type `make thing` than `poetry run command args`, and I benefit from make's autocompletion.

Yet I find it quite difficult to decide on which tool should be the entry point:
`make`? `bake`? `poetry`? `invoke`? which one is a dependency of the other?

For my future copier templates, I think I'll give these instructions to users/contributors:

Creation of a new project (between 1 and 2 lines, depending if you already have pipx or copier installed):
```bash
pip install --user pipx
pipx run copier https://github.com/pawamoy/copier-template project
# copier task calls make setup
# make setup installs poetry with pipx (if not found)
# and runs poetry install
```

Contributing on a copier-generated project (3 lines):
```bash
git clone https://github.com/pawamoy/project
cd project
make setup  # same as before
# copier is a dependency of the project, for its "update" feature
```

The Makefile would of course provide many other useful rules: build, check, clean, format, docs, help, publish, release, setup, test, etc.

[...]

PR #261, opened 2020-08-17
The new `[tool.poe.tasks]` section and new poetry based workflow are …

nat-n:
I've been working on a task runner that I think based on the discussion in #129 would be a good fit for this project :)

I also took a stab at #154 to keep things coherent. I'm happy to iterate if there are things we'd prefer to differently.

### Key features

- easy, succinct, and flexible toml based config 
- all tasks are executed via `poetry run` unless already inside a poetry shell
- for tasks defined as a single command, all additional arguments are passed to that command
- executing the poe tool without args displays help including available tasks
- supports [tab completion for zsh](https://github.com/nat-n/poethepoet#enable-tab-completion-for-zsh)

## Summary of changes in this PR

1. The new `[tool.poe.tasks]` section in the pyproject.toml functionally replicates the old make file.
  - Everything is implemented as simple cmd tasks that just basically just run the given content as a subprocess, except *clean* which is defined as a sequence of the two subtasks.
  - the two subtasks of *clean* are prefixed with `_` which means they're hidden and not directly runnable, though this is not necessary if we think using them directly would be useful
  - removed install task since I believe poetry makes it redundant 
2. Update CONTRIBUTING.md for the poetry + poe based workflow

## Other Features

Note that poe also supports:
- shell tasks which can use posix shell features such as multiple commands, pipes, boolean operators, background jobs, etc (but require a posix shell to be available so won't always work on windows)
- script tasks which reference a python function exactly like poetry scripts

It's also possible to define new tasks using other toml syntax forms like:

- inline tables
  ```toml
  run = { script = "copier.cli:CopierApp.run", help = "run the copier cli"}
  ```
- sub-tables
  ```toml
  [tool.poe.tasks.check-all]
  help = "Run all checks"
  sequence = [
    "check",
    "coverage"
  ]
  ```

## Try it out:

```bash
pip install poethepoet
poe test
```


.github/workflows/ci.yml
```diff
@@ -61,9 +61,14 @@ jobs:
...
-      - run: python -m pip install poetry poetry-dynamic-versioning
+      - name: Install poetry
         shell: bash
-      - run: poetry install
+        run: |
+          python -m pip install poetry poetry-dynamic-versioning
+          echo "::set-env name=PATH::$HOME/.poetry/bin:$PATH"
+      - run: |
+          poetry run python -m pip install pip -U
+          poetry install
```
Yajo, line 71: Please don't modify these lines.

First of all, You don't need `echo "::set-env name=PATH::$HOME/.poetry/bin:$PATH"` because you're installing poetry with pip. That is only used when installing it like in https://python-poetry.org/docs/#osx-linux-bashonwindows-install-instructions

Second, it doesn't make sense to upgrade pip after you've used it.

Finally, these steps are tightly coupled with cache, so I beg you to not touch them unless strictly required to have faster builds.

.github/workflows/ci.yml
```diff
@@ -99,8 +104,14 @@ jobs:
...
-      - run: python -m pip install poetry poetry-dynamic-versioning
-      - run: poetry install
+      - name: Install poetry
+        shell: bash
+        run: |
+          python -m pip install poetry poetry-dynamic-versioning
+          echo "::set-env name=PATH::$HOME/.poetry/bin:$PATH"
+      - run: |
+          poetry run python -m pip install pip -U
+          poetry install
```
Yajo, line 114: Again, please revert this.

CONTRIBUTING.md
```diff
@@ -40,6 +40,14 @@ If you are proposing a feature:
 -   Remember that this is a volunteer-driven project, and that contributions are welcome
     :)
 
+## Dev environment setup
+
+We use some tools as part of our development workflow which you'll need to install into
+your host environment:
+
+-   [poetry](https://python-poetry.org/) for packaging and dependency management
+-   [poethepoet](https://github.com/nat-n/poethepoet) for running development tasks
```
Yajo, line 49: Do you really need to install this? Isn't it installed automatically by poetry?

CONTRIBUTING.md
```diff
@@ -70,8 +77,8 @@ Now you can make your changes locally.
 6.  When you're done making changes, check that your changes pass all tests
 
 ```bash
-pytest -x .
-flake8 .
+poe test
+poe check
```
Yajo, line 81:
```suggestion
poetry run poe test
poetry run poe check
```

... just in case the guy forgot to activate poetry shell.

pyproject.toml
```diff
@@ -69,6 +69,36 @@ mkdocs-material = "^5.5.5"
...
+       pip-wheel-metadata
+       *.egg-info
+"""
+_clean-pyc = """
+rm -rf .pytest_cache
+       ./**/__pycache__
+       ./**/*.pyc
+       ./**/*.pyo
+       ./**/*~
+"""
```
Yajo, line 90: I don't think this is very useful because builds are done in CD jobs. Also these would fail on Windows:

```suggestion
```

pyproject.toml
```diff
@@ -69,6 +69,36 @@ mkdocs-material = "^5.5.5"
...
+
+docs = "mkdocs serve"
+
+test.cmd = "pytest -x"
```
Yajo, line 94:
```suggestion
test.cmd = "pytest -n auto --color=yes"
```

pyproject.toml
```diff
@@ -69,6 +69,36 @@ mkdocs-material = "^5.5.5"
...
+test.help = "run tests"
+
+check.cmd = "pre-commit run --all-files"
+check.help = "check (and auto-fix) style with pre-commit"
```
Yajo, line 98:
```suggestion
lint.cmd = "pre-commit run --all-files"
lint.help = "check (and auto-fix) style with pre-commit"
```

pyproject.toml
```diff
@@ -89,6 +119,18 @@ line_length = 88
...
+[tox]
+isolated_build = true
+envlist = py36, py37, py38
+[testenv]
+whitelist_externals = poetry
+commands =
+    poetry install -v
+    poe test
+"""
+
```
Yajo, line 133: We're not using tox because our build matrix is built at CI, so please remove this part too.

pyproject.toml
```diff
@@ -69,6 +69,36 @@ mkdocs-material = "^5.5.5"
 # HACK https://readthedocs.org/api/v2/build/11653917.txt
 # TODO Remove this dependency when mkdocstrings supports pymdown-extensions==8.0
 pymdown-extensions = "<8.0,>=6.3"
+tox = "^3.19.0"
```
Yajo, line 72: Remove tox (explained below), but add poethepoet please.

Also, add it above of the `HACK` line (around line 64)

Yajo (changes requested):

CONTRIBUTING.md
nat-n: not necessary, poe does it for you if you (assuming it's installed in the host environment).

CONTRIBUTING.md
nat-n: It's possible to have poe installed inside the poetry managed environment, and use it with poetry run/shell, but I've not added it as a dependency because I think it's more useful this way.

Having poe installed in the host environment (such as via pip, or pipx) is also necessary for tab completion to work.

.github/workflows/ci.yml
nat-n: I can revert all the changes except the pip upgrade which is necessary for `poetry install` to work if there is anything more than basic toml syntax used for anything in pyproject.toml, for example dotted paths cause pip to crash.

This is because pip is used inside `poetry install` (I believe there's a plan to change this in future) and the default version of pip uses pytoml for toml parsing which is deprecated and quite limited.

pyproject.toml
nat-n: This is meant to match the functionality in the makefile, which I guess wasn't very windows friendly either.

If it's really not wanted I can remove it, though I'll add that in my experience having an easy way to clean up various artefacts from the project sometimes comes in handy, as does occasionally creating full builds as part of the dev workflow.

pyproject.toml
nat-n: that command does a lot more than linting?

pyproject.toml
nat-n: OK, updating the CONTRIBUTING.md to match as well then.

pyproject.toml
nat-n: Done.

pyproject.toml
Yajo: Well, it lints and reformats IIRC.

pyproject.toml
Yajo: Well, if you add it, please use Python code, so it can be used on Windows. Windows support las only started being tested recently. Otherwise, I prefer to remove it and add it later if needed.

.github/workflows/ci.yml
```diff
@@ -61,9 +61,13 @@ jobs:
...
             cache|${{ runner.os }}|${{ env.PY }}|${{ hashFiles('pyproject.toml') }}|${{
             hashFiles('poetry.lock') }}|${{ hashFiles('.pre-commit-config.yaml') }}
-      - run: python -m pip install poetry poetry-dynamic-versioning
+      - name: Install poetry
         shell: bash
-      - run: poetry install
+        run: |
+          python -m pip install poetry poetry-dynamic-versioning
+      - run: |
+          poetry run python -m pip install pip -U
```
Yajo, line 69: I still don't think we need this line. Did you add it for some specific reason?

.github/workflows/ci.yml
```diff
@@ -61,9 +61,13 @@ jobs:
           key:
             cache|${{ runner.os }}|${{ env.PY }}|${{ hashFiles('pyproject.toml') }}|${{
             hashFiles('poetry.lock') }}|${{ hashFiles('.pre-commit-config.yaml') }}
-      - run: python -m pip install poetry poetry-dynamic-versioning
+      - name: Install poetry
         shell: bash
-      - run: poetry install
+        run: |
+          python -m pip install poetry poetry-dynamic-versioning
```
Yajo, line 67: Here you should add poethepoet right? So we can leverage our lovely brand new task manager to run the CI?

pyproject.toml
nat-n: My feeling is that putting "code" in a config file isn't great, so it's not directly supported. Still there're currently two ways to make it work:

```toml
_clean-build = """
python -c "import os;\
etc\
"
"""
```

which is not so nice,

Or we could put a separate python module in the project e.g. `./devtasks.py` or what have you and reference functions from it like:

```toml
clean.script = "devtasks:clean"
```

which is clean enough in the toml, but comes with the overhead of another free floating python module in this case.

Also for the sake of discussion: it's fairly easy to add a new task type, I'm even thinking poe should support task types defined as third party plugins, so a plumbum or raw python or whatever task type is conceivable. Though it would take a bit more analysis and development to support.

What do you think?

.github/workflows/ci.yml
nat-n: The following poetry install like will fail without it, at least for the python 3.6 environment. I guess it could be made to only apply in that part of the test matrix, though I don't know if it's worth the complexity.

This build demonstrates what I mean: https://github.com/copier-org/copier/runs/995092646?check_suite_focus=true

pyproject.toml
Yajo: > we could put a separate python module in the project e.g. `./devtasks.py` or what have you and reference functions from it

Seems like a good option!

.github/workflows/ci.yml
Yajo: But that message:

```
pip._vendor.pytoml.core.TomlError: /home/runner/work/copier/copier/pyproject.toml(75, 1): msg
```

Doesn't it mean there's a syntax error in the toml?

nat-n: well sort of; from the perspective of the pytoml implementation which doesn't properly support all of the toml spec. This seems to be the biggest problem with the PEP-517/pyproject.toml design; different libraries use different toml parsers with different limitations but they all have to parse the same file.

AFAIK the most compete toml parsing library available for python is tomlkit which poetry and poe both use.

.github/workflows/ci.yml
nat-n: good idea, how do you like it now?

I also removed the redundant poetry install in the build stage.

pyproject.toml
```diff
@@ -1,8 +1,7 @@
-[tool]
 [tool.poetry]
 name = "copier"
 # This version is a placeholder autoupdated by poetry-dynamic-versioning
-version = "0.0.0"
+version = "5.0.0"
```
Yajo, line 4:
```suggestion
version = "0.0.0"
```

Read comment just one line above :stuck_out_tongue_winking_eye:

PR #267, opened 2020-09-05
Document the update process

### LINKED ISSUE #239: Docs: draw the logic graph
Even after navigating through the code quite some times, I'm still not sure what happens when generating or updating a project.
I think having this graph would help tremendously new users to understand how Copier works, but developers as well, and help a lot when thinking about design and new features implementation.

For example, it would help resolve these issues: #196, #197, #238 and #240.

Of course we should wait for #203, and then maybe use a mermaid-like Markdown extension to draw the graph in the docs 🙂

pawamoy:
Resolves #239.

I think I managed to get a nice mermaid graph:

Screenshot_2020-09-05_12-59-48

One problem though: prettier does not understand markdown admonitions and dedents/breaks them.
Not sure if we should just drop the admonition, or find a way to make prettier happy. 

EDIT: ref: https://github.com/prettier/prettier/issues/9154

### DIFF
--- a/docs/updating.md
+++ b/docs/updating.md
@@ -42,3 +42,72 @@ repos:
             language: fail
             files: "\\.rej$"
 ```
+
+## How the update works
+
+To understand how the updating process works, take a look at this diagram:
+
+<!-- prettier-ignore-start -->
+<div class="mermaid">
+graph TD
+
+%% nodes ----------------------------------------------------------
+template_repo("template repository")
+template_current("/tmp/template<br>(current tag)")
+template_latest("/tmp/template<br>(latest tag)")
+
+project_regen("/tmp/project<br>(fresh, current version)")
+project_current("current project")
+project_half("half migrated<br>project")
+project_updated("updated project")
+project_applied("updated project<br>(diff applied)")
+project_full("fully updated<br>and migrated project")
+
+update["upate current<br>project in-place<br>(prompting)<br>+ run tasks again"]
+compare["compare to get diff"]
+apply["apply diff"]
+
+diff("diff")
+
+%% edges ----------------------------------------------------------
+        template_repo --> |git clone| template_current
+        template_repo --> |git clone| template_latest
+
+     template_current --> |generate and run tasks| project_regen
+      project_current --> compare
+      project_current --> |apply pre-migrations| project_half
+        project_regen --> compare
+         project_half --> update
+      template_latest --> update
+               update --> project_updated
+              compare --> diff
+                 diff --> apply
+      project_updated --> apply
+                apply --> project_applied
+      project_applied --> |apply post-migrations| project_full
+
+%% style ----------------------------------------------------------
+classDef grey fill:#e8e8e8;
+class compare,update,apply grey;
+</div>
+
+<!-- prettier-ignore-end -->
+
+As you can see here, `copier` does several things:
+
+-   it regenerates a fresh project from the current template version
+-   then it compares both version, to get the diff from "fresh project" to "current
+    project"
+-   now it applies pre-migrations to your project, and updates the current project with
+    the latest template changes (asking confirmation)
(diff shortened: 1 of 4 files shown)


pawamoy:
Hmm there was an update of the mermaid plugin and now it doesn't work as expected. Investigating.

docs/updating.md
```diff
@@ -42,3 +42,70 @@ repos:
...
+
+<!-- prettier-ignore-start -->
+!!! important
+    The diff obtained by comparing the fresh, regenerated project to your
+    current project can cancel the modifications applied by the update from the latest
+    template version. During the process, `copier` will ask you confirmation to overwrite or
+    skip modifications, but in the end, it is possible that nothing has changed (except for
+    the version in `.copier-answers.yml` of course). This is not a bug: although it can be
+    quite surprising, this behavior is correct.
+<!-- prettier-ignore-end -->
```
Yajo, line 111: I have managed to do this before this way:

```suggestion
!!! important

    The diff obtained by comparing the fresh, regenerated project to your
    current project can cancel the modifications applied by the update from the latest
    template version. During the process, `copier` will ask you confirmation to overwrite or
    skip modifications, but in the end, it is possible that nothing has changed (except for
    the version in `.copier-answers.yml` of course). This is not a bug: although it can be
    quite surprising, this behavior is correct.
```

You can find examples of that around the codebase.

Obviously, prettier thinks the inner block is a code block, but mkdocs seems to get it fine.

Yajo (changes requested):
Rebase please.

Also, a rebase should trigger actions, which seem to be forever waiting now.

pawamoy:
It seems ReadTheDocs is breaking mkdocs.yml by adding contents to it. See how `format` becomes `null` instead of `!!python/name:mermaid2.fence_mermaid`:

```yaml
 - pymdownx.superfences:
    custom_fences:
    - class: mermaid
      format: null
```

https://readthedocs.org/api/v2/build/11817460.txt

Yajo:
> Also, a rebase should trigger actions, which seem to be forever waiting now.

It's nice that now I see the builds got triggered, finally!



> And re-opened here: [readthedocs/readthedocs.org#7461](https://github.com/readthedocs/readthedocs.org/issues/7461)

Yikes, so then we cannot merge this until that is fixed, right? Or we'd end up breaking our docs...

pawamoy:
> Yikes, so then we cannot merge this until that is fixed, right? Or we'd end up breaking our docs...

Indeed :cry: 

I don't mind letting this PR sit here until we get some news on that RTD issue :slightly_smiling_face:

Yajo:
OK I set it to draft until then, to avoid further reviews.

pawamoy:
Waiting for https://github.com/readthedocs/readthedocs.org/pull/7507. In the meantime if we really want this now, there's a workaround: we can use `<div class="mermaid">...</div>` instead of a code-block with fences.

Yajo:
Great idea!

pawamoy:
Damn, prettier refuses to ignore the div when I wrap it inside `<!-- prettier-ignore-start -->` comments...

Yajo:
This looks pretty good, but you need to fix the docs build and run pre-commit for it to be merged.

Thanks again!

pawamoy:
Apparently `prettier` expects a blank line before the `<!-- prettier-ignore-end -->` comment for it to work.

Yajo (approved):

Yajo:
So how shall we proceed now that https://github.com/readthedocs/readthedocs.org/pull/7507 is fixed?

pawamoy:
We can now configure superfences for mermaid blocks: https://squidfunk.github.io/mkdocs-material/reference/diagrams/#superfences

And in the Markdown page, use a proper triple-backticks code block instead of a div with mermaid class. The prettier comments can also be removed I think.

However I don't know when RTD will release this just-merged PR!

Issue #240, opened 2020-07-29
Allow finer control over execution of tasks


pawamoy:
I would like to discuss about a way to improve the usability of tasks. In short, I think it would be great to have control over when (and if) a task is executed.

Related to #196 and #238.

Yajo:
Yes, I think that it should work similar to migrations: pre and post. We could even have pre-copy, pre-update, post-copy, post-update. That should be enough, right?

Yajo:
@baurmatt had a good idea at https://github.com/copier-org/copier/issues/238#issuecomment-665732541: add `$OPERATION={copy,update}` to the environment. With that, you can simply include any needed `if` in your task code and perform it or not, without needing to overcomplicate the `copier.yml` file. Also it'd be fully backwards compatible.

IMHO the path tho follow.

pawamoy:
> pre-copy, pre-update, post-copy, post-update

I'd love to have those indeed. Would this be accomplished with both `STAGE` and `OPERATION` env vars then? The four combinations would be mutually exclusive, meaning each one if only executed when the conditions are met.

`STAGE` / `OPERATION` | copy | update
--- | --- | ---
before |  |
after | |

Yajo:
Yes, well... definitely this couldn't be backwards-compatible after all.

pawamoy:
I'm OK with implementing this incrementally if you'd like to make it backwards-compatible first, then add functionality like more stages.

pawamoy:
Hey @Yajo, I'd like to start tackling this. Basically, I want to add an `OPERATION` environment variable when running tasks, as was discussed previously.

Do you already have an idea of how this could be implemented?
I was wondering if it would make sense to set the environment variable with `os.environ["COPIER_OPERATION"] = "update"` to later on pass it in the `extra_env` argument:

```python
run_tasks(
        conf, render, [{"task": t, "extra_env": {"STAGE": "task", "OPERATION": os.environ["COPIER_OPERATION"]}} for t in conf.tasks]
    )
```

...but I'm not sure because it's like using a global variable :sweat_smile: 

What do you think?

Yajo:
What if we apply a logic like the one explained at #326 but for tasks too?

pawamoy:
Isn't it exactly like what is discussed here :slightly_smiling_face: ? Passing everything needed (env vars) to tasks when executing them?

Yajo:
Lol yes...

[...]

PR #287, opened 2020-10-07
Add operation environment variable

pawamoy:
First pass at #240.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -208,7 +208,18 @@ def copy_local(conf: ConfigData) -> None:
         print("")  # padding space
 
     run_tasks(
-        conf, render, [{"task": t, "extra_env": {"STAGE": "task"}} for t in conf.tasks]
+        conf,
+        render,
+        [
+            {
+                "task": t,
+                "extra_env": {
+                    "STAGE": "task",
+                    "OPERATION": os.environ.get("OPERATION", "copy"),
+                },
+            }
+            for t in conf.tasks
+        ],
     )
     if not conf.quiet:
         print("")  # padding space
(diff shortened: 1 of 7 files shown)


copier/main.py
```diff
@@ -290,6 +301,9 @@ def update_diff(conf: ConfigData) -> None:
             )
         }
     )
+    # At this point we'll run tasks in the current project:
+    # copy_local() needs to pass "update" as operation
+    os.environ["OPERATION"] = "update"
```
Yajo, line 306: This is too much, you're altering the environment variables for all the process.

You better do:

```python
with local.env(OPERATION='update'):
   ...
```

Docs: https://plumbum.readthedocs.io/en/latest/local_machine.html#environment

Yajo (commented):
It looks like a good first approach.

However I was hoping for a deeper refactoring of how tasks work. My idea was to run tasks always: pre-copy, post-copy, pre-update, post-update. The updatediff case is a special one that I'd like to review deeper, because tasks need to run there in the temporary subproject generated to obtain the diff.

Also I'm wondering if having a complex `_tasks` configuration is needed, given any script can just operate based on env variables.

What are your thoughts?

copier/main.py
pawamoy: We could also consider passing an `operation` argument to `copy_local`, defaulting to `copy`, as it's the only place we use `os.environ.get("OPERATION")`.

pawamoy:
I've open #288 after thinking more about this.

Yajo:
Thanks, I'll keep it in mind when fixing #240.

PR #293, opened 2020-10-17
Explain how to use multiple templates for the same project

Yajo:
This is very cool and it doesn't seem to be clearly explained anywhere.


pawamoy:
Well indeed that is extremely interesting! I often wondered about partial templates and it seems to be the solution I was looking for!

Yajo:
Yeah I think it makes Copier stand out in the crowd but it wasn't clearly documented until now... I'm happy you like it :blush:

PR #295, opened 2020-10-19
Remove YAML casting for bool types

### LINKED ISSUE #280: Disallow str-parseable input by API
Copier allows API input such as `"true"` for a bool question, and converts that to `True`. And the same for all data types.

The reasoning behind this is that CLI always sends strings.

But this overcomplicates the implementation and makes CLI arguments more surprising, as they can get casted to different answers depending on the question type. 

Gotta remove that nonsense. Instead, make the CLI app transform all data before sending it to the API. And use always the same transform method: YAML.

Yajo:
~~If calling Copier through CLI, it's normal to convert all strings to expected data, but if calling through API, you should know what kind of data you're passing in, so it shouldn't try to be smarter than you.~~

This was going to fix #280, but after all that's not a real issue.

I'm just going to merge the improvements done while attempting to fix that, without changing Copier behavior.

### DIFF
--- a/copier/config/user_data.py
+++ b/copier/config/user_data.py
@@ -485,9 +485,6 @@ def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
     # Skip casting None into "None"
     if type_fn is str and answer is None:
         return answer
-    # Parse correctly bools as 1, true, yes...
-    if type_fn is bool and isinstance(answer, str):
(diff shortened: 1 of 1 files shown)


Yajo:
Finally the only non-breaking meaningful change has been the removal of the specific YAML casting for bool types. Not as big as expected, but good change anyway.

PR #300, opened 2020-10-31
fixed typo

nschlemm:
"-f" makes more sense and matches https://github.com/copier-org/copier/blob/master/copier/cli.py#L138 ;)


Yajo:
Thank you!

PR #314, opened 2020-12-03
Refactor

### LINKED ISSUE #110: More class, less dangling methods

Yajo:
Fix #110.

Important changes (the list is a self-reminder WIP, to not forget updating docs and release notes):

- OOP. Haters gonna hate, but makes all easier to reason about.
- `skip` removed. Now there's only `skip_if_exists`.
- `skip_if_exists` extends the template's skip, instead of replacing it.
- Many tests are using module fixtures instead of random demo dirs.


pawamoy:
...aaaaaaaaand it's merged. Congrats!

Yajo:
Me too! 🤣🤣🤣

It still needs lots of polish, but other stuff con be done on smaller steps.

I won't be able to publish an alpha until one friend releases new versions of some dependencies (ahem.. 😛).

Issue #315, opened 2020-12-15
Confusing subdirectory flag behavior.


PhilReinhold:
I'm trying to render templates from a git repository where the template is stored in a subdirectory.

I thought I might be able to use the subdirectory flag to do this, but I'm not getting any rendered output. The program exits without any output.

I would ideally like to have the copier.yml in the subdirectory, but I tried moving it to the repo root, and at this point the config questions occur, but so long as the subdirectory flag is present, no files are created.

Finally, I notice that if I omitted the subdirectory flag, and had the copier file in a subdirectory, files were created, but all of the templated values were rendered as empty strings. If config values are missing, I would expect an error rather than rendering as empty string.

pawamoy:
Hello! It seems we simply had different use-cases in mind for a "subdirectory" feature. While you want to use the subdirectory of a repository as the whole template directory (containing copier.yml), I wanted to use the subdirectory of a repository to locate the files to generate only (copier.yml still being at the root of the repo). In fact, I wanted to mimic cookiecutter's behavior: the repo is the template, but the rendered files/dirs are inside a subdirectory.

Yajo:
Hi there. Thanks for your report and PR 😊 

However, before merging this, I'm wondering if this is a feature we really want to support. I understand that you were expecting a different thing from the subdirectory flag, but that can be fixed better by just improving the docs.

So my question is what would be the benefit of having several templates in a single repo, vs. having one template per repo? Why do we actually need this?

I can imagine, though, some arguments in the opposite direction: why would it be better to have a single template per repo?

1. **Tags.** We support updating from a repo, but updates are based on tags. If I start a subproject based on your repo's template1, and then your repo evolves template2 and makes a new release, I'd have to update my subproject even if there haven't been any changes in template1.
2. **Simplicity.** 1 repo - 1 template. Simple. I like simple things... 🤷🏼‍♂️
3. **Git submodules.** If you _really_ need to have one metarepo containing your templates, you could just have a git repo with submodules, each one pointing to another template. Then, you have all your code together, while other templates still have 1 repo each.

balrok:
Not the author, but maybe it helps the case if you get multiple opinions. Because when I started with copier, I also missed that feature.

* At my workplace its very **complicated** to get a new repository (don't ask why ;-)). When I just want to quickly sketch out a copier-template I would need to register another repo, which usually stops me from doing so. (right now I only implemented one template exactly because of that aspect)
[...]

PR #316, opened 2020-12-16
Instantiate template from subdirectory of repo

PhilReinhold:
Pursuant to problems described in issue #315. The use case is that we have a template in a subdirectory of a git repo. While I originally believed this is what the subdirectory flag is for, it became apparent reading through the tests that this option was conceived for other uses.

This pull request adds a command line option "--repo-subdir", which is added to the src_path as soon as the repository is checked out, allowing the copier.yml to be pulled from the subdirectory.

In addition, this subdirectory is stored in the copier-answers (as "_repo_subdir") allowing one to `copier update` without re-specifying the command line option.

I'll note that I'm on windows, where many of the tests are marked xfail, so I can't check that these haven't broken.

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -131,6 +131,17 @@ class CopierApp(cli.Application):
         ),
     )
 
+    repo_subdir: cli.SwitchAttr = cli.SwitchAttr(
+        ["-c", "--repo-subdir"],
+        str,
+        help=(
+            "If using a VCS repository as a source, indicates that the template"
+            " (i.e. the directory containing copier.yml) is found in the given"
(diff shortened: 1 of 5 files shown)


pawamoy:
ReadTheDocs failure comes from `poetry_dynamic_versioning`. It seems it tries to import a Poetry command, that I guess is not present in the `poetry_core` package?

Yajo:
After last coments in #315, @PhilReinhold  would you like to redo this PR to improve the docs instead? Docs written by newcomers are very valuable!

Thanks again.

Yajo:
Closing as stale.

PR #357, opened 2021-02-14
Migrate from poetry to pdm

Yajo:
This is an attempt to start using pdm as package management tool.

https://frostming.com/2021/01-22/introducing-pdm/ explains why it is conceptually better, as it doesn't require the existence of a virtualenv.

Some extra nice things:

- Bundles task support, no need for poethepoet.
- Bundles dynamic version support.
- Bundles dynamic python version classifier support (we were forgetting py3.9!).

Blockers:

- [x] https://github.com/frostming/pdm/issues/264
- [x] https://github.com/frostming/pdm/issues/270 - Applied workaround.
- [x] https://github.com/frostming/pdm/issues/272
- [ ] Fix CI workflow (no sense doing this until previous blocker is gone)


pawamoy:
I see you're a man of culture! I'm currently trying PDM out in my projects as well :smile:  Very promising :slightly_smiling_face:

Yajo:
Well, right now it can't solve dependencies properly... 🤷🏼‍♂️ 

But I like its idea.

frostming:
Thank you guys for adopting pdm in your project. I will cut a release after the mentioned bugs are fixed.

Another note for using PDM in CI: PDM can only be installed in Python 3.7+ while it can still manage environments under Python 3.6 and lower. Refer to [this workflow](https://github.com/frostming/marko/blob/master/.github/workflows/ci.yml) to see how to make it work.

Yajo:
Thanks @frostming for your comments (and for pdm).

I'll think about what to do with copier on py 3.6 once all other versions builds are green.

I managed to install it locally (py 3.9) but it seems that on CI all py3.7+ builds fail with [the same error](https://github.com/copier-org/copier/pull/357/checks?check_run_id=1993312371#step:10:504):

    pip._internal.exceptions.DistributionNotFound: No matching distribution found for dataclasses==0.8; python_version < "3.7"

It doesn't make much sense to me, as these versions are not supposed to install that dependency. Is it a bug on my dependencies definition or on pdm? 🤔 Thank you.

frostming:
@Yajo Please retry, also please regenerate the lock file to be consistent with pyproject.toml so that CI don't need to generate in each run

frostming:
FWIW, I wrote a document on how to use PDM in CI, especially for testing Python 3.6 projects:

https://pdm.fming.dev/advanced/#use-pdm-in-continuous-integration

pawamoy:
@Yajo PDM was improved a lot since then, maybe we could try it again for Copier :slightly_smiling_face:

Yajo:
Hmm I'm not sure. These days I'm more convinced about using something based on nix. Besides dependabot has support for poetry but not for pdm AFAIK. Any thoughts about those points?

pawamoy:
Using nix would be a drastic (but very interesting) change. It would need a well-written contribution guide. Why not. But: is nix supported by dependabot? By ReadTheDocs? Can it build sdists and wheels? Publish them on pypi.org?

About PDM support in dependabot: https://github.com/dependabot/dependabot-core/issues/3190. So maybe stick with Poetry until dependabot supports standards way of declaring dependencies.

Yajo:
Your points are valid and https://www.tweag.io/blog/2020-08-12-poetry2nix/ convinced me. 

That's however another piece of the puzzle missing for pdm. And more important: proper vscode/gitpod integration. 

I love pdm but I guess we'll have to stick with poetry for now. 🤷‍♂️

PR #364, opened 2021-02-28
ci: Stop using Poetry debug verbosity

pawamoy:
I think it comes from this: https://github.com/python-poetry/poetry/issues/3663


pawamoy:
Ah, I forgot that you track poetry.lock. I need to update it.

Yajo (approved):

PR #365, opened 2021-03-02
implement git branch

hparfr:
Hi,

This PR allows to specify which branch the template should use.

My use case : the template is a git repo with different branches (like "apache" and "nginx")
When a project is started with branch = nginx, I want to
 1) specify it in `copier -b nginx git://path/to/repo a_dir`
 2) keep it after `copier update`

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -54,6 +54,7 @@ class CopierApp(cli.Application):
         answers_file: Set [answers_file][] option.
         exclude: Set [exclude][] option.
         vcs_ref: Set [vcs_ref][] option.
+        vcs_branch: Set [vcs_branch][] option.
         pretend: Set [pretend][] option.
         force: Set [force][] option.
         skip: Set [skip_if_exists][] option.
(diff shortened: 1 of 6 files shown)


hparfr:
I couldn't run the tests on my machine because of the sames issues as in the CI.

Yajo:
I'm sad because you already did the work, but this is not going to be accepted because it goes against Copier's workflow, by design. 

To be able to update a project properly, supporting migrations and all the bells and whistles, we need to know the template version. It comes from git tags. Since git tags are something that git stores per repo, and not per branch, we simply cannot support a single repo with different templates in different branches. If we were to do so, we'd lose the ability to update them, and that's a key feature of Copier.

There are other products that also lose features in those scenarios. One example is Github itself, where if you write `Fix #1` in a commit message, it will only close the issue if committed to the main branch.

You have alternatives available:

1. Use 1 repo per template. You can share common code with git submodules.
2. Use subdirectory (which is a bit misleading in Copier 5, but its nonsense has been removed for Copier 6). One question can be used to let the user choose the subdirectory.

I hope you don't take it badly, but I'll close it because it's technically not feasible. 😅  Thanks!

hparfr:
>I'm sad because you already did the work, but this is not going to be accepted because it goes against Copier's workflow, by design.

My bad, It was not wise to not start with an issue :)

I dislike the tag only approach because it has his limitations too. It supposes you follow a strict discipline and won't prevent you from mistakes.

While I think both way of working can be in the same tool, it will require more code to ensure no conflicts between these features, I understand you point of view.

> I hope you don't take it badly,

No problem @Yajo  ;-)

Issue #208, opened 2020-05-15
Add a slugify filter


pawamoy:
I cannot think of a template I wrote that didn't need such a filter 🙂 

I usually write a very naive one myself:

```yaml
repository_name:
  type: str
  help: Your repository name
  default: "[[ project_name|lower|replace('_', '-')|replace(' ', '-') ]]"
```

But it will break in so many cases.

As an alternative, if we don't want to support more and more filters in `copier`, maybe we could think of allowing custom filters to be used. Take a look at https://github.com/mattrobenolt/jinja2-cli/pull/86, it allows this use of `jinja2-cli`:

```
jinja2 -f ansible.plugins.filter.ipaddr.ipaddr test.txt
```

In `copier`'s case, we would not be able to load the filter using its Python dotted-path, so we would have to think of another way to add it to the Jinja environment.

Yajo:
Makes sense.

I also wonder if we could reuse Ansible's [filters](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html) and [tests](https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html), which are amazing.

pawamoy:
I recently saw that `pyyaml` allows to load Python objects directly: https://pyyaml.org/wiki/PyYAMLDocumentation#yaml-tags-and-python-types

This could be useful. The user could pass `!!python/module:my_filters` and we could add the all the functions defined in this module as filters in the jinja env, or something similar :slightly_smiling_face:

Yajo:
cool!

baurmatt:
As described in #277: I'd like to be able to implement custom filters for e.g. generating a secret with hmac:

```
import binascii, hashlib, hmaci
hmac.new(binascii.unhexlify(key), msg.encode(), hashlib.sha256).hexdigest()
```

Yajo:
Until today, although copier is made with python, it remained language-agnostic.

Adding custom filters or tests imported through Python seems a very good option, but I'm wondering if an exec function that runs a system executable would be better... so someone that only knows js can still drop in an executable js file and get its stdout into the template...

Opinions?

Yajo:
BTW I opened https://github.com/copier-org/copier/projects/1. I hope you get some hype 😂😂

If you want to collaborate on any issue... thanks!

baurmatt:
[...]

PR #370, opened 2021-03-12
feat: Add extensions option allowing to load additional Jinja2 extensions

pawamoy:
Issue #208

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -392,7 +392,7 @@ class Worker:
             raise ExtensionNotFoundError(
                 f"Copier could not load some Jinja extensions:\n{error}\n"
                 "Make sure to install these extensions alongside Copier itself.\n"
-                "See the docs at https://copier.readthedocs.io/en/latest/configuring/extensions"
+                "See the docs at https://copier.readthedocs.io/en/latest/configuring/jinja_extensions"
             )
         return env
 
--- a/copier/main.py
+++ b/copier/main.py
@@ -559,7 +559,6 @@ class Worker:
 
         See [generating a project][generating-a-project].
         """
-        os.environ["COPIER_TEMPLATE_PATH"] = str(self.template.local_abspath)
         was_existing = self.subproject.local_abspath.exists()
         if not self.quiet:
             # TODO Unify printing tools
--- a/copier/main.py
+++ b/copier/main.py
@@ -377,13 +377,9 @@ class Worker:
         paths = [str(self.template.local_abspath)]
         loader = FileSystemLoader(paths)
         default_extensions = [
-            "cookiecutter.extensions.JsonifyExtension",
-            "cookiecutter.extensions.RandomStringExtension",
-            "cookiecutter.extensions.SlugifyExtension",
-            "jinja2_time.TimeExtension",
             "jinja2_ansible_filters.AnsibleCoreFiltersExtension",
         ]
-        extensions = default_extensions + list(self.template.extensions)
+        extensions = default_extensions + list(self.template.jinja_extensions)
         # We want to minimize the risk of hidden malware in the templates
         # so we use the SandboxedEnvironment instead of the regular one.
         # Of course we still have the post-copy tasks to worry about, but at least
--- a/copier/main.py
+++ b/copier/main.py
@@ -28,7 +28,7 @@ from questionary import unsafe_prompt
 from .errors import UserMessageError
 from .subproject import Subproject
 from .template import Template
-from .tools import Style, TemporaryDirectory, printf, to_nice_yaml
+from .tools import Style, TemporaryDirectory, printf
 from .types import (
     AnyByStrDict,
     JSONSerializable,
@@ -376,13 +376,21 @@ class Worker:
         """
         paths = [str(self.template.local_abspath)]
         loader = FileSystemLoader(paths)
+        default_extensions = [
+            "cookiecutter.extensions.JsonifyExtension",
+            "cookiecutter.extensions.RandomStringExtension",
+            "cookiecutter.extensions.SlugifyExtension",
+            "jinja2_time.TimeExtension",
+            "jinja2_ansible_filters.AnsibleCoreFiltersExtension",
+        ]
+        extensions = default_extensions + list(self.template.extensions)
         # We want to minimize the risk of hidden malware in the templates
         # so we use the SandboxedEnvironment instead of the regular one.
         # Of course we still have the post-copy tasks to worry about, but at least
         # they are more visible to the final user.
-        env = SandboxedEnvironment(loader=loader, **self.template.envops)
-        default_filters = {"to_nice_yaml": to_nice_yaml}
-        env.filters.update(default_filters)
+        env = SandboxedEnvironment(
+            loader=loader, extensions=extensions, **self.template.envops
+        )
         return env
 
     @cached_property
--- a/copier/main.py
+++ b/copier/main.py
@@ -1,6 +1,5 @@
 """Main functions and classes, used to generate or update projects."""
 
-import os
 import json
 import platform
 import subprocess
@@ -26,7 +25,7 @@ from pydantic.dataclasses import dataclass
 from pydantic.json import pydantic_encoder
 from questionary import unsafe_prompt
 
-from .errors import UserMessageError
+from .errors import ExtensionNotFoundError, UserMessageError
 from .subproject import Subproject
 from .template import Template
 from .tools import Style, TemporaryDirectory, printf
@@ -390,7 +389,7 @@ class Worker:
                 loader=loader, extensions=extensions, **self.template.envops
             )
         except ModuleNotFoundError as error:
-            raise UserMessageError(
+            raise ExtensionNotFoundError(
                 f"Copier could not load some Jinja extensions:\n{error}\n"
                 "Make sure to install these extensions alongside Copier itself.\n"
                 "See the docs at https://copier.readthedocs.io/en/latest/configuring/extensions"
--- a/copier/main.py
+++ b/copier/main.py
@@ -392,7 +392,7 @@ class Worker:
             raise ExtensionNotFoundError(
                 f"Copier could not load some Jinja extensions:\n{error}\n"
                 "Make sure to install these extensions alongside Copier itself.\n"
-                "See the docs at https://copier.readthedocs.io/en/latest/configuring/jinja_extensions"
+                "See the docs at https://copier.readthedocs.io/en/latest/configuring/#jinja_extensions"
             )
         return env
 
--- a/copier/main.py
+++ b/copier/main.py
@@ -1,5 +1,6 @@
 """Main functions and classes, used to generate or update projects."""
 
+import os
 import json
 import platform
 import subprocess
@@ -384,9 +385,16 @@ class Worker:
         # so we use the SandboxedEnvironment instead of the regular one.
         # Of course we still have the post-copy tasks to worry about, but at least
         # they are more visible to the final user.
-        env = SandboxedEnvironment(
-            loader=loader, extensions=extensions, **self.template.envops
-        )
+        try:
+            env = SandboxedEnvironment(
+                loader=loader, extensions=extensions, **self.template.envops
+            )
+        except ModuleNotFoundError as error:
+            raise UserMessageError(
+                f"Copier could not load some Jinja extensions:\n{error}\n"
+                "Make sure to install these extensions alongside Copier itself.\n"
+                "See the docs at https://copier.readthedocs.io/en/latest/configuring/extensions"
+            )
         return env
 
     @cached_property
@@ -551,6 +559,7 @@ class Worker:
 
         See [generating a project][generating-a-project].
         """
+        os.environ["COPIER_TEMPLATE_PATH"] = str(self.template.local_abspath)
         was_existing = self.subproject.local_abspath.exists()
         if not self.quiet:
             # TODO Unify printing tools
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -14,7 +14,6 @@ from typing import Any, Callable, Optional, TextIO, Union
 import colorama
 from packaging.version import Version
 from pydantic import StrictBool
-from yaml import safe_dump
 
 from .types import ExcInfo, IntSeq
 
@@ -117,17 +116,6 @@ def cast_str_to_bool(value: Any) -> bool:
 
 def copy_file(src_path: Path, dst_path: Path, follow_symlinks: bool = True) -> None:
     shutil.copy2(src_path, dst_path, follow_symlinks=follow_symlinks)
-
-
-def to_nice_yaml(data: Any, **kwargs) -> str:
-    """Dump a string to pretty YAML."""
-    # Remove security-problematic kwargs
-    kwargs.pop("stream", None)
-    kwargs.pop("Dumper", None)
-    result = safe_dump(data, **kwargs)
-    if isinstance(result, str):
-        result = result.rstrip()
-    return result or ""
 
 
 def force_str_end(original_str: str, end: str = "\n") -> str:
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -421,7 +421,7 @@ _jinja_extensions:
     your users must install the `jinja2-time` Python package.
 
     ```bash
-    # with pip
+    # with pip, in the same virtualenv where Copier is installed
     pip install jinja2-time
 
     # if Copier was installed with pipx
@@ -435,7 +435,7 @@ _jinja_extensions:
     - [`jinja2_ansible_filters.AnsibleCoreFiltersExtension`](https://gitlab.com/dreamer-labs/libraries/jinja2-ansible-filters/):
       this extension adds most of the [Ansible filters](https://docs.ansible.com/ansible/2.3/playbooks_filters.html) to the environment.
 
-    You don't need to tell your templates users to install these extensions:
+    You don't need to tell your template users to install these extensions:
     Copier depends on them, so they are always installed when Copier is installed.
 
 ### `migrations`
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -365,21 +365,6 @@ copier --exclude '*' --exclude '!file-i-want' copy template destination
 
     Instead, CLI/API definitions **will extend** those from `copier.yml`.
 
-### `jinja_extensions`
-
--   Format: `List[str]`
--   CLI flags: N/A
--   Default value: []
-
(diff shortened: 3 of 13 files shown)


copier/cli.py
```diff
@@ -59,6 +59,7 @@ class CopierApp(cli.Application):
         skip: Set [skip_if_exists][] option.
         prereleases: Set [use_prereleases][] option.
         quiet: Set [quiet][] option.
+        extensions: Set [extensions][] option.
```
pawamoy, line 62: Ah it's not needed here in fact. There will be no CLI option for this.

copier/main.py
```diff
@@ -376,13 +382,21 @@ def jinja_env(self) -> SandboxedEnvironment:
...
         paths = [str(self.template.local_abspath)]
         loader = FileSystemLoader(paths)
+        default_extensions = [
+            "cookiecutter.extensions.JsonifyExtension",
+            "cookiecutter.extensions.RandomStringExtension",
+            "cookiecutter.extensions.SlugifyExtension",
+            "jinja2_time.TimeExtension",
+            "jinja2_ansible_filters.AnsibleCoreFiltersExtension",
+        ]
+        extensions = default_extensions + list(self.extensions)
```
pawamoy, line 392: I chose to set default extensions here, so they are *always* loaded, whatever the template writer puts in `_extensions`. But maybe you think we should let writers override these?

copier/tools.py
```diff
@@ -119,17 +118,6 @@ def copy_file(src_path: Path, dst_path: Path, follow_symlinks: bool = True) -> N
...
 
 
-def to_nice_yaml(data: Any, **kwargs) -> str:
-    """Dump a string to pretty YAML."""
-    # Remove security-problematic kwargs
-    kwargs.pop("stream", None)
-    kwargs.pop("Dumper", None)
-    result = safe_dump(data, **kwargs)
-    if isinstance(result, str):
-        result = result.rstrip()
```
pawamoy, line 129: Contrary to this function, the `to_nice_yaml` from Ansible filters does not seem to `rstrip()` the result. This is the reason the tests fail, but for the life of me I was not able to fix the test! I tried adding or removing newlines with multiple methods, but pytest would not have it. I'll continue trying of course, but if you want to to take a look I don't say no :sweat_smile:

Yajo: It's weird, but it doesn't seem like a big loss. You can modify the tests if you want.

pawamoy: Pffff, I was tired I guess, it's just that the same assert is repeated three times and I didn't get that until now... Tests now passing. I will add some more.

pawamoy:
I've added tests, but I'm not sure I took the right approach. With this implementation:

1. default extensions (ansible, cookiecutter, etc.) are always loaded
2. extensions in copier.yml are also loaded (written as import strings)
3. extensions given through the `extensions` argument of `copier.copy` are also loaded

It means that a template *writer* can add extensions, or override the default ones, while a template *user* (given they run Copier programmatically, in Python) can add extensions (which seems useless?), or override the default or template ones.

I can see the use-case of point 3 (overriding a filter for example), but I'm not sure it makes sense. What are your thoughts?

pawamoy:
Arf I messed up the rebase. Will fix it another day.

Yajo:
> I can see the use-case of point 3 (overriding a filter for example), but I'm not sure it makes sense. What are your thoughts?

I think it's better to _not_ support this use case. It would produce confusion for template users and unexpected results for template designers.

This feature should exist *only* in the template (copier.yaml).

Thanks for your effort! I'll review deeper when you end the draft phase 😊

Yajo:
> Arf I messed up the rebase. Will fix it another day.

I see some conflicts with current master. You might prefer to do a merge instead, which is usually easier. I can squash at merge later, so don't worry too much about git history right now (if that makes your life easier 😉 )

pawamoy:
>  think it's better to not support this use case. It would produce confusion for template users and unexpected results for template designers.

Done! I removed the `extensions` parameter from `copier.copy`, templates users won't be able to mess with extensions :slightly_smiling_face: 

Side note: I have an unrelated test failing locally that I'm not sure how to fix? `test_jinja_defaults_v6_min_version`

```
copier.errors.UnknownCopierVersionWarning: Cannot check Copier version constraint.
```

Yajo:
> Side note: I have an unrelated test failing locally that I'm not sure how to fix: `test_jinja_defaults_v6_min_version`

Maybe you need to pull with git tags.

pawamoy:
This is ready for review. Maybe we could add more documentation? And a TODO for the `make_secret` as mentioned previously? The changelog also lags behind at v4, not sure if I should take care of it or if it's autogenerated. Waiting for your review (no rush)!

Yajo:
> The changelog also lags behind at v4, not sure if I should take care of it or if it's autogenerated

huh... that's a problem I have to fix at some point. I'll open a separate issue, I need some strategy on that.

I'll take a deep review when I have some time, thank you!

copier/main.py
```diff
@@ -376,13 +376,21 @@ def jinja_env(self) -> SandboxedEnvironment:
         """
         paths = [str(self.template.local_abspath)]
         loader = FileSystemLoader(paths)
+        default_extensions = [
+            "cookiecutter.extensions.JsonifyExtension",
+            "cookiecutter.extensions.RandomStringExtension",
```
Yajo, line 381: These 2 extensions are not needed, because `AnsibleCoreFiltersExtension` includes:

- `ans_random`
- `to_json`
- `to_nice_json`

copier/main.py
```diff
@@ -376,13 +376,21 @@ def jinja_env(self) -> SandboxedEnvironment:
...
+            "cookiecutter.extensions.SlugifyExtension",
```
Yajo, line 382: The point above makes me think about this one extension...

Cookiecutter is a big dependency. Since you're adding a way to easily include external extensions... Would it be acceptable to you to remove the whole cookiecutter thing?

After all, if you just `pipx inject copier cookiecutter` and add this extension to your template (or maybe [jinja2-slug](https://pypi.org/project/jinja2-slug/) if that's all you want), it should be enough for most cases.

copier/main.py
```diff
@@ -376,13 +376,21 @@ def jinja_env(self) -> SandboxedEnvironment:
...
+            "jinja2_time.TimeExtension",
```
Yajo, line 383: This is a convenient extension too, but again `AnsibleCoreFiltersExtension` would do the trick with `{{ '%Y-%m-%d %H:%M:%S' | strftime }}`.

So, in this case also, please remove the whole dependency.

I think we can live with just ansible-core-filters shipped by default.

---

😢 Sorry! I know it was me who asked you to include those dependencies... but after seeing the code in place, seeing how ansible-core-filters does the job, and also seeing how easy it would be to add extra extensions, I changed my mind. I'm very sorry! 🙏🏼

docs/configuring.md
```diff
@@ -365,6 +365,25 @@ copier --exclude '*' --exclude '!file-i-want' copy template destination
...
+filters, global variables and functions, or tags to the environment.
+
+!!! info
+
+    The following extensions are *always* loaded:
+
+    - [`cookiecutter.extensions.JsonifyExtension`](https://cookiecutter.readthedocs.io/en/latest/advanced/template_extensions.html#jsonify-extension)
+    - [`cookiecutter.extensions.RandomStringExtension`](https://cookiecutter.readthedocs.io/en/latest/advanced/template_extensions.html#random-string-extension)
+    - [`cookiecutter.extensions.SlugifyExtension`](https://cookiecutter.readthedocs.io/en/latest/advanced/template_extensions.html#slugify-extension)
+    - [`jinja2_time.TimeExtension`](https://pypi.org/project/jinja2-time/)
```
Yajo, line 384: So, this should disappear.

pyproject.toml
```diff
@@ -28,9 +28,11 @@ copier = "copier.cli:CopierApp.run"
 python = ">=3.6.1,<3.10"
 "backports.cached-property" = { version = "^1.0.0", python = "<3.8" }
 colorama = "^0.4.3"
+cookiecutter = "^1.7.2"
```
Yajo, line 31: This too should disappear.

tests/demo_extensions_additional/copier.yml
```diff
@@ -0,0 +1,3 @@
+_extensions:
```
Yajo, line 1: These days I'm more in favor of building the code tree in the test module. It seems cleaner and easier to reference the test with the template. I've done that for you.

docs/configuring.md
```diff
@@ -365,6 +365,25 @@ copier --exclude '*' --exclude '!file-i-want' copy template destination
 
     Instead, CLI/API definitions **will extend** those from `copier.yml`.
 
+### `extensions`
```
Yajo, line 368: Do you think it makes sense to call this `jinja_extensions`? 🤔 

That maybe makes it more self-explanatory, and separates it from maybe copier extensions or other kind of extensions that might come in the future? Disclaimer: I'm not particularly thinking on doing that right now.

tests/test_updatediff.py
```diff
@@ -132,11 +132,11 @@ def test_updatediff(tmp_path_factory):
...
-            # Changes here will be overwritten by Copier
-            _commit: v0.0.1
-            _src_path: {bundle}
-            author_name: Guybrush
-            project_name: to become a pirate
+        # Changes here will be overwritten by Copier
+        _commit: v0.0.1
+        _src_path: {bundle}
+        author_name: Guybrush
+        project_name: to become a pirate\n
```
Yajo, line 139: I think all you really needed was to add the last `\n`, right? If so, please do that (to reduce the diff).

tests/test_updatediff.py
```diff
@@ -175,11 +175,11 @@ def test_updatediff(tmp_path_factory):
...
-                # Changes here will be overwritten by Copier
-                _commit: v0.0.2
-                _src_path: {bundle}
-                author_name: Guybrush
-                project_name: to become a pirate
+            # Changes here will be overwritten by Copier
+            _commit: v0.0.2
+            _src_path: {bundle}
+            author_name: Guybrush
+            project_name: to become a pirate\n
```
Yajo, line 182: Same about the diff.

tests/test_updatediff.py
```diff
@@ -204,11 +204,11 @@ def test_updatediff(tmp_path_factory):
...
-                # Changes here will be overwritten by Copier
-                _commit: {last_commit}
-                _src_path: {bundle}
-                author_name: Guybrush
-                project_name: to become a pirate
+            # Changes here will be overwritten by Copier
+            _commit: {last_commit}
+            _src_path: {bundle}
+            author_name: Guybrush
+            project_name: to become a pirate\n
```
Yajo, line 211: And here too.

Yajo (changes requested):
I like very much how you developed this. Thank you! You'll have to revert some things though, see below (please don't hate me).

Also, some points that I couldn't comment inline:

- `now` and `make_secret` need to bubble a `FutureWarning` that indicates how these will disappear in a future major Copier release, and how to upgrade each of them. (`now` is commented below. `make_secret` can be replaced with `999999999999999999999999999999999|ans_random|hash("sha512")`)
- You need to document how to actually install more extensions. You can include some common ones in a small example catalog (in docs).
- The exception raised when an extension isn't found is not much helpful for template end users. Maybe that could be intercepted when creating the environment to display a more helpful error?

tests/test_updatediff.py
pawamoy: Sure, I deemed it was not that much noise to bring consistency on multiline-strings indentation. Feel free to do that in another commit/PR.

docs/configuring.md
pawamoy: We already have `envops` and not `jinja_envops`, but yeah, `jinja_extensions` is fine, it's more explicit :slightly_smiling_face:

tests/demo_extensions_additional/copier.yml
pawamoy: > I've done that for you.

Not sure to understand! I don't any new commit from you on this branch?

copier/main.py
pawamoy: You're totally right, I was a bit troubled with adding such a dependency anyway!

copier/main.py
pawamoy: Hahaha it's totally fine don't worry! :joy:

pawamoy:
> I like very much how you developed this.

Well, your great refactor made this really easy!

I'll now work on the three additional points you mentioned :slightly_smiling_face:

pawamoy:
It seems Ansible provides a `password_hash` filter that is not available in the extension: https://docs.ansible.com/ansible/2.3/playbooks_filters.html#hashing-filters. Too bad. I worry a bit that we won't be able to send PRs to this project as I can't find its sources on the web :confused:

Yajo:
> I worry a bit that we won't be able to send PRs to this project as I can't find its sources on the web confused

They're not linked in pypi.org for some reason, but they are at https://gitlab.com/dreamer-labs/libraries/jinja2-ansible-filters/ as it seems.

tests/demo_extensions_additional/copier.yml
Yajo: Ah! I forgot to upload it.

Yajo (commented):
Just one little extra detail.

docs/configuring.md
```diff
@@ -365,6 +365,21 @@ copier --exclude '*' --exclude '!file-i-want' copy template destination
...
+-   Default value: []
+
+Additional Jinja2 extensions to load in the Jinja2 environment. Extensions can add
+filters, global variables and functions, or tags to the environment.
+
+!!! info
+
+    The following extensions are *always* loaded:
+
+    - [`jinja2_ansible_filters.AnsibleCoreFiltersExtension`](https://pypi.org/project/jinja2-ansible-filters/)
```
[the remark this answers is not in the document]
pawamoy: Huh, I seem to have starred this repo two weeks ago :O what am I doing X)
Thanks :)

pawamoy:
Alright, just a bit of docs now, and we should be good.

pawamoy:
Exception OK:

> Copier could not load some Jinja extensions:
No module named 'extension_not_installed'
Make sure to install these extensions alongside Copier itself.
See the docs at https://copier.readthedocs.io/en/latest/configuring/extensions

copier/main.py
```diff
@@ -376,13 +377,24 @@ def jinja_env(self) -> SandboxedEnvironment:
...
         # they are more visible to the final user.
-        env = SandboxedEnvironment(loader=loader, **self.template.envops)
-        default_filters = {"to_nice_yaml": to_nice_yaml}
-        env.filters.update(default_filters)
+        try:
+            env = SandboxedEnvironment(
+                loader=loader, extensions=extensions, **self.template.envops
+            )
+        except ModuleNotFoundError as error:
+            raise UserMessageError(
```
Yajo, line 393: Please use a specific exception class. See the errors module.

copier/main.py
```diff
@@ -547,6 +559,7 @@ def run_copy(self) -> None:
 
         See [generating a project][generating-a-project].
         """
+        os.environ["COPIER_TEMPLATE_PATH"] = str(self.template.local_abspath)
```
Yajo, line 562: And this? Were you unable to obtain it from elsewhere? Maybe the file system loader, or `_copier_conf.src_path`?

pawamoy: Oops, forgot to remove that! Yes I was able to get that from the loader!

copier/main.py
pawamoy: Do you mean something like this? 

```python
class ExtensionNotFoundError(UserMessageError, CopierError):
    """Extensions listed in the configuration could not be loaded."""
```

By the way, why do you inherit from both `UserMessageError` *and* `CopierError` when the first already inherits from the second?

Yajo: > Do you mean something like this?

Yes



> By the way, why do you inherit from both `UserMessageError` _and_ `CopierError` when the first already inherits from the second?

Oops, that's a bug.

pawamoy: I wrote like that anyway, for the sake of consistency. You'll be able to fix all error classes at once in a later commit/PR :slightly_smiling_face:

pawamoy:
That's a lot of fixups, sorry :sweat_smile: 

I've fixed the last few bits like comments and docs. It's good on my end.

docs/configuring.md
```diff
@@ -380,6 +380,64 @@ Also don't ask questions to the user; just use default values
...
+Examples of extensions you can use:
+
+-   [`jinja_markdown.MarkdownExtension`](https://github.com/jpsca/jinja-markdown):
+    provides a `markdown` tag that will render Markdown to HTML using
+    [PyMdown extensions](https://facelessuser.github.io/pymdown-extensions/).
+-   [`jinja2_slug.SlugExtension`](https://pypi.org/project/jinja2-slug/#files): provides
+    a `slug` filter using [unicode-slugify](https://github.com/mozilla/unicode-slugify).
+-   [`jinja2_time.TimeExtension`](https://github.com/hackebrot/jinja2-time): adds a
+    `now` tag that provides convenient access to the
+    [arrow.now()](http://crsmithdev.com/arrow/#arrow.factory.ArrowFactory.now) API.
```
Yajo, line 401: I think it would be very helpful to list here cookiecutter extensions, because those won't appear while searching pypi, and would help people coming from cookiecutter to preserve some features they might be using.

docs/configuring.md
```diff
@@ -380,6 +380,64 @@ Also don't ask questions to the user; just use default values
...
+
+Search for more extensions on GitHub using the
+[jinja2-extension topic](https://github.com/topics/jinja2-extension) or
+[other Jinja2 topics](https://github.com/search?q=jinja&type=topics).
```
Yajo, line 405: https://pypi.org/search/?q=jinja+extension would help too.

Yajo (commented):
This looks extremely good, thanks for your valuable contribution.

There are a little couple of additions I'd like you to add to the docs before merging, please.

docs/configuring.md
pawamoy: Totally. But maybe we could show how to do the same with Ansible filters, without a dependency to cookiecutter, since it's possible.

docs/configuring.md
pawamoy: Done.

Yajo (approved):
Awesome!

pawamoy:
Just wanted to say thank you again for taking the time to review this with me @Yajo! I really appreciate it 🙂

Yajo:
I appreciate your time and effort too, thanks 😄

Issue #186, opened 2020-04-16
Better handle empty suffix


pawamoy:
I tried to bypass (:smiling_imp:) using a suffix for templates with `_templates_suffix: ""` in my `copier.yml`. Unfortunately it didn't work as expected: only directories were rendered (see #185), and no file was copied/rendered at all, even files with no extension.

Not sure what the code is doing with the suffix, but surely an error/warning could be displayed if the suffix is empty.

Yajo:
Why would you like to resolve the suffix before answering? 🤔

pawamoy:
For the same reason as in #197, i.e. to avoid prompting the users for nothing. We could argue however that, supposedly, the template writer already made sure the template is working, and didn't set an empty suffix as it doesn't work. In that case, maybe we could simply tell in the documentation that an empty suffix is invalid.

Yajo:
Ah I see. Well, I think that the case of empty suffix, meaning that all files are templates, should be valid.

pawamoy:
Alright! Then I'll try to send a PR with a fix 🙂

PR #397, opened 2021-05-04
fix: Handle empty templates suffix

pawamoy:
Issue #186

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -463,7 +463,8 @@ class Worker:
         templated_sibling = (
             self.template.local_abspath / f"{relpath}{self.template.templates_suffix}"
         )
-        if templated_sibling.exists():
+        # With an empty suffix, the templated sibling always exists.
+        if templated_sibling.exists() and self.template.templates_suffix:
             return None
         rendered_parts = []
         for part in relpath.parts:
@@ -473,7 +474,9 @@ class Worker:
                 return None
             rendered_parts.append(part)
         with suppress(IndexError):
-            if is_template:
+            # With an empty suffix, the next instruction
+            # would erroneously empty the last rendered part
+            if is_template and self.template.templates_suffix:
                 rendered_parts[-1] = rendered_parts[-1][
                     : -len(self.template.templates_suffix)
                 ]
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -721,6 +721,15 @@ Suffix that instructs which files are to be processed by Jinja as templates.
     _templates_suffix: .my-custom-suffix
     ```
 
+An empty suffix is also valid, and will instruct Copier to copy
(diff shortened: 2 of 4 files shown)


copier/main.py
```diff
@@ -463,7 +463,7 @@ def _render_path(self, relpath: Path) -> Optional[Path]:
         templated_sibling = (
             self.template.local_abspath / f"{relpath}{self.template.templates_suffix}"
         )
-        if templated_sibling.exists():
+        if templated_sibling.exists() and self.template.templates_suffix:
```
pawamoy, line 466: With an empty suffix, the templated sibling *always* exists.

copier/main.py
```diff
@@ -473,7 +473,7 @@ def _render_path(self, relpath: Path) -> Optional[Path]:
                 return None
             rendered_parts.append(part)
         with suppress(IndexError):
-            if is_template:
+            if is_template and self.template.templates_suffix:
```
pawamoy, line 476: With an empty suffix, the next instruction would actually empty the last rendered part (`= rendered_parts[-1][0:-0]`)

tests/demo_empty_suffix/hello
```diff
@@ -0,0 +1 @@
+hello {{ name }}!
```
pawamoy, line 1: Tests not ready, could you show me again how to generate them from the tests directly rather than creating `demo` folders? :slightly_smiling_face:

Yajo: Yes, see for example https://github.com/copier-org/copier/blob/1f24b5a02e33960cd7d71c998e475b253efd62ae/tests/test_tasks.py#L9

Yajo:
Thanks for those comments, so I understand better the patch. I think they fit well as code comments. I like the approach. What's missing:

- [ ] Test
- [ ] Docs
- [ ] Note in roadmap
- [ ] Commit name (it's feat, not fix; empty suffixes were not supported before).

pawamoy:
Tests OK, docs OK, commit type OK, I'm not sure what/where the roadmap is!

docs/configuring.md
```diff
@@ -721,6 +721,15 @@ Suffix that instructs which files are to be processed by Jinja as templates.
     _templates_suffix: .my-custom-suffix
     ```
 
+Starting at version 6, an empty suffix is also valid, and will instruct Copier to copy
```
Yajo, line 724:
```suggestion
An empty suffix is also valid, and will instruct Copier to copy
```


Version changes details belong to the Changelog (that's what I meant by "roadmap", Sorry for the mistake 😅).

Yajo (changes requested):
Great contribution! Please just add the changelog note.

Yajo (approved):

PR #407, opened 2021-05-15
Replace merging approach to a three-way merge

oprypin:
It always succeeds and, instead of leaving `.rej` files, it puts merge markers in the destination.

In my view the three-way merge conflict markers are easy to understand, while `.rej` files are super annoying and not very useful.

And, well, the merging approach here changes from doing a two-way diff and applying it to the destination, to just doing a three-way diff. Which I believe may lead to better results and not suffer from weirdness trying to save and restore a unified diff.

For example, whereas before you'd get an "overwrite?" prompt and a `.rej` file saying that `python = "^3.6"` changed to `python = "^3.7"`, now none of that happens and this gets written into the file:

Screenshot from VS Code:
image

Example is based on https://github.com/oprypin/mkdocs-same-dir/blob/b929337215ebe3f469b11b59b44052bb2546ac41/.tools/copier-answers.yml and changing `min_python_version: '3.7'`.

---

For now this PR is just an example so you can try it out (please do try it out!) and see if you like the approach.

This pull request is not in a state that's for merging. Moreover, it is based on an old state of the codebase because I found that https://github.com/copier-org/copier/pull/314 adds a lot of assumptions that the destination directory is the same as the source directory, making this very difficult to achieve (at least I couldn't).

I recommend to [view this PR with ignoring whitespace](https://github.com/copier-org/copier/pull/407/files?diff=split&w=1)

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -7,10 +7,11 @@ import subprocess
 import sys
 import tempfile
 from pathlib import Path
+from shutil import copyfile
 from typing import Callable, Dict, List, Optional, Sequence, Tuple
 
 from packaging.version import InvalidVersion, Version
-from plumbum import ProcessExecutionError, colors, local
+from plumbum import colors, local
 from plumbum.cli.terminal import ask
 from plumbum.cmd import git
 
@@ -246,9 +247,13 @@ def update_diff(conf: ConfigData) -> None:
                     "Downgrades are not supported."
                 )
     # Copy old template into a temporary destination
-    with tempfile.TemporaryDirectory(prefix=f"{__name__}.update_diff.") as dst_temp:
+    with tempfile.TemporaryDirectory(
+        prefix=f"{__name__}.update_diff.old."
+    ) as old_temp, tempfile.TemporaryDirectory(
+        prefix=f"{__name__}.update_diff.new."
+    ) as new_temp:
         copy(
-            dst_path=dst_temp,
+            dst_path=old_temp,
             data=last_answers,
             force=True,
             quiet=True,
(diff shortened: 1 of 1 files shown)


pawamoy:
That screenshot is really selling the point.

I agree on the rej files: I almost never look at them and simply delete them just after an update (my `make clean` runs a `find . type -name "*.rej" -delete`).

And for medium-to-big updates, I always end up staging parts of the modifications and reverting some others, or merging manually (which is error prone: easy to miss a line that was deleted and should have been kept) directly in VSCode, so actual markers like presented in the screenshot would definitely help during this process!

Yajo:
For me, those files help to know if the update was successful, while combined with a pre-commit hook that forbids them. Do you know of any hooks that forbid committing these kind of diffs?

Yajo:
Answering my own question, yes there's this hook: https://github.com/pre-commit/pre-commit-hooks#check-merge-conflict

So this would be a totally acceptable and great improvement. The PR is _very_ outdated now, and it needs tests (well... actually I think it will *break* tests, so fixing those should be enough probably). May you continue with it @oprypin?

oprypin:
@Yajo Thanks for consideration.

The PR was "outdated" from the start: I intentionally worked on an old version of the project because, in my opinion, with the more recent refactors the code is harder to work with in general and also has more assumptions about file states that go against this approach.

I never really fully learned the codebase and by now I probably forgot everything, so it seems unlikely that I will pick up this PR myself.

I also had some trouble dealing with the fact that the "Overwrite? y/n" prompt doesn't make sense anymore and how to replace it. And maybe other similar decisions.

The `merge` function is the part that's "interesting" and also quite universal. Maybe it would be productive if *you* take only that function and implement the other parts of integrating it with the current code.

Yajo:
> I intentionally worked on an old version

If you never wanted to produce a mergeable PR, it would have been more productive to open an issue. Closing.

PR #413, opened 2021-06-15
feat: Add additional flags for using default questions or overwriting files

### LINKED ISSUE #338: `--no-input` flag
Shouldn't there be a `--no-input` flag to tell Copier to not ask questions and use default values?  
Available during creation and update.

The `--force` should be just as it says: “Overwrite files that already exist, without asking”… so nothing to do with questions/answers.

dennisroche:
Fixes #338 as discussed

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -56,6 +56,8 @@ class CopierApp(cli.Application):
         vcs_ref: Set [vcs_ref][] option.
         pretend: Set [pretend][] option.
         force: Set [force][] option.
+        defaults: Set [defaults][] option.
+        overwrite: Set [overwrite][] option.
         skip: Set [skip_if_exists][] option.
         prereleases: Set [use_prereleases][] option.
         quiet: Set [quiet][] option.
@@ -120,7 +122,16 @@ class CopierApp(cli.Application):
         ["-n", "--pretend"], help="Run but do not make any changes"
     )
     force: cli.Flag = cli.Flag(
-        ["-f", "--force"], help="Overwrite files that already exist, without asking"
+        ["-f", "--force"],
+        help="Same as `--defaults --overwrite`.",
+    )
+    defaults: cli.Flag = cli.Flag(
+        ["-l", "--defaults"],
+        help="Use default answers to questions, which might be null if not specified.",
+    )
+    overwrite: cli.Flag = cli.Flag(
+        ["-w", "--overwrite"],
+        help="Overwrite files that already exist, without asking.",
     )
     skip: cli.Flag = cli.SwitchAttr(
         ["-s", "--skip"],
@@ -167,7 +178,8 @@ class CopierApp(cli.Application):
             dst_path=dst_path,
             answers_file=self.answers_file,
             exclude=self.exclude,
-            force=self.force,
+            defaults=self.force or self.defaults,
+            overwrite=self.force or self.overwrite,
             pretend=self.pretend,
             quiet=self.quiet,
             src_path=src_path,
--- a/copier/main.py
+++ b/copier/main.py
@@ -626,7 +626,7 @@ class Worker:
                 self,
                 dst_path=dst_temp,
                 data=self.subproject.last_answers,
-                force=True,
+                defaults=True,
                 quiet=True,
                 src_path=self.subproject.template.url,
                 vcs_ref=self.subproject.template.commit,
--- a/copier/main.py
+++ b/copier/main.py
@@ -107,10 +107,15 @@ class Worker:
 
             See [cleanup_on_error][].
 
-        force:
-            When `True`, disable all user interactions.
+        defaults:
+            When `True`, use default answers to questions, which might be null if not specified.
 
-            See [force][].
+            See [defaults][].
+
+        overwrite:
+            When `True`, Overwrite files that already exist, without asking.
+
+            See [overwrite][].
 
         pretend:
             When `True`, produce no real rendering.
@@ -132,7 +137,8 @@ class Worker:
     use_prereleases: bool = False
     skip_if_exists: StrSeq = ()
     cleanup_on_error: bool = True
-    force: bool = False
+    defaults: bool = False
+    overwrite: bool = False
     pretend: bool = False
     quiet: bool = False
 
@@ -230,9 +236,9 @@ class Worker:
                 file_=sys.stderr,
             )
             return False
-        if self.force:
+        if self.overwrite:
             printf(
-                "force",
+                "overwrite",
                 dst_relpath,
                 style=Style.WARNING,
                 quiet=self.quiet,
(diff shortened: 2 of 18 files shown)


copier/cli.py
```diff
@@ -120,7 +122,15 @@ class CopierApp(cli.Application):
         ["-n", "--pretend"], help="Run but do not make any changes"
     )
     force: cli.Flag = cli.Flag(
-        ["-f", "--force"], help="Overwrite files that already exist, without asking"
+        ["-f", "--force"],
+        help="Overwrite files that already exist, without asking. Use default answers to questions, which might be null if not specified.",
```
Yajo, line 126: To avoid repetition, I think I prefer:

```suggestion
        help="Same as `--default --overwrite`.",
```

copier/cli.py
```diff
@@ -120,7 +122,15 @@ class CopierApp(cli.Application):
...
+    )
+    defaults: cli.Flag = cli.Flag(
+        ["--defaults"],
+        help="Use default answers to questions, which might be null if not specified.",
+    )
+    overwrite: cli.Flag = cli.Flag(
+        ["--overwrite"], help="Overwrite files that already exist, without asking."
```
Yajo, line 133: Please add short flags

```suggestion
    defaults: cli.Flag = cli.Flag(
        ["-l", "--defaults"],
        help="Use default answers to questions, which might be null if not specified.",
    )
    overwrite: cli.Flag = cli.Flag(
        ["-w", "--overwrite"], help="Overwrite files that already exist, without asking."
```

docs/configuring.md
```diff
@@ -490,6 +490,33 @@ Overwrite files that already exist, without asking.
...
+
+### `overwrite`
+
+-   Format: `bool`
+-   CLI flags: `--overwrite`
+-   Default value: `False`
+
+Overwrite files that already exist, without asking.
+
+Also don't ask questions to the user; just use default values
```
Yajo, line 517: I think this statement is wrong.

copier/cli.py
```diff
@@ -168,6 +178,8 @@ def _worker(self, src_path: OptStr = None, dst_path: str = ".", **kwargs) -> Wor
             answers_file=self.answers_file,
             exclude=self.exclude,
             force=self.force,
+            defaults=self.defaults,
+            overwrite=self.overwrite,
```
Yajo, line 182: I'm OK about keeping the `-f` shortcut for CLI calls, but I'd prefer to remove it from API, now that we have them split.

So, this call should become:

```suggestion
            defaults=self.defaults or self.force,
            overwrite=self.overwrite or self.force,
```

copier/main.py
```diff
@@ -133,6 +143,8 @@ class Worker:
     skip_if_exists: StrSeq = ()
     cleanup_on_error: bool = True
     force: bool = False
```
Yajo, line 145: This attribute should disappear and stop being used.

However, it can be kept in helper methods (run_auto, run_copy and run_update) if you want.

Next release will be major, so there's no problem in breaking compatibility (it's broken already anyway).

I understand that making this change can turn many tests red. Do you feel like doing it?

Yajo (changes requested):
Thanks for your contribution!

One suggestion about removing the flag at API level.

Also, please, we need some test of this new feature. Both CLI and API tests would be best.

copier/main.py
```diff
@@ -320,7 +332,7 @@ def answers(self) -> AnswersMap:
             questions.append(
                 Question(
                     answers=result,
-                    ask_user=not self.force,
+                    ask_user=not self.force or self.defaults,
```
pawamoy, line 335: I'm wondering: shouldn't it be `not self.force and not self.defaults`?

```python
>>> force = False
>>> defaults = True
>>> ask = not force or defaults
>>> ask
True
```

Tell me if I'm wrong but we don't want to ask anything if force is false and we have defaults, right?

docs/configuring.md
pawamoy: Seems it was copied from the `force` description.

dennisroche:
Thanks for the reviews; will action the comments soon.

copier/main.py
dennisroche: yes, this should be ` or not self.defaults`

copier/main.py
dennisroche: I was wanting to keep this as a non-breaking change, perhaps release sooner as a 5.1.1 release. 

> I understand that making this change can turn many tests red. Do you feel like doing it?

I like breaking tests. However, would prefer to that in a new PR e.g. add new feature then remove depreciated in 2 steps. That ok?

copier/cli.py
dennisroche: Agree, that makes sense.

docs/configuring.md
dennisroche: yes, copy/paste error.

copier/main.py
dennisroche: as we're removing `force` from the API (keeping the cli), this will become `ask = not defaults`

copier/main.py
Yajo: If you want, I can release a new v 6 alpha when merged, but not a V 5.x. Currently, the master branch has diverged too much from v 5, so I'm not getting patches for it, and next release will have to be 6. However, it's still far from stable.

Regarding the split on 2 PR, we can do it if you feel more confident, but given our current situation I'm not sure it's worth it. You can just split on various commits here. But i'll let you decide 😄

dennisroche: Yes, understand that `master` is very different. Have you considered a `develop` branch for v6 so that patches could still be accepted for v5?

>I understand that making this change can turn many tests red. Do you feel like doing it?

I have made the change. Currently on Windows 29 tests are failing with:

```
result = Path(self.url)
        if self.vcs == "git":
            result = Path(clone(self.url_expanded, self.ref))
            if self.ref is None:
                checkout_latest_tag(result, self.use_prereleases)
        if not result.is_dir():
>           raise ValueError("Local template must be a directory.")
E           ValueError: Local template must be a directory.
```

Spent way too much time trying to understand this one. Pushed the changes up and going to see how it fails on CI.

Yajo: > Have you considered a `develop` branch for v6 so that patches could still be accepted for v5?

Yes, but I have very constrained resources so as to maintain 2 branches, and I have no time pressure to publish v6, so for the time being I prefer to stick to a single branch.



> I have made the change. Currently on Windows 29 tests are failing with:

It seems there's a different bug that doesn't let it get there.

dennisroche: thanks for pushing the fix for the tests. to be honest, i was going to abandon this work as moved onto other projects.

Yajo (approved):

copier/main.py
Yajo: You were so close... 😄 But the last mile needed some deeper knowledge on the code base, so I couldn't ask for more. Thanks for your contribution.

PR #446, opened 2021-09-17
Add config-file option

ArturCC:
This feature allows to specify different config file instead of hardcoded copier.yaml/copier.yml in template directory. The file can be in any other place (ie. in destination dir) and allows to use one copier.yaml for multiple templates in my case.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -131,6 +131,7 @@ class Worker:
     src_path: Optional[str] = None
     dst_path: Path = field(default=".")
     answers_file: Optional[RelativePath] = None
+    config_file: Optional[str] = None
     vcs_ref: OptStr = None
     data: AnyByStrDict = field(default_factory=dict)
     exclude: StrSeq = ()
@@ -363,6 +364,17 @@ class Worker:
         return self.answers_file or self.template.answers_relpath
 
     @cached_property
+    def config_file(self) -> Path:
+        """Obtain the proper path for the config file.
+
+        It comes from:
+
+        1. User choice.
+        2. Copier default.
+        """
+        return self.config_file
+
+    @cached_property
     def all_exclusions(self) -> StrSeq:
         """Combine default, template and user-chosen exclusions."""
         return self.template.exclude + tuple(self.exclude)
@@ -540,7 +552,12 @@ class Worker:
             if self.subproject.template is None:
                 raise TypeError("Template not found")
             url = self.subproject.template.url
-        return Template(url=url, ref=self.vcs_ref, use_prereleases=self.use_prereleases)
+        return Template(
+            url=url,
+            ref=self.vcs_ref,
+            use_prereleases=self.use_prereleases,
+            config_file=self.config_file,
+        )
 
     @cached_property
     def template_copy_root(self) -> Path:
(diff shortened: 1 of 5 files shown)


copier/template.py
```diff
@@ -170,23 +170,32 @@ class Template:
 
             Helpful if you want to test templates before doing a proper release, but you
             need some features that require a proper PEP440 version identifier.
+
+        config_file:
+            Path to the file to be used instead copier.y?ml
```
pawamoy, line 175: Typo `y?ml`?

docs/configuring.md
```diff
@@ -365,6 +365,14 @@ Don't forget to read [the docs about the answers file](#the-copier-answersyml-fi
     _answers_file: .my-custom-answers.yml
     ```
 
+### `config_file`
+
+-   Format: `str`
+-   CLI flags: `-c`, `--config-file`
+-   Default value: `copier.y?ml`
```
pawamoy, line 372: OK might not be a typo, but the `?` wildcard is usually put *after* the character it affects, not instead of it. Just a nit pick,

copier/template.py
ArturCC: Yes, it had to be `ya?ml`, but now I have changed the description to be more straightforward

docs/configuring.md
ArturCC: You are right, I have changed it to explicit two values

copier/main.py
```diff
@@ -362,6 +363,17 @@ def answers_relpath(self) -> Path:
...
+    def config_file(self) -> Path:
+        """Obtain the proper path for the config file.
+
+        It comes from:
+
+        1. User choice.
+        2. Copier default.
+        """
+        return self.config_file
+
```
Yajo, line 376: This code is buggy because the computed property and the field are named the same. Just remove it.

Yajo (changes requested):

Yajo:
> allows to use one copier.yaml for multiple templates in my case

Wouldn't a symlink work for you? 🤔

copier/main.py
ArturCC: Removed

ArturCC:
I tried the symlink but it wasn't enough elastic for me

Yajo:
I find it confusing that, if the path specified is relative, it becomes relative to the CWD and not to the template. It seems to me that what you're trying to fix would be covered already by including an external config file. Have you tried that? Does it work for you? If not, why?

I'm trying to avoid becoming feature creep (again), so please be patient 😁

ArturCC:
Path is a path, can be both relative or absolute in this case ;) "Config-file" option is for me absolutely basic feature for any application.
Include from yaml will not work for me.

I want to keep in one repository multiple config files lets call them:
- copier-app-a.yaml
- copier-app-b.yaml
- copier-app-c.yaml
- ...

And then use them for templates from other repository.

Particularly my use-case is templating environments for multiple K8S apps.

pawamoy:
So, you would have to clone your repository containing the config files first, and then run a `copier -c /path/to/config/on/disk.yml` command to generate projects?

Yajo:
Why include is not enough?

ArturCC:
I achieved my purpose with answers-file and some includes. Thank you for your time.

This PR can be closed.

Issue #451, opened 2021-09-25
Support updating a subproject when it isn't in the root of the git repo


Yajo:
### Discussed in https://github.com/copier-org/copier/discussions/449

<div type='discussions-op-text'>

<sup>Originally posted by **flippedAben** September 22, 2021</sup>
I have a Git repo at `~/repo`. I have another Git repo at `~/my_copier_template`. I tell `copier` to use `~/my_copier_template` to generate a subproject at `~/repo/path/to/S` (call it __S__ for short). I track my new subproject in `~/repo` by adding and committing the changes.

When I try to update __S__, it fails at: https://github.com/copier-org/copier/blob/b8ff4d8c711cf16683dd14f4bf3768df2b85e07e/copier/main.py#L607-L610

This is because `copier` assumes that a subproject is tracked by Git if and only if it is in the root of a git repo, as stated here:
https://github.com/copier-org/copier/blob/b8ff4d8c711cf16683dd14f4bf3768df2b85e07e/copier/subproject.py#L77-L80

But, __S__ is not at the root of `~/repo`. I would like to change copier to support this scenario, but wanted to get some thoughts before proceeding.

Part of the change will involve changing how we update:
https://github.com/copier-org/copier/blob/b8ff4d8c711cf16683dd14f4bf3768df2b85e07e/copier/main.py#L670-L673

How do we feel about not doing the `fetch`, and doing a `git diff` between __R__:`/path/to/S` and the cleanly generated __T__, instead of `git diff-tree`.
</div>

PR #452, opened 2021-09-25
fix: Fix issue with update subdirectory given an answers file path

bburgin:
Fixes https://github.com/copier-org/copier/issues/451

## Justification
During update, the destination _subdirectory is joined with the rendered answers_file value. We want to avoid including the same folders twice in the path.

## Implementation
Write the answers filename to the config, but without the path.

## Testing
Added unit test for this case.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -191,9 +191,15 @@ class Worker:
         # Backwards compatibility
         # FIXME Remove it?
         conf = asdict(self)
+
+        # During update, the destination _subdirectory is joined with the rendered answers_file value.
+        # To avoid including the same folders twice in the path, write to the config the
+        # answers filename (without the path).
+        answers_file = self.answers_relpath.name
+
         conf.update(
             {
-                "answers_file": self.answers_relpath,
+                "answers_file": answers_file,
                 "src_path": self.template.local_abspath,
             }
         )
--- a/tests/test_subdirectory.py
+++ b/tests/test_subdirectory.py
@@ -1,6 +1,8 @@
 import os
+from pathlib import Path
 
 import pytest
+import yaml
 from plumbum import local
 from plumbum.cmd import git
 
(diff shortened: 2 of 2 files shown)


Yajo:
This one doesn't seem to fix the linked issue.

Keep in mind that the subdirectory option belongs to the template, and indicates *what subdirectory of the template should I render in `dst`*. Whereas the answers_file option belongs to the destination and indicates *where the answers file is located inside `dst`*. I think you're mixing subjects.

The expected fix is to allow updates in destination when destination is inside a git repo, no matter if it's the root of the repo or not.

bburgin:
@Yajo:

Thank you for the clarifications. I am still trying to understand the various use cases and what we can do in the code to meet them. I see two use cases:
* subdirectory use case 1: The user goes to the subdirectory of a repo and runs update from there. The subdirectory contains the answer file.
* subdirectory use case 2: The user goes to the root of the repo and runs update from there, without cd'ing into the subdirectory that contains the answer file.

I think we should support both use cases. Otherwise, how would the user know which is the correct usage?

> The expected fix is to allow updates in destination when destination is inside a git repo, no matter if it's the root of the repo or not.

I think you are correct. The issue (subdirectory use case 1) that I linked in this PR may be the wrong issue for the changes I made. I meant my changes to fix subdirectory use case 2.

For subdirectory use case 2
Without the fix from this PR, the update call fails. The user gets doubled paths for the answer file and an error that the file could not be found. See the unit test in this PR for an example of the use case and inputs to copier. You can see the error by running the unit test but without the fix I made in main.py.

Potential action items
I could remove the mention of the issue (subdirectory use case 1) from this PR and proceed forward with it to fix subdirectory use case 2. What do you think?

> Keep in mind that the subdirectory option belongs to the template, and indicates what subdirectory of the template should I render in dst. Whereas the answers_file option belongs to the destination and indicates where the answers file is located inside dst. I think you're mixing subjects.

I am open to suggestions on how to improve the changes in this PR, to still fix subdirectory use case 2. What changes do you want me to make, so that it better aligns with the copier design and code?

Thank you.

Yajo:
> subdirectory use case 2: The user goes to the root of the repo and runs update from there, without cd'ing into the subdirectory that contains the answer file.

Just forget about this use case. Once you fix case 1, this will work automatically with:

    copier update path/to/subfolder

We can't guess where the subfolder is because a single project can have more than 1 subfolders that each one use different templates.

The copier-answers path provided with `-a` is always relative to the root of the destination (`path/to/subfolder` in above example).

bburgin:
Thanks. Are you saying that subdirectory use case 2 should not be supported?

I think it could be supported with this PR change that I have made. The user can provide enough information in the -a argument so that the path down to the answers file is not ambiguous. The -a argument can contain the relative path folders down to the answer file from the destination path, as I have shown in the unit test.

If this is still not acceptable, then can we make copier return a better error for this case? Something like:
"This case is not supported. Please always pass folder information via the destination path not the -a argument. The -a argument must be a filename only, not a path."

bburgin:
I am currently working on the subdirectory use case 1 and will post a separate PR for it.

Yajo:
Yes, case 2 is conceptually wrong, because -a is not meant to explain where the root of your project is. Instead it just tells where to find the answers, and those could perfectly exist in a subdirectory.

The thing that tells the project root is `dst_path`.

If your project has several subsections that apply templates, you should indicate that in dst_path. If a subsection uses more than 1 template, then you'll need different -a inside that subsection.



> I am currently working on the subdirectory use case 1 and will post a separate PR for it.

Ok, closing this one as explained. Thanks!

PR #453, opened 2021-10-01
fix: Fix git error for subdirectory case

bburgin:
Fixes https://github.com/copier-org/copier/issues/451

## Justification
Allow the user to run update from a repo but passing in a subdirectory for the dst_path.

## Implementation
During update, when we check that a folder is a git repo, ask git if it is a git repo or not. Use the repo root returned from git for a given subdirectory.

## Testing
Added unit test for this case.


tests/test_subdirectory.py
```diff
@@ -86,6 +88,68 @@ def test_update_subdirectory(demo_template, tmp_path):
...
+        git("commit", "-m1")
+        copier.run_update(
+            component_reldir,
+            defaults=True,
+            overwrite=True,
+            answers_file=answers_filename,
+        )
+    answers = yaml.safe_load(dst.joinpath(answers_file_relpath).read_bytes())
+    assert answers["_commit"] == "2"
+    assert dst.joinpath(component_reldir, "test.txt").read_text() == "False"
```
Yajo, line 150: I find the patch OK, but the test is very complex to understand. Don't worry about that, I'll simplifiy it later when I have time. 😉

Yajo (changes requested):

tests/test_subdirectory.py
bburgin: okay thanks

PR #465, opened 2021-10-27
Keep trailing newlines for Jinja templates by default. Closes #464

### LINKED ISSUE #464: feat: set jinja `keep_trailing_newline` to `true` by default

No

I know that with v6 you want to use the Jinja default for the templates (which I appreciate :) ). 
However, I doubt that it makes sense to trim trailing newlines in template files for copier. 
IMHO: the sensible default would be to keep the trailing newlines for template files.

I'm aware that one can change it in the copier config.

**Additional context**

timofurrer:


Yajo (changes requested):
Thanks! It's great 😊 

Would you please add a new test that ensures that when a template sets it to `False`, it works as expected?

timofurrer:
@Yajo done 🎉

Yajo (approved):

PR #476, opened 2021-11-11
raise CopierAnswersInterrupt on partial answers

jacobstr:
Use case: I'm using the copier api and would like to capture partially completed answers. When completing long questionnaires it's sometimes :sadpanda: to ctrl-c and repeat your work. I plan on:

- catching this exception
- writing the `e.answers.user` out to a file.
- on subsequent runs, checking for it's existence and populating `run_copy(defaults=partial_answers)`.

I suspect the above procedure could likely be generalized as a first class feature in the project e.g. copier could handle writing a `~/.copier_answers_partial.json`  or `~/.coper_{{template_name}}_answers_partial.json`, or even write out to `SubProject.last_answers`.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -24,7 +24,7 @@ from pydantic.dataclasses import dataclass
 from pydantic.json import pydantic_encoder
 from questionary import unsafe_prompt
 
-from .errors import ExtensionNotFoundError, UserMessageError
+from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
 from .subproject import Subproject
 from .template import Template
 from .tools import Style, TemporaryDirectory, printf
@@ -337,13 +337,16 @@ class Worker:
             )
         for question in questions:
             # Display TUI and ask user interactively only without --defaults
-            new_answer = (
-                question.get_default()
-                if self.defaults
-                else unsafe_prompt(
-                    question.get_questionary_structure(), answers=result.combined
-                )[question.var_name]
-            )
+            try:
+                new_answer = (
+                    question.get_default()
+                    if self.defaults
+                    else unsafe_prompt(
+                        question.get_questionary_structure(), answers=result.combined
+                    )[question.var_name]
+                )
+            except KeyboardInterrupt as err:
+                raise CopierAnswersInterrupt(result, question, self.template) from err
             previous_answer = result.combined.get(question.var_name)
             if new_answer != previous_answer:
                 result.user[question.var_name] = new_answer
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -441,3 +441,33 @@ def test_multiline_defaults(tmp_path_factory, spawn):
         == answers["json_multi"]
         == ["one", "two", {"three": ["four"]}]
     )
+
+
+def test_partial_interrupt(tmp_path_factory, spawn):
+    template, subproject = (
+        tmp_path_factory.mktemp("template"),
+        tmp_path_factory.mktemp("subproject"),
+    )
+    build_file_tree(
+        {
+            template
+            / "copier.yml": yaml.safe_dump(
+                {
+                    "question_1": "answer 1",
+                    "question_2": "answer 2",
+                }
+            ),
+        }
+    )
+    tui = spawn(COPIER_PATH + (str(template), str(subproject)), timeout=10)
+    deque(map(tui.expect_exact, ["question_1?", "Format: str", "answer 1"]))
+    # Answer the first question using the default.
+    tui.sendline()
+    # Send a ctrl-c.
+    tui.send(Keyboard.ControlC + Keyboard.Enter)
+    # This string (or similar) should show in the output if we've gracefully caught the
+    # interrupt.
+    # We won't want our custom exception (CopierAnswersInterrupt) propagating
+    # to an uncaught stacktrace.
+    tui.expect_exact("Execution stopped by user\n")
+    tui.expect_exact(pexpect.EOF)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ Summary:
     `$VERSION_PEP440_CURRENT` and `$VERSION_PEP440_TO`, which will always get a valid
     PEP440 version identifier, without the `v` prefix, allowing your migration scripts
     to have a valid standard where to base their logic.
+-   Raise a CopierAnswersInterrupt instead of a bare KeyboardInterrupt to
+    provide callers with additional context - such as the partially completed
+    AnswersMap.
 
 ### Changed
 
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -7,6 +7,11 @@ from pydantic.errors import _PathValueError
 from .tools import printf_exception
 from .types import PathSeq
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # always false
+    from .user_data import AnswersMap, Question, Template
+
 
 # Errors
 class CopierError(Exception):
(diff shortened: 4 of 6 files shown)


pawamoy:
Nice. But I wonder: if you type a wrong answer, and want to start again to change it, then you can't? because the wrong answer is reused? How will you deal with this?

jacobstr:
Almost! See #477. The idea is that you override the template-provided defaults with the dict of previous answers.

Additionally, if you made it partway through and had a few answers to change you can edit the file (see #477 for a snippet that shows the CopierAnswersInterrupt + default_overrides working side by side.

pawamoy:
> you have the option of editing the answer - it's not skipped.

Ah, of course! OK, that's great then :slightly_smiling_face:

copier/main.py
```diff
@@ -44,6 +44,32 @@
     from backports.cached_property import cached_property
 
 
+class CopierAnswersInterrupt(Exception):
```
Yajo, line 47: Please move it to errors.py

copier/main.py
```diff
@@ -337,16 +363,21 @@ def answers(self) -> AnswersMap:
...
+            try:
+                new_answer = (
+                    question.get_default()
+                    if self.defaults
+                    else unsafe_prompt(
+                        question.get_questionary_structure(), answers=result.combined
+                    )[question.var_name]
+                )
+            except KeyboardInterrupt:
+                raise CopierAnswersInterrupt(result, question, self.template)
```
Yajo, line 375: Right now in this case, the last question is nicely grayed out on the TUI. By changing the exception I'm afraid if it would explode on the user hands uglyly. Is that controlled?

Yajo (changes requested):
Nice addition! Some remarks. Besides, lt needs a test.

Yajo:
BTW if upstream ever implements https://github.com/tmbo/questionary/issues/81, the problem you're trying to avoid wouldn't even exist. Have you considered it?

copier/main.py
jacobstr: Ack. May lose the type hinting to avoid circular imports (user_data imports errors, errors would need to import user_data).

jacobstr:
> BTW if upstream ever implements [tmbo/questionary#81](https://github.com/tmbo/questionary/issues/81), the problem you're trying to avoid wouldn't even exist. Have you considered it?

WRT to this - wasn't aware of this issue, but it doesn't appear to have lots of upstream traction. My concern for _this_ project would be potentially ending up with a redundant way of doing things vis-a-vis questionary.

But in practice, this MR is decoupled from any sort of "resume" functionality in questionary. We're - just tacking on some context to the KeyboardInterrupt. I think the salient thing is maintaining as much compatibility as possible with folks that may already be handling plain-old `KeyboardInterrupt` so that CopierAnswersInterrupt has a similar shape.

I still imagine regardless of what questionary does, it's not going to suppress the KeyboardInterrupt - it's consumers want a chance to handle that signal as well.

copier/main.py
jacobstr: I'm modifying the CopierAnswersInterrupt to extend KeyboardInterrupt to address this. Will add a test case, but local putzing shows it works as intended. Stacktrace without, No stackstrace with.

copier/main.py
pawamoy: You can use this:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:  # always false
    from copier... import user_data
```

tests/test_prompt.py
```diff
@@ -441,3 +441,33 @@ def test_multiline_defaults(tmp_path_factory, spawn):
...
+            ),
+        }
+    )
+    tui = spawn(COPIER_PATH + (str(template), str(subproject)), timeout=10)
+    deque(map(tui.expect_exact, ["question_1?", "Format: str", "answer 1"]))
+    # Answer the first question using the default.
+    tui.sendline()
+    # Send a ctrl-c.
+    tui.send(Keyboard.ControlC + Keyboard.Enter)
+    # This string (or similar) should show in the output if we've gracefully caught the
```
jacobstr, line 468: @Yajo here's the verification that our stacktrace is being handled gracefully.

copier/main.py
jacobstr: Ack, done. 

Note to folks that have hibernated on modern python roughly boiled down to this:

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:  # always false
    from .user_data import AnswersMap, Question, Template

...
# In the constructor, use string references to the type names instead of the imported Objects.
class CopierAnswersInterrupt(CopierError, KeyboardInterrupt):
    def __init__(self, answers: 'AnswersMap', last_question: 'Question', template: 'Template') -> None:
```

The stringly references could be removed in 3.7+ as per this SO comment: https://stackoverflow.com/a/39757388 

I couldn't find any references to `from __future__ import annotations` to backport that functionality to 3.6 (listed as one of the pyproject.toml classifiers currently). Not worth saving a couple of quotes IMO.

tests/test_prompt.py
jacobstr: Incidentally - I enjoyed this test rig after wrapping my head around it. Found it quirky though to get established. It looks like `tui.expect_exact()` consumes stdout until it reads those characters _somewhere_.  Had to dig around the API in pdb to find `readlines()` on the POpenSpawn object which made grokking what was going on under the hood easier.

I'd like to incorporate this in some of my own stuff. Pretty good ROI for contributing to this project :)

copier/main.py
```diff
@@ -337,16 +337,21 @@ def answers(self) -> AnswersMap:
...
-            )
+            try:
+                new_answer = (
+                    question.get_default()
+                    if self.defaults
+                    else unsafe_prompt(
+                        question.get_questionary_structure(), answers=result.combined
+                    )[question.var_name]
+                )
+            except KeyboardInterrupt:
```
Yajo, line 348:
```suggestion
            except KeyboardInterrupt as err:
```

copier/main.py
```diff
@@ -337,16 +337,21 @@ def answers(self) -> AnswersMap:
...
+                raise CopierAnswersInterrupt(result, question, self.template)
```
Yajo, line 349:
```suggestion
                raise CopierAnswersInterrupt(result, question, self.template) from err
```

AFAIK and According to docs this is the correct way of converting one exception into another.

copier/main.py
```diff
@@ -337,16 +337,21 @@ def answers(self) -> AnswersMap:
...
+
```
Yajo, line 350: Meaningless change, please undo.

copier/main.py
```diff
@@ -337,16 +337,21 @@ def answers(self) -> AnswersMap:
...
             previous_answer = result.combined.get(question.var_name)
             if new_answer != previous_answer:
                 result.user[question.var_name] = new_answer
+
```
Yajo, line 354: Meaningless change, please undo.

tests/test_prompt.py
```diff
@@ -441,3 +441,33 @@ def test_multiline_defaults(tmp_path_factory, spawn):
...
+
+def test_partial_interrupt(tmp_path_factory, spawn):
+    template, subproject = (
+        tmp_path_factory.mktemp("template"),
+        tmp_path_factory.mktemp("subproject"),
+    )
+    build_file_tree(
+        {
+            template
+            / "copier.yml": yaml.dump(
```
Yajo, line 454:
```suggestion
            / "copier.yml": yaml.safe_dump(
```

tests/test_prompt.py
Yajo: I'm happy to see you're having a pleasant contribution experience 😊

Yajo (changes requested):
Very good now! Just a couple of minor improvements on code to get ready to merge.

Thanks for your contribution!

Yajo (approved):

Issue #235, opened 2020-07-28
Add support for user config file


achaussier:
Hello,

First, thanks for this useful project I begin to use, with really great features !

I've checked quickly, but I haven't see the management of a user config file (as `.cookiecutterrc`) to manage user default answers on top of template default answers.

The order of override could be (from top to bottom):
1. Template default values
2. If existing user values (from `~/.copierrc` ?)
3. Last used values if already deployed and `.copier-answers.yml` exists

What do you think about that ?

Yajo:
Absolutely needed!

achaussier:
Thanks for your work @noirbizarre !

Have a nice day

PR #477, opened 2021-11-11
allow overriden defaults in questionnaire

jacobstr:
This works in conjunction with https://github.com/copier-org/copier/pull/476, but is independent of it. If partial answers are written out to a file, they can be re-loaded as default overrides to whatever is specified in the template allowing one to resume progress.

Some code that uses the combination of these two MRs:

```
PARTIAL_ANSWERS = Path('~/.copier_partial.json').expanduser()

kwargs = {}

if Path(PARTIAL_ANSWERS).exists():
    with open(PARTIAL_ANSWERS) as f:
        kwargs['default_overrides'] = json.load(f)

try:
    result = copier.run_auto(TEMPLATES_SRC, os.getcwd(), **kwargs)
except copier.CopierAnswersInterrupt as e:
    with open(PARTIAL_ANSWERS, "w") as f:
        f.write(json.dumps(e.answers.user))
    print("Wrote partial answers to %s" % PARTIAL_ANSWERS)
```

Demo of these two features working together: https://asciinema.org/a/iVJvC8wAb8s3nSVKkmnhDjeZE
- First pass, I make some modifications and bail.
- Second pass, you can see the prior inputs being used.
- Prior to the third pass I modify the previous answers that were saved ala the snippet above.
- Third pass it's using the template defaults again because I emptied the javascript object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Summary:
 -   Raise a CopierAnswersInterrupt instead of a bare KeyboardInterrupt to
     provide callers with additional context - such as the partially completed
     AnswersMap.
+-   Support for `user_defaults`, which take precedence over template defaults.
 
 ### Changed
 
--- a/copier/main.py
+++ b/copier/main.py
@@ -111,6 +111,9 @@ class Worker:
 
             See [defaults][].
 
+        user_defaults:
+            Specify user defaults that may override a template's defaults during question prompts.
+
         overwrite:
             When `True`, Overwrite files that already exist, without asking.
 
@@ -137,6 +140,7 @@ class Worker:
     skip_if_exists: StrSeq = ()
     cleanup_on_error: bool = True
     defaults: bool = False
+    user_defaults: AnyByStrDict = field(default_factory=dict)
     overwrite: bool = False
     pretend: bool = False
     quiet: bool = False
@@ -317,6 +321,7 @@ class Worker:
         """
         result = AnswersMap(
             default=self.template.default_answers,
+            user_defaults=self.user_defaults,
             init=self.data,
             last=self.subproject.last_answers,
             metadata=self.template.metadata,
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -97,6 +97,11 @@ class AnswersMap:
             Default data from the template.
 
             See [copier.template.Template.default_answers][].
+
+        user_defaults:
+            Default data from the user e.g. previously completed and restored data.
+
+            See [copier.main.Worker][].
     """
 
     # Private
@@ -107,6 +112,7 @@ class AnswersMap:
     init: AnyByStrDict = field(default_factory=dict)
     metadata: AnyByStrDict = field(default_factory=dict)
     last: AnyByStrDict = field(default_factory=dict)
+    user_defaults: AnyByStrDict = field(default_factory=dict)
     default: AnyByStrDict = field(default_factory=dict)
 
     @cached_property
@@ -118,6 +124,7 @@ class AnswersMap:
             self.init,
             self.metadata,
             self.last,
+            self.user_defaults,
             self.default,
             DEFAULT_DATA,
         )
@@ -208,7 +215,10 @@ class Question:
             try:
                 result = self.answers.last[self.var_name]
             except KeyError:
-                result = self.render_value(self.default)
+                try:
+                    result = self.answers.user_defaults[self.var_name]
+                except KeyError:
+                    result = self.render_value(self.default)
         result = cast_answer_type(result, cast_fn)
         return result
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -344,9 +344,10 @@ def test_user_defaults(tmp_path_factory, user_defaults, data, expected):
 
 
 @pytest.mark.parametrize(
-    "user_defaults_initial, user_defaults_updated, data_initial, data_updated, expected",
+    "user_defaults_initial, user_defaults_updated, data_initial, data_updated, expected_initial, expected_updated",
     [
-        # User defaults + secondary defaults.
+        # Initial user defaults and updated used defaults. The output
+        # should remain unchanged following the update operation.
         (
             {
                 "a_string": "foo",
@@ -361,8 +362,38 @@ def test_user_defaults(tmp_path_factory, user_defaults, data, expected):
             A string: foo
             """
             ),
+            dedent(
+                """\
+            A string: foo
+            """
+            ),
         ),
-        # User defaults + data provided.
+        # User defaults + data provided. Provided data should take precedence
+        # and the resulting content unchanged post-update.
+        (
+            {
+                "a_string": "foo",
+            },
+            {
+                "a_string": "foobar",
+            },
+            {
+                "a_string": "yosemite",
+            },
+            {},
+            dedent(
+                """\
+            A string: yosemite
+            """
+            ),
+            dedent(
+                """\
+            A string: yosemite
+            """
+            ),
+        ),
+        # User defaults + secondary defaults + data overrides. `data_updated` should
+        # override user and template defaults.
         (
             {
                 "a_string": "foo",
@@ -381,6 +412,11 @@ def test_user_defaults(tmp_path_factory, user_defaults, data, expected):
             A string: yosemite
             """
             ),
+            dedent(
+                """\
+            A string: red rocks
+            """
+            ),
         ),
     ],
 )
@@ -390,7 +426,8 @@ def test_user_defaults_updated(
     user_defaults_updated,
     data_initial,
     data_updated,
-    expected,
+    expected_initial,
+    expected_updated,
 ):
     src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
     with local.cwd(src):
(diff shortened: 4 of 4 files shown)


copier/main.py
```diff
@@ -326,6 +330,10 @@ def answers(self) -> AnswersMap:
             if var_name in result.init:
                 # Do not ask again
                 continue
+
+            if self.default_overrides.get(var_name) is not None:
+                details['default'] = self.default_overrides[var_name]
```
Yajo, line 335: Just in case you want a specific `None` default, I think it's better to:

```suggestion
            with supress(KeyError):
                details['default'] = self.default_overrides[var_name]
```

Yajo (changes requested):
A little comment on code.

This feature is cool. With a bit of rework, it could fix https://github.com/copier-org/copier/issues/235.

Copier could support loading these custom defaults from `$XDG_CONFIG_DIR/copier/user-defaults.yaml`. It could also feature a CLI flag `--user-defaults` to allow the execution environment to override that path. Then, the custom defaults get loaded from there automatically.

Implementation would probably go around https://github.com/copier-org/copier/blob/9f05fdf77d725933be3f779f1c3215467851f92a/copier/user_data.py#L70

Also, this still needs docs, tests, and a CHANGELOG entry to be merged.

Good work! Thanks 🙂

jacobstr:
> A little comment on code.
> 
> This feature is cool. With a bit of rework, it could fix #235.
> 
> Copier could support loading these custom defaults from `$XDG_CONFIG_DIR/copier/user-defaults.yaml`. It could also feature a CLI flag `--user-defaults` to allow the execution environment to override that path. Then, the custom defaults get loaded from there automatically.
> 
> Implementation would probably go around
> 
> https://github.com/copier-org/copier/blob/9f05fdf77d725933be3f779f1c3215467851f92a/copier/user_data.py#L70
> 
> Also, this still needs docs, tests, and a CHANGELOG entry to be merged.
> 
> Good work! Thanks 🙂

Cheers! 

To confirm - you believe this is an incremental step towards getting that functionality? I.E. it simplifies a subsequent MR (likely in cli.py?) that leverages the default-overrides available in the Worker class to merge these `rc` file defaults?

Walking that out a bit, we'll at some point want to merge (in order):

* user defaults (e.g. aws_project = teamname-dev)
* partially completed forms (e.g. aws_project = otherteamname-dev)

I think that logic could live entirely in cli.py - i.e. the caller would be responsible for reconciling these alternate default sources.

jacobstr:
* Tests added.
* Changelog updated.

WRT to docs, the docstrings have been updated. Please point me to where  we'd want additional docs.

Yajo:
> I think that logic could live entirely in cli.py - i.e. the caller would be responsible for reconciling these alternate default sources.

In general terms, copier behaves the same through API and CLI. However it's true that this can be an exception because it's more ergonomic for an API call to provide user defaults through data than through a local file. So: yes, this is a valid intermediate step forward.

copier/main.py
```diff
@@ -326,6 +330,10 @@ def answers(self) -> AnswersMap:
             if var_name in result.init:
                 # Do not ask again
                 continue
+
+            with suppress(KeyError):
+                details['default'] = self.default_overrides[var_name]
```
Yajo, line 335: Sorry, I just realized that this logic should be in the right spot: 

https://github.com/copier-org/copier/blob/fefc5921f989bb602fecbff8a48b89bfa2ab7266/copier/user_data.py#L70

That class is in charge of choosing the correct answer from different sources, you're just introducing a new source.

Yajo (changes requested):

copier/main.py
jacobstr: Just occurred to me that merging default_overrides and template_defaults near here: https://github.com/copier-org/copier/blob/overriden-defaults/copier/main.py#L323 and simply passing `defaults` feels best.

It feels odd having multiple sources in AnswersMap. At least, from the POV of the AnswersMap API having both `defaults` and `default_overrides` feels odd. Do you get default_override_overrides as a third tier at some point?

jacobstr: LMK how you feel about that. IMO reconciliation of these overrides feels like glue code well suited main / Worker. AnswersMap is a pretty straightforward data class at the moment.

If you believe this is an `AnswersMap` thing, it would make sense to generalize / expand `defaults` into some multi-tiered thing such as an ordered list of `DefaultSources` rather than a single default. 

E.g. tidy up this implementation by doing it here (as mentioned) https://github.com/copier-org/copier/blob/overriden-defaults/copier/main.py#L323

If AnswersMap gets multiple default_sources in the future, then main.py no longer has to merge template defaults and default_overrides and could just pass something like:

```python
AnswersMap(
    default=self.template.defaults, 
    additional_defaults=[
        DefaultSource('user_configuration', self.user_defaults),
        DefaultSource('default_overrides', self.default_overrides),
    ]
)
```

jacobstr: Of course, as I try it I see it won't work because the `Question` is created independently using different `question_data`.

Yajo: Well, I still feel `AnswersMap` is the place where it should go, as its purpose for existence is to prioirze and merge different answer sources.

Regarding the change, I'd just change `defaults` for `template_defaults`, and add a new `user_defaults` that takes priority over `template_defaults`.

jacobstr: Alright, I see it. I'll have something.

jacobstr: https://github.com/copier-org/copier/pull/477/commits/8d3e90523b34572bbaa0e1874433511e2ba342ae

This look about right? Not entirely clear where we want the `self.user_defaults` in the `AnswersMap:combined`. Existing tests pass but `AnswersMap` might benefit from some more finagling now.

jacobstr:
Random test failure appears spurious but I can't re-run.

copier/user_data.py
```diff
@@ -203,12 +210,15 @@ def get_default(self) -> Any:
         """Get the default value for this question, casted to its expected type."""
         cast_fn = self.get_cast_fn()
         try:
-            result = self.answers.init[self.var_name]
+            result = self.answers.user_defaults[self.var_name]
```
Yajo, line 213: The correct order would be: init, last, user defaults, template defaults.

tests/test_config.py
```diff
@@ -227,3 +227,73 @@ def test_worker_config_precedence(tmp_path, test_input, expected_exclusions):
...
+            A string: foo
+            A number: 42
+            A boolean: False
+            A list: four, five, six
+            """
+            )
+        )
+    ]
+)
+def test_user_defaults(tmp_path_factory, user_defaults, expected):
```
Yajo, line 265: This test is a bit incomplete. For example, there are bugs above that are not detected. Namely: 

- The test should pass data directly, asserting that init data gets more priority than user defaults.

- Also it should  update a project and assert that the last answers are used instead of user defaults.

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ Summary:
 -   Raise a CopierAnswersInterrupt instead of a bare KeyboardInterrupt to
     provide callers with additional context - such as the partially completed
     AnswersMap.
+-   New property on Worker, and run_* variants: `user_defaults` allows
```
Yajo, line 34: I'd simplify this with: "support for user defaults"

Yajo (changes requested):
Thanks friend! 🙂 The code looks nice, but it has some bugs. That also reveals the test is nice but not enough (as it should have detected those bugs). See code comments below.

tests/test_config.py
jacobstr: * Added `data` to this test case as well as changes suggested. Interesting, it didn't seem to make a difference in whether the test passed or not so I'm still not exercising the ordering changes.
* Any prior art in the existing tests to point me towards would be appreciated here.

CHANGELOG.md
jacobstr: Met halfway with: Support for `user_defaults`, which precedence over template defaults.

tests/test_config.py
```diff
@@ -227,3 +236,207 @@ def test_worker_config_precedence(tmp_path, test_input, expected_exclusions):
...
+            },
+            dedent(
+                """\
+            A string: red rocks
+            """
+            ),
+        ),
+    ],
+)
+def test_user_defaults_updated(
```
jacobstr, line 387: @Yajo here's my go at updating the template asserting copier answers + default interactions.

tests/test_config.py
```diff
@@ -227,3 +236,207 @@ def test_worker_config_precedence(tmp_path, test_input, expected_exclusions):
...
+            },
+            {
+                "a_string": "yosemite",
+            },
+            {
+                "a_string": "red rocks",
+            },
+            dedent(
+                """\
+            A string: yosemite
```
Yajo, line 381: If it passes it's weird. It should be this AFAICS:

```suggestion
            A string: red rocks
```

CHANGELOG.md
Yajo: The provided tests are OK 👍

tests/test_config.py
jacobstr: Ack. Fixed the test in a subsequent test case:


```python
        (
            {
                "a_string": "foo",
            },
            {
                "a_string": "foobar",
            },
            {
                "a_string": "yosemite",
            },
            {
                "a_string": "red rocks",
            },
            dedent(
                """\
            A string: yosemite
            """
            ),
            dedent(
                """\
            A string: red rocks
            """
            ),
        ),
```

On the first pass we provide: 

```
user_default: foo
data: yosemite
```

result: yosemite

Second pass:

```
user_default: foobar
data: yosemite
```

result: red rocks (overriding the copier answers from the initial copy).

Yajo (approved):
🎉

PR #485, opened 2021-11-20
feat: add support to parse yaml files as data input

Ket3r:
The `--data/-d` option now accepts yaml files instead of only key
value pairs. It parses the the yaml file and updates the data
dictionary with the content.

This shall make it easier for the user to configure more complex
projects. It also allows sending someone your project template
configuration.

# From the documentation:

As alternative you can also pass the path to a yaml file that contains your data. This
cannot be used in the API.

You have a input.yml file that contains:

```yml
user_name: Alice Bob
age: 7
height: 1.83
```

Calling this:
```sh
copier -d input.yml copy template destination
```

does the same thing as:
```sh
copier -d "user_name=Alice Bob" -d "age=7" -d "height=1.83" copy template destination
```

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -43,16 +43,18 @@ Basically, there are 3 different commands you can run:
 
 Below are the docs of each one of those.
 """
-
+import re
 import sys
 from functools import wraps
 from io import StringIO
+from pathlib import Path
 from textwrap import dedent
 from unittest.mock import patch
 
 import yaml
 from plumbum import cli, colors
 
+from copier.template import load_template_config
 from copier.tools import copier_version
 
 from .errors import UserMessageError
@@ -189,6 +191,17 @@ class CopierApp(cli.Application):
             values: The list of values to apply.
                 Each value in the list is of the following form: `NAME=VALUE`.
         """
+        files = [e for e in values if "=" not in e]
+        values = [e for e in values if "=" in e]
+        print(files)
+        print(values)
+        for file in files:
+            if not Path(file).exists():
+                print(colors.red | f"File: {file} doesn't exist", file=sys.stderr)
+            if not re.match(r".+\.ya?ml$", file, re.IGNORECASE):
+                print(colors.red | f"File: {file} isn't a yaml file", file=sys.stderr)
+            d = load_template_config(Path(file), True)
+            self.data.update(d)
         for arg in values:
             key, value = arg.split("=", 1)
             value = yaml.safe_load(value)
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -385,7 +385,7 @@ running `copier update`, this setting has no effect.
 
 ### `data`
 
--   Format: `dict|List[str=str]`
+-   Format: `dict|List[str=str]|yamlfile`
 -   CLI flags: `-d`, `--data`
 -   Default value: N/A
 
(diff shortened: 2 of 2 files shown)


copier/cli.py
```diff
@@ -189,6 +191,17 @@ def data_switch(self, values: List[str]) -> None:
...
                 Each value in the list is of the following form: `NAME=VALUE`.
         """
+        files = [e for e in values if "=" not in e]
+        values = [e for e in values if "=" in e]
+        print(files)
+        print(values)
+        for file in files:
+            if not Path(file).exists():
+                print(colors.red | f"File: {file} doesn't exist", file=sys.stderr)
+            if not re.match(r".+\.ya?ml$", file, re.IGNORECASE):
```
Yajo, line 201: I find it more ergonomic to avoid regexp and use `Path(file).suffix not in {"yaml", "yml"}`

copier/cli.py
```diff
@@ -189,6 +191,17 @@ def data_switch(self, values: List[str]) -> None:
             values: The list of values to apply.
                 Each value in the list is of the following form: `NAME=VALUE`.
         """
+        files = [e for e in values if "=" not in e]
+        values = [e for e in values if "=" in e]
+        print(files)
+        print(values)
```
Yajo, line 197: I guess this is just debugging garbage, right?

copier/cli.py
```diff
@@ -189,6 +191,17 @@ def data_switch(self, values: List[str]) -> None:
             values: The list of values to apply.
                 Each value in the list is of the following form: `NAME=VALUE`.
         """
+        files = [e for e in values if "=" not in e]
+        values = [e for e in values if "=" in e]
```
Yajo, line 195: Please avoid using meaningles var names such as `e`, for the sake of readability.

copier/cli.py
```diff
@@ -189,6 +191,17 @@ def data_switch(self, values: List[str]) -> None:
...
+        print(values)
+        for file in files:
+            if not Path(file).exists():
+                print(colors.red | f"File: {file} doesn't exist", file=sys.stderr)
+            if not re.match(r".+\.ya?ml$", file, re.IGNORECASE):
+                print(colors.red | f"File: {file} isn't a yaml file", file=sys.stderr)
+            d = load_template_config(Path(file), True)
+            self.data.update(d)
         for arg in values:
             key, value = arg.split("=", 1)
```
Yajo, line 206: I'd like it more if you used this same loop instead of creating a new one which re-defines the `values` argument. All would be more obvious to the reader. Besides, you can use this line to know if there's a `=` in the argument, as there'll be a `ValueError` otherwise.

docs/configuring.md
```diff
@@ -385,7 +385,7 @@ running `copier update`, this setting has no effect.
 
 ### `data`
 
--   Format: `dict|List[str=str]`
+-   Format: `dict|List[str=str]|yamlfile`
```
Yajo, line 388: Please undo this change, as it is only meant for API AFAICR (you can't pass a `dict` through CLI).

Yajo (changes requested):
❤️ Great!

Yajo:
Sadly this seems to have become stale. Lovely feature though. I hope somebody attend the comments and reopen in the future.

Issue #389, opened 2021-04-17
feat: make {{ _copier_conf|to_json }} work


Yajo:
Gotta remove this weird thing:

https://github.com/copier-org/copier/blob/8dd9a1ba20afba204bfba40cf71d963f461974b8/docs/creating.md#L38-L39

https://github.com/copier-org/copier/blob/8dd9a1ba20afba204bfba40cf71d963f461974b8/copier/main.py#L203

It should be JSON-dumpable without quirks.

Yajo:
Fixed in #508

PR #508, opened 2021-12-17
feat: Make the `to_json` filter support the conf object

pawamoy:
Fixes issue #389

Any idea for a test :smile:? I guess I'll just run the filter in a template.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from shutil import rmtree
 from typing import Callable, List, Mapping, Optional, Sequence
 from unicodedata import normalize
+from warnings import warn
 
 import pathspec
 from jinja2.loaders import FileSystemLoader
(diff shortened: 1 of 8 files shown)


copier/main.py
```diff
@@ -400,6 +410,11 @@ def jinja_env(self) -> SandboxedEnvironment:
                 "Make sure to install these extensions alongside Copier itself.\n"
                 "See the docs at https://copier.readthedocs.io/en/latest/configuring/#jinja_extensions"
             )
+        # patch the `to_json` filter to support Pydantic dataclasses
+        def to_json(value, *args, **kwargs):
+            return json.dumps(value, *args, **kwargs, default=pydantic_encoder)
+
+        env.filters["to_json"] = to_json
```
Yajo, line 417: I think you can simplify it with a partial:

```suggestion
        env.filters["to_json"] = partial(env.filters["to_json"], default=pydantic_encoder)
```

Yajo (commented):
It looks awesome! Thanks, this was one of the blockers for v6.

The test could be simple: just create a `whatever.json.jinja` file with `{{ _copier_conf | to_json }}" inside and check it dumps a valid json object.

copier/main.py
pawamoy: Nice, thanks!

pawamoy:
I'm afraid keeping the `.json` method for backward-compatibility is problematic: the pydantic encoder then fails because it cannot serialize the `.json` method on the conf object. Should we simply drop it? Otherwise we must derive the pydantic encoder to ignore functions (or copy the conf object and remove the `.json` attribute from the copy).

Yajo (approved):
It's not my favorite solution because it can introduce big incompatibilities with v5, but after all it's a major release, so nevermind. If it's a real problem, somebody can fix it later.

BTW this is still draft. Anything left behind?

pawamoy:
It's ready now.

PR #509, opened 2021-12-17
build: Accept Python 3.11 as requirement

pawamoy:
Also see https://iscinumpy.dev/post/bound-version-constraints/ for why I didn't cap with `<4.0` or `<3.12`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
     runs-on: ${{ matrix.os }}
     steps:
       # HACK https://github.com/actions/cache/issues/315
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
           - windows-latest
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.python-version == '3.11-dev' }}
     steps:
       # HACK https://github.com/actions/cache/issues/315
       - name: Enable msys binaries
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
     runs-on: ${{ matrix.os }}
     steps:
       # HACK https://github.com/actions/cache/issues/315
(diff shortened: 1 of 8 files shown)


pawamoy:
Ah, the failures come from apipkg, which I know fixed the issue already. That illustrate the benefit of avoiding upper bounds on dependencies :joy:

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
```
Yajo, line 34: You shouldn't set a prerelease as required. Let the CI be happy if 3.11-dev fails. See https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

Also, drop 3.6.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ copier = "copier.cli:CopierApp.run"
 "Bug Tracker" = "https://github.com/copier-org/copier/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.6.2,<3.11"
+python = ">=3.6.2"
```
Yajo, line 30: According to https://endoflife.date/python, 3.6 has 6 days left of support, and I can safely assume we won't have copier v6 before that time, so you can put 3.7 as the lower limit.

Yajo (changes requested):
Good idea, but some comments on it.

I had to go to the section "If you need a TL;DR for the TL;DR" of that link you passed (it's weekend, my friend 😆). Seems legit. 👍🏼

.github/workflows/ci.yml
pawamoy: Right!

pyproject.toml
pawamoy: OK!

README.md
```diff
@@ -21,7 +21,7 @@ A library and CLI app for rendering project templates.
 
 ## Installation
 
-1. Install Python 3.6.1 or newer (3.8 or newer if you're on Windows).
+1. Install Python 3.7 or newer (3.8 or newer if you're on Windows).
```
Yajo, line 24: This deserves something in the changelog

Yajo (changes requested):

README.md
pawamoy: Oh yes, the changelog

pawamoy:
By the way, there are two HACK/FIXME comments in the code related to Python 3.6, not sure if you'd like to fix them in this PR or another.

README.md
pawamoy: Should it go in Deprecated, Removed, Changed?

Yajo:
Feel free to do it, if CI goes green I'll merge 😄

README.md
Yajo: That's removed, right?

pawamoy:
Could you re-run the CI jobs? I'm not sure the one displayed here are the right ones.

Yajo:
It seems you're hitting https://github.com/mhammond/pywin32/issues/1716. Weird because the fix is in version 302, which is the one being installed in CI.

Well, nevermind. I guess it's normal to get these kind of failures.

We were having pending builds of 3.6 because the master branch was configured to expect those builds. I fixed its expectations with github UI and now it allows me to merge.

Do you want to try to fix the 3.11 build, or do I merge now?

pawamoy:
OK thanks! You can merge now, I'll see if I can do something for 3.11 later :)

Issue #461, opened 2021-10-20
bug: copier not deleting a file on update


joao-p-marques:
**Describe the problem**

When you have a file in the template that is rendered depending on a certain variable from copier data, it doesn't get deleted in an update. When you render the template for the first time, it works as expected, but not on an update.

**Template**

The 2 files currently in https://github.com/Tecnativa/doodba-copier-template/tree/main/odoo/custom/build.d should only be rendered if the variable `odoo_version < 11`

(The same happens with https://github.com/OCA/oca-addons-repo-template/issues/77)

**To Reproduce**

1. Run `copier -f -d odoo_version=11 copy gh:Tecnativa/doodba-copier-template .`
2. The 2 files mentioned above are not created, as expected
3. In another folder, run `copier -f -d odoo_version=10 copy gh:Tecnativa/doodba-copier-template .`
4. The 2 files are created, as expected.
5. Run `copier -f -d odoo_version=11 update`
6. The 2 files are not deleted.

**Logs**
You can see here `copier` interprets them as `identical`:
<details>

```
 identical  .
  conflict  .copier-answers.yml
 overwrite  .copier-answers.yml
 identical  LICENSE
 identical  test.yaml
  conflict  tasks.py
 overwrite  tasks.py
  conflict  setup-devel.yaml
 overwrite  setup-devel.yaml
 identical  prod.yaml
 identical  odoo
 identical  odoo/custom
 identical  odoo/custom/ssh
 identical  odoo/custom/ssh/known_hosts
 identical  odoo/custom/ssh/id_rsa.pub
 identical  odoo/custom/ssh/id_rsa
 identical  odoo/custom/ssh/config
 identical  odoo/custom/src
 identical  odoo/custom/src/repos.yaml
 identical  odoo/custom/src/private
 identical  odoo/custom/src/private/.editorconfig
 identical  odoo/custom/src/addons.yaml
 identical  odoo/custom/entrypoint.d
 identical  odoo/custom/entrypoint.d/.empty
 identical  odoo/custom/dependencies
  conflict  odoo/custom/dependencies/pip.txt
      skip  odoo/custom/dependencies/pip.txt
 identical  odoo/custom/dependencies/npm.txt
 identical  odoo/custom/dependencies/gem.txt
 identical  odoo/custom/dependencies/apt_build.txt
 identical  odoo/custom/dependencies/apt.txt
 identical  odoo/custom/conf.d
 identical  odoo/custom/conf.d/.empty
[...]

PR #520, opened 2021-12-29
feat: include local dirty changes by default. fixes #184

### LINKED ISSUE #184: Don't checkout anything by default on local templates
I'm currently translating a cookiecutter template to a copier one, and each time I change something, I have to commit the change for copier to pick it up. At first it was confusing ("why did files that I removed were copied??"), and now it's annoying (a "wip" commit for every change, and then I'll have to squash them or something).

I think it's because copier tries to checkout the latest tag? I'd like it not to do anything if I don't specify the `-r, --vcs-ref` option.

sabard:
A quick attempt at implementing the changes described in [this comment](https://github.com/copier-org/copier/issues/184#issuecomment-861639395).
 - if a template is vcs-tracked and dirty, the changes are committed automatically with a "wip" commit and propagated to the subproject
 - user is warned in this scenario
 - tests for copy and update on dirty templates

I made this the default behavior, but would be happy to add a flag for this behavior and maintain the current default.

Also, think this will have to rebase off of #519 to pass CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,8 @@ Summary:
 -   Raise a CopierAnswersInterrupt instead of a bare KeyboardInterrupt to provide
     callers with additional context - such as the partially completed AnswersMap.
 -   Support for `user_defaults`, which take precedence over template defaults.
--   Copies dirty changes from template to subdirectory by default.
+-   Copy dirty changes from a git-tracked template to the project by default, to make
+    testing easier.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ Summary:
 -   Raise a CopierAnswersInterrupt instead of a bare KeyboardInterrupt to provide
     callers with additional context - such as the partially completed AnswersMap.
 -   Support for `user_defaults`, which take precedence over template defaults.
--   Copies dirty changes from template to subdirectory by defaultl.
+-   Copies dirty changes from template to subdirectory by default.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Summary:
 -   Raise a CopierAnswersInterrupt instead of a bare KeyboardInterrupt to provide
     callers with additional context - such as the partially completed AnswersMap.
 -   Support for `user_defaults`, which take precedence over template defaults.
+-   Copies dirty changes from template to subdirectory by defaultl.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -42,12 +42,12 @@ To create a template:
 ```yaml title="copier.yml"
 # questions
 project_name:
-    type: str
-    help: What is your project name?
+  type: str
+  help: What is your project name?
 
 module_name:
-    type: str
-    help: What is your Python module name?
+  type: str
+  help: What is your Python module name?
 ```
 
 ```python+jinja title="{{project_name}}/{{module_name}}.py.jinja"
--- a/README.md
+++ b/README.md
@@ -42,12 +42,12 @@ To create a template:
 ```yaml title="copier.yml"
 # questions
 project_name:
-  type: str
-  help: What is your project name?
+    type: str
+    help: What is your project name?
 
 module_name:
-  type: str
-  help: What is your Python module name?
+    type: str
+    help: What is your Python module name?
 ```
 
 ```python+jinja title="{{project_name}}/{{module_name}}.py.jinja"
--- a/copier/main.py
+++ b/copier/main.py
@@ -1,4 +1,5 @@
 """Main functions and classes, used to generate or update projects."""
+
 import platform
 import subprocess
 import sys
@@ -676,7 +677,6 @@ class Worker:
                         file=sys.stderr,
                     )
                     diff = diff_cmd("--inter-hunk-context=0")
-                print(diff)
         # Run pre-migration tasks
         self._execute_tasks(
             self.template.migration_tasks("before", self.subproject.template)
--- a/copier/main.py
+++ b/copier/main.py
@@ -1,5 +1,4 @@
 """Main functions and classes, used to generate or update projects."""
-
 import platform
 import subprocess
 import sys
@@ -677,6 +676,7 @@ class Worker:
                         file=sys.stderr,
                     )
                     diff = diff_cmd("--inter-hunk-context=0")
+                print(diff)
         # Run pre-migration tasks
         self._execute_tasks(
             self.template.migration_tasks("before", self.subproject.template)
--- a/copier/template.py
+++ b/copier/template.py
@@ -17,7 +17,6 @@ from pydantic.dataclasses import dataclass
 from yamlinclude import YamlIncludeConstructor
 
 from .errors import (
-    DirtyLocalWarning,
     InvalidConfigFileError,
     MultipleConfigFilesError,
     OldTemplateWarning,
--- a/copier/template.py
+++ b/copier/template.py
@@ -426,11 +426,17 @@ class Template:
     def local_abspath(self) -> Path:
         """Get the absolute path to the template on disk.
 
-        This may clone it if `url` points to a
-        VCS-tracked template.
+        This may clone it if `url` points to a VCS-tracked template.
+        Dirty changes for local VCS-tracked templates will be copied.
         """
         result = Path(self.url)
         if self.vcs == "git":
+            if result.is_dir():
+                with local.cwd(result):
+                    if bool(git("status", "--porcelain").strip()):
+                        git("add", ".")
+                        git("commit", "-am", "wip")
+                        # warning msg to user?!
             result = Path(clone(self.url_expanded, self.ref))
             if self.ref is None:
                 checkout_latest_tag(result, self.use_prereleases)
--- a/copier/template.py
+++ b/copier/template.py
@@ -17,6 +17,7 @@ from pydantic.dataclasses import dataclass
 from yamlinclude import YamlIncludeConstructor
 
 from .errors import (
+    DirtyLocalWarning,
     InvalidConfigFileError,
     MultipleConfigFilesError,
     OldTemplateWarning,
--- a/copier/template.py
+++ b/copier/template.py
@@ -431,12 +431,6 @@ class Template:
         """
         result = Path(self.url)
         if self.vcs == "git":
-            if result.is_dir():
-                with local.cwd(result):
-                    if bool(git("status", "--porcelain").strip()):
-                        git("add", ".")
-                        git("commit", "-am", "wip")
-                        # warning msg to user?!
             result = Path(clone(self.url_expanded, self.ref))
             if self.ref is None:
                 checkout_latest_tag(result, self.use_prereleases)
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -111,6 +111,30 @@ def checkout_latest_tag(local_repo: StrOrPath, use_prereleases: OptBool = False)
         return latest_tag
 
 
+def _perform_clone(src: str, dst: str, ref: OptStr = None) -> str:
+    """Clone from src into dst.
+
+    Args:
+        src:
+            Git-parseable repo clone source.
+        dst:
+            Git-parseable repo clone destination.
+        ref:
+            Reference to checkout. For Git repos, defaults to `HEAD`
+    """
+    _clone = git["clone", "--no-checkout", src, dst]
+    # Faster clones if possible
+    if GIT_VERSION >= Version("2.27"):
+        _clone = _clone["--filter=blob:none"]
+    _clone()
+
+    with local.cwd(dst):
+        git("checkout", ref or "HEAD")
+        git("submodule", "update", "--checkout", "--init", "--recursive", "--force")
+
+    return dst
+
+
 def clone(url: str, ref: OptStr = None) -> str:
     """Clone repo into some temporary destination.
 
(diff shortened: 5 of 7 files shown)


copier/template.py
```diff
@@ -426,11 +427,20 @@ def templates_suffix(self) -> str:
...
-        VCS-tracked template.
+        This may clone it if `url` points to a VCS-tracked template.
+        Dirty changes for local VCS-tracked templates will be copied.
         """
         result = Path(self.url)
         if self.vcs == "git":
+            if result.is_dir():
+                with local.cwd(result):
+                    if bool(git("status", "--porcelain").strip()):
+                        git("add", ".")
```
pawamoy, line 438: Do we want `-A` as well? Without it, removed files would not be added, leaving some dirtiness.

copier/template.py
```diff
@@ -426,11 +427,20 @@ def templates_suffix(self) -> str:
...
+                        git("commit", "-am", "wip")
```
pawamoy, line 439: If everything was added by `git add`, `-a` is not necessary.

tests/test_dirty_local.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # test updating a file
+        with open("aaaa.txt", "a") as f:
+            f.write("dolor sit amet")
+
+    # dst must be vcs-tracked to use run_update
+    with local.cwd(dst):
+        git("init")
+        git("add", ".")
+        git("commit", "-am", "first commit on dst")
```
pawamoy, line 74: Same two remarks as above (+`-A` and -`-a`)

pawamoy:
So, just to be clear: this change creates the wip commits in the temporary copy/clone, right? Not in the original template repository?

copier/template.py
sabard: Yep, good catch. Although removed files are currently not staged anyway during update—looks like the same issue as in #461. I took a stab fixing it by removing files that were present in `dst` and not `src`, but this is difficult when you factor in templating and it was causing errors in other tests.

sabard:
I originally was creating the commit in the original template repository because that was easier, but I agree that it is too intrusive. 

The current code modifies `vcs::clone` instead of `template::local_abspath` and copies a dirty repo into a temporary directory so that it can add, commit, and clone.

copier/vcs.py
```diff
@@ -111,15 +114,36 @@ def checkout_latest_tag(local_repo: StrOrPath, use_prereleases: OptBool = False)
...
             Reference to checkout. For Git repos, defaults to `HEAD`.
     """
+
+    src = url
+    if Path(src).is_dir():
+        is_dirty = False
+        with local.cwd(url):
+            is_dirty = bool(git("status", "--porcelain").strip())
+        if is_dirty:
+            src = tempfile.mkdtemp(prefix=f"{__name__}.dirty.")
```
pawamoy, line 134: The temporary directory should probably be removed once we're done using it. Maybe by using a `with` statement and rearranging the code so that `_clone` is inside?

pawamoy:
Docs failure is because mkdocs-autorefs 0.3.1 was installed, while mkdocstrings 0.17.0 was not. I should probably have tagged mkdocs-autorefs 0.4.0 to prevent that. Or better yet, upper bounds could be removed on Copier's dependencies 😛

copier/vcs.py
sabard: great, fixed and moved the duplicate `_clone` logic to `_perform_clone`

sabard:
Got it—should I just add `mkdocs-autorefs = "^0.4.0"`  to `pyproject.toml`? I always use pip-tools and not too familiar with poetry.

pawamoy:
I wasn't clear: *mkdocstrings* should be updated (up to 0.17.0), not mkdocs-autorefs. But it should probably be done in another PR.

sabard:
I saw this was fixed in #518, so I just rebased instead.

pawamoy:
Oh yes, good catch.

tests/test_dirty_local.py
```diff
@@ -0,0 +1,107 @@
...
+
+    generated = (dst / "pyproject.toml").read_text()
+    control = (Path(__file__).parent / "reference_files/pyproject.toml").read_text()
+    assert generated == control
+
+    # assert template still dirty
+    assert bool(git("status", "--porcelain").strip())
+
+
+# Will fail due to lingering deleted file until #461 is fixed
```
Yajo, line 35: I use the keyword HACK to track... well, hacks 😄 

Also TODO to make clear what to do after the HACK is solved.

```suggestion
# Will fail due to lingering deleted file.
# HACK https://github.com/copier-org/copier/issues/461
# TODO Remove xfail decorator when fixed.
```

Yajo (changes requested):
Hi @sabard! Welcome to Copier 😊 

This PR is very nice already. There are some improvements to do below, but overall it's almost ready to merge.

Thank you very much for your contribution!

copier/main.py
```diff
@@ -677,6 +676,7 @@ def run_update(self) -> None:
                         file=sys.stderr,
                     )
                     diff = diff_cmd("--inter-hunk-context=0")
+                print(diff)
```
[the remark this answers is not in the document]
sabard: yes, thanks for catching this!

copier/vcs.py
```diff
@@ -108,23 +111,61 @@ def checkout_latest_tag(local_repo: StrOrPath, use_prereleases: OptBool = False)
...
+    # Faster clones if possible
+    if GIT_VERSION >= Version("2.27"):
+        _clone = _clone["--filter=blob:none"]
+    _clone()
+
+    with local.cwd(dst):
+        git("checkout", ref or "HEAD")
+        git("submodule", "update", "--checkout", "--init", "--recursive", "--force")
+
+    return dst
```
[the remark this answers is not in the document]
sabard: think this is needed so that `clone` returns `location` as is used [here](https://github.com/copier-org/copier/blob/3fe1f42728a4d3ce40b564d63420764f43d98ef1/copier/template.py#L434). I could also change `clone` to call `_perform_clone` and separately return `location`—lmk what you prefer

README.md
```diff
@@ -42,12 +42,12 @@ To create a template:
...
-  type: str
-  help: What is your project name?
+    type: str
+    help: What is your project name?
 
 module_name:
-  type: str
-  help: What is your Python module name?
+    type: str
+    help: What is your Python module name?
```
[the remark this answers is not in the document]
sabard: this was happening automatically from `poe lint`. will see if i can commit without it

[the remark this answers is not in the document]
sabard: was only able to get this pushed out with a `--no-verify`. are there any local options for prettier i need to set to make this work?

sabard:
Removed Python 3.7 incompatible `dirs_exist_ok` option in `shutil.copytree` and fixed a bug in `test_dirty_local.py` causing CI to error.

Should be ready for another CI run.

README.md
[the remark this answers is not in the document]
sabard: this started working again all of a sudden—not sure what changed

tests/test_dirty_local.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # assert template still dirty
+    with local.cwd(src):
+        assert bool(git("status", "--porcelain").strip())
+
+
+# Will fail due to lingering deleted file.
+# HACK https://github.com/copier-org/copier/issues/461
+# TODO Remove xfail decorator when fixed.
+@pytest.mark.xfail
```
Yajo, line 46:
```suggestion
@pytest.mark.xfail(strict=True)
```

This way we won't forget to do what the comment says.

Yajo (changes requested):

Yajo:
Please also add a note in the changelog.

copier/vcs.py
```diff
@@ -108,23 +112,73 @@ def checkout_latest_tag(local_repo: StrOrPath, use_prereleases: OptBool = False)
...
+            url_abspath = Path(url).absolute()
+            with TemporaryDirectory(prefix=f"{__name__}.dirty.") as src:
+                with local.cwd(src):
+                    for item in os.listdir(url_abspath):
+                        s = os.path.join(url_abspath, item)
+                        d = os.path.join(src, item)
+                        if os.path.isdir(s):
+                            copytree(s, d)
+                        else:
+                            copy2(s, d)
```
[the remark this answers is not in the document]
sabard: This turned out to be not so simple for a few reasons:
- `git diff` does not, by default, give a diff of the untracked files (of interested here doing a dirty copy)
- `HEAD` might not exist (e.g., `git init` was just run)

That said, I was able to get something closer to what you are asking for working by using `git diff --no-index` and then excluding `*/.git/*` on the apply. This will of course copy over other unwanted files to the temporary clone directory, but they shouldn't make it into the subdirectory since they will be in the clone's .gitignore. I think this is arguably just as inefficient in the worst-case (large .gitignored file getting needlessly copied).

I tried getting something like ` git --git-dir=<clone_dst>/.git --work-tree=<template_src> diff HEAD` [SO source](https://stackoverflow.com/a/32840462/916969) to work to no avail. Seems that anytime `git-diff` is used without `--no-index`, it ignores untracked files.

Need to make a couple fixes for CI still, but pushed latest changes to avoid losing progress.

sabard:
The most recent commit should probably pass CI. However:

I'm seeing `error: git diff header lacks filename information when removing 1 leading pathname component` from the `test_worker_config_precedence` test (attempting to copy the copier template itself). This is only triggered when the working tree is dirty, of course, but would appreciate any help figuring out how to fix this. Should be able to reproduce the error by checking out the latest commit here, making a change, and running the tests.

sabard:
After some thinking, I figured it would be a better solution to create a new `.git` dir in a temporary location, but point it to the files in the template dir. This lets us add and commit the dirty (and untracked!) changes while avoiding the extra copy. Please see the latest code and let me know if you have any questions.

copier/vcs.py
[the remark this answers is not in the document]
Yajo: AFAIK `git stash` works both with staged and untracked files. Have you tried with it?

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ Summary:
 -   Raise a CopierAnswersInterrupt instead of a bare KeyboardInterrupt to provide
     callers with additional context - such as the partially completed AnswersMap.
 -   Support for `user_defaults`, which take precedence over template defaults.
+-   Copies dirty changes from template to subdirectory by default.
```
Yajo, line 34:
```suggestion
-   Copy dirty changes from a git-tracked template to the project by default, to make testing easier.
```

tests/test_dirty_local.py
```diff
@@ -0,0 +1,117 @@
...
+            copy2(item_src_path, item_dst_path)
+
+    with local.cwd(src):
+        git("init")
+
+    with pytest.warns(DirtyLocalWarning):
+        copier.copy(str(src), str(dst), data=DATA, quiet=True)
+
+    generated = (dst / "pyproject.toml").read_text()
+    control = (Path(__file__).parent / "reference_files/pyproject.toml").read_text()
```
Yajo, line 35:
```suggestion
    control = (Path(__file__).parent / "reference_files" / "pyproject.toml").read_text()
```

copier/vcs.py
```diff
@@ -108,23 +111,68 @@ def checkout_latest_tag(local_repo: StrOrPath, use_prereleases: OptBool = False)
...
+        is_dirty = False
+        with local.cwd(url):
+            is_dirty = bool(git("status", "--porcelain").strip())
+        if is_dirty:
+            url_abspath = Path(url).absolute()
+            with TemporaryDirectory(prefix=f"{__name__}.dirty.") as src:
+                with local.cwd(src):
+                    git("--git-dir=.git", f"--work-tree={url_abspath}", "init")
+                    git("--git-dir=.git", f"--work-tree={url_abspath}", "add", "-A")
+                    git(
```
Yajo, line 164: This is a smart solution!

Have you tried this?:

1. Clone.
2. Do this same hack, but on the cloned directory.

sabard: This is possible, but cloning a dirty git-tracked template seems more error prone to me. For example, it may not have a HEAD. I believe the submodule update would also have to be avoided and then called after the hack.

Yajo: Ok, but in that case I still have some doubts about if this will work fine for updates, because if I'm understanding the code correctly, this is not copying the git history correctly, so it wouldn't work as expected for updates. So please could you update the test? (See below)

Note: git has no problem in cloning empty repos AFAIK.

tests/test_dirty_local.py
```diff
@@ -0,0 +1,117 @@
...
+            / "copier.yml": """\
+                _envops:
+                    "keep_trailing_newline": True
+            """,
+            src
+            / "aaaa.txt": """
+                Lorem ipsum
+            """,
+            src
+            / "to_delete.txt": """
```
Yajo, line 67: Since the failure from this test is in this file, then please remove this part of the test and remove the xfail decorator. If the algorithm is fine, the test will pass. Later, when #461 is fixed, it will include the needed tests.

sabard: kept file, but removed the assert. added a node referencing the linked issue

copier/vcs.py
sabard: yeah, you're totally right. had to move things around a little bit to account for submodules, but this looks to be working now after cloning first

copier/vcs.py
```diff
@@ -150,29 +126,36 @@ def clone(url: str, ref: OptStr = None) -> str:
...
-                    )
-                    warn(
-                        "Dirty template changes included automatically.",
-                        DirtyLocalWarning,
-                    )
-                    return _perform_clone(src, location, ref)
-
-    return _perform_clone(url, location, ref)
+            with local.cwd(location):
+                # git("--git-dir=.git", f"--work-tree={url_abspath}", "init")
```
Yajo, line 142: Please remove this line.

copier/vcs.py
```diff
@@ -150,29 +126,36 @@ def clone(url: str, ref: OptStr = None) -> str:
...
+                    "--no-verify",
+                )
+                warn(
+                    "Dirty template changes included automatically.",
+                    DirtyLocalWarning,
+                )
+
+    with local.cwd(location):
+        git("checkout", ref or "HEAD")
+        git("submodule", "update", "--checkout", "--init", "--recursive", "--force")
```
Yajo, line 159: I've got the feeling that this block should be done before importing the ditty changes. However, if the test goes green I guess I'm just wrong.

Yajo (commented):
Looks much nicer now!

Just remove that leftover line and one question to merge!

Thanks again, great job!

sabard:
Any tips for figuring out what's wrong with CI on windows? I cloned and ran the tests on my windows box and they passed, so having trouble replicating to get more info on the error.

copier/vcs.py
sabard: I think it shouldn't matter if it runs before or after. But if there is no HEAD in the case of a recently intialized repo, then this fails and would need to be called after as well. I also realized that specifying `ref` means that dirty changes won't be considered and so added that to the dirty checking logic.

tests/test_dirty_local.py
```diff
@@ -0,0 +1,116 @@
...
+    with pytest.warns(DirtyLocalWarning):
+        run_update(dst, defaults=True, overwrite=True)
+
+    # make sure changes propagate after update
+    assert os.path.exists(dst / "test_file.txt")
+
+    p1 = str(src / "aaaa.txt")
+    p2 = str(dst / "aaaa.txt")
+    filecmp.clear_cache()
+    assert filecmp.cmp(p1, p2)
```
Yajo, line 112: Convert these to Path and do: 

```suggestion
    assert p1.read_text() == p2.read_text()
```

It should fail equally, but the error will be more descriptive.

Yajo (approved):
Looks good! Just make it green to merge. 

Thanks.

tests/test_dirty_local.py
Yajo: It seems you were hitting some weird scenario with fliecmp.cmp and it was ok.

Yajo (approved):

PR #544, opened 2022-01-15
docs: Change diagram colors when switching between dark and light themes

### LINKED ISSUE #541: bug: Diagram not displaying correctly in dark mode in latest documentation
**Describe the problem**
In the latest documentation (not in the stable one) it is observed that the arrows of the update diagram are not visible in the dark mode. See https://copier.readthedocs.io/en/latest/updating/#how-the-update-works.

**Expected behavior**

Arrows must be visible on all themes. 

**Screenshots/screencasts/logs**

dark mode:

Screenshot_20220113_211119

light mode:

Screenshot_20220113_211106

pawamoy:
Fixes #541 

Not as pretty as the Material for MkDocs Insiders integration, but I guess users don't usually switch theme that often, even less so while looking at a diagram, so I think that is totally acceptable (also, glitches in the GIF just come from my way of recording, using peek).

mermaid_theme_switch

### DIFF
--- a/docs/js/refresh_on_theme_switch.js
+++ b/docs/js/refresh_on_theme_switch.js
@@ -1,0 +1,10 @@
+var paletteSwitcher1 = document.getElementById("__palette_1");
+var paletteSwitcher2 = document.getElementById("__palette_2");
+
+paletteSwitcher1.addEventListener("change", function () {
+    location.reload();
+});
+
+paletteSwitcher2.addEventListener("change", function () {
+    location.reload();
+});
--- a/docs/js/refresh_on_theme_switch.js
+++ b/docs/js/refresh_on_theme_switch.js
@@ -1,6 +1,6 @@
-const ref = document.querySelector("[data-md-component=palette]")
-component$.subscribe(component => {
+const ref = document.querySelector("[data-md-component=palette]");
+component$.subscribe((component) => {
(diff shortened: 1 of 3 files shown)


eyllanesc:
@pawamoy As pointed out in [this discussion](https://github.com/squidfunk/mkdocs-material/discussions/3429#discussioncomment-1910541) you could use: 

```js
const ref = document.querySelector("[data-md-component=palette]")
component$.subscribe(component => {
  if (component.ref === ref) {
    alert$.next(`Changed to "${component.color.scheme}"`)
    console.log(component.index, component.color)
  }
})
```

pawamoy:
You mean, inside the extra JS file and instead of the `theme` callback in mkdocs.yml? Feel free to suggest updates through the GitHub review system :slightly_smiling_face:

eyllanesc:
When your PR is accepted I will try to improve the logic a bit. My idea is to use the public API of mkdocs-material so that if there is a change in the future, then the logic will not be broken so easily.

pawamoy:
OK so the callback in mkdocs.yml is still needed. But the JS code is cleaner.

eyllanesc:
@pawamoy  It seems that the pre-commit requires that you format the JavaScript code

Yajo (approved):

PR #553, opened 2022-01-28
Update configuring.md "Conditional files" section

flippedAben:
Using the syntax stated in the current documentation, I get this error when running `copier`:

```
jinja2.exceptions.TemplateSyntaxError: expected token 'end of print statement', got <variable>
```

Using `{%` instead of `{{`, it works as expected.


Yajo (approved):
Good catch, thank you!

PR #557, opened 2022-02-03
Print header specify template version being used.

andrewleech:
This PR simply adds a header message upon running copier to indicate to the user the template version being copied / updated from, eg.
```
~/copier$ python -m copier --force ~/my_template ~/new_project

Copying from template version v0.1.1
 identical  .
 identical  update.py
 identical  src
 identical  src/unix
...
```

I'm a relatively new user of copier and did not realise that its normal behavior is to always use the latest tag of a template repo, rather than the current master.

I found this particularly confusing when working from a local clone of a template under development. I'd make changes to the template, then re-run copier to check they all worked. Sometimes copier used my new changes, other times it seemed to copy really old files instead. It took me far longer than you'd think to figure out that when there were local changes it would use current head (as printed in the warning message) but once the template repo is clean it would ignore the current commit checked out in the template repo and skip back to an old tag.

I can see how this tag / release behaviour would often be sensible / desiable when pulling from a url, and there's the --vcs-ref arg available to override that. Once this had all clicked in my head it did make sense.

While there's a warning header message that shows when it is using the current dirty changes, there's currently no indication of what version / commit of the template is being used when the template repo is clean.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -585,11 +585,14 @@ class Worker:
         See [generating a project][generating-a-project].
         """
         was_existing = self.subproject.local_abspath.exists()
-        if not self.quiet:
-            # TODO Unify printing tools
-            print("")  # padding space
         src_abspath = self.template_copy_root
         try:
+            if not self.quiet:
+                # TODO Unify printing tools
+                print(
+                    f"\nCopying from template version {self.template.version}",
+                    file=sys.stderr,
+                )
             self._render_folder(src_abspath)
             if not self.quiet:
                 # TODO Unify printing tools
@@ -642,6 +645,11 @@ class Worker:
             raise UserMessageError(
                 f"Your are downgrading from {self.subproject.template.version} to {self.template.version}. "
                 "Downgrades are not supported."
+            )
+        if not self.quiet:
+            # TODO Unify printing tools
+            print(
+                f"Updating to template version {self.template.version}", file=sys.stderr
             )
         # Copy old template into a temporary destination
         with TemporaryDirectory(prefix=f"{__name__}.update_diff.") as dst_temp:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Summary:
 -   Support for `user_defaults`, which take precedence over template defaults.
 -   Copy dirty changes from a git-tracked template to the project by default, to make
     testing easier.
+-   Advertise clearly which version is being copied or updated in the CLI.
 
 ### Changed
 
--- a/docs/generating.md
+++ b/docs/generating.md
@@ -32,6 +32,32 @@ generating the project.
 
 [All the available options][copier.cli] are described with the `--help-all` option.
 
+## Templates versions
+
+By default, copier will copy from the last release found in template git tags, sorted as
+[PEP 440][], regardless of whether the template is from a URL or a local clone of a git
+repository.
+
+The exception to this is if you use a local clone of a template repository that has had
+any modifications made, in this case Copier will use this modified working copy of the
+template to aid development of new template features.
+
+If you would like to override the version of template being installed, the
+[`--vcs-ref`](../configuring/#vcs_ref) argument can be used to specify a branch, tag or
+other reference to use.
+
+For example to use the latest master branch from a public repository:
+
+```
+copier --vcs-ref master https://github.com/foo/copier-template.git ./path/to/destination
+```
+
+Or to work from the current checked out revision of a local template:
+
+```
+copier --vcs-ref HEAD path/to/project/template path/to/destination
+```
+
 ## Regenerating a project
 
 When you execute `copier copy $template $project` again over a preexisting `$project`,


Yajo:
Thanks for the explanation and the PR! It makes total sense.

I'll comment on code later when I have more time. For the time being, have you considered adding a note somewhere in the docs, so the next newbie doesn't have to dazzle around to grasp this concept?

andrewleech:
@Yajo the documentation is a good idea, I've added a section to readme in a second commit in the PR, let me know what you think of it.

README.md
```diff
@@ -122,6 +122,28 @@ Templates have these goals:
 Copier tries to have a smooth learning curve that lets you create simple templates that
 can evolve into complex ones as needed.
 
+## Templates versions
+
+Copier generally uses the most recent release / tag of a template if the template 
+is from either a url or a local clone of a git repository.
```
pawamoy, line 128:
```suggestion
is from either a URL or a local clone of a git repository. 
```

README.md
```diff
@@ -122,6 +122,28 @@ Templates have these goals:
 Copier tries to have a smooth learning curve that lets you create simple templates that
 can evolve into complex ones as needed.
 
+## Templates versions
+
+Copier generally uses the most recent release / tag of a template if the template 
+is from either a url or a local clone of a git repository. 
+
+The exception to this rule is if you use a local clone of a template repository that has 
+had any modifications made, in this case copier will use this modified working copy of
```
pawamoy, line 131:
```suggestion
had any modifications made, in this case Copier will use this modified working copy of 
```

README.md
```diff
@@ -122,6 +122,28 @@ Templates have these goals:
...
+## Templates versions
+
+Copier generally uses the most recent release / tag of a template if the template 
+is from either a url or a local clone of a git repository. 
+
+The exception to this rule is if you use a local clone of a template repository that has 
+had any modifications made, in this case copier will use this modified working copy of 
+the template to aid development of new template features.
+
+If you would like to override the version of template being installed the `--vcs-ref`
```
pawamoy, line 134:
```suggestion
If you would like to override the version of template being installed, the `--vcs-ref` 
```

README.md
```diff
@@ -122,6 +122,28 @@ Templates have these goals:
 Copier tries to have a smooth learning curve that lets you create simple templates that
 can evolve into complex ones as needed.
 
+## Templates versions
```
Yajo, line 125: These docs look great, although I don't think this subject is so important as to document it directly in the README. It's more an implementation/usage detail that is only useful when you start using copier more deeply.

There are other places where this would fit better:

- https://copier.readthedocs.io/en/latest/generating/
- https://copier.readthedocs.io/en/latest/updating/
- https://copier.readthedocs.io/en/latest/configuring/#vcs_ref (BTW, here it mentions something related):

  > By default, copier will copy from the last release found in template git tags, sorted as PEP 440.

Please pick the one you think fits best and move these docs there.

copier/main.py
```diff
@@ -590,6 +590,7 @@ def run_copy(self) -> None:
             print("")  # padding space
         src_abspath = self.template_copy_root
         try:
+            print(f"Copying from template version {self.template.commit}")
```
Yajo, line 593: Don't you think the version (after processing to the PEP440 equivalent) would be more helpful?

Besides, please print to STDERR:

```suggestion
            print(f"Copying from template version {self.template.version}", file=sys.stderr)
```

copier/main.py
```diff
@@ -643,6 +644,7 @@ def run_update(self) -> None:
                 f"Your are downgrading from {self.subproject.template.version} to {self.template.version}. "
                 "Downgrades are not supported."
             )
+        print(f"Updating from template version {self.template.commit}")
```
Yajo, line 647: Same as above, plus removing that "from" which I think is a bit misleading:

```suggestion
        print(f"Updating to template version {self.template.version}", file=sys.stderr)
```

Yajo (changes requested):
Finally I got time to review.

Thanks a lot for the contribution, looks quite helpful indeed.

Some comments though.

copier/main.py
andrewleech: Ah, I hadn't seen that attribute, it probably does make more sense. I'm not sure what that contains when copying from a modified commit, or from a vcs-ref like HEAD, I'll double check them myself.

README.md
andrewleech: This versioning behavior tripped up both myself and a couple of my colleages on unrelated occasions before I figured out what was going on. 
It seemed particularly confusing when following an example of using a local clone of a template - the assumed behavior is that the template that's checked out in the specified folder would be used as-is, rather than copier using something from its git history.
As such I feel it does warrant being in a fairly prominent place. 

That being said, perhaps the "Generating a project" docs page just after the first initial usage section makes sense. 
Also, with the print() noting the version used when running, that should inform users quickly enough to avoid future confusion.

To be honest I don't think I'd read any of those docs previously... the readme felt like it had enough information really. When I started to get confused by it I really should have read through the docs, maybe I might have found that note on the `--vcs-ref` arg.

Yajo: The 1st part you say was #184 and got fixed recently. I'm not sure if it was released already though.



> That being said, perhaps the "Generating a project" docs page just after the first initial usage section makes sense.
> Also, with the print() noting the version used when running, that should inform users quickly enough to avoid future confusion.

I agree. Would you please move this there then please?

Yajo: Oh sorry, I didn't notice you already did it.

Yajo (changes requested):
Thanks!

This PR looks fantastic as it is. However, pre-commit seems to be telling that code is not following our standards.

The good part is that most of those standards are automated, so all you have to do is:

```sh
pre-commit run -a
git commit -a --amend
git push -f
```

Check out https://copier.readthedocs.io/en/latest/contributing/#dev-environment-setup to know the recommended dev env setup.

andrewleech:
Thanks I'll fix that. I've been meaning to get into using precommit hooks for this sort of thing on my projects...

Yajo (approved):

Yajo (changes requested):
Our tests detected that this feature is not respecting the `quiet` flag. Can you fix that please?

andrewleech:
Ah, that makes sense!

Yajo (approved):

Yajo:
It seems you forgot the pre-commit thing again.

Did you know you can do `pre-commit install` and it will check always automatically before doing the commit?

In any case, can you please follow the steps from https://github.com/copier-org/copier/pull/557#pullrequestreview-878025875 again?

andrewleech:
Ah geez, yep sorry for the noise! I had installed pre-commit but I did the last fix-up on a different computer and forgot. 
I'd even shared your dev docs on pre-commit usage with a colleage as a fantastic example of best practice we should aim to replicate!

PR #561, opened 2022-02-08
test: fix date comparison not working in my timezone

dstanek:
The now() template builtin is being used in the test and is an alias
for datetime.utcnow(). The now however, just uses datetime.now(). This
means that my datetime.now() will always make this test fail.

Signed-off-by: David Stanek <dstanek@dstanek.com>

### DIFF
--- a/tests/test_templated_prompt.py
+++ b/tests/test_templated_prompt.py
@@ -217,7 +217,7 @@ def test_templated_prompt_builtins(tmp_path_factory):
     )
     Worker(str(src), dst, defaults=True, overwrite=True).run_copy()
     that_now = datetime.fromisoformat((dst / "now").read_text())
-    assert that_now <= datetime.now()
+    assert that_now <= datetime.utcnow()
     assert len((dst / "make_secret").read_text()) == 128
 
 


Yajo (approved):

copier/tools.py
```diff
@@ -172,3 +173,56 @@ def cleanup(self):
...
+            colorama.Fore.YELLOW, "diff unavailable: unable to determine encoding"
+        )
+        return
+    try:
+        new_lines = get_lines(new_content)
+    except Exception:
+        color_print(
+            colorama.Fore.YELLOW, "diff unavailable: unable to determine encoding"
+        )
+        return
```
pawamoy, line 209: I know some flake8 plugins would say: "only one line in a try block", but this could probably be factorized to this:

```suggestion
    try:
        original_lines = get_lines(original_content)
        new_lines = get_lines(new_content)
    except Exception:
        color_print(
            colorama.Fore.YELLOW, "diff unavailable: unable to determine encoding"
        )
        return
```

pawamoy: Also, is there a reason you catch `Exception` rather than the more specific `UnicodeDecodeError`?

Yajo:
It seems you have pushed here the commits for #562. Can you remove them please?

pawamoy:
Ah!! I was wondering where my comments went! I just commented on the wrong PR (where the commits shouldn't be haha).

PR #562, opened 2022-02-08
feat: add ability to show unified diffs for changed files

dstanek:
Signed-off-by: David Stanek <dstanek@dstanek.com>

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -12,7 +12,6 @@ from shutil import rmtree
 from typing import Callable, Iterable, List, Mapping, Optional, Sequence
 from unicodedata import normalize
 
-import colorama
 import pathspec
 from jinja2.loaders import FileSystemLoader
 from jinja2.sandbox import SandboxedEnvironment
@@ -27,7 +26,7 @@ from questionary import unsafe_prompt
 from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
 from .subproject import Subproject
 from .template import Template
-from .tools import Style, TemporaryDirectory, color_print, is_binary, print_diff, printf
+from .tools import Style, TemporaryDirectory, is_binary, print_diff, printf
 from .types import (
     AnyByStrDict,
     JSONSerializable,
@@ -263,7 +262,7 @@ class Worker:
             return True
         if self.diff and original_content and new_content:
             if is_binary(original_content) or is_binary(new_content):
-                color_print(colorama.Fore.YELLOW, "    diff not availalbe for binaries")
+                print(colors.warn | "    diff not availalbe for binaries")
             else:
                 print_diff(dst_relpath, original_content, new_content)
         return bool(ask(f" Overwrite {dst_relpath}?", default=True))
--- a/copier/main.py
+++ b/copier/main.py
@@ -12,6 +12,7 @@ from shutil import rmtree
 from typing import Callable, Iterable, List, Mapping, Optional, Sequence
 from unicodedata import normalize
 
+import colorama
 import pathspec
 from jinja2.loaders import FileSystemLoader
 from jinja2.sandbox import SandboxedEnvironment
@@ -26,7 +27,7 @@ from questionary import unsafe_prompt
 from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
 from .subproject import Subproject
 from .template import Template
-from .tools import Style, TemporaryDirectory, printf
+from .tools import Style, TemporaryDirectory, color_print, is_binary, print_diff, printf
 from .types import (
     AnyByStrDict,
     JSONSerializable,
@@ -128,6 +129,12 @@ class Worker:
             When `True`, disable all output.
 
             See [quiet][].
+
+        diff:
+            When `True` show a unified diff for changed files.
+
+            See [diff][].
+
     """
 
     src_path: Optional[str] = None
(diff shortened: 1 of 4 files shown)


dstanek:
I'm still working on this, but I wanted to get the idea out there.

copier/tools.py
```diff
@@ -172,3 +173,56 @@ def cleanup(self):
...
+        lineterm="",
+    ):
+        if line.startswith("---") or line.startswith("+++"):
+            color_print(colorama.Fore.YELLOW, f"{indent}{line}")
+        elif line.startswith("@@"):
+            color_print(colorama.Fore.MAGENTA, f"{indent}{line}")
+        elif line.startswith("-"):
+            color_print(colorama.Fore.RED, f"{indent}{line}")
+        elif line.startswith("+"):
+            color_print(colorama.Fore.GREEN, f"{indent}{line}")
```
Yajo, line 226: https://python-prompt-toolkit.readthedocs.io/en/master/pages/printing_text.html#pygments-token-text-tuples with a diff lexer should save a lot of code around here.

copier/tools.py
```diff
@@ -172,3 +173,56 @@ def cleanup(self):
...
+    return content.decode("utf-8").split(os.linesep)
+
+
+def is_binary(content: bytes) -> bool:
+    """Return a boolean indicating if the content is a binary file."""
+    return b"\x00" in content or b"\xff" in content
+
+
+def color_print(color: int, msg: str):
+    print(f"{color}{msg}{colorama.Fore.RESET}")
```
Yajo, line 192: You have https://python-prompt-toolkit.readthedocs.io/en/master/pages/printing_text.html#html and https://plumbum.readthedocs.io/en/latest/colors.html for printing colorized output. You probably won't need this function.

copier/main.py
```diff
@@ -144,6 +151,7 @@ class Worker:
     overwrite: bool = False
     pretend: bool = False
     quiet: bool = False
+    diff: bool = False
```
Yajo, line 154: This is a nice addition. I don't think it's needed to add a new flag for it.

yajo (commented):
the tricky part are updates, they apply some diff without asking. 

https://github.com/copier-org/copier/blob/e4eccc26c2eaec0d84b8037e849d093f9ae7a2b1/copier/main.py#L705

Gotta do something about that, but anyways that's already a problem, so this won't make it worse.

some code comments too.

very nice addition, thanks!

copier/tools.py
dstanek: So I was going down this path first, but it seemed very difficult to get the result I wanted. I went ahead and finished that work a put together a [gist showing the differences](https://gist.github.com/dstanek/aad702dbc18ff54df6c41f54fe782f33). `orig.diff` is the code from this PR and `new.diff` uses `pygments` and `prompt_toolkit`.

One reason the new code is much larger is that I wanted to have the two-line file header yellow like it is with git. This meant that I had to extend the lexer. In order to add the extra indentation, I had to also extend the `PygmentsTokens` class to inject whitespace. There may be better ways to accomplish these tasks, but I'm never had to dip into the bowels of these libraries before.

copier/main.py
dstanek: Just default to always diff when we can?

copier/tools.py
```diff
@@ -172,3 +173,56 @@ def cleanup(self):
...
+            colorama.Fore.YELLOW, "diff unavailable: unable to determine encoding"
+        )
+        return
+    try:
+        new_lines = get_lines(new_content)
+    except Exception:
+        color_print(
+            colorama.Fore.YELLOW, "diff unavailable: unable to determine encoding"
+        )
+        return
```
pawamoy, line 209: I know some flake8 plugins would say: "only one line in a try block", but this could probably be factorized as:

```suggestion
    try:
        original_lines = get_lines(original_content)
        new_lines = get_lines(new_content)
    except Exception:
        color_print(
            colorama.Fore.YELLOW, "diff unavailable: unable to determine encoding"
        )
        return
```

Also, is there a reason you catch `Exception` rather than the more specific `UnicodeDecodeError`?

dstanek: I could do that.

pawamoy: I see, thanks!

copier/tools.py
dstanek: I'll update this implementation here since that's what I'm actively using and I think I can make it cleaner and more concise. If you like the other way better I can switch to that later.

Yajo:
Hi folks! This was stiill under my radar, although quite deferred because it needs some testing and deep analyzing. 

However, recently we got #627 which proposes to remove `--overwrite` and instead let the user solve the diff after updating. IMHO that seems a simpler approach, both in code and in behavior. 

I'd love to hear your opinions.

dstanek:
> Hi folks! This was stiill under my radar, although quite deferred because it needs some testing and deep analyzing.
> 
> However, recently we got #627 which proposes to remove `--overwrite` and instead let the user solve the diff after updating. IMHO that seems a simpler approach, both in code and in behavior.
> 
> I'd love to hear your opinions.

I'll take a look at that in more detail, but it doesn't seem related to this PR. In this PR I wanted the ability do what Ansible calls "check mode." The ability to run a command and see what would change if it were to be applied.

Just a little background on my use case. I have a set of a few dozen microservices (likely to keep growing) that are based on the same configuration files and general structure. I want to convert it to use copier as the base for management. New services would use copier to start off. With this feature, we could make changes to our base configuration repo (say a logging format change) and have an easy way for a service owner to see what is different without creating new files. We could also run this over all services to see which is out of sync (run a weekly report from CICD like dependabot).

PR #588, opened 2022-02-27
build: Relax dependencies constraints (remove upper bounds)

### LINKED ISSUE #584: Remove upper bounds on dependencies
> About the issue itself: it's true that Copier sets upper bounds on dependencies. I agree that it should relax these constraints, most of them are probably not needed. @Yajo what do you think? Would you agree on changing `^x.x.x` to `>=x.x.x` on Copier's dependencies?

_Originally posted by @pawamoy in https://github.com/copier-org/copier/issues/581#issuecomment-1052628910_

Instead of putting upper bounds, it's better to not put them and instead just skip the buggy versions, if any.

pawamoy:
Fixes #584

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -91,7 +91,7 @@ python-versions = ">=3.6.1"
 
 [[package]]
 name = "click"
-version = "8.0.3"
+version = "8.0.4"
 description = "Composable command line interface toolkit"
 category = "main"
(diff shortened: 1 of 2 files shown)


Yajo (approved):

Yajo:
Cool, thanks!

Issue #324, opened 2021-01-12
Problems getting started on Windows 10 (WAS: Why is Copier not running with pipx's python?)


Cecron:
First time user on Windows 10 here. I installed copier using pipx according to the instructions. I did not expect courier to fail based on the fact that Python is not in my PATH. I assumed courier would be supplied with Python from within pipx...

If this is intentionally, I'm curious of the reason.
Or perhaps this is a template specific thing? If so, are there any basic "poetry-based" templates available I could start with?

```
PS C:\proj\python> pipx install copier
  installed package copier 5.1.0, Python 3.9.1
  These apps are now globally available
    - copier.exe
PS C:\proj\python> copier "gh:pawamoy/copier-poetry" firstProject
...
    create  tests\__init__.py

 > Running task 1 of 1: python scripts/post_generation.py
'python' is not recognized as an internal or external command,
operable program or batch file.
...
```

I keep my python virtual environments in the project directory, so this looks like a chicken and the egg problem to me... ;-)

pawamoy:
It could be because the Windows subprocess (when running the Python task) does not have the same PATH as your shell.

I see three potential solutions:

- passing `shell=True` to the subprocess command
- resolving `python` to its absolute path before running the subprocess
- on the user side, properly set the PATH environment variable (for the user account preferably, not the system), by adding the directory where your Python executable is

> If so, are there any basic "poetry-based" templates available I could start with?

I was going to advertise my template but I see you are currently using it and it's failing :sweat_smile: 
Hopefully one of the solutions above will work!

Cecron:
Note that I purposely don't have Python in my PATH, since I use a pipx installed poetry and keep my virtual environments inside my project directories.

For now I went with solution 3, by creating and activating a virtual environment and my firstProject was created successfully. :-)

However, now I have the same questions about make, is it a mandatory requirement I missed?
(Sphinx e.g. is supplying a make.bat for the Windows platform.)
I do have git installed and in the path, but that is not shipped with make. :-(

```
PS C:\proj\python\copier> py -3 -m venv --prompt . .venv
PS C:\proj\python\copier> .\.venv\Scripts\Activate.ps1
(copier) PS C:\proj\python\copier> copier "gh:pawamoy/copier-poetry" firstProject
...
 > Running task 1 of 1: python scripts/post_generation.py
Project sucessfully generated!
Run `make` to show the available actions.

(copier) PS C:\proj\python\copier> make
[...]

PR #612, opened 2022-03-17
feat: Add jinja variable "_copier_python" to get sys.executable.

andrewleech:
This PR proposes adding a new built-in jinja variable `_copier_python` that provides the `sys.executable` path such that it can be used to run a script in a repeatable fashion, always using the same python envinroment that copier is installed into.

### Background
In my usage of `copier` I want to run a python script as a `task` to manage some post-template operations. I'd prefer not to use invoke or any other third party tool that requires dependencies to be manually pre-installed.

``` yaml 
_tasks:
  - python update.py
```

Initially this appeared to be working fine, but then some other people re-using the template ran into a range of issues; `import yaml` failing, or `python` cannot be found etc. This issue: https://github.com/copier-org/copier/issues/324 is closely related.

I realised pretty quickly that yes, I'm at the mercy of the users PATH and which ever `python` it finds first (if any). Particularly the Windows (`py.exe`) vs Linux (`python3`) difference makes it very hard to have one template that works everywhere. 

This PR resolves the issue by allowing users to define their task like:
``` yaml 
_tasks:
  - "{{ _copier_python }} update.py"
  - ["{{ _copier_python }}", update.py]
```
guaranteeing the copier python will always be used on any platform.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Summary:
 -   Copy dirty changes from a git-tracked template to the project by default, to make
     testing easier.
 -   Advertise clearly which version is being copied or updated in the CLI.
+-   Add jinja variable `_copier_python` to provide python `sys.executable`.
 
 ### Changed
 
--- a/copier/main.py
+++ b/copier/main.py
@@ -207,6 +207,7 @@ class Worker:
             _copier_answers=self._answers_to_remember(),
             _copier_conf=conf,
             _folder_name=self.subproject.local_abspath.name,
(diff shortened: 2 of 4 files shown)


pawamoy:
I had the exact same issues, so big +1 for me.

Yajo:
Cool one. Please could you add a changelog entry?

andrewleech:
> Cool one. Please could you add a changelog entry?

I've added a line, hopefully in the right place :-)

Yajo:
Thanks!

Issue #613, opened 2022-03-18
Improve conflict resolution


tpluscode:
Would it be possible to change how copier handles conflicts?

Presently, conflicting changes are saved as `*.rej` files. Their contents may not be immediately obvious to interpret to decide what to do about them.

I suppose that what copier does is essentially a 3-way merge between the baseline generated from `vN` of a template, the newer version generated from `vN+1`, and the actual project files.

Would it be possible to instead modify the actual template files so that they include the `>>>>` markings for conflicting changes? Maybe this way it would be possible to handle conflicts with `git difftool` or IDE?

pawamoy:
Relevant: https://github.com/copier-org/copier/pull/407

Yajo:
It would be a nice improvement indeed.

tguillemot:
I have some time during my weekend to work on this. You can find my proposal in #627.

PR #627, opened 2022-03-27
Improve conflict resolution

tguillemot:
I propose this work to solve #613: it replaces the merging approach using the `rej` files during an update by a two-way merge. 

This PR is a based on the work of #407 with several differences:
- using two-way merge instead of three-way merge
- update the documentation to tell about how the conflict works and add an example of the `check-merge-conflict` pre-commit hook
- update the existing tests 

Why two-way instead of three : IMHO, I think the majority of the git users are used to the two-way merge instead of the three-way. I can be easily convince to change this behaviour. Maybe we can add an option to let the user choose what he wants. 

I still have some remaining questions :
1. How do we manage the retro compatibility ? The best way would be to add an option to keep the previous behaviour with warning to inform that it will be removed later.
2. What can we do with the `overwrite` option ?

Thanks in advance for your help and time.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -45,7 +45,7 @@ else:
     from backports.cached_property import cached_property
 
 # Backport of `shutil.copytree` for python 3.7 to accept `dirs_exist_ok` argument
-if sys.version_info <= (3, 8):
+if sys.version_info >= (3, 8):
     from shutil import copytree
 else:
     from distutils.dir_util import copy_tree
--- a/copier/main.py
+++ b/copier/main.py
@@ -6,11 +6,12 @@ import subprocess
 import sys
 from contextlib import suppress
 from dataclasses import asdict, field, replace
+from distutils.dir_util import copy_tree
 from functools import partial
 from itertools import chain
 from pathlib import Path
-from shutil import copyfile, copytree, rmtree
-from typing import Callable, Iterable, List, Mapping, Optional, Sequence
+from shutil import copyfile, rmtree
+from typing import Callable, Iterable, List, Mapping, Optional, Sequence, Set
 from unicodedata import normalize
 
 import pathspec
@@ -40,8 +41,22 @@ from .user_data import DEFAULT_DATA, AnswersMap, Question
 # HACK https://github.com/python/mypy/issues/8520#issuecomment-772081075
 if sys.version_info >= (3, 8):
     from functools import cached_property
+
 else:
     from backports.cached_property import cached_property
+
+# Backport of `shutil.copytree` for python 3.7 to accept `dirs_exist_ok` argument
+if sys.version_info <= (3, 8):
+    from shutil import copytree
+else:
+    from distutils.dir_util import copy_tree
+
+    def copytree(src: Path, dst: Path, dirs_exist_ok: bool = False):
+        """Backport of `shutil.copytree` with `dirs_exist_ok` argument.
+
+        Can be remove once python 3.7 dropped.
+        """
+        copy_tree(str(src), str(dst))
 
 
 @dataclass
@@ -694,7 +709,7 @@ class Worker:
             self.run_copy()
 
             # Extract the list of files to merge
-            participating_files: set[Path] = set()
+            participating_files: Set[Path] = set()
             for src_dir in (original_dst_temp, reference_dst_temp):
                 for root, dirs, files in os.walk(src_dir, topdown=True):
                     if root == src_dir and ".git" in dirs:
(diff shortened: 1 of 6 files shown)


docs/updating.md
```diff
@@ -22,25 +22,18 @@ other git ref you want.
...
-          - id: forbidden-files
-            name: forbidden files
-            entry: found copier update rejection files; review them and remove them
-            language: fail
-            files: "\\.rej$"
+    - repo: https://github.com/pre-commit/pre-commit-hooks
+        rev: v4.0.1
+        hooks:
+        - id: check-merge-conflict
+            args: [--assume-in-merge]
```
Yajo, line 36: These lines have wrong indentation and are not valid yaml. Can you fix that please?

Yajo (commented):
Please can you [add co-autorship info](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) with the author of #407 to credit him?

Yajo:
Thank you so much for this PR. It's a nice enhancement and the
implementation looks much better!

Regarding backwards compatibility, it has been broken in so many ways now
that I don't really care that much 🤣

About overwrite... I think I somehow get the problem by looking at the
code, but I'd appreciate it if you could explain it and any solutions you
may have thought.

Finally, please remember to add a note in the changelog.

Yajo:
BTW have you seen #562? It displays the diff interactively to accept or reject before continuing.

tguillemot:
> Please can you [add co-autorship info](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) with the author of https://github.com/copier-org/copier/pull/407 to credit him?

Of course. It's done.

> I'd appreciate it if you could explain it and any solutions you may have thought

You're perfectly right, sorry for that. 
In the current behaviour of copier the `--overwrite` option overwrite the files that already exist without asking. With the new behaviour we do not to overwrite things in the "old way", but it still make sense to give an option to the user to accept automatically merge upstream (a kind of `git checkout --theirs`). IMHO, I prefer to remove the `overwrite` option and let the user manage the conflict as he wants on his side.

tguillemot:
> BTW have you seen https://github.com/copier-org/copier/pull/562? It displays the diff interactively to accept or reject before continuing.

Indeed I missed that PR.

copier/main.py
```diff
@@ -686,37 +694,57 @@ def run_update(self) -> None:
...
+                    subfile_names.append(subfile_kind)
+
+                with local.cwd(merge_dst_temp):
+                    output = git("merge-file", "-p", *subfile_names, retcode=None)
+
+                dest_path = Path(self.dst_path, basename)
+                # Remove the file if it was already removed in the project
+                if not output and "modified" not in subfile_names:
+                    try:
+                        dest_path.unlink()
```
Yajo, line 741: Are you fixing #461 too? 😍

Yajo:
I'll have to review more in depth, because the part you're modifying is one of the most important in copier. Just be patient please.

tguillemot:
@Yajo Don't worry for the time, I understand perfectly :)

And I have to find a way to fix windows problems first.

copier/main.py
```diff
@@ -686,37 +694,57 @@ def run_update(self) -> None:
...
+                    ("new upstream", self.dst_path),
+                ]:
+                    path = Path(src_dir, basename)
+                    if path.is_file():
+                        copyfile(path, Path(merge_dst_temp, subfile_kind))
+                    else:
+                        subfile_kind = os.devnull
+                    subfile_names.append(subfile_kind)
+
+                with local.cwd(merge_dst_temp):
```
bkmeneguello, line 734: Maybe this is the perfect spot to check if the file has changed and show a brief diff to the user (`difflib.unified_diff(old_upstream, new_upstream)`)

barrywhart:
@yajo, @tguillemot: What's the status of this PR? If you don't have time to work on it, I can work on it.

I personally prefer 3-way merge: That's the behavior that gives you conflict markers, right? E.g.
```
<<<<<<< HEAD
foo
=======
bar
>>>>>>> cb1abc6bd98cfc84317f8aa95a7662815417802d
```

@yajo: You mentioned needing to review it carefully due to the risk of breaking copier. Should we consider making the behavior configurable?
* Old way: `.rej` files
* New way: Inline conflict markers
* Potentially: New way 2: 2-way merge

yajo:
Continuing in #807, which has no conflicts

Issue #629, opened 2022-03-29
Internal keys / non-prompted values


verdaatt:
**Is your feature request related to a problem? Please describe.**

Trying to build a template for Terraform structure, which means a lot of the file tree structure depends on one or more of the inputs. Using a "choice" type here makes a lot more sense from a usability standpoint than using multiple competing boolean keys. For example:

*Example copier.yaml*

```yaml
flavor:
  type: str
  choices:
    - Docker
    - Instances
    - Kubernetes
    - None
```

But that means having to do string comparison if-statements in directory names to create conditional directories. And everywhere else when this value is evaluated. If multiple answers are correct for some conditional directories then that results in even longer if-and statements:

*Example filetree*

```jinja2
.
├── {% if flavor == 'docker' %}ecs{% endif %}
├── {% if flavor == 'kubernetes' %}eks{% endif %}
├── {% if flavor == 'instance' %}bastion{% endif %}
├── {% if flavor == 'docker' or flavor == 'kubernetes' %}portainer{% endif %}
└── {% if flavor != 'none' %}vpc{% endif %}
```

This can make the directory names quite long and makes everything poorly readable. It also results in a lot of Jinja repetition across the template.

**Describe the solution you'd like**

I was hoping to create keys derived from the choice key behind the scenes, e.g. without prompting the user for input. In my scenario these would be booleans, but it could also be useful for concatenating strings or other post-input processing purposes that deduplicates a lot of code from the template files.

Example of how I'd love for this to work for my use case:

*Future copier.yaml*
```
flavor:
  type: str
  choices:
    - Docker
    - Instances
    - Kubernetes
    - None

# Internal variables
isContainer:
  type: bool
  value: "{% if flavor == 'docker' or flavor == 'kubernetes' %}true{% else %}false{% endif %}"

isDocker:
  type: bool
  value: "{% if flavor == 'docker' %}true{% else %}false{% endif %}"

isInstance:
  type: bool
  value: "{% if flavor == 'instance' %}true{% else %}false{% endif %}"

isKubernetes:
  type: bool
  value: "{% if flavor == 'kubernetes' %}true{% else %}false{% endif %}"

isLite:
  type: bool
  value: "{% if flavor == 'none' %}true{% else %}false{% endif %}"
```

*Future filetree*

```jinja2
.
├── {% if isDocker %}ecs{% endif %}
├── {% if isKubernetes %}eks{% endif %}
├── {% if isInstance %}bastion{% endif %}
[...]

PR #641, opened 2022-04-14
docs: Document the context hook extension

pawamoy:
Closes issue #629.


docs/configuring.md
```diff
@@ -1035,6 +1036,151 @@ _commit: v1.0.0
...
+├── {% if flavor == 'instance' %}bastion{% endif %}
+├── {% if flavor == 'docker' or flavor == 'kubernetes' %}portainer{% endif %}
+└── {% if flavor != 'none' %}vpc{% endif %}
+```
+
+Each of these folders then have multiple templated files that must also repeat again and
+again these Jinja conditions to render the contents correctly.
+
+You can reduce repetitions with
+[macros](https://jinja.palletsprojects.com/en/2.10.x/templates/#macros) or updated
```
Yajo, line 1082: Old link!

docs/configuring.md
```diff
@@ -1035,6 +1036,151 @@ _commit: v1.0.0
...
+└── {% if isLite %}vpc{% endif %}
+```
+
+We have good news for you: this is possible. A Copier/Jinja extension was written
+specially for this use-case:
+[copier-templates-extensions](https://github.com/copier-org/copier-templates-extensions)
+and its
+[`ContextHook` extension](https://github.com/copier-org/copier-templates-extensions#context-hook-extension).
+
+### Using a context hook
```
Yajo, line 1106: I feel like this chapter is too much. These docs should be in the extension itself, don't you think?

docs/configuring.md
```diff
@@ -1035,6 +1036,151 @@ _commit: v1.0.0
...
+└── {% if flavor != 'none' %}vpc{% endif %}
+```
+
+Each of these folders then have multiple templated files that must also repeat again and
+again these Jinja conditions to render the contents correctly.
+
+You can reduce repetitions with
+[macros](https://jinja.palletsprojects.com/en/2.10.x/templates/#macros) or updated
+context objects that you can import from special templated files
+([see issue #229](https://github.com/copier-org/copier/issues/229#issuecomment-661686740)).
```
Yajo, line 1084: I'd recommend using a [namespace](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-globals.namespace) instead.

docs/configuring.md
```diff
@@ -1035,6 +1036,151 @@ _commit: v1.0.0
...
+[`ContextHook` extension](https://github.com/copier-org/copier-templates-extensions#context-hook-extension)
+lets you modify the context used to render templates, so that you can add, change or
+remove variables.
+
+In order for Copier to be able to load and use the extension when generating a project,
+it must be installed alongside Copier itself. If Copier was installed with pip in a
+virtualenv, just `pip install copier-templates-extensions` as well, and tell your users
+to do the same. If Copier was installed with
+[pipx](https://github.com/pipx-project/pipx), inject it with
+`pipx inject copier copier-templates-extensions`, and tell your users to do the same.
```
Yajo, line 1118: This paragraph is repeated above in the jinja extensions section. Maybe just link to it? Unless you move these docs to the extension.

Yajo (commented):
It seems you have to run pre-commit to fix the builds

docs/configuring.md
pawamoy: Ah, good catch!

docs/configuring.md
pawamoy: Yeah I was wondering the same. Do you think we should just drop everything above "Using a context hook"? I'd be totally fine with it, no worries.

docs/configuring.md
pawamoy: Ah, yes, we can probably just point to it then.

docs/configuring.md
pawamoy: I'd like to have @verdaat's opinion as well, since you and I are probably less used to read docs like new comers would :slightly_smiling_face:

pawamoy: And I don't think this chapter would be useful in the extensions docs themselves. And if they're not useful there, they're probably not useful here either.

Yajo: Maybe we can just open a new page called FAQ and drop the note there. Then put the rest of the docs in the extension repo itself.

pawamoy: OK I just did that :slightly_smiling_face:

docs/configuring.md
pawamoy: Irrelevant now that this section is removed.

docs/comparisons.md
```diff
@@ -26,6 +26,7 @@ docs! We don't want to be biased, but it's easy that we tend to be:
 | Requires programming                     | No                               | No                              | Yes, JS       |
 | Requires templates to have a suffix      | Yes by default, configurable[^3] | No, not configurable            | You choose    |
 | Task hooks                               | Yes                              | Yes                             | Yes           |
+| Context hooks                            | Yes[^5]                          | Yes                             | ?             |
```
pawamoy, line 29: I searched through the docs but honestly have no idea if it's possible to alter the user's answers (and/or add variables) before rendering with Yeoman generators.

Yajo: Well, if somebody with more experience in Yeoman comes in, let them fix this.

Yajo (approved):

pawamoy:
Please do squash commits when merging :slightly_smiling_face:

docs/faq.md
```diff
@@ -0,0 +1,87 @@
...
+You can then configure your Jinja extensions in Copier's configuration file:
+
+```yaml title="copier.yaml"
+_jinja_extensions:
+    - copier_templates_extensions.TemplateExtensionLoader
+    - extensions/context.py:ContextUpdater
+```
+
+Following this example, you are supposed to provide a `context.py` file in the
+`extensions` folder at the root of your template. This file contains your context hook:
```
verdaatt, line 34:
```suggestion
`extensions` folder at the root of your template to modify the context. If for
example your `copier.yaml` contains a multiple-choice variable like this:

```yaml title="copier.yaml"
flavor:
  type: str
  choices:
    - Docker
    - Instances
    - Kubernetes
    - None
\```

The `context.py` file contains your context hook which could look like:
```

verdaatt (commented):
Looks good. Couple of small suggestions

pawamoy:
Suggestions applied! Thanks @verdaatt

yajo (approved):

PR #667, opened 2022-05-16
fix release data

seebi:
small typo


seebi:
... but important :)

yajo:
Ah! Butterfingers me 🤦‍♂️

PR #682, opened 2022-05-26
Refactor the tests to only depend on the prompt format in one place

pfmoore:


tests/test_complex_questions.py
```diff
@@ -140,50 +141,39 @@ def test_api(tmp_path, template_path):
 
 def test_cli_interactive(tmp_path, spawn, template_path):
     """Test copier correctly processes advanced questions and answers through CLI."""
-    invalid = ["Invalid input"]
+
+    def check_invalid(tui, name, format, invalid, help=None):
+        """Check that invalid input is reported correctly"""
+        expect_prompt(tui, name, format, help)
+        tui.sendline(invalid)
+        tui.expect_exact("Invalid input")
```
yajo, line 149: This imitates more closely what we were doing. It refers to the fact that when you send some input, it is displayed on screen:

```suggestion
        tui.expect_exact(invalid)
        tui.expect_exact("Invalid input")
```

yajo (changes requested):
Thanks! The refactor is indeed much helpful and makes total sense. Just one detail to merge.

tests/test_complex_questions.py
pfmoore: Ah, does that work? If you send text to the application, you can then check it was printed? Cool. I'll make that change, and also change the argument to be called `invalid_value` because I think that's a little clearer (I was getting confused as to whether it was the input or the error message!)

I'll push the fix later this evening.

pfmoore: By the way, what's the practice here? Should I amend the commit and force push, or should I just commit the fix (and if so, do I need to use the commitizen format of commit message)?

pawamoy: I believe you can simply push the fix, with whatever commit message, as we can then squash all the commits when merging. I'll let @yajo confirm.

pfmoore: I've pushed the fix with a simple commit message. If that's not right, I can easily enough fix it up. Hope that's OK!

PR #689, opened 2022-06-02
feat: change the format of the question prompt

### LINKED ISSUE #674: Add the ability to customise how questions are asked
### Discussed in https://github.com/copier-org/copier/discussions/673

<div type='discussions-op-text'>

<sup>Originally posted by **pfmoore** May 23, 2022</sup>
The prompt currently appears to be a 🎤emoji, followed by the help (if specified), then the item name and a format prompt. I'd like to be able to change the prompt as I find it difficult to read (in particular, the format prompt is very distracting, and I'd like to omit it). Is that possible?</div>

Having checked the source code, it seems that this isn't possible. But the `questionary` library looks very flexible, so it would be possible to do this if there was a way to describe the format in the template configuration. Would such a change be welcome? I'd be more than happy to work on a PR for this, but I'm very new to the project (I'm currently trying to migrate my existing cookiecutter templates over to copier) so I don't know if that's something that would be considered a reasonable feature for the project.

I'd imagine doing this via settings in the `copier.yml` file, so the format would be decided by the template, rather than by the user. Would that make sense?

pfmoore:
Fixes #674


pfmoore:
I tried to update the documentation, but I couldn't check the change as `poe docs` doesn't work for me (maybe a problem because I'm running Python 3.10, I'm not sure - it failed in the `markdown` module).

pfmoore:
Actually, I just checked on ReadTheDocs for the rendered docs for this PR, and it looks OK.

pfmoore:
Not sure I understand why the ubuntu-3.8 test failed. It doesn't look related to my change, as far as I can tell...

yajo:
Thanks! The patch looks perfect.



> I couldn't check the change as `poe docs` doesn't work for me

There's something weird with the docs dependencies. I actually still don't understand how it's possible that readthedocs still works fine 😆 

I guess it'll eventually fix by itself on some dependabot update.



> Not sure I understand why the ubuntu-3.8 test failed. It doesn't look related to my change, as far as I can tell...

Weird thing, I never saw it before. I restarted the build, let's see if it goes green.

Thanks!

pfmoore:
Awesome, thanks :-)

Issue #398, opened 2021-05-05
feat: questionnaire validation


RoyalTS:
I'd love a possibility to validate questionnaire inputs. Being able to validate `str` against regex would already be a huge help, e.g. by specifying a field appropriately in `copier.yml`:

```yaml
project_name:
  type: str
  help: Your project name
  validate: ^[a-zA-Z][_a-zA-Z0-9]+$
```

Yajo:
This was the feature I was hoping nobody asked 🤣

Yes, gotta do it. 

However, I'm more o the side of allowing a Jinja template to validate, so you can validate any data type. It'd be interesting for YAML questions for example.

eyllanesc:
my 2 cents: I think "validate" should accept jinja filters, and there should be filters like regex ones, range validators, etc.

andrew-glenn:
@yajo Just to clarify - you're thinking an inline jinja template, versus a path to a jinja template on the local filesystem, correct? 

FWIW my team is discussing this and considering a contribution; Just want to align with maintainer desires to (hopefully) fast-track any PR that we submit. :)

yajo:
I was thinking on an in-line template that, if renders to an empty (or whitespace-only) string means OK, but if it renders to some text, that text is considered the error message. Does that make sense for you?

yajo:
Correct!

Another example:

```yaml
your_age:
  type: int
  validator: "{% if your_age <= 0 %}Must be positive{% endif %}"

**Comment by barrywhart** [comment]
@yajo: This feature looks really useful! If I understand correctly, it hasn't been released yet. Will there be a new release soon? 🙏

**Comment by yajo** [comment]
I wanted to release sonner, but saw there's some incompatibility with dependencies that you can see in #785. I want to at least fix that before the next release, or I'll get a flood of issues.

I still didn't have time though... if someone volunteers, the next release will be sooner.

**Comment by barrywhart** [comment]
Thanks! I may be able to look at it early next week.

Great project, BTW. We are switching from cookiecutter soon. 🙌🏽

**Comment by pawamoy** [comment]
I can take a look this week as well.

**Comment by barrywhart** [comment]
@yajo, @pawamoy: I have a possible fix for the test failures in #785. See PR #804.

**Comment by nils-borrmann-y42** [comment]
@yajo I'm a bit confused about this feature. This doesn't seem to be able to solve the original usecase of validating a regex. Obviously the validator could look something like this - but this fails since `re` is not available:

```
[...]

PR #694, opened 2022-06-10
Add: regex answer validation

ironhouzi:
Question entries can contain a keyword: `regex`, which can be used to
specify a regular expression for validating user input.

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -1,6 +1,7 @@
 """Functions used to load user data."""
 import datetime
 import json
+import re
 import sys
 import warnings
 from collections import ChainMap
@@ -15,6 +16,7 @@ from typing import (
     ChainMap as t_ChainMap,
     Dict,
     List,
+    Optional,
     Union,
 )
 
@@ -154,6 +156,9 @@ class Question:
             Additional text printed to the user, explaining the purpose of
             this question. Can be templated.
 
+        regex:
+            Regular expression for validating user input.
+
         multiline:
             Indicates if the question should allow multiline input. Defaults
             to `True` for JSON and YAML questions, and to `False` otherwise.
@@ -189,6 +194,7 @@ class Question:
     default: Any = None
     help: str = ""
     ask_user: bool = False
+    regex: Optional[str] = None
     multiline: Union[str, bool] = False
     placeholder: str = ""
     secret: bool = False
@@ -323,6 +329,8 @@ class Question:
         if self.choices:
             questionary_type = "select"
             result["choices"] = self._formatted_choices
+        if self.regex is not None:
+            self._re = re.compile(self.regex)
         if questionary_type == "input":
             if self.secret:
                 questionary_type = "password"
@@ -359,6 +367,11 @@ class Question:
 
     def validate_answer(self, answer) -> bool:
         """Validate user answer."""
+        if self.regex is not None:
+            return (
+                self._re.match(answer) is not None
+                or f'Does not match regex: `{self.regex}`'
+            )
         cast_fn = self.get_cast_fn()
         try:
             cast_fn(answer)


ironhouzi:
Closes #398

Will create tests over the weekend.

Currently tested with the following `copier.yml`:

```
project_description:
  type: str
  help: Describe your project
  regex: '^.{10,79}\.'
```
Screenshot for invalid input prompt:
image
Screenshot for valid input prompt:
image

copier/user_data.py
```diff
@@ -359,6 +367,11 @@ def get_multiline(self) -> bool:
 
     def validate_answer(self, answer) -> bool:
         """Validate user answer."""
+        if self.regex is not None:
+            return (
+                self._re.match(answer) is not None
```
pawamoy, line 372: I think a search would allow for less strict regex, while still allowing exact matches with `^` and `$` at the boundaries of the regex.

ironhouzi: Thanks. I'll take a look while writing the tests.

yajo:
Hello @ironhouzi!

Thank you very much for the PR! And also thanks for willing to contribute to this project. 😊 

As you can see in https://github.com/copier-org/copier/issues/398#issuecomment-1145399410, the design proposal is quite different from the one you have implemented. Instead of just dropping a regex to validate, the proposal is to add a new attribute to the question (maybe `validation`?) that can contain a Jinja template. If the template renders to some text, then that text will be the error message printed to the user.

This is much more powerful than just a regex. It is also easier to use for template designers.

I have to admit it feels a bit weird that you propose here a totally different approach without even asking. 😕 I guess maybe you didn't read the whole thread... Well, don't worry, it's not a big problem, but I encourage you to read it before continuing, in case you didn't before.

In any case, I have reviewed the PR and still find that the jinja template approach would be better. Would you like to share some thoughts about the subject? Are there good reasons why you think this alternative solution is better?

Thanks again.

ironhouzi:
Hello @yajo.

Sorry for approaching this inappropriately. I just happened to find myself in the situation where I needed a max line length validator and saw that the author of the issue was in a very similar situation.

In stead of just adding a +1 to the discussion, stating that I would also be seeing value in such a feature, I thought I'd roll up my sleeves and create something that would work for me straight away. In stead of just going on about my day, I thought it would be nice to share the work with anyone else who was waiting for the proper design to be implemented.

I did not have any expectations of the PR being accepted, since I do not know the code base well and have only spent an hour or two on the hack. However, I did hope to provide some value by sharing what I had already done for myself and I must say that the hack works really well for my immediate needs, although I fully see how the proposed design will be a much better implementation for the project. Leveraging jinja2 is a really great move and it has already enabled me to create templates for reading author name and email from git, which I use as default values for these questioners. I am very glad that you care so much for ensuring that this tool is backed by great design decisions. Thank you for the great work!

I am not sure how I should have done this more appropriately and I really appreciate your feedback, so that I can learn how to improve on benefiting others when in a similar situation in the future. If you could please help me understand how I should have gone about this situation, I would appreciate that tremendously now that you know the intention behind my actions.

Thank you very much for your time on giving thoughtful feedback and reviewing the PR!

yajo:
Thanks for your explanation. I'm very glad to see you could get the message in good mood. 😊 I appreciate your efforts and your willingness to contribute. Don't give up!

The proposed design is quite simple, as you can see in https://github.com/copier-org/copier/issues/398#issuecomment-1145399410. Not much more to add above of that.

So the 1st thing I recommend you to do is to write the test. The test suite is simple and extensive, so I find it faster to do TDD with Copier. At first, I'd go for the non-interactive test (calling copier just from API, similar to how you would do it when passing `--force` from CLI). The test could be something similar to this one, but catching a `ValidationError` instead of a warning: https://github.com/copier-org/copier/blob/5ab752c11f2c0884b6595821d040f94919ada680/tests/test_normal_jinja2.py#L87-L107

Also I would parametrize the test, checking various error messages and also checking that a valid value goes ok. Find an example of parametrization with pytest here: https://github.com/copier-org/copier/blob/db9a81d08851ec093874f9bdcd682e26f0e03d8b/tests/test_updatediff.py#L501-L503

Also [see the docs](https://docs.pytest.org/en/6.2.x/parametrize.html).

Once the test is written and failing, let's go for the implementation.

You already found most of the interesting blocks regarding the implementation. Just override the `validate_answer` method here: https://github.com/copier-org/copier/blob/db9a81d08851ec093874f9bdcd682e26f0e03d8b/copier/user_data.py#L356-L363

It should support a new attribute in the [Question](https://github.com/copier-org/copier/blob/db9a81d08851ec093874f9bdcd682e26f0e03d8b/copier/user_data.py#L139-L196). I'd call it `validator`. It should contain an inline jinja template, if the attribute is set.

Since we already have a basic validator for the type, that one can go first. If the answer doesn't match the type, there isn't much left to validate.

After type is validated, then go for the custom validator if defined. We already render inline jinja templates elsewhere. Actually any value in a question can be templated, and it goes through this function: https://github.com/copier-org/copier/blob/db9a81d08851ec093874f9bdcd682e26f0e03d8b/copier/user_data.py#L379-L392

So, just pass the input value and the template to that method and raise a `ValidationError` if it returns a non-empty string.

Once that's done, just run the test until it goes green.

After finishing with that, if you want to go the extra mile, write an interactive test. This feature will have an impact in interactive sessions, as the error message has to be properly rendered while doing the wrong input. Here you can find a [pexpect](https://pexpect.readthedocs.io/en/stable/)-based test that can inspire you. It's long but easy to understand: https://github.com/copier-org/copier/blob/db9a81d08851ec093874f9bdcd682e26f0e03d8b/tests/test_complex_questions.py#L139-L247

I'll now close this PR, so this patch can stay here in case somebody else wants to cherry-pick it for their immediate validation needs while we get to the final solution. Please feel free to open a new PR when ready, or to continue the discussion.

Thanks!

PR #698, opened 2022-06-14
feat: questionier validation

### LINKED ISSUE #398: feat: questionnaire validation

ironhouzi:
Use inline jinja2 templates for validating answers. The template must
return an empty string if valid and a non-empty string used as the error
message if invalid.

Closes #398 .


ironhouzi:
Continuing from our [previous discussion](https://github.com/copier-org/copier/pull/694#issuecomment-1154153542).

I've tried to follow your advice @yajo . However, I am not sure I understand the copier tool in all the ways it is used. I only use it as an interactive application, so that's the use case I've tried to implement.

I did write a few interactive tests, although they are quite simple due to my hardship using the pexpect library to check the tty buffer. I tried to have one test where both invalid cases where checked before finally inserting the valid string, but these intermediary checks would fail and I found it hard to reason with the buffer check, even when dumping the buffer to a log file. So instead I just copied the simple testing scheme that was already laid out and checked both cases individually.

I also tried to implement the library feature, but it is not clear to me how raising a ValidationError can be implemented while also ensuring that the interactive use case works, where validation either returns `True` or a string used as the interactive error message. This is the interface expected by the [library](https://pypi.org/project/questionary/).

copier/user_data.py
```diff
@@ -193,6 +198,7 @@ class Question:
     placeholder: str = ""
     secret: bool = False
     type: str = ""
+    validation: Optional[str] = None
```
yajo, line 201: You can just use this:

```suggestion
    validation: str = ""
```

Makes it easier to reason about the value, because it will always be a `str`. Just it will be `""` if the user supplies nothing.

copier/user_data.py
```diff
@@ -357,14 +363,19 @@ def get_multiline(self) -> bool:
...
             cast_fn(answer)
-            return True
         except Exception:
             return False
+        else:
+            if self.validation is None:
+                return True
+
+        validation = self.render_validation(self.validation, answer)
+        return validation == '' or validation
```
yajo, line 378: I think this is better, if we just add the line commented below to `render_value` (see following comments):

```suggestion
        try:
            validation = self.render_value(self.validation).strip()
        except UserMessageError as error:
            raise ValidationError(str(error)) from UserMessageError
        if validation:
            raise ValidationError(validation)
        return True
```

copier/user_data.py
```diff
@@ -395,6 +406,24 @@ def render_value(self, value: Any) -> str:
...
+        template = self.jinja_env.from_string(template_string)
+        answers = {**self.answers.combined, **{self.var_name: value}}
+        try:
+            # strip whitespace from indented multiline template text
+            result = template.render(**answers).strip()
+        except UndefinedError as error:
+            raise UserMessageError(str(error)) from error
+
+        return result
+
```
yajo, line 426: You can remove all this code.

```suggestion
```

copier/user_data.py
```diff
@@ -395,6 +406,24 @@ def render_value(self, value: Any) -> str:
...
+    def render_validation(self, template_string: str, value: str) -> str:
+        """Render a single jinja validation template value.
+
+        Validation templates will return an empty string on success and a
+        non-empty string as the error message. The error message defined in a
+        template might contain indents, which will be stripped from the returned
+        error message.
+        """
+        template = self.jinja_env.from_string(template_string)
+        answers = {**self.answers.combined, **{self.var_name: value}}
```
yajo, line 418: Instead of adding this whole new method, just add the `**{self.var_name: value}` part to [the call to `template.render` inside `render_value()`](https://github.com/copier-org/copier/pull/698/files#diff-520654b89873468fcdb6a0317f022436ab8f3bc467e56ea2d7cc0102c2ed4b00L394). That should be enough.

yajo (commented):
Thank you! This looks really much better than #694!

I see the code formatting is failing, so please just do this locally:

```sh
poetry run pre-commit install -t pre-commit -t commit-msg
poetry run pre-commit run -a
git commit -a --amend
```

That should reformat your code and let you push-force it again to get things green.


I have added some comments. You code looks pretty nice, but please attend them so it fits better the overall design of the software.

Thank you also for the tests. Don't worry about the library/non-interactive cases then. If you're not familiar with those, just let's try to get all your proposal good, and then I'll add another commit with the missing tests.

PR #733, opened 2022-08-08
Feat: Adding force_overwrite config option.

andrew-glenn:
This PR adds the `force_overwrite` config option, which allows template authors more granularity in how their templates are consumed downstream. 

The intended use case (and reason for this PR) is the ability to centralize core information, such as license info, legal notices, code of conduct, etc. With this PR, copier will force-overwrite the content of the given files in the downstream consumer repository, without prompting. 

This is considered a relative edge case, but is critical for situations where more granularity is required. 

_Edit to add: this differs from the existing `overwrite` functionality, which indiscriminately overwrites without prompting. This adds pattern matching, allowing more flexibility_

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -102,6 +102,11 @@ class Worker:
         skip_if_exists:
             User-chosen additional [file skip patterns][skip_if_exists].
 
+        force_overwrite:
+            Specific files to force-overwrite. Allows more granularity than the `overwrite` option.
+
+            See [force_overwrite][].
+
         cleanup_on_error:
             Delete `dst_path` if there's an error?
 
@@ -139,6 +144,7 @@ class Worker:
     exclude: StrSeq = ()
     use_prereleases: bool = False
     skip_if_exists: StrSeq = ()
+    force_overwrite: StrSeq = ()
     cleanup_on_error: bool = True
     defaults: bool = False
     user_defaults: AnyByStrDict = field(default_factory=dict)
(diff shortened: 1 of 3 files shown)


yajo:
Thanks for the contribution! 

I'm not sure I like this feature. I think it's the user who must decide if files should or not be overwritten. Why would the template designer choose that? 🤔

andrew-glenn:
In my use-case, there are certain files that need to be fully managed by the
upstream template. Examples would be a software license, legal notices, etc. 

A user could choose to change the software license of a repo, which has
significant implications for both the project, and the company that's providing
the repository for public consumption. 

A more concrete example would be the open-source "department" (for lack of a
better term) of BigCorp Inc. Management decides that all repositories must have
an Apache2.0 license. The hope / intention was to use copier as a way to enforce
this policy.

Thinking about the bigger purpose and scope of the copier project, I would have
the same opinion in your position as a maintainer. I think this _may_ be better
suited for a post-migration strategy. Assuming that's the case, I think a
secondary PR is needed to make certain private methods public so they can be
used in a post-migration script.

Looking forward to your thoughts.

yajo:
The case that you mention seems to be better fixed with:
1. Multiple templates. This is already supported by copier. One template for legal stuff, another for the normal project lifecycle. This way, updating the normal project will never touch legal files, and vice-versa. 
2. CODEOWNERS file. This is supported by github and gitlab. Blocks changes into certain files by unauthorized users. It would forbid non-legal users from changing legal files.

See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners and https://docs.gitlab.com/ee/user/project/code_owners.html

WDYT?

andrew-glenn:
Anticipating that it wouldn't be good fit for upstream, I've accomplished what I need via wrapping the Worker class, invoking it twice, and playing with excludes.

PR #736, opened 2022-08-09
Sorting match patterns

### LINKED ISSUE #735: Unordered pattern matching can lead to ignored negate conditions
https://github.com/copier-org/copier/blob/fa3515eff2cc6b69048ce5c66b5baef3efd5543f/copier/main.py#L216-L221

Pathspec.match file leverages `pathspec.util.match_file`[^1] 

```python
	def match_file(self, file, separators=None):
		# type: (Union[Text, PathLike], Optional[Collection[Text]]) -> bool
		"""
		Matches the file to this path-spec.
		*file* (:class:`str` or :class:`~pathlib.PurePath`) is the file path
		to be matched against :attr:`self.patterns <PathSpec.patterns>`.
		*separators* (:class:`~collections.abc.Collection` of :class:`str`)
		optionally contains the path separators to normalize. See
		:func:`~pathspec.util.normalize_file` for more information.
		Returns :data:`True` if *file* matched; otherwise, :data:`False`.
		"""
		norm_file = util.normalize_file(file, separators=separators)
		return util.match_file(self.patterns, norm_file)
```

`Pathspec.util.match_file`[^2] will iterate through each pattern until finished, and return a boolean (pattern match or not).

```python
def match_file(patterns, file):
	# type: (Iterable[Pattern], Text) -> bool
	"""
	Matches the file to the patterns.
	*patterns* (:class:`~collections.abc.Iterable` of :class:`~pathspec.pattern.Pattern`)
	contains the patterns to use.
	*file* (:class:`str`) is the normalized file path to be matched
	against *patterns*.
	Returns :data:`True` if *file* matched; otherwise, :data:`False`.
	"""
	matched = False
	for pattern in patterns:
		if pattern.include is not None:
			if file in pattern.match((file,)):
				matched = pattern.include
	return matched
```

This leads to a situation where a negate-pattern placed before a wildcard is ignored.

Example:

```
>>> x = ["!foo.txt", "*.txt", "bar.env", "blah.other"]
```

The most straightforward way to handle this is to sort the patterns before passing them to the pathspec library (modifying `copier.worker._path_matcher`). 

Previous example, sorted:

```
['*.txt', 'bar.env', 'blah.other', '!foo.txt']
```

A PR is forthcoming.

[^1]: https://github.com/cpburnz/python-path-specification/blob/7ca27e2300ced6364a3816820eb2532c89c7d01c/pathspec/pathspec.py#L129-L144
[^2]: https://github.com/cpburnz/python-path-specification/blob/7ca27e2300ced6364a3816820eb2532c89c7d01c/pathspec/util.py#L270-L288

andrew-glenn:
This PR closes #735

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -215,10 +215,28 @@ class Worker:
 
     def _path_matcher(self, patterns: Iterable[str]) -> Callable[[Path], bool]:
         """Produce a function that matches against specified patterns."""
+        patterns = self._sort_patterns(patterns)
         # TODO Is normalization really needed?
         normalized_patterns = (normalize("NFD", pattern) for pattern in patterns)
         spec = pathspec.PathSpec.from_lines("gitwildmatch", normalized_patterns)
         return spec.match_file
+
+    def _sort_patterns(self, patterns: Iterable[str]) -> Iterable[str]:
+        # See https://github.com/copier-org/copier/issues/735
+        """
+        Properly sort patterns so that wildcards are first and negation patterns are last.
+        """
+
+        def _sort(i1, i2):
+            if (("!" in i1) and ("!" in i2)) or (("*" in i1) and ("*" in i2)):
+                return 0
+            if ("!" in i1) or ("*" in i2):
+                return 1
+            if ("*" in i1) or ("!" in i2):
+                return -1
+            return 0
+
+        return sorted(patterns, key=cmp_to_key(_sort))
 
     def _solve_render_conflict(self, dst_relpath: Path):
         """Properly solve render conflicts.


yajo (approved):

copier/main.py
```diff
@@ -215,11 +215,29 @@ def _render_context(self) -> Mapping:
...
         return spec.match_file
 
+    def _sort_patterns(self, patterns: Iterable[str]) -> Iterable[str]:
+        # See https://github.com/copier-org/copier/issues/735
+        """
+        Properly sort patterns so that wildcards are first and negation patterns are last.
+        """
+
+        def _sort(i1, i2):
+            if (("!" in i1) and ("!" in i2)) or (("*" in i1) and ("*" in i2)):
```
yajo, line 231: Shouldn't you use `i1.startswith("!")` and similar?

yajo (changes requested):
Thanks for your contribution! 

However it seems to be a wrong fix, after double-checking.

If you read [upstream docs](https://git-scm.com/docs/gitignore), you will find this example:


    # exclude everything except directory foo/bar
    /*
    !/foo
    /foo/*
    !/foo/bar

It shows that processing patterns in order makes sense and is the expected behavior. OTOH, if you find the library is not matching inn the same fashion as a gitignore, then you should probably file this issue upstream, to the pathspec project. 

The correct fix would be to document it instead, or just point to upstream docs, unless I'm misunderstanding something. 

Thanks!

andrew-glenn:
Since `copier` is inheriting the behavior of pathspec, I think for the project, a doc change is probably the best option.

yajo:
Don't you think that just reordering the exclusions should be enough for your use case?

PR #752, opened 2022-08-26
build(deps): remove upper bound from Python version specifier

sisp:
Capping the Python version (e.g. `<4`) is a bad practice, so I've removed it.

* https://github.com/pypa/packaging.python.org/pull/850
* https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special
* https://discuss.python.org/t/requires-python-upper-limits/12663


pawamoy:
We originally left the Python 4 upper bound for reasons mentioned in #509. Great to learn it's not needed anymore

PR #754, opened 2022-08-26
chore: use isort profile "black" and remove default settings

sisp:
I've simplified the `isort` config in `pyproject.toml` by using the [profile "black"](https://pycqa.github.io/isort/docs/configuration/profiles.html#black). I've also removed redundant settings that explicitly set `isort`'s default values.


pawamoy:
Interesting, I didn't know Isort implemented a Black profile! Thanks

yajo (approved):

PR #756, opened 2022-08-26
chore: remove Black's default line length setting

sisp:
I've removed the `line-length` setting for Black because 88 characters is the default value anyway, so it's redundant to set it explicitly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,6 @@ style = "pep440"
 vcs = "git"
 
 [tool.black]
-line-length = 88
 target-version = ['py36']
 
 [tool.isort]


pawamoy:
Explicit is better than implicit :smile:?

sisp:
I agree in general and don't insist on this change. It just felt a bit arbitrary because, e.g., `skip-string-normalization = false` isn't set explicitly either, not to mention the huge number of settings `isort` has most of which are left at their implicit defaults. :thinking:

yajo (approved):
Thanks for helping keeping the house clean.

sisp:
I'm glad to help. 🙂

PR #761, opened 2022-08-28
chore(pre-commit): remove "poetry run" wrapper for local hooks

sisp:
I think wrapping local hook commands with `poetry run <command>` in `.pre-commit-config.yaml` is not necessary because `pre-commit` is run within the context of the Poetry-managed virtual env already:

* Case 1: The Poetry-managed virtual env is activated:
   ```sh
   poetry shell

   git commit
   # or
   pre-commit run ...
   ```

* Case 2: `pre-commit` is run with the `poetry run <command>` wrapper explicitly:
   ```sh
   poetry run git commit
   # or
   poetry run pre-commit run ...
   ```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,19 +23,19 @@ repos:
     hooks:
       - id: autoflake
         name: autoflake
-        entry: poetry run autoflake
+        entry: autoflake
         language: system
         types: [python]
         args: ["-i", "--remove-all-unused-imports", "--ignore-init-module-imports"]
       - id: black
         name: black
-        entry: poetry run black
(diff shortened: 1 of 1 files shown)


yajo:
I'm gonna keep it as it is. Using `poetry run` within a poetry shell doesn't harm. OTOH if the dev is using `pre-commit` as installed outside the venv, it should make it work without problem too. Thanks anyway!

sisp:
> OTOH if the dev is using pre-commit as installed outside the venv, it should make it work without problem too. 

I believe so, too, but only if the dev already ran `poetry install` before. In any case, I'm fine with your decision. 👍

PR #759, opened 2022-08-28
build(isort): install with poetry

yajo:
pre-commit started failing with:

```
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: AssertionError: BUG: expected environment for python to be healthy immediately after install, please open an issue describing your environment

more info:

virtualenv python version did not match created version:
- actual version: 3.10.6.final.0
- expected version: 3.10.4.final.0
```

Quick fix while https://github.com/copier-org/copier/pull/696 gets on track.

Sadly I had to rollback https://github.com/copier-org/copier/pull/752 due to https://github.com/PyCQA/isort/issues/1945.

PR #762, opened 2022-08-28
chore(pre-commit): use language "system" for local isort hook

sisp:
I think local pre-commit hooks should use `language: system` because they are installed via Poetry instead of `pre-commit`. The way I read the [`pre-commit` docs on `language: python`](https://pre-commit.com/#python), the `language` tells `pre-commit` how to _install_ a hook. Since `isort` is now installed via Poetry, it's already installed and directly available to `pre-commit`.

Follow-up of #759.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,7 +48,7 @@ repos:
         require_serial: true
       - id: poetry_isort
         name: isort
-        entry: isort
+        entry: poetry run isort
         require_serial: true
         language: system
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -50,7 +50,7 @@ repos:
         name: isort
         entry: isort
```
yajo, line 51: You're right, and besides this is needed too:

```suggestion
        entry: poetry run isort
```

yajo (approved):

PR #763, opened 2022-08-29
chore(pre-commit): add pyupgrade hook

sisp:
I've added a pre-commit hook that runs [`pyupgrade`](https://github.com/asottile/pyupgrade) which automatically checks and updates Python syntax to newer versions of the language, so we can make sure we don't use unnecessarily old syntax which is often less convenient. I've configured `pyupgrade` to check for and update to Python 3.7 syntax which is Copier's minimum Python version at this point.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,6 +54,12 @@ repos:
         types_or: [cython, pyi, python]
         args: ["--filter-files"]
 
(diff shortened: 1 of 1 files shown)


sisp:
I'm not sure why Codecov is having some issue in so many of the checks. But I think those failures are unrelated to the change made in this PR.

PR #764, opened 2022-08-29
chore(pre-commit): rename isort hook ID

sisp:
I've renamed the `isort` pre-commit hook ID from `poetry_isort` to `isort` for better consistency with the other hook IDs. It's just a minor change, but I like consistency. :wink:


sisp:
I'm not sure why Codecov is having some issue in so many of the checks. But I think those failures are unrelated to the change made in this PR. Can you perhaps trigger the checks again?

PR #765, opened 2022-08-29
chore(commitizen): move config to pyproject.toml

sisp:
I've moved `commitizen`'s config from `.cz.toml` to `pyproject.toml` as that's [recommended for Python projects](https://commitizen-tools.github.io/commitizen/config/#pyprojecttoml-or-cztoml) and also nicer in my opinion.


sisp:
I think it's still worth it, even when it's just an intermediate config consolidation until #696 is ready.

PR #766, opened 2022-08-29
docs: update commitizen usage in contributing guide

sisp:
I've updated the instructions for using `commitizen` in `CONTRIBUTING.md`. Before, `commitizen` was listed as a prerequisite to be installed in the host environment. Also, the `cz commit` command was shown for making a commit. At the same time, `commitizen` is configured as pre-commit hook and the steps for making a contribution include enabling also the `commit-msg` hooks and activating the Poetry shell, so the `commitizen` pre-commit hook is run when executing the `git commit` command.

I think the purely pre-commit-based workflow is smoother than additionally requesting `commitizen` to be installed in the host environment. Also, if the `commitizen` version in the host environment differs from the pre-commit hook version, inconsistent behavior may occur which would be confusing for a developer.

Therefore, I've removed `commitizen` as a prerequisite to be installed in the host environment and replaced the `cz commit` command by `git commit` in `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,8 +52,6 @@ your host environment:
 
 -   [Poetry](https://python-poetry.org/) for packaging and dependency management
 -   [poethepoet](https://github.com/nat-n/poethepoet) for running development tasks
--   [commitizen](https://commitizen-tools.github.io/commitizen/) for writing Git commits
-    easily
 
 Or you can use
 [![Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/copier-org/copier)
@@ -109,7 +107,7 @@ Ready to contribute? Here's how to set up the project for local development.
 
     ```sh
     git add .
-    cz commit  # use `git commit` if you prefer, but this helps
+    git commit
     git push origin name-of-your-bugfix-or-feature
     ```
 


CONTRIBUTING.md
```diff
@@ -109,7 +107,7 @@ Ready to contribute? Here's how to set up the project for local development.
 
     ```sh
     git add .
-    cz commit  # use `git commit` if you prefer, but this helps
+    git commit
```
yajo, line 110: Currently we don't launch the commit message linter in our actions definitions.

I think it's much more simpler to ask contributors to run `cz commit` than asking them to read https://www.conventionalcommits.org/en/v1.0.0/ and produce a properly-formatted commit.

So I'd leave this as it was. It was also mentioning that you can use `git commit` if you prefer, so I think previous doc was better.

yajo (changes requested):

CONTRIBUTING.md
sisp: I'm not sure I can follow your line of argument. :confused:

> Currently we don't launch the commit message linter in our actions definitions.

How is this related to the contributor workflow? :thinking:

> I think it's much more simpler to ask contributors to run `cz commit` than asking them to read https://www.conventionalcommits.org/en/v1.0.0/ and produce a properly-formatted commit.

I agree it's better to lint the commit message than trust people to get the formatting right, reading documentation is still necessary though to make the linter pass. But when contributors follow the guide and install the `commit-msg` hooks, then `commitizen` will run as part of the pre-commit run also when using `git commit`. What's the advantage of using `cz commit` then?

yajo: > > Currently we don't launch the commit message linter in our actions definitions.
> 
> How is this related to the contributor workflow? 🤔 

If I get commits from contributors correctly formatted in conventional commits format, it makes my life easier. But since I still don't have a proper CI that checks that, I have to rely on contributors doing that properly by themselves. Otherwise I just squash on merge and reword the commit, but I prefer not to do that if possible.



> pass. But when contributors follow the guide and install the `commit-msg` hooks, then `commitizen` will run as part of the pre-commit run also when using `git commit`. What's the advantage of using `cz commit` then?

It's a matter of UX. I want it to be as much simple as possible to contribute appropriately.

When you rely on the pre-commit hook, you get a message like this buried into the rest of pre-commit checks:

```
commitizen check.........................................................Failed
- hook id: commitizen
- exit code: 14

commit validation: failed!
please enter a commit message in the commitizen format.
commit "": "wrong format"
pattern: (?s)(build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)(\(\S+\))?!?:( [^\n\r]+)((\n\n.*)|(\s*))?$
```

That doesn't help much.

However, when you use `cz c`, you get this:



It's IMHO a much better newbie experience. I don't personally use it because I write conventional commits directly, but I don't want newbies to have to read a full spec on how to write commits when there's this nice tool that makes it for them.

sisp: Thanks for the detailed explanation! :pray: Now I get what you mean. Alright, I agree, `cz commit` is worth using over `git commit`.

We _could_ consider making `commitizen` a dev dependency instead of assuming it's installed in the host environment, and then change the pre-commit hook to use the "system" binary. What do you think?

yajo: Seems good to me.

sisp:
I'm closing this PR in favor of #779.

PR #772, opened 2022-08-31
build(deps): add missing dependency "prompt-toolkit"

sisp:
I've added `prompt-toolkit` as a top-level dependency because it's imported in two modules, so Copier directly depends on it. Before, Copier happened to work fine despite this missing top-level dependency because `prompt-toolkit` is a dependency of `questionary`:

```shell-session
$ poetry show -a -t
...
questionary 1.10.0 Python library to build pretty command line user prompts ⭐️
└── prompt-toolkit >=2.0,<4.0
...
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -939,8 +939,8 @@ docs = ["mkdocs-material", "mkdocstrings"]
 
 [metadata]
 lock-version = "1.1"
-python-versions = ">=3.7,<4.0"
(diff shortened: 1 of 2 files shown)


yajo:
If questionary ever decided to use a different toolkit, a failure in our import would help us diagnose that, as our own code would have to change too. I mean that we only use prompt-toolkit because it's used by questionary. Also I don't want to care at all about the version, as long as it's compatible with questionary's version definition. So I don't think we should take the burden of maintaining it in our direct dependencies.

Thanks for the proposal anyway!

PR #776, opened 2022-09-01
refactor: add "flake8-simplify" plugin and simplify code

sisp:
I've added the Flake8 plugin [`flake8-simplify`](https://github.com/MartinThoma/flake8-simplify) and applied its simplification suggestions.

There's one exception though: [`SIM117`](https://github.com/MartinThoma/flake8-simplify#rules) which suggests merging several `with` statements that use the same scope. Applying this rule leads to some ugly formatting by Black (https://github.com/psf/black/issues/412, https://github.com/psf/black/issues/664).

I think this Flake8 plugin has identified some nice small improvements and contributes to maintaining a clean code base. [Poetry uses it, too](https://github.com/python-poetry/poetry/blob/a613347bf0a9fd7b7ec782cd45222b3a0f0e7401/.pre-commit-config.yaml#L40). :wink:


yajo (approved):
Thanks!

PR #779, opened 2022-09-02
build(deps-dev): add "commitizen" dev dependency

sisp:
I've added `commitizen` as a dev dependency, made the pre-commit hook use the locally installed `cz` tool, and adapted the contributing guide accordingly.

Supersedes #766 (see the conclusion of the discussion in https://github.com/copier-org/copier/pull/766#discussion_r961298028).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,7 +106,7 @@ Ready to contribute? Here's how to set up the project for local development.
 
     ```sh
     git add .
-    poetry run cz commit  # use `git commit` if you prefer, but this helps
+    cz commit  # use `git commit` if you prefer, but this helps
     git push origin name-of-your-bugfix-or-feature
     ```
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -108,7 +106,7 @@ Ready to contribute? Here's how to set up the project for local development.
 
     ```sh
     git add .
-    cz commit  # use `git commit` if you prefer, but this helps
+    poetry run cz commit  # use `git commit` if you prefer, but this helps
```
yajo, line 109: Following your own line of reasoning in https://github.com/copier-org/copier/pull/767, this should be left as it was, isn't it?

yajo (changes requested):

CONTRIBUTING.md
sisp: You're right, of course! :bow: I've reverted this change. Thanks! :+1:

yajo (approved):
Thanks!

PR #792, opened 2022-09-08
fix: Resolve path before checking Git bundle

### LINKED ISSUE #791: bug: Git bundle support breaks with relative paths
**Describe the problem**
Trying to pass a relative path to Git bundle results in an exception: 

```python
  File "/home/user/dev/project/__pypackages__/3.8/lib/copier/template.py", line 443, in local_abspath
    raise ValueError("Local template must be a directory.")
```

**To Reproduce**

1. create a Git bundle of a template:
    ```bash
    git clone https://github.com/pawamoy/copier-pdm
    cd copier-pdm
    git bundle create template.bundle --all
    ```
2. try to generate a project from it:
    ```bash
    copier template.bundle newproj
    ```

**Expected behavior**
Relative paths to Git bundles supported.

**Environment**
 - OS: Linux
 - Copier version: 6.1.0
 - Python version: 3.8
 - Installation method: pipx

**Additional context**
The issue was pretty easy to debug:

```python
def is_git_bundle(path: Path) -> bool:
    with TemporaryDirectory(prefix=f"{__name__}.is_git_bundle.") as dirname:
        with local.cwd(dirname):
            git("init")
            # BUG: here path is a relative path, so it fails because we just entered another directory
            return bool(git["bundle", "verify", path] & TF)
```

The solution is easy as well: just resolve path before passing it to `is_git_bundle`, or before entering the temporary directory (not sure which solution makes the most sense):

```diff
 def is_git_bundle(path: Path) -> bool:
+    path = path.resolve()
     with TemporaryDirectory(prefix=f"{__name__}.is_git_bundle.") as dirname:
         with local.cwd(dirname):
             git("init")
             return bool(git["bundle", "verify", path] & TF)
```

I'll send a PR right away since that's a very small change.

pawamoy:
Closes #791.

I won't be able to add a test right now, but can do later this week.

### DIFF
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -3,6 +3,7 @@ import os
 import re
 import sys
 import tempfile
+from contextlib import suppress
 from pathlib import Path
 from warnings import warn
 
@@ -46,6 +47,8 @@ def is_in_git_repo(path: StrOrPath) -> bool:
 
 def is_git_bundle(path: Path) -> bool:
     """Indicate if a path is a valid git bundle."""
+    with suppress(OSError):
+        path = path.resolve()
     with TemporaryDirectory(prefix=f"{__name__}.is_git_bundle.") as dirname:
         with local.cwd(dirname):
             git("init")
--- a/tests/test_vcs.py
+++ b/tests/test_vcs.py
@@ -50,6 +50,10 @@ def test_get_repo():
     assert get("git.myproject.org/MyProject") is None
     assert get("https://google.com") is None
 
+    assert (
+        get("tests/demo_updatediff_repo.bundle") == "tests/demo_updatediff_repo.bundle"
+    )
+
 
 def test_clone():
     tmp = vcs.clone("https://github.com/copier-org/copier.git")


pawamoy:
Ah, the test suite found a breaking test case 🚀 
Fixing it right now.

copier/vcs.py
```diff
@@ -46,6 +46,7 @@ def is_in_git_repo(path: StrOrPath) -> bool:
 
 def is_git_bundle(path: Path) -> bool:
     """Indicate if a path is a valid git bundle."""
+    path = path.resolve()
     with TemporaryDirectory(prefix=f"{__name__}.is_git_bundle.") as dirname:
         with local.cwd(dirname):
             git("init")
```
yajo, line 52: You can also do:

```suggestion
    with TemporaryDirectory(prefix=f"{__name__}.is_git_bundle.") as dirname:
        git("-C", dirname, "init")
        return bool(git["-C", dirname, "bundle", "verify", path] & TF)
```

Both solutions are good, so do as you prefer 😄

yajo (commented):
I'll set as draft until you get the test stuff. Thanks!

copier/vcs.py
pawamoy: Oh well, I think I prefer yours: the least directory changes, the better. And that removes the exception on Windows :+1:

pawamoy:
Failure seems unrelated to the change? Also, let me know if you'd like that I add a test for this (not trying to avoid it: I just can't do it from work and will have to do from home).

yajo:
Possibly a rebase will fix it.

Please add the test when you have time. No hurry.

pawamoy:
Well I found a way to do this from work 😄 

Turns out the `-C` flag changes the current directory as well so I reverted to the `path.resolve()` solution.
Also added a test, and confim it successfully failed without the patch.

pawamoy:
Butched up the rebase... that's what happens when I click on buttons instead of running git commands 😩

PR #802, opened 2022-09-14
fix: Make sure to delete local temporary clones

### LINKED ISSUE #709: Temp files being left over
Hi,
I'm noticing some temp files being left over when running copier
Typically under windows under the `C:\Users\username\AppData\Local\Temp` directory
such as `copier.vcs.clone.sxn86nzf`

I'm running copier using a template that's located on the disk as a directory
But I'm noticing it only happens for templates / directories that are setup as a git repo.
If there's no .git sub-directory inside the template directory then no temp directory is created (I'm guessing since it doesn't have to check the git tags)

Is this expected behaviour? or should copier be cleaning up it's temp dirs after running?

**Environment**
 - OS: Windows
 - Copier version: 6.1.0
 - Python version: 3.9.13
 - Installation method: pipx+pypi,

pawamoy:
Closes #709

I implemented that on GitPod.io but have a lot of test failures due to pexpect timeouts, and `pkg_resources` module not being found (it seems the env on gitpod is kinda broken). Opening to see how CI behaves.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -53,7 +53,7 @@ class Worker:
     This class represents the state of a copier work, and contains methods to
     actually produce the desired work.
 
-    To use it properly, instantiate it by filling properly all dataclass fields.
+    To use it properly, use it as a context manager and fill all dataclass fields.
 
     Then, execute one of its main methods, which are prefixed with `run_`:
 
@@ -61,6 +61,12 @@ class Worker:
     -   [run_update][copier.main.Worker.run_update] to update a subproject.
     -   [run_auto][copier.main.Worker.run_auto] to let it choose whether you
         want to copy or update the subproject.
+
+    Example:
+        ```python
+        with Worker(src_path="https://github.com/copier-org/autopretty.git", "output") as worker:
+            worker.run_copy()
+        ```
 
     Attributes:
         src_path:
--- a/copier/main.py
+++ b/copier/main.py
@@ -28,7 +28,7 @@ from questionary import unsafe_prompt
 from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
 from .subproject import Subproject
 from .template import Task, Template
-from .tools import Style, TemporaryDirectory, printf, handle_remove_readonly
+from .tools import Style, TemporaryDirectory, printf
 from .types import (
     AnyByStrDict,
     JSONSerializable,
@@ -153,13 +153,8 @@ class Worker:
     def __exit__(self, type, value, traceback):
         self._cleanup()
 
-    def _cleanup(self) -> None:
-        if self.template.is_temp_clone:
-            rmtree(
-                self.template.local_abspath,
-                ignore_errors=False,
-                onerror=handle_remove_readonly,
-            )
+    def _cleanup(self):
+        self.template._cleanup()
 
     def _answers_to_remember(self) -> Mapping:
         """Get only answers that will be remembered in the copier answers file."""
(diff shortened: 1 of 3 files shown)


pawamoy:
OK, I was able to fix the env on gitpod by upgrading pip and setuptools in both the pyenv version and the poetry venv.

pawamoy:
Network failure in job ubuntu 3.8, re-running it should fix it.

yajo:
Don't you think we need to fix this too? https://github.com/copier-org/copier/blob/890ac866a09bcb7dcc448e6360eeaf7e0f563c13/copier/vcs.py#L133

pawamoy:
Oh, that's exactly what gets deleted by the added code. Sorry for not making it clear. I'll update the PR description accordingly.

yajo:
Maybe a cleaner solution would be to add a `destination` argument to the `clone()` function. When calling it, we already provisioned the temporary directory. This way, the ownership of that variable belongs to the worker, and it would clean it up automatically, as I feel the cleanup should be in the temporary directory class better than in the worker.

pawamoy:
I'm not sure to follow.

> When calling it, we already provisioned the temporary directory.

Where would it be? I can't seem to find an attribute or property pointing to a temporary directory in the Template class.
If you meant to create the temporary directory above `clone`, in `Template.local_abspath`, then the temporary directory is still bound to the template and not the worker.

> the cleanup should be in the temporary directory class

That would be best indeed, but I don't see how we could use it in methods/properties as it's a context manager, and the temporary clone needs to exist for a longer time than the scope of the Template class' methods/properties.

copier/main.py
```diff
@@ -146,6 +146,20 @@ class Worker:
...
+    def __exit__(self, type, value, traceback):
+        self._cleanup()
+
+    def _cleanup(self) -> None:
+        if self.template.is_temp_clone:
+            rmtree(
+                self.template.local_abspath,
+                ignore_errors=False,
+                onerror=handle_remove_readonly,
+            )
```
yajo, line 161: I think the cleanup should exist in `Template`, which is the one that creates the directory. Then, here:

```suggestion
        self.template._cleanup()
```

copier/template.py
```diff
@@ -443,6 +443,10 @@ def local_abspath(self) -> Path:
             raise ValueError("Local template must be a directory.")
         return result.absolute()
 
+    @cached_property
+    def is_temp_clone(self) -> bool:
+        return self.local_abspath != Path(self.url).absolute()
```
yajo, line 448: I think you can just have a normal property called `_temp_clone`, which is filled up with the clone path inside `local_abspath()`, only if required.

Then add `_cleanup()` and remove that path recursively if it exists, just like you were already doing.

yajo (changes requested):
I've reviewed this deeper and now I understand better what you said and why this is a good design.

I think it's good for a class to provision the paths it needs and clean them up, so let's move that logic to `Template` itself. It should be:

1. Define `Template._cleanup()`.
2. Call it in `Worker.__exit__()`.
3. Define `Template._temp_clone` as explained below.

This way, the template creates the paths and provides the way to remove it.

Adding context manager support to `Worker` also seems to be the way to go. TBH I was thinking that implementing `__del__()` would be better, but I've now learned that can lead to other problems.

Please don't forget to add a couple of tests:
1. The temporary clone is removed.
2. Copying from a local repo is not removed.

Also the API docs to use `Worker` as a context manager.

Thanks!

PR #804, opened 2022-09-16
Fix bug with pydantic 1.10.2

barrywhart:
**NOTE:** All tests are passing locally, except for this one (`tests/test_normal_jinja2.py::test_jinja_defaults_v6_min_version - copier.errors.UnknownCopierVersionWarning: Cannot check Copier version constraint.
`), which I think is because I installed `copier` in editable mode.

This is a fix for the dependabot pydantic 1.10.2 issue surfaced in PR #785. I branched from the dependabot branch in order to see the overall GitHub CI results, but if tests pass, I can create a new PR with _just_ my changes.

The main change I made was to modify `Worker` class to use `pydantic.BaseModel` rather than `pydantic.dataclass`. Rationale:
  * The failing test in the dependabot PR occurred because `pydantic.dataclass` was not raising a `TypeError` when extra fields were passed to the constructor.
  * I attempted to forbid extra fields as described [here](https://stackoverflow.com/questions/71837398/pydantic-validations-for-extra-fields-that-not-defined-in-schema), but I was seeing behavior similar to that described [here](https://github.com/pydantic/pydantic/issues/986), where this does not work correctly for `pydantic.dataclass`.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -191,7 +191,7 @@ def test_flags_bad_data(data):
 def test_flags_extra_fails():
     key = "i_am_not_a_member"
     conf_data = {"src_path": "..", "dst_path": ".", key: "and_i_do_not_belong_here"}
-    with pytest.raises(TypeError):
+    with pytest.raises(ValidationError):
         copier.Worker(**conf_data)
 
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -206,7 +206,7 @@ def is_subdict(small, big):
 
 def test_worker_good_data(tmp_path):
     # This test is probably useless, as it tests the what and not the how
-    conf = copier.Worker("./tests/demo_data", tmp_path)
+    conf = copier.Worker(src_path="./tests/demo_data", dst_path=tmp_path)
     assert conf._render_context()["_folder_name"] == tmp_path.name
     assert conf.all_exclusions == ("exclude1", "exclude2")
     assert conf.template.skip_if_exists == ["skip_if_exists1", "skip_if_exists2"]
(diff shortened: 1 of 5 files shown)


barrywhart:
Some tests are failing. This one seems tricky: `tests/test_updatediff.py::test_updatediff`. Apparently, copier is not overwriting existing files when it should. Any ideas?

tests/test_config.py
```diff
@@ -191,7 +191,7 @@ def test_flags_bad_data(data):
 def test_flags_extra_fails():
     key = "i_am_not_a_member"
     conf_data = {"src_path": "..", "dst_path": ".", key: "and_i_do_not_belong_here"}
-    with pytest.raises(TypeError):
+    with pytest.raises(ValidationError):
```
barrywhart, line 194: Note the error type changed. Is this a concern?

barrywhart:
@yajo: Ready for review!

barrywhart:
About the failed Mac build, it says something about a dependent package being "yanked". If that's the error, it's likely related to the updated dependencies, not the code change.

Would you like me to recreate this PR without the package version changes? Or remove them from this PR?

yajo:
Thanks! Yes, the 3.11 builds break all the time, but forget about them because it's still not stable.

Thanks to your PR and clear description, I saw this is [a known breaking change in pydantic 1.10](https://github.com/pydantic/pydantic/releases/tag/v1.10.0). I think I've found a simpler fix I'm testing in #795. If that works out, we'll be able to close this PR. I couldn't get to that without your help, so thanks a lot anyways!

Let's see what the CI there says.

yajo:
Well, it seems to have worked, so let me close this one. Thanks everyone!

PR #812, opened 2022-09-19
refactor: model a task to execute using a dataclass

sisp:
I've introduced a Data Class `Task` to model a task to execute (used internally for [tasks](https://copier.readthedocs.io/en/v6.2.0/configuring/#tasks) and [migrations](https://copier.readthedocs.io/en/v6.2.0/configuring/#migrations)).

Before, a task was represented by a dict which is less descriptive and less type-safe. Also, there was some inconsistency between tasks and migrations in terms of where additional environment variables are specified that shall be set while executing a task command. Specifically, migrations had their additional environment variables set in the `Template.migration_tasks` property while tasks had their additional environment variables set in the `Worker.run_copy(...)` method.

The slightly less pretty way of [setting the `use_shell` flag in each branch of the if-else condition](https://github.com/copier-org/copier/pull/812/files#diff-51a5fc56dc37de31d15f74ed1458469179fe72a8b697cbbc83b2a8e8e508f90bL179-R184) in the `Worker._execute_tasks(...)` method cannot be avoided without introducing a type cast

```diff
-task_cmd = self._render_string(task_cmd)
+task_cmd = self._render_string(cast(str, task_cmd))
```

because it seems Mypy doesn't support control flow analysis of aliased conditions unlike, e.g., [TypeScript v4.4+](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#control-flow-analysis-of-aliased-conditions-and-discriminants).

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -27,7 +27,7 @@ from questionary import unsafe_prompt
 
 from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
 from .subproject import Subproject
-from .template import Template
+from .template import Task, Template
 from .tools import Style, TemporaryDirectory, printf
 from .types import (
     AnyByStrDict,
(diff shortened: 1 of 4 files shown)


yajo (approved):
Cool!

PR #820, opened 2022-09-24
docs: add `_secret_questions` setting

### LINKED ISSUE #809: `_secret_questions` setting is undocumented
The `_secret_questions` setting

https://github.com/copier-org/copier/blob/608654a03e748a3ed1b032c089e20b2811a0cfd3/tests/test_answersfile.py#L34-L35

is not documented. I think it's a redundant setting since a question supports the `secret` flag as well. In any case, the docs and capabilities should be in sync.

Which solution do you prefer?

1. Document this setting.
2. Remove support for this setting in the code.

sisp:
I've added documentation for the `_secret_questions` setting.

Closes #809.

### DIFF
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -834,7 +834,7 @@ Suppress status output.
 
 -   Format: `List[str]`
 -   CLI flags: N/A
--   Default value: N/A
+-   Default value: `[]`
 
 Question variables to mark as secret questions. This is especially useful when questions
 are provided in the [simplified prompt format](#questions). It's equivalent to
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -830,6 +830,26 @@ Suppress status output.
 
     Not supported in `copier.yml`.
 
+### `secret_questions`
+
+-   Format: `List[str]`
+-   CLI flags: N/A
+-   Default value: N/A
+
+Question variables to mark as secret questions. This is especially useful when questions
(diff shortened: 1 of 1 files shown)


docs/configuring.md
```diff
@@ -830,6 +830,26 @@ Suppress status output.
 
     Not supported in `copier.yml`.
 
+### `secret_questions`
+
+-   Format: `List[str]`
+-   CLI flags: N/A
+-   Default value: N/A
```
yajo, line 837:
```suggestion
-   Default value: `[]`
```

yajo (changes requested):
I think you need to change this little thing.

docs/configuring.md
sisp: Sounds good.

But then I think [`migrations`](https://copier.readthedocs.io/en/latest/configuring/#migrations), [`skip_if_exists`](https://copier.readthedocs.io/en/latest/configuring/#skip_if_exists), and [`tasks`](https://copier.readthedocs.io/en/latest/configuring/#tasks) also need an empty list as the default value. By the way, [`jinja_extensions`](https://copier.readthedocs.io/en/latest/configuring/#jinja_extensions) already has an empty list as the default value.

I can create a separate PR for the other corrections, if you agree that's how it should be.

sisp: WDYT about the other settings, @yajo?

yajo: Yes, we should correct them too 👍

PR #826, opened 2022-09-28
refactor: remove unreachable code

sisp:
I've removed unreachable code in an `if` condition that handles a [question choice format that isn't supported](https://github.com/copier-org/copier/blob/e8d6239f8711054cbe30ec0472f147e04479a199/tests/test_complex_questions.py#L61-L89).

The handled format is:

```yaml
choices:
  - name: ...  # <--
    value: ... # <--
  - ...

# or

choices:
  key:
    name: ...  # <--
    value: ... # <--
  ...
```

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -273,14 +273,10 @@ class Question:
         if isinstance(self.choices, dict):
             choices = list(self.choices.items())
         for choice in choices:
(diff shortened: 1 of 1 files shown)


yajo (approved):
Oops, thanks!

PR #832, opened 2022-09-30
refactor(typing): use abstract container types where possible

sisp:
I've replaced a few occurrences of concrete container types (`Dict` and `List`) by abstract container types (`Mapping` and `Sequence`) which is slightly better practice AFAIK. See, e.g.,

* https://blog.logrocket.com/understanding-type-annotation-python/#adding-type-hints-dictionaries
* https://blog.logrocket.com/understanding-type-annotation-python/#adding-type-hints-lists

for an interesting article on this topic.

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -58,7 +58,7 @@ from copier.tools import copier_version
 
 from .errors import UserMessageError
 from .main import Worker
-from .types import AnyByStrDict, List, OptStr
+from .types import AnyByStrDict, OptStr, StrSeq
 
 
 def handle_exceptions(method):
@@ -183,7 +183,7 @@ class CopierApp(cli.Application):
         list=True,
         help="Make VARIABLE available as VALUE when rendering the template",
     )
-    def data_switch(self, values: List[str]) -> None:
+    def data_switch(self, values: StrSeq) -> None:
         """Update [data][] with provided values.
 
         Arguments:
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -13,9 +13,9 @@ from typing import (
     Any,
     Callable,
     ChainMap as t_ChainMap,
-    Dict,
-    List,
+    Mapping,
     Optional,
+    Sequence,
     Union,
 )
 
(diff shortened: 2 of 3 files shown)


sisp:
Hm, I'll need to check why the tests are failing for Linux and macOS. Surprisingly, they're passing for Windows. 🤔

yajo:
I agree with the change, once the CI goes ✔️ .

Keep in mind that probably those annotations have a runtime effect with Pydantic. It might be the cause of the test problems.

sisp:
I think you're right that the problem originates from Pydantic. Pydantic v1.10.2 handles the `Mapping` type incorrectly the way I see it:

```python
from typing import Any, Dict, Mapping

from pydantic import parse_obj_as


@dataclass
class Model:
    m: Mapping[Any, Any]
    d: Dict[Any, Any]


v = [["key", "value"]]
x = parse_obj_as(Model, {"m": v, "d": v})
print(x.m)  # ['key']
print(x.d)  # {'key': 'value'}
```

This behavior of `Mapping` vs. `Dict` is inconsistent and unexpected IMO.

sisp:
I've reverted one usage of `Mapping` to `Dict` where Pydantic's behavior is odd. The tests are passing now; the one test that's failing is because of a Codecov upload issue.

sisp:
I've created an issue over at Pydantic to indicate the `Mapping` bug: https://github.com/pydantic/pydantic/issues/4588

sisp:
I've submitted a PR to Pydantic which fixes the `Mapping` bug: https://github.com/pydantic/pydantic/pull/4589

I suggest to proceed here without waiting for that PR to get merged because it's uncertain if and when it'll get merged and when a patch release for Pydantic will become available since they're working on Pydantic v2.

sisp:
It seems this bug won't be fixed in Pydantic v1 and parsing a list of pairs as a dict seems to be a coincidental feature (see https://github.com/pydantic/pydantic/issues/4588#issuecomment-1268180778).

yajo:
> the one test that's failing is because of a Codecov upload issue

Yet again... I can't believe how much often that fails... 🤦🏼‍♂️

PR #835, opened 2022-10-01
Expand tilde

### LINKED ISSUE #825: Expand ~ in answers' _src_path
**Describe the problem**
I have a template in a private git repository ($work's policy). I can't give access to the repository to others, so I went with the solution of packaging the git repo as a bundle inside of a Python package. Users can then install the Python package and use wrapper commands to generate or update projects.

The issue is then that `_src_path` in `.copier-answers.yml` is a local path **specific to the current user** (the Python package + git bundle are installed through pipx, and therefore end up under the user's home).

I tried to replace the user-specific part (`/home/user`) with `~`, but then the path is not detected as a git repo/bundle anymore.
Running an update results in `Updating is only supported in git-tracked templates.`

**Expected behavior**
`~` in `_src_path` should be expanded.

**Environment**
 - OS: Linux
 - Copier version: 6.2.0
 - Python version: 3.8
 - Installation method: pipx + pypi

**Additional context**
I'll send a PR. The fix is quite easy: use `.expanduser()` on paths received in `is_git_repo_root` and `is_git_bundle`. I don't think there's a way to expand higher in the call chain because we're not yet sure if it's a URL or a local path.
Oh and this could be considered a feature request rather than a bug, so feel free to change the associated label 🙂

pawamoy:
Closes #825

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -158,9 +158,12 @@ class Worker:
 
     def __exit__(self, type, value, traceback):
         if value is not None:
+            # exception was raised from code inside context manager:
+            # try to clean up, ignoring any exception, then re-raise
             with suppress(Exception):
                 self._cleanup()
             raise value
+        # otherwise clean up and let any exception bubble up
         self._cleanup()
 
     def _cleanup(self):
--- a/copier/main.py
+++ b/copier/main.py
@@ -157,6 +157,10 @@ class Worker:
         return self
 
     def __exit__(self, type, value, traceback):
+        if value is not None:
+            with suppress(Exception):
+                self._cleanup()
+            raise value
         self._cleanup()
 
     def _cleanup(self):
--- a/copier/template.py
+++ b/copier/template.py
@@ -207,9 +207,12 @@ class Template:
 
     @property
     def _temp_clone(self) -> Optional[Path]:
+        clone_path = self.local_abspath
+        original_path = Path(self.url).expanduser()
         with suppress(OSError):
-            if self.local_abspath != Path(self.url).expanduser().resolve():
-                return self.local_abspath
+            original_path = original_path.resolve()
+        if clone_path != original_path:
+            return clone_path
         return None
 
     @cached_property
--- a/copier/template.py
+++ b/copier/template.py
@@ -209,7 +209,7 @@ class Template:
     def _temp_clone(self) -> Optional[Path]:
         clone_path = self.local_abspath
         original_path = Path(self.url).expanduser()
-        with suppress(OSError):
+        with suppress(OSError):  # triggered for URLs on Windows
             original_path = original_path.resolve()
         if clone_path != original_path:
             return clone_path
--- a/copier/template.py
+++ b/copier/template.py
@@ -207,8 +207,9 @@ class Template:
 
     @property
     def _temp_clone(self) -> Optional[Path]:
-        if self.local_abspath != Path(self.url).expanduser().resolve():
-            return self.local_abspath
+        with suppress(OSError):
+            if self.local_abspath != Path(self.url).expanduser().resolve():
+                return self.local_abspath
         return None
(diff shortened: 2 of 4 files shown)


copier/vcs.py
```diff
@@ -68,23 +68,30 @@ def get_repo(url: str) -> OptStr:
...
-        url = "".join((url, GIT_POSTFIX))
-    return url
+    if is_git_url:
+        if url.startswith("git+"):
+            url = url[4:]
+        elif url.startswith("https://") and not url.endswith(GIT_POSTFIX):
+            url = "".join((url, GIT_POSTFIX))
+        return url
+
+    return url_path.as_posix()
```
pawamoy, line 94: Previously, `url_path` was never returned, and we needed to return it to make the rest of the flow work.
I've refactored a bit to make things more readable, and to return either `url` or `url_path` depending on whether the URL is a Git URL or a local path.

tests/test_vcs.py
```diff
@@ -75,3 +79,21 @@ def test_dont_remove_local_clone(tmp_path):
     with Worker(src_path=src_path, dst_path=tmp_path, defaults=True) as worker:
         worker.run_copy()
     assert exists(src_path)
+
+
+def test_update_using_local_source_path_with_tilde(tmp_path):
+    src_path = vcs.clone("https://github.com/copier-org/autopretty.git")
+    fake_user_path = f"~/{'/'.join(['..'] * len(Path.home().parts))}{src_path}"
```
pawamoy, line 86: Here we trick Copier by using `~` in the path but climbing up as much as needed to fall back on `/tmp`, where the temporary directory is. I believe we could have used `len(Path.home().parts) - 1`  but an extra `../` doesn't hurt.

pawamoy (commented):
Let me know if some parts aren't clear and if you'd like that I add my comments directly in the code.

pawamoy:
OMG Windows is such a PITA to work with :raised_eyebrow: 

I've added comments to the code to make it more clear. Ready for review.

copier/main.py
```diff
@@ -157,6 +157,10 @@ def __enter__(self):
         return self
 
     def __exit__(self, type, value, traceback):
+        if value is not None:
+            with suppress(Exception):
+                self._cleanup()
+            raise value
```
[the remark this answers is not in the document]
pawamoy: Comment added to the code.

copier/vcs.py
pawamoy: Refactored again to avoid computing unnecessary things.

tests/test_vcs.py
pawamoy: Comments added to the test itself.

copier/template.py
```diff
@@ -207,7 +207,7 @@ def _cleanup(self) -> None:
 
     @property
     def _temp_clone(self) -> Optional[Path]:
-        if self.local_abspath != Path(self.url).absolute():
+        if self.local_abspath != Path(self.url).expanduser().resolve():
```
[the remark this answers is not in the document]
pawamoy: Nothing to do here, it was just FYI.

pawamoy:
Tested locally:

```console
% poetry run copier ~/data/dev/copier-pdm output
<copier prompts and output>

% cd output 
% cat .copier-answers.yml 
_commit: 0.10.2
_src_path: /home/pawamoy/data/dev/copier-pdm
<rest of the file>

% sed -i 's#/home/pawamoy#~#' .copier-answers.yml
% git init .
% git add -A
% git commit -m initial

% cd ..
% poetry run copier -f update output
<copier output>

% cat output/.copier-answers.yml
_commit: 0.10.2
_src_path: ~/data/dev/copier-pdm
<rest of the file>
```

Tidle preserved!

yajo (approved):

PR #846, opened 2022-10-19
Merge answerfile options

stereobutter:
Hi there 👋 
This is my first PR here and it adds code, tests and docs for merging copier settings from multiple yaml documents included into `copier.yml`.

# Example

Given the following project layout
```

|-- copier.yml
|-- foo.txt
|-- component
     |-- component.yml
     |-- bar.txt
```

and the files 
```yaml
# copier.yml
...
_skip_if_exists:
  - foo.txt
---
!include ./component/component.yml
```

```yaml
# component.yml
...
_skip_if_exists:
  - component/bar.txt
...
```

the resulting value for `_skip_if_exists` will be the concatenated values from both documents i.e. `["foo.txt", "/component/bar.txt"]` instead the latest document overwriting the previous value i.e. `["/component/bar.txt"]`. 

# Details
This PR only adds logic for merging settings that are not order dependent i.e. `skip_if_exists`, `jinja_extensions`, `secret_questions` and `exclude`. For order dependent settings (such as  `tasks` or `migrations`) it would have to be decided what the order of execution should be. Do tasks from multiple document run in the order the documents are defined or in reverse? In any case this should probably go in another PR and I'd be happy to work on this.

### DIFF
--- a/copier/template.py
+++ b/copier/template.py
@@ -1,7 +1,7 @@
 """Tools related to template management."""
 import re
 import sys
-from collections import ChainMap
+from collections import ChainMap, defaultdict
 from contextlib import suppress
 from dataclasses import field
 from pathlib import Path
@@ -94,12 +94,28 @@ def load_template_config(conf_path: Path, quiet: bool = False) -> AnyByStrDict:
 
     try:
         with open(conf_path) as f:
-            flattened_result = deepflatten(
-                yaml.load_all(f, Loader=yaml.FullLoader),
-                depth=2,
-                types=(list,),
+            flattened_result = list(
+                deepflatten(
+                    yaml.load_all(f, Loader=yaml.FullLoader),
+                    depth=2,
+                    types=(list,),
+                )
             )
-            return dict(ChainMap(*reversed(list(flattened_result))))
+            merged_options = defaultdict(list)
+            for option in (
+                "_exclude",
+                "_jinja_extensions",
+                "_secret_questions",
+                "_skip_if_exists",
+            ):
+                for result in flattened_result:
+                    try:
+                        values = result[option]
+                    except KeyError:
+                        pass
+                    else:
+                        merged_options[option].extend(values)
+            return dict(ChainMap(dict(merged_options), *reversed(flattened_result)))
     except yaml.parser.ParserError as e:
         raise InvalidConfigFileError(conf_path, quiet) from e
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -341,14 +341,17 @@ cannot use Jinja templating in your answers.
 
 ### Include other YAML files
 
-The `copier.yml` file supports multiple documents. When found, they are merged (**not**
-deeply merged; just merged) and the latest one defined has priority.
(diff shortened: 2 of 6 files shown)


yajo:
Thanks for the contribution. May I am what's the problem you find with current behavior? It seems more advanced to me at first sight.

stereobutter:
> Thanks for the contribution. May I am what's the problem you find with current behavior? It seems more advanced to me at first sight.

A limitation of the current behavior (that this PR partially addresses) is that you cannot split a large `copier.yml` into multiple modular files where settings are concerned. You can split questions across multiple files but settings have to be globally defined.

At work I use `copier` to define base templates for our infrastructure from modular components (that are used in multiple such base templates). The top level `copier.yml` then just includes all the components' `copier.yml` files.

```
|-- copier.yml
|-- componentA  # 
|     |-- copier.yml 
|     |-- ...
|-- componentB
|     |-- copier.yml 
|     |-- ...
|-- ...
|-- componentN
```

A concrete problem we have is that each component may use different jinja extensions and that with the current behavior one needs to maintain the list of extensions needed in the top level `copier.yml` instead of each component defining the extensions it needs. This is also true for files we like to exclude or skip. Some definitions make sense at a global level (say excluding all `readme.md` files) but for component specific files it gets messy if they all need to maintained in the global `copier.yml`. This is especially true when components change (and multiple people are responsible) for some of the components. 

I hope that explanation makes sense? I also spotted an issue with the PR where the auto formatter seems to have messed up the docs changes I added. I will try and fix this tomorrow. Just let me know if you have any other questions and/or see any issue with the PR?

yajo:
Ah sorry, I seem to have understood your initial comment on the contrary. So, you mean that the *current* behavior is to replace lists, whereas the *new* behavior is to concatenate them, right?

stereobutter:
> 



> Ah sorry, I seem to have understood your initial comment on the contrary. So, you mean that the _current_ behavior is to replace lists, whereas the _new_ behavior is to concatenate them, right?

Yes, that is correct. The new behavior is to concatenate the lists for the settings `jinja_extensions`, `exclude`, `skip_if_exists` and `secret_questions`. For other settings the current behavior of replacing the settings is retained. I'd like to add merging of other settings, especially `tasks` and `migrations` in another PR if this one is accepted.

stereobutter:
I've fixed the docs formatting and added clearer language and an example for merging settings from multiple files. I've also changed the test slightly so that it checks that options are concatenated in the order they were given.

stereobutter:
@yajo thank you for your review and sorry to bother you again but could you trigger a rerun of the CI? Some of the required checks raise 
```Error: Codecov: Failed to properly upload: The process '/Users/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255``` 

[It appears there is no clear cause for this issue](https://github.com/codecov/codecov-action/issues/598) and I hope this is just a fluke that disappears on a rerun.

yajo:
Ah yes... codecov fails all the time, sorry. Retrying.

Issue #444, opened 2021-09-12
Dockerfile and a wrapper script to mount template and output directories as volumes...


masavini:
Hi,
the Dockerfile could look like this:
```dockerfile
FROM python:3.9-alpine

ARG USER=copier
ARG GROUP=copier
ARG UID=1000
ARG GID=1000

RUN addgroup -g "${GID}" "${GROUP}" && \
    adduser -h /home/"${USER}" -u "${UID}" -G "${GROUP}" -s /bin/bash -D "${USER}"


RUN set -ex; \
    apk add --no-cache --virtual .build-deps \
        gcc \
        musl-dev; \
    apk add --no-cache \
        bash \
        git \
        py3-virtualenv;

USER "${USER}"

ENV PATH=/home/${USER}/.local/bin:$PATH


RUN set -ex; \
    python3 -m pip install --user pipx; \
    pipx install copier; \
    pipx install invoke; \
    pipx install pre-commit; \
    pipx ensurepath;


USER root
RUN set -ex; \
    apk del .build-deps;

USER "${USER}"

ENTRYPOINT ["copier"]
```

I already built it and performed a few tests, it works pretty good.

It would be also nice to have a wrapper script to parse arguments and automatically mount template and output directories as volumes...

What do you think about?

Yajo:
Seems good to me, but I'm not eager to work on this. However if you want to contribute it, I will review it. Regarding the Dockerfile itself, before somebody opens the PR I'll ask you to change it:

- According to https://pythonspeed.com/articles/alpine-docker-python/ alpine isn't the best choice. Use `python:3.9-slim` instead.
- Also I wouldn't care about users and uids/gids. Whoever cares, can just `docker run --user 1000:1000 copier`
- The PR opened must apply https://github.com/Tecnativa/image-template to this repo. It provides lots of good stuff out of the box, like multiarch builds, auto-tagged builds, :edge, :latest... etc.
- Images should be pushed to ghcr.io, not docker hub (to avoid boring setup).
- Docs will have to be updated also.

mbwhite:
I was just thinking of doing something similar, but specifically to run with the alpha (that I'm assuming is the master branch?)

On the `image-template` is the process to clone this (copier) repo, create a branch.. then apply the template?

Yajo:
Yes, basically.

BTW since my last comment lots of things happened. I'd probably prefer to end using a nix-based image, once #585 is fixed. 

Of course that can land later.

yajo:
[...]

PR #847, opened 2022-10-20
feat: docker integration for copier

taiyeoguns:
Regarding issue #444 

- added Dockerfile
- updated Github actions
- updated documentation


yajo:
Thank you very much for the contribution! However, as explained in https://github.com/copier-org/copier/issues/444#issuecomment-1094354019, my plans are to fix nix support first and, later, use it to provide the images. 

Providing the image now with this different system could give more problems later on, so for now I'll reject this. I hope you understand. 

Thanks again!

PR #856, opened 2022-11-02
Disable git GPG commit signatures on temporary commits

lastmikoi:
Hi !

First time getting around using the tool, which I'm very excited about the lifecycle management side of things.

My git is globally configured to sign all my commits using GPG by default, as I want committing unsigned code to be an 'opt-out' in my environments.

However, since `copier` creates new temporary commits for internal reasons, it triggers a few GPG signature popups, so I looked into disabling that behaviour, as I have no wish to sign temporary commits and the commit commands fail if I don't.

It's not elegant - *especially* in the tests, where even more `init`s happen - but this patch basically inject a `commit.gpgsign=False` in each copier-created temporary repositories.

I'd welcome any feedback, as this is mostly a quick fix to allow me to continue working on my code template, but I figured it could be helpful to others !

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -716,6 +716,7 @@ class Worker:
                 git("add", ".")
                 git("config", "user.name", "Copier")
                 git("config", "user.email", "copier@copier")
+                git("config", "commit.gpgsign", "False")
                 # 1st commit could fail if any pre-commit hook reformats code
                 git("commit", "--allow-empty", "-am", "dumb commit 1", retcode=None)
                 git("commit", "--allow-empty", "-am", "dumb commit 2")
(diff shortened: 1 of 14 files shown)


yajo:
Why don't you just use a gpg agent? That way, you unlock the key once and the rest just works.

I use gpg signing and never had this problem.

lastmikoi:
Yeah I'm using gpg-agent, but for Signature it's configured to ask my smartcard's pin for each signature action (for Authentication, namely ssh, I put a longer cache time). I manually review all patch chunks before signature and use those signatures to secure some of my deployment's supply chains, so having my agent sign things blindly doesn't fit my needs.

yajo (approved):
OK, that makes total sense.

yajo:
I rebased it to get the CI fixes from #861.

PR #870, opened 2022-11-16
Add support for answers files in subdirectories

nilsdebruin:
Draft pull request, based on the discussion: https://github.com/copier-org/copier/discussions/860.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -725,6 +725,7 @@ class Worker:
             prefix=f"{__name__}.recopy_diff."
         ) as new_copy:
             old_worker = self._make_old_worker(old_copy)
+            old_worker.update_context = True
             old_worker.run_copy()
             recopy_worker = replace(
                 self,
@@ -762,6 +763,7 @@ class Worker:
             self._execute_tasks(
                 self.template.migration_tasks("before", self.subproject.template)
             )
+            self.update_context = True
             self._uncached_copy()
             # Try to apply cached diff into final destination
             with local.cwd(self.subproject.local_abspath):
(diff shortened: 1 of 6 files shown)


copier/main.py
```diff
@@ -154,6 +154,11 @@ class Worker:
 
         conflict:
             One of "rej" (default), "inline" (still experimental).
+
+        update_context:
+            When `True`, ignores answers file directory when updating.
+
+            See [update_context][].
```
yajo, line 161: Hello! I've reviewed this, but I cannot understand the patch. Quoting from the discussion:

> to be able to store the answer files in a sub-directory

So this is a rather small scope AFAICS.

What's this flag for, `update_context`? What relation does it have with the fact that an answers file is or not in a directory?

I'm a bit confused with this patch... 😅

yajo (changes requested):

copier/main.py
nilsdebruin: Hi Yajo! Sorry for the late response! When I was developing this, I saw that when updating an already created template, the paths for the answer file did not resolve to the correct directory compared to generating a new project. Therefore I included the update_context to account for this difference. Maybe I did not understand the difference and made the wrong choice, but let me know how I can help / improve!

yajo: I appreciate the effort you put into creating a patch for this issue. However, upon review, it seems that your patch does not address the issue at hand. It might be more effective to start fresh and consider writing a test first before implementing a solution. This can help ensure that the solution effectively addresses the issue.

nilsdebruin: Hi @yajo thanks for your feedback! I will close this pull request and re-think how I could potentially solve this.

Issue #889, opened 2022-12-06
`pip install copier` fails on windows


wilsonwaters:
**Describe the problem**
Running `pip install copier` with python 3.10 fails on Windows when the Microsoft C++ build tools package is not installed. I would prefer not install the heavy (7GB) c++ compiler installed just to use copier.

This seems to be caused by the [iteration_utilities package](https://pypi.org/project/iteration-utilities) lacking a python 3.10 wheel file in pypi and it [doesn't look like they are planning to generate wheel files for python 3.10](https://github.com/MSeifert04/iteration_utilities/issues/333) any time soon.

**To Reproduce**
1. Run `pip install copier` on Windows with python 3.10 and *without* visual c++ compiler installed

**Logs**

Running `pip install copier`
<details>

```
PS C:\Users\wilson> pip install copier
Collecting copier
  Using cached copier-7.0.1-py3-none-any.whl (34 kB)
Collecting iteration_utilities>=0.11.0
  Using cached iteration_utilities-0.11.0.tar.gz (124 kB)
  Preparing metadata (setup.py) ... done
Collecting pygments>=2.7.1
  Downloading Pygments-2.13.0-py3-none-any.whl (1.1 MB)
     ---------------------------------------- 1.1/1.1 MB 6.3 MB/s eta 0:00:00
Requirement already satisfied: plumbum>=1.6.9 in c:\users\wilson\appdata\local\programs\python\python310\lib\site-packages (from copier) (1.8.0)
Collecting pyyaml-include>=1.2
  Downloading pyyaml_include-1.3-py3-none-any.whl (20 kB)
Collecting questionary>=1.8.1
  Downloading questionary-1.10.0-py3-none-any.whl (31 kB)
Collecting jinja2-ansible-filters>=1.3.1
  Downloading jinja2_ansible_filters-1.3.2-py3-none-any.whl (18 kB)
Requirement already satisfied: packaging>=21.0 in c:\users\wilson\appdata\local\programs\python\python310\lib\site-packages (from copier) (21.3)
Collecting jinja2>=3.1.1
  Downloading Jinja2-3.1.2-py3-none-any.whl (133 kB)
     ---------------------------------------- 133.1/133.1 kB 4.0 MB/s eta 0:00:00
Collecting pydantic>=1.10.2
  Downloading pydantic-1.10.2-cp310-cp310-win_amd64.whl (2.1 MB)
     ---------------------------------------- 2.1/2.1 MB 6.3 MB/s eta 0:00:00
[...]

PR #890, opened 2022-12-06
Remove dependency on iteration utilities

wilsonwaters:
Solving #889 

Installing copier on Windows with python 3.10 and no c++ compiler fails. This is due to the dependent package iteration_utilities having no python 3.10 wheel file available on pypi.

Solution is to remove the dependency on iteration_utilities. Fortunately the only use of this package is for the `deepflatten` function which I've re-implemented directly in copier.

### DIFF
--- a/copier/template.py
+++ b/copier/template.py
@@ -12,7 +12,6 @@ from warnings import warn
 import dunamai
 import packaging.version
 import yaml
-from iteration_utilities import deepflatten
 from packaging.version import Version, parse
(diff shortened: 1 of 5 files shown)


wilsonwaters:
Added tests for new deepflatten function. Don't want to reduce the coverage %!

yajo:
Closing as explained in https://github.com/copier-org/copier/issues/889#issuecomment-1354250750.

Thank you very much! ❤️

PR #895, opened 2022-12-17
docs: add example of regular expression validator

mvanholsteijn:
As requested in https://github.com/copier-org/copier/discussions/892


yajo (approved):
Thanks! ❤️

PR #902, opened 2022-12-23
build(deps): fix `funcy` version specifier

sisp:
I've fixed `funcy`'s version specifier to use only a lower bound. Although it's not explicitly mentioned, it looks to me like `funcy` uses SemVer, so only a lower bound should be used.

While at it, I've also placed the `funcy` dependency in alphabetic order with the other dependencies.


yajo (approved):
Thanks!

PR #909, opened 2023-01-09
docs: update note on Jinja defaults for Copier 7+

sisp:
I've updated the docs related to backwards compatibility of the Jinja syntax defaults in older Copier versions which was dropped in Copier 7.


yajo (approved):
❤️ thanks again!

PR #910, opened 2023-01-09
fix: do not set git filter for local shallow clones

adrien-berchet:
When using `git >= 2.27`, the filter `--filter=blob:none` is used to make the cloning process faster. Nevertheless, this can cause troubles when cloning local repositories.

### DIFF
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -147,7 +147,12 @@ def clone(url: str, ref: OptStr = None) -> str:
     # Faster clones if possible
     if GIT_VERSION >= Version("2.27"):
         file_url = re.match("(file://)?(.*)", url).groups()[-1]
-        if not is_git_shallow_repo(file_url):
+        if is_git_shallow_repo(file_url):
+            warn(
+                f"The repository '{url}' is a shallow clone, this might lead to unexpected "
+                "failure or unusually high resource consumption."
+            )
+        else:
             _clone = _clone["--filter=blob:none"]
     _clone()
 
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -136,7 +136,11 @@ def clone(url: str, ref: OptStr = None) -> str:
     location = mkdtemp(prefix=f"{__name__}.clone.")
     _clone = git["clone", "--no-checkout", url, location]
     # Faster clones if possible
-    if GIT_VERSION >= Version("2.27"):
+    if (
+        GIT_VERSION >= Version("2.27")
+        and not url.startswith("file://")
+        and not Path(url).exists()
+    ):
         _clone = _clone["--filter=blob:none"]
     _clone()
 
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -12,7 +12,7 @@ from packaging.version import Version
 from plumbum import TF, ProcessExecutionError, colors, local
 from plumbum.cmd import git
 
-from .errors import DirtyLocalWarning
+from .errors import DirtyLocalWarning, ShallowCloneWarning
 from .tools import TemporaryDirectory
 from .types import OptBool, OptStr, StrOrPath
 
@@ -150,7 +150,8 @@ def clone(url: str, ref: OptStr = None) -> str:
         if is_git_shallow_repo(file_url):
             warn(
                 f"The repository '{url}' is a shallow clone, this might lead to unexpected "
-                "failure or unusually high resource consumption."
+                "failure or unusually high resource consumption.",
+                ShallowCloneWarning,
             )
         else:
             _clone = _clone["--filter=blob:none"]
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -45,6 +45,15 @@ def is_in_git_repo(path: StrOrPath) -> bool:
         return False
 
 
+def is_git_shallow_repo(path: StrOrPath) -> bool:
+    """Indicate if a given path is a git shallow repo directory."""
+    try:
+        git("-C", path, "rev-parse", "--is-shallow-repository")
+        return True
+    except (OSError, ProcessExecutionError):
+        return False
+
+
 def is_git_bundle(path: Path) -> bool:
     """Indicate if a path is a valid git bundle."""
     with suppress(OSError):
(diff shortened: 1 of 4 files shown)


yajo (approved):

adrien-berchet:
Hi @yajo 
Thanks for supporting this change.
I am not able to reproduce the error of the CI locally, do you have any idea about this?

adrien-berchet:
@yajo it should be good now (I tested with `git==2.39.0` and `git==2.25.1`). The tests failed because the initial repos were cloned with `--filter=blob:none` (since it is not a local repo), then they were cloned again locally (so without the `--filter=blob:none), but during the test the process has to be able to fetch the blobs, which was not possible since they were not fetched by the 1st clone. (not sure it's clear :thinking: )

tests/test_vcs.py
```diff
@@ -67,6 +67,17 @@ def test_clone():
...
 
+def test_local_clone():
+    tmp = vcs.clone("https://github.com/copier-org/copier.git")
+    assert tmp
+    assert exists(join(tmp, "README.md"))
+
+    local_tmp = vcs.clone(str(tmp))
+    assert local_tmp
+    assert exists(join(local_tmp, "README.md"))
+    shutil.rmtree(local_tmp, ignore_errors=True)
```
yajo, line 78: Don't you need to clear `tmp`  too?

copier/vcs.py
```diff
@@ -137,7 +137,11 @@ def clone(url: str, ref: OptStr = None) -> str:
     location = mkdtemp(prefix=f"{__name__}.clone.")
     _clone = git["clone", "--no-checkout", url, location]
     # Faster clones if possible
-    if GIT_VERSION >= Version("2.27"):
+    if (
+        GIT_VERSION >= Version("2.27")
+        and not url.startswith("file://")
+        and not Path(url).exists()
+    ):
```
yajo, line 144: What kind of problems were caused by local blobless clones?

adrien-berchet: I use Copier to test a custom template in CI and the blobless clones make `git clone` to use like 10GB of memory, which is quite annoying for a simple CI job (it uses about 1GB with this branch). This CI runs in a docker container on a local cluster, so the setup is a bit complex, and I don't really know why git does not like blobless repos in this case... But fetching blobs for local repos should not cause perf issues as local transfers are usually very fast in this case.

tests/test_vcs.py
adrien-berchet: I don't think so, the last error in CI is caused by paths in the Windows case

copier/vcs.py
yajo: That's weird because local cloning operations are common across all our test codebase. 

Usually `/tmp` is mounted as `tmpfs`, which means that its files exist in memory. Maybe your repo is too big? Could you try setting a different tmp directory? Or mounting another kind of filesystem there?

adrien-berchet: I performed a few more tests and I think I got it. The real issue is when copier tries to clone a shallow clone. This happens in CI because, by default, Gitlab creates a shallow clone of the tested repository and then I run Copier on this cloned repository.
To reproduce you can try the following commands:
```shell
rm -rf /tmp/test_copier;
mkdir /tmp/test_copier;
cd /tmp/test_copier;

git clone --depth=2 https://github.com/copier-org/autopretty.git autopretty;

git clone --no-checkout autopretty /tmp/test_copier/copy_1 --filter=blob:none;
```
As you will see, many warnings are reported and the process is much longer than expected. It also takes much more memory, though it's harder to monitor.
If you run the same script removing the `--filter=blob:none`, you will see that the second clone does not report any warning and is much faster.

adrien-berchet: I proposed a new patch to just check if the repository is a shallow clone or not.

adrien-berchet: Any comment on this @yajo ?

adrien-berchet: Yeah I see your point.
Nevertheless, I think it would be nice to at least raise a warning in such cases because:
* it can still work (if the shallow clone contains all the required commits).
* even if it works it can lead to unusually high resource consumption (that was my initial issue, which was quite hard to understand).
* if it fails it can be hard to debug.

adrien-berchet: (the warning would be in addition to the doc ofc)

yajo: Good idea! Wanna update this PR with those changes?

adrien-berchet: Sure, I pushed new commits with a warning and a new FAQ entry

copier/vcs.py
```diff
@@ -138,7 +147,14 @@ def clone(url: str, ref: OptStr = None) -> str:
     _clone = git["clone", "--no-checkout", url, location]
     # Faster clones if possible
     if GIT_VERSION >= Version("2.27"):
-        _clone = _clone["--filter=blob:none"]
+        file_url = re.match("(file://)?(.*)", url).groups()[-1]
+        if is_git_shallow_repo(file_url):
+            warn(
```
yajo, line 152: Please can you add a new warning based on this one and use it? That helps assertions in tests

https://github.com/copier-org/copier/blob/f8ae8d6e8ed8ec5333abe4b20ad53bc87123bff0/copier/errors.py#L102

tests/test_vcs.py
```diff
@@ -67,6 +69,42 @@ def test_clone():
...
+def test_shallow_clone(tmp_path, recwarn):
+    # This test should always work but should be much slower if `is_git_shallow_repo()` is not
+    # checked in `vcs.clone()`.
+    src_path = str(tmp_path / "autopretty")
+    git("clone", "--depth=2", "https://github.com/copier-org/autopretty.git", src_path)
+    assert exists(join(src_path, "README.md"))
+
+    if vcs.GIT_VERSION >= Version("2.27"):
+        with pytest.warns(
+            UserWarning,
```
yajo, line 92: Here too, if you match by class, there's no need to match by msg

copier/vcs.py
adrien-berchet: Sure, done

yajo:
Please could you rebase or merge with master and fix the conflicts? Thanks.

adrien-berchet:
> Please could you rebase or merge with master and fix the conflicts? Thanks.

Sure, I rebased.

yajo (approved):

adrien-berchet:
Thanks!

PR #927, opened 2023-01-18
Feature/add subdirectory for answersfile

nilsdebruin:
The following test has been added to be able to validate if a subdirectory for an answersfile is possible


tests/test_cli.py
```diff
@@ -39,6 +61,39 @@ def test_good_cli_run(tmp_path, template_path):
     assert answers["_src_path"] == str(template_path)
 
 
+def test_good_cli_run_dot_config(tmp_path, template_path_with_dot_config):
+    with local.cwd(str(template_path_with_dot_config)):
+        git_commands()
+
+    run_result = CopierApp.run(
+        ["--quiet", str(template_path_with_dot_config), str(tmp_path)],
```
yajo, line 69: You need this for the test to be correct.

```suggestion
        ["--quiet", "-a", ".config/altered-answers.yml", str(template_path_with_dot_config), str(tmp_path)],
```

yajo (changes requested):
There's a little bug in the test code. Other than that, this seems like a good starting point to develop the solution! 😉 

Could you please mark the PR as draft in the mean time?

nilsdebruin:
@yajo, many thanks for your check and feedback!

nilsdebruin:
I have added a new approach to solve the issue

copier/main.py
```diff
@@ -558,6 +558,13 @@ def _render_path(self, relpath: Path) -> Optional[Path]:
             if not part:
                 return None
             rendered_parts.append(part)
+
+        # The following logic was added based on the discussion: https://github.com/copier-org/copier/discussions/859
```
yajo, line 562: This is not really needed, `git blame` will give you that!

If you want, you can add that kind of comment in the commit message though.

copier/main.py
```diff
@@ -558,6 +558,13 @@ def _render_path(self, relpath: Path) -> Optional[Path]:
...
+        if len(rendered_parts) > 1 and str(self.answers_relpath).startswith(
+            rendered_parts[0]
+        ):
+            rendered_parts.pop(0)
```
yajo, line 566: This also doesn't seem like the fix you're looking for. If there were more files under `.config/`, they'd be affected too, right?

copier/main.py
nilsdebruin: Ok, I will remove this!

copier/main.py
nilsdebruin: Good point, I will see if I can make it more specific tailored to the answersfile!

copier/main.py
nilsdebruin: Done!

copier/main.py
nilsdebruin: I have changed the implementation and added more tests, so different config folder levels are tested as well

yajo:
Can you fix the conflicts please? So I can trigger the ci

nilsdebruin:
> Can you fix the conflicts please? So I can trigger the ci

Done!

yajo:
It took a bit longer because I couldn't understand why the fix worked. I pushed another commit with some cleanup and a useful comment for my future me. Merging, thanks!

Issue #929, opened 2023-01-20
Copier tries to render file even if the name is a conditional template that renders empty


rafalkrupinski:
**Describe the problem**

When:
- template file name is a condition,
- template contents is only valid if the condition is truthy (depends on a missing variable)
- the condition is falsy

Copier still tries to render the file and fails.

**Template**
filename: `{% if conditional %}file{% endif %}.jinja`
```jinja
{{ conditional.variable }}
```

**To Reproduce**

```python
run_copy(
  'template',
  'target_dir',
  'data':{'other_variables':...}
)
```

**Logs**

<details>

```
self = <jinja2.sandbox.SandboxedEnvironment object at 0x103ada380>, obj = Undefined, attribute = 'variable'

    def getattr(self, obj: t.Any, attribute: str) -> t.Union[t.Any, Undefined]:
        """Subscribe an object from sandboxed code and prefer the
        attribute.  The attribute passed *must* be a bytestring.
        """
        try:
>           value = getattr(obj, attribute)
E           jinja2.exceptions.UndefinedError: 'conditional' is undefined


```

</details>

**Expected behavior**

Copier shouldn't try to render a template file if a condition in name is falsy.

**Environment**
 - OS: Linux 5
 - Copier version: 7.0.1 & master
 - Python version: 3.10.7
 - Installation method: poetry, as a library

PR #930, opened 2023-01-20
Fix: remove template_suffix from rendered name before checking for em…

rafalkrupinski:
Copier allows template file names to render as empty, in which case it skips rendering the file.
However, before this patch, it does render the file, and only skips writing it.
When the file name checks for the presence of a variable, and that variable is then used inside the template content, Copier fails (as described in #929).


copier/main.py
```diff
@@ -551,20 +551,18 @@ def _render_path(self, relpath: Path) -> Optional[Path]:
         # With an empty suffix, the templated sibling always exists.
         if templated_sibling.exists() and self.template.templates_suffix:
             return None
+        if (
+            self.template.templates_suffix
+            and str(relpath).endswith(self.template.templates_suffix)
+        ):
```
yajo, line 557:
```suggestion
        if self.template.templates_suffix and is_template:
```

tests/test_conditional.py
```diff
@@ -0,0 +1,76 @@
+from pytest import TempPathFactory
```
yajo, line 1: Would you mind to please name the file `test_conditional_file_name.py`?

Just `test_conditional.py` is not much specific (we have a lot of "conditionals" to test).

yajo (approved):
It looks quite fine. Just a couple of details.

Also, it seems to me like this won't pass the linters. Please run `nix-shell --run 'pre-commit run -a'` to autofix it.

yajo (approved):

Issue #692, opened 2022-06-08
Add an OS specific helper for replacing characters for file separators


jkleve:
From [a discussion](https://github.com/copier-org/copier/discussions/683) we had about creating a directory structure, it was proposed to use `replace(".", "/")` in a task script. As this is specific to unix like filesystems it would be nice to have a helper that is OS specific. On Windows it will be `\` and on unix like OS' it'll be `/`.

The working example (unix specific) is

```sh
# task.sh.jinja
templates_dir=templates
src_dir=src/main/{{ package.replace(".", "/") }}
test_dir=src/test/{{ package.replace(".", "/") }}

mkdir -p "$src_dir"
mkdir -p "$test_dir"
```

A nice alternative, as proposed by @yajo & @pawamoy  would be 
```sh
# task.sh.jinja
templates_dir=templates
src_dir=src/main/{{ package.replace(".", copier_sep) }}
test_dir=src/test/{{ package.replace(".", copier_sep) }}

mkdir -p "$src_dir"
mkdir -p "$test_dir"
```

pawamoy:
I don't think the distinction between Linux and Windows is relevant: indeed, the point of this issue is to provide a way to create directory structures without tasks, and the suggestion is to provide a context variable (for example `copier_sep` or `os_sep`) that is *not* a real OS separator (because they are not allowed in filenames), and that Copier will handle as if it was an actual separator. So in any case, users will use a single variable, whatever the OS is, and Copier will run Python's mkdir, where `/` works on all systems anyway.

jkleve:
I see. You were pushing to add a feature where we can do this as a task in `copier.yml` instead of in a shell script?

pawamoy:
IIUC, what @yajo suggested is that Copier natively supports creating arbitrarily deep directories without tasks, yes.

rafalkrupinski:
umm... this works for me. You can use forward slash on Windows, no problem.
I just made a custom jinja extension with a global DIR_SEP='/' and Copier creates directory structure.

rafalkrupinski:
Oh, and I had to add one line to copier, as it doesn't create parent directories for rendered files.
I'll prepare a patch

rafalkrupinski:
In the PR 6e06df7 there's a test for this behaviour. It passes `sep: '/'` as part of `data`

rafalkrupinski:
@yajo @pawamoy ,
Do you think Copier should provide a separator variable to facilitate this?

[...]

PR #932, opened 2023-01-21
Make sure the parent directory of the target file exists

rafalkrupinski:
When rendering files, the filename template may hold forward slash characters, in which case Copier attempts to write a directory structure. But it doesn't ensure the directory structure exists before writing the file.
This patch fixes it.

Related to #692

### DIFF
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -352,15 +352,14 @@ def test_value_with_forward_slash(tmp_path_factory):
     src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
     build_file_tree(
         {
-            src / "{{ filename.replace('.', sep) }}.txt"
-            : "This is template.",
+            src
+            / "{{ filename.replace('.', _copier_conf.sep) }}.txt": "This is template.",
         }
     )
     copier.run_auto(
         str(src),
         dst,
         data={
-            "sep": "/",
             "filename": "a.b.c",
         },
     )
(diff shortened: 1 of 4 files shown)


tests/test_copy.py
```diff
@@ -346,3 +346,25 @@ def test_commit_hash(src_repo, tmp_path):
     copier.run_copy(str(src_repo), tmp_path)
     assert tmp_path.joinpath("tag").read_text() == "1.0"
     assert tmp_path.joinpath("commit").read_text() == commit
+
+
+def test_value_with_forward_slash(tmp_path_factory):
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    build_file_tree(
+        {
+            src / "{{ filename.replace('.', sep) }}.txt"
```
yajo, line 355:
```suggestion
            src / "{{ filename.replace('.', _copier_conf.sep) }}.txt"
```

tests/test_copy.py
```diff
@@ -346,3 +346,25 @@ def test_commit_hash(src_repo, tmp_path):
...
+            : "This is template.",
+        }
+    )
+    copier.run_auto(
+        str(src),
+        dst,
+        data={
+            "sep": "/",
```
yajo, line 363:
```suggestion
```

yajo (changes requested):
Thanks! Neat enhancement!

Just like suggested in the linked issue, please add a line below this one: https://github.com/copier-org/copier/blob/3d211e1dae25550a719edffd7cd92d79efb98339/copier/main.py#L244

Add:

```py
"sep": os.sep,
```

Then apply the suggestions to tests so it actually tests this new feature.

Finally, please drop a line or two in the docs to tell people `_copier_conf.sep` exists. I think this is the place: https://github.com/copier-org/copier/blob/3d211e1dae25550a719edffd7cd92d79efb98339/docs/creating.md?plain=1#L92-L100

tests/test_copy.py
rafalkrupinski: I mean to send the separator as a separate PR, after getting some feedback in #692 
But this works too

yajo (approved):

PR #938, opened 2023-01-26
feat: Preserve symlinks when copying templates

### LINKED ISSUE #937: allow to choose if symlinks should be copied as symlinks
**Describe the problem**
I have a copier template that has a `CHANGELOG.rst` file in the template root and a symlink to it in its `docs/` folder. After running `copier copy` the generated project ends up with a second copy of `CHANGELOG.rst` in the `docs/` folder instead of a symlink to it.

**Template**

https://github.com/AdrianFreundQC/copier-symlink-demo
Also as a zip here: [copier-symlink-demo-main.zip](https://github.com/copier-org/copier/files/10511794/copier-symlink-demo-main.zip)

**To Reproduce**

1. Run `copier copy https://github.com/AdrianFreundQC/copier-symlink-demo out`
2. Run `ls -l out/`
3. See that `symlink.txt` is not a symlink

**Logs**

<details>

```
$ copier copy https://github.com/AdrianFreundQC/copier-symlink-demo out
No git tags found in template; using HEAD as ref

Copying from template version 0.0.0.post1.dev0+9ac1db8
    create  .
    create  symlink.txt
    create  original_file.txt

$ ls -l out/
total 16
-rw-r--r--@ 1 adrian  staff  12 Jan 26 17:33 original_file.txt
-rw-r--r--@ 1 adrian  staff  12 Jan 26 17:33 symlink.txt
```

</details>

**Expected behavior**

Copier should preserve the symlink:
```
$ ls -l out/
total 8
-rw-r--r--@ 1 adrian  staff  12 Jan 26 17:33 original_file.txt
lrwxr-xr-x@ 1 adrian  staff  17 Jan 26 17:36 symlink.txt -> original_file.txt
```

**Environment**
 - OS: macOS Ventura 13.1
 - Copier version: copier 7.0.1
 - Python version: Python 3.11.0
 - Installation method: mamba+conda-forge

AdrianFreundQC:
Closes #937.

This PR makes copier preserve symlinks when copying a template instead of copying the file the symlink points to.

The old behavior can be restored by setting `_preserve_symlinks` to false in `copier.yml`.

If the name of a symlink is templated it will be rendered. The targets of symlinks will only be rendered if the symlink name ends with the template suffix (`.jinja` by default).

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -11,7 +11,7 @@ from functools import partial
 from itertools import chain
 from pathlib import Path
 from shutil import rmtree
-from typing import Callable, Iterable, Mapping, Optional, Sequence
+from typing import Callable, Iterable, Mapping, Optional, Sequence, Union
 from unicodedata import normalize
 
 from jinja2.loaders import FileSystemLoader
@@ -29,7 +29,7 @@ from questionary import unsafe_prompt
 from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
 from .subproject import Subproject
 from .template import Task, Template
-from .tools import Style, TemporaryDirectory, printf
+from .tools import Style, TemporaryDirectory, printf, readlink
 from .types import (
     AnyByStrDict,
     JSONSerializable,
@@ -302,7 +302,11 @@ class Worker:
         return bool(ask(f" Overwrite {dst_relpath}?", default=True))
 
     def _render_allowed(
-        self, dst_relpath: Path, is_dir: bool = False, expected_contents: bytes = b""
+        self,
+        dst_relpath: Path,
+        is_dir: bool = False,
+        is_symlink: bool = False,
+        expected_contents: Union[bytes, Path] = b"",
     ) -> bool:
         """Determine if a file or directory can be rendered.
 
@@ -311,6 +315,8 @@ class Worker:
                 Relative path to destination.
             is_dir:
                 Indicate if the path must be treated as a directory or not.
+            is_symlink:
+                Indicate if the path must be treated as a symlink or not.
             expected_contents:
                 Used to compare existing file contents with them. Allows to know if
                 rendering is needed.
@@ -321,7 +327,11 @@ class Worker:
         if dst_relpath != Path(".") and self.match_exclude(dst_relpath):
             return False
         try:
-            previous_content = dst_abspath.read_bytes()
+            previous_content: Union[bytes, Path]
+            if is_symlink:
+                previous_content = readlink(dst_abspath)
+            else:
+                previous_content = dst_abspath.read_bytes()
         except FileNotFoundError:
             printf(
                 "create",
@@ -519,6 +529,45 @@ class Worker:
             dst_abspath.write_bytes(new_content)
             dst_abspath.chmod(src_mode)
 
+    def _render_symlink(self, src_abspath: Path) -> None:
+        """Render one symlink.
+
+        Args:
+            src_abspath:
+                Symlink to be rendered. It must be an absolute path within
+                the template.
+        """
+        assert src_abspath.is_absolute()
+        src_relpath = src_abspath.relative_to(self.template_copy_root)
+        dst_relpath = self._render_path(src_relpath)
+        if dst_relpath is None:
+            return
+        dst_abspath = Path(self.subproject.local_abspath, dst_relpath)
+
+        src_target = readlink(src_abspath)
+        if src_abspath.name.endswith(self.template.templates_suffix):
+            dst_target = Path(self._render_string(str(src_target)))
+        else:
+            dst_target = src_target
+
+        if not self._render_allowed(
+            dst_relpath,
+            expected_contents=dst_target,
+            is_symlink=True,
+        ):
+            return
+
+        if not self.pretend:
+            # symlink_to doesn't overwrite existing files, so delete it first
+            if dst_abspath.is_symlink() or dst_abspath.exists():
+                dst_abspath.unlink()
+            dst_abspath.symlink_to(dst_target)
+            if sys.platform == "darwin":
+                # Only macOS supports permissions on symlinks.
+                # Other platforms just copy the permission of the target
+                src_mode = src_abspath.lstat().st_mode
+                dst_abspath.lchmod(src_mode)
+
     def _render_folder(self, src_abspath: Path) -> None:
         """Recursively render a folder.
 
@@ -540,6 +589,8 @@ class Worker:
         for file in src_abspath.iterdir():
             if file.is_dir():
                 self._render_folder(file)
+            elif file.is_symlink() and self.template.preserve_symlinks:
+                self._render_symlink(file)
             else:
                 self._render_file(file)
 
--- a/copier/template.py
+++ b/copier/template.py
@@ -447,6 +447,14 @@ class Template:
         return result
 
     @cached_property
+    def preserve_symlinks(self) -> bool:
+        """Know if Copier should preserve symlinks when rendering the template.
+
+        See [preserve_symlinks][].
+        """
+        return bool(self.config_data.get("preserve_symlinks", False))
+
+    @cached_property
     def local_abspath(self) -> Path:
         """Get the absolute path to the template on disk.
 
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -178,3 +178,17 @@ class TemporaryDirectory(tempfile.TemporaryDirectory):
     @staticmethod
     def _robust_cleanup(name):
         shutil.rmtree(name, ignore_errors=False, onerror=handle_remove_readonly)
+
+
+def readlink(link: Path) -> Path:
+    """A custom version of os.readlink/pathlib.Path.readlink.
+
+    pathlib.Path.readlink is what we ideally would want to use, but it is only available on python>=3.9.
+    os.readlink doesn't support Path and bytes on Windows for python<3.8
+    """
+    if sys.version_info >= (3, 9):
+        return link.readlink()
+    elif sys.version_info >= (3, 8) or os.name != "nt":
+        return Path(os.readlink(link))
+    else:
+        return Path(os.readlink(str(link)))
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1022,6 +1022,18 @@ Run but do not make any changes.
 
     Not supported in `copier.yml`.
 
+### `preserve_symlinks`
+
+-   Format: `bool`
+-   CLI flags: N/A
+-   Default value: `False`
+
+Keep symlinks as symlinks. If this is set to `False` symlinks will be replaced with the
+file they point to.
+
+When set to `True` and the symlink ends with the template suffix (`.jinja` by default)
+the target path of the symlink will be rendered as a jinja template.
+
 ### `quiet`
 
 -   Format: `bool`
--- a/tests/demo/symlink.txt
+++ b/tests/demo/symlink.txt
@@ -1,0 +1,1 @@
+aaaa.txt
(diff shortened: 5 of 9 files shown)


AdrianFreundQC:
I have just force-pushed some fixed. I can now confirm that this works using the template in the linked issue and breaks no existing tests.

I haven't yet written automated tests for this.

yajo:
I think we'll break a lot of workflows if this is not optional (even if it's enabled by default, I think there should be a way to disable it).

@sisp @pawamoy what do you think?

pawamoy:
I don't know about a lot of workflows, but just recently we had someone with a use-case for symlinks being rendered.

AdrianFreundQC:
Pushed again. Tests should no longer fail on python <3.10.

I can try making this an option. Probably only next week though.

sisp:
> This is actually a feature that allows you to reuse files in some ways and reduce code duplication. However indeed it would make sense to make it optional.
>
> &mdash; https://github.com/copier-org/copier/issues/937#issuecomment-1405775180

TBH, I would have expected that symlinks remain symlinks after generating a project. Would it be possible to "reuse files [...] and reduce code duplication" via [Jinja "include"](https://jinja.palletsprojects.com/en/3.1.x/templates/#include) instead? If it works, it only works for text files though but not for binary files. Or [Copier "tasks"](https://copier.readthedocs.io/en/latest/configuring/#tasks) could be used to copy files, although I admit it's less declarative than the current symlink approach or the Jinja "include" idea.

If making it optional is the way to go, I wonder whether it's sufficient to have a global setting or whether it should be configurable at a more granular level. :thinking:

sisp:
By the way, the current symlink behavior seems to be undocumented. This leads me to wonder:

1. How many Copier users actually rely on the current symlink behavior?
2. How bad would it be to introduce a breaking change to this undocumented feature? (Although it's certainly not a good idea to break people's templates, but relying on an undocumented feature is known to be dangerous.)

AdrianFreundQC:
My suggestion would be the following:
Preserve symlinks by default. This is what I would expect and what is in line with other tools such as cookiecutter. If this is too disruptive it might be possible to have it not preserve by default and change the default once 8.0 releases.

jinja templating in both the symlink name and the symlink target should be rendered (So you can for example have a symlink that points to a file with a templated name).

There should be an option in `copier.yml` to change whether or not to preserve symlinks. When it is off everything should behave as before.
I'm not sure if a CLI option should exist, but I tend towards no, as it would probably mostly lead to users breaking templates that were designed with one of the options in mind.

AdrianFreundQC:
I also just checked the top 20 public copier templates on github and found exactly one that relies on symlinks being replaced with the file they point to.

https://github.com/OCA/oca-addons-repo-template uses a symlink to copy the `.gitignore` from the template repo to  generated project.

Maybe a third option would be useful for some (Preserve all symlinks, preserve no symlinks, preserve only symlinks pointing to inside the generated project). Or maybe that would just be confusing and lead to more problems...

yajo:
> https://github.com/OCA/oca-addons-repo-template uses a symlink to copy the `.gitignore` from the template repo to generated project.

I happen to maintain that template, and some others with at least that same purpose. However that can be easily changed to using import.

I'd just add an option to copier.yaml: `_symlinks`  with choices `as_symlinks` or `as_files`.



> jinja templating in both the symlink name and the symlink target should be rendered (So you can for example have a symlink that points to a file with a templated name).

Makes total sense.

copier/main.py
```diff
@@ -513,6 +517,34 @@ def _render_file(self, src_abspath: Path) -> None:
...
+            return
+        dst_abspath = Path(self.subproject.local_abspath, dst_relpath)
+        if dst_abspath.is_dir():
+            return
+        src_mode = src_abspath.lstat().st_mode
+
+        src_target = readlink(src_abspath)
+        dst_target = self._render_path(src_target)
+        if dst_target is None:
+            return
```
yajo, line 534: imho the symlink should be copied raw always, when `preserve_symlinks == True`.

If it is absolute, it should be copied absolute. If it is relative, then relative. Template maintainers can create the symlink as they want, so it's better to lend them that responsability. Just document it and test it.

copier/template.py
```diff
@@ -435,6 +435,15 @@ def templates_suffix(self) -> str:
...
         return result
 
+    @cached_property
+    def preserve_symlinks(self) -> bool:
+        """Get whether to preserve symlinks
+
+        By default: `True`
+
+        See [preserve_symlinks][]"""
+        return bool(self.config_data.get("preserve_symlinks", True))
```
yajo, line 445: > I'd just add an option to copier.yaml: `_symlinks` with choices `as_symlinks` or `as_files`.

Sorry, I didn't see this code. Your option name seems better to me. And I like it more that it is a boolean. Please leave it as you've done, thanks!

copier/tools.py
```diff
@@ -178,3 +178,17 @@ def cleanup(self):
     @staticmethod
     def _robust_cleanup(name):
         shutil.rmtree(name, ignore_errors=False, onerror=handle_remove_readonly)
+
+
+def readlink(link: Path) -> Path:
+    """A custom version of os.readline/pathlib.Path.readline.
+
+    pathlib.Path.readline is what we ideally would want to use, but it is only available on python>=3.9.
+    os.readline doesn't support Path and bytes on Windows for python<3.8
```
yajo, line 187:
```suggestion
    """A custom version of os.readline/pathlib.Path.readlink.

    pathlib.Path.readlink is what we ideally would want to use, but it is only available on python>=3.9.
    os.readlink doesn't support Path and bytes on Windows for python<3.8
```

yajo (changes requested):

AdrianFreundQC:
I think this PR should be ready now.

copier/main.py
```diff
@@ -320,7 +321,10 @@ def _render_allowed(
         if dst_relpath != Path(".") and self.match_exclude(dst_relpath):
             return False
         try:
-            previous_content = dst_abspath.read_bytes()
+            if is_symlink:
+                previous_content = bytes(readlink(dst_abspath))
```
yajo, line 325: Seems less error-prone:

```suggestion
                previous_content = readlink(dst_abspath)
```

copier/main.py
```diff
@@ -300,6 +300,7 @@ def _render_allowed(
         self,
         dst_relpath: Path,
         is_dir: bool = False,
+        is_symlink: bool = False,
         expected_contents: bytes = b"",
```
yajo, line 304:
```suggestion
        expected_contents: Union[bytes, str] = b"",
```

copier/main.py
```diff
@@ -513,6 +517,36 @@ def _render_file(self, src_abspath: Path) -> None:
...
+
+        src_target = readlink(src_abspath)
+        if src_abspath.name.endswith(self.template.templates_suffix):
+            dst_target = Path(self._render_string(str(src_target)))
+        else:
+            dst_target = src_target
+
+        if not self._render_allowed(
+            dst_relpath,
+            expected_contents=bytes(dst_target),
```
yajo, line 536:
```suggestion
            expected_contents=dst_target,
```

copier/main.py
```diff
@@ -513,6 +517,36 @@ def _render_file(self, src_abspath: Path) -> None:
             dst_abspath.write_bytes(new_content)
             dst_abspath.chmod(src_mode)
 
+    def _render_symlink(self, src_abspath: Path) -> None:
+        assert src_abspath.is_absolute()
+        src_relpath = src_abspath.relative_to(self.template_copy_root)
```
yajo, line 522: I think I commented it previously... Absolute links should be kept absolute IMHO.

copier/main.py
```diff
@@ -513,6 +517,36 @@ def _render_file(self, src_abspath: Path) -> None:
             dst_abspath.write_bytes(new_content)
             dst_abspath.chmod(src_mode)
 
+    def _render_symlink(self, src_abspath: Path) -> None:
```
yajo, line 520: Please add a docstring to the method, explaining clearly what it does. It will help my future self when maintaining this 😄

copier/main.py
```diff
@@ -513,6 +517,36 @@ def _render_file(self, src_abspath: Path) -> None:
...
+        assert src_abspath.is_absolute()
+        src_relpath = src_abspath.relative_to(self.template_copy_root)
+        dst_relpath = self._render_path(src_relpath)
+        if dst_relpath is None:
+            return
+        dst_abspath = Path(self.subproject.local_abspath, dst_relpath)
+
+        src_target = readlink(src_abspath)
+        if src_abspath.name.endswith(self.template.templates_suffix):
+            dst_target = Path(self._render_string(str(src_target)))
```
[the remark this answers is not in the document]
pawamoy: Yep, docs absolutely needed, especially since there's a new config option `_preserve_symlinks`.

copier/main.py
AdrianFreundQC: This doesn't change the target of the symlink. This is the path where the symlink itself is.
The handling of the symlink target is in the code block below

copier/main.py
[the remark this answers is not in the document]
AdrianFreundQC: Is it enough to add it to the `preserve_symlinks` option documentation or should I also add it somewhere else in the docs? If so, where would it fit best?

copier/main.py
AdrianFreundQC: I used `Union[bytes, Path]` instead.

copier/template.py
```diff
@@ -435,6 +435,15 @@ def templates_suffix(self) -> str:
             return DEFAULT_TEMPLATES_SUFFIX
         return result
 
+    @cached_property
+    def preserve_symlinks(self) -> bool:
+        """Get whether to preserve symlinks
+
+        By default: `True`
+
+        See [preserve_symlinks][]"""
```
yajo, line 444: To fit with style:

```suggestion
        """Know if Copier should preserve symlinks when rendering the template.

        See [preserve_symlinks][].
    """
```

copier/template.py
yajo, line 445: I think that we should default to `False`.

There are a lot of templates that use this feature as it was implemented before. At least I have a few myself. I also think it's a good default.

```suggestion
        return bool(self.config_data.get("preserve_symlinks", False))
```

This will affect some of the tests.

docs/configuring.md
```diff
@@ -854,6 +854,18 @@ Run but do not make any changes.
 
     Not supported in `copier.yml`.
 
+### `preserve_symlinks`
+
+-   Format: `bool`
+-   CLI flags: N/A
+-   Default value: `True`
```
yajo, line 861: As above:

```suggestion
-   Default value: `False`
```

copier/main.py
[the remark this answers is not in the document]
yajo: I see it OK with the docs you already added. Thanks!

yajo (changes requested):
Thanks for this enhancement. It is very well structured! And special thanks for the tests!

Some comments below.

copier/template.py
AdrianFreundQC: Changed default to false and adjusted tests. Most tests test the default (False) behavior. All tests testing the True behavior are in `test_symlinks.py`

AdrianFreundQC:
I addressed all the review comments and rebased because of a conflict. I'm not yet sure why one test fails when run with `flake check`, but passes when run  with `poe test`, but I can reproduce it locally and am looking into it.

AdrianFreundQC:
Tests pass again. nix flake didn't preserve the symlink in the demo template while copying it to the nix environment. The test couldn't preserve the symlink because it wasn't there in the first place.

I removed the symlink from the demo template again and instead added more fine-grained tests, so this has 100% diff coverage again.

I'm not familiar enough with nix to know if this is a bug or expected behavior.

AdrianFreundQC:
I just rebased this to fix the merge conflicts. Any update on when this might get merged or get another review round?

yajo (approved):
Excellent, thanks!

Issue #934, opened 2023-01-23
Updatable templates for template hierarchies and reuse


sisp:
**Is your feature request related to a problem? Please describe.**

I'd like to create a Copier base template (e.g. a general Python project template) and Copier child templates (e.g. a Python web app project template using FastAPI, a Python ML library template using PyTorch, etc.). The Copier child templates would mostly extend the Copier base template (e.g. add new files, add content to existing files from the Copier base template) but might also overwrite files/folders in the Copier base template (e.g. modify content in existing files from the Copier base template) or even delete files from the Copier base template (although this might be a rare case). Naively, I could create independent Copier templates, but then maintenance wouldn't scale because I'd have to sync the common parts manually. Instead, I would like to update a Copier child template with the changes of its Copier base template in the same way as I can update a generated project with the changes of its associated Copier template.

**Describe the solution you'd like**

I know this topic has been discussed a few times already (#79, #402, #416), but I haven't found a satisfactory solution and came up with a (to my knowledge) new approach that I'd like to suggest and discuss as it doesn't seem possible out of the box.

To clarify possibly confusing terminology upfront, I'd like to explicitly distinguish between *Copier templates* (i.e. project templates) and *Jinja templates* (i.e. content templates used in files or file/folder names).

So far, I've seen this topic under the term "template inheritance" which (to me) sounds similar to the idea of, e.g., [OverlayFS](https://en.wikipedia.org/wiki/OverlayFS) where multiple filesystem sources are merged, possibly enhanced with [Jinja inheritance](https://jinja.palletsprojects.com/en/3.1.x/templates/#template-inheritance) such that an overwritten file may actually be an extension of the base file rather than a complete replacement. I believe #79 tried to implement such an approach, but it was not merged for apparently good reason.

I've been giving this problem some thought for a while and believe that the word "inheritance" might be misleading. Instead of trying to inherit from a Copier base template, I think a Copier child template should be a (modified) copy of the Copier base template containing a reference to the Copier base template similar to how a generated project has a reference to its corresponding Copier template in `.copier-answers.yml`. Then, a very similar update mechanism as for updating a generated project could be used to update a Copier child template with the changes of its associated Copier base template:

```mermaid
graph TD
%% nodes ----------------------------------------------------------
base_template_repo("base template repository")
base_template_current("/tmp/base_template<br>(current tag)")
base_template_latest("/tmp/base_template<br>(latest tag)")
child_template_regen("/tmp/child_template<br>(fresh, current version)")
[...]

PR #941, opened 2023-01-30
feat: ignore default excludes when using subdirectory

sisp:
I've changed the behavior of the default values of the [`exclude`](https://copier.readthedocs.io/en/latest/configuring/#exclude) list such that the default values are not applied when a subdirectory is specified in `copier.yml`. See https://github.com/copier-org/copier/issues/934#issuecomment-1407444680 for the original discussion.

Note that the default values remain active when the subdirectory is explicitly set to `""` or `"."` because I think this is identical to omitting the `subdirectory` setting and seems like a sensible behavior to me. WDYT, @yajo?

Also, I think it's actually a breaking change because the implicit behavior is changed and Copier template developers might need to add an explicit `exclude` list if they rely on the default values while using a subdirectory. Should this PR be held back until a major release is planned? WDYT, @yajo?

### DIFF
--- a/copier/template.py
+++ b/copier/template.py
@@ -300,7 +300,12 @@ class Template:
 
         See [exclude][].
         """
-        return tuple(self.config_data.get("exclude", DEFAULT_EXCLUDE))
+        return tuple(
+            self.config_data.get(
+                "exclude",
+                [] if Path(self.subdirectory) != Path(".") else DEFAULT_EXCLUDE,
+            )
+        )
 
     @cached_property
     def jinja_extensions(self) -> Tuple[str, ...]:
--- a/copier/template.py
+++ b/copier/template.py
@@ -303,7 +303,7 @@ class Template:
         return tuple(
             self.config_data.get(
                 "exclude",
-                [] if Path(self.subdirectory) != Path(".") else DEFAULT_EXCLUDE,
+                DEFAULT_EXCLUDE if Path(self.subdirectory) == Path(".") else [],
             )
         )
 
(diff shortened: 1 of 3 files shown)


copier/template.py
```diff
@@ -300,7 +300,12 @@ def exclude(self) -> Tuple[str, ...]:
 
         See [exclude][].
         """
-        return tuple(self.config_data.get("exclude", DEFAULT_EXCLUDE))
+        return tuple(
+            self.config_data.get(
+                "exclude",
+                [] if Path(self.subdirectory) != Path(".") else DEFAULT_EXCLUDE,
```
yajo, line 306: Would you please invert this condition? I've got [something personal](https://www.recallstack.icu/en/2020/09/01/do-not-use-negative-booleans-please/) against negative conditionals 😂

yajo (approved):
Great! Just one little nitpicking request, I hope you don't mind 🙃

copier/template.py
sisp: Sure thing, makes sense! :+1: I'll try to remember it. :wink:

yajo (changes requested):
The patch looks good to me. However I just realized that this needs a docs update. Could you please add that last step? Sorry for not realizing earlier...

sisp:
No worries, I had forgotten about the docs as well. I've pushed a commit with a short info admonition in the `exclude` section. Do you think it's sufficient? Or would you like me to elaborate further, perhaps with an example, and/or by extending

> - Default value: `["copier.yaml", "copier.yml", ...]`

to something like

> - Default value: `["copier.yaml", "copier.yml", ...]`, or `[]` if `subdirectory` is set

?

sisp:
I think the failing CI job isn't related to the changes in this PR.

yajo (approved):

PR #942, opened 2023-02-02
fix: Remove useless is_dir check

AdrianFreundQC:
This check is no longer needed after 7e0def672998 (fix: don't attempt to render a file if its name is empty) has been added. That commit also made this line no longer be covered by the tests.

This check made copier skip copying a file with an empty name (The path would then point to the directory containing the file).


yajo (approved):
Thanks!

Issue #956, opened 2023-02-08
Empty answer to `type: str` text question without default value is `None` but should be `""`


sisp:
**Describe the problem**

The result of an empty answer to a `type: str` text question without a default value is `None`. This behavior is unexpected and inconsistent with the behavior other (text input) questions.

**Template**

```yaml
string_question:
  type: str
```

*Note:* If a default value is set

```yaml
string_question:
  type: str
  default: the default value
```

then the result of an empty answer is `""`.

**To Reproduce**

1. Run `copier /path/to/template /path/to/project`
2. Answer nothing, i.e. just press "Enter".
3. See the content of `.copier-answers.yml`:

    ```yaml
    _src_path: /path/to/template
    string_question: null
    ```
4. If you add a `.jinja` file to the template with content like

    ```
    answer: {% if string_question == None %}None{% else %}"{{ string_question }}"{% endif %}
    ```

    then the content of the rendered file will be:

    ```
    answer: None
    ```

**Expected behavior**

An empty answer should result in an empty string `""`.

**Environment**
 - OS: Ubuntu 20.04
 - Copier version: `copier 7.0.1`
 - Python version: 3.9.13
 - Installation method: pipx+pypi

**Additional context**

I think the current behavior is incorrect for several reasons:
1. It is inconsistent with other question types (`int` and `float`) where an empty answer is not accepted with an error `Invalid input`.
2. It is inconsistent with all other question types that have no answer value with the semantics of "not answered" (if that's the intention of `None` for `type: str` questions).
3. It is inconsistent with a `type: str` question that has a default value as the result of an empty answer is `""` and not `None`.
4. It is not possible to answer a `type: str` question with an empty string which may be a perfectly valid answer.

sisp:
I think this issue also questions the definition of the `-l, --default` flag

```
-l, --defaults                     Use default answers to questions, which might be null if not specified.
```

along with Copier's current behavior when the no default answer is available and the value is `null`. This behavior is inconsistent with with the interactive questionnaire where it isn't possible to provide a `null` answer, i.e. questions without a default value *must* be answered. The way I see it, the `-l, --default` flag only makes sense to use when _all_ questions have default answers, and when it is used while some questions don't have default answers then Copier should raise an error.

WDYT, @yajo?

sisp:
If a default value was required (which is the case for the short question syntax), then this problem would be solved, too. But requiring default values would be a breaking change, and there might not always be natural default values.

yajo:
[...]

PR #958, opened 2023-02-10
fix: require answer for questions without default value

### LINKED ISSUE #355: Mandatory answer and weird behavior with no default value
The documentation says: `default: Leave empty to force the user to answer.`

When not specifying the `default` key, the default value is `None`; perfect.  
But the prompt shows `🎤 [None]:`.  
I think it should be `🎤 :` maybe with above that `VAR_NAME? Required. Format: str`

When pressing enter, it should not accept an empty answer and should ask again (i.e. force the user to enter a value… as mandatory, and as described in the doc)

Currently, if pressing enter, that variable is `None` and using it in another default value "as a string", if becomes the string`"None"`… which is a known "problem" with Jinja (or Django templates at least).  
Wouldn't it be better to use an empty string `""` rather than `None`?

sisp:
I've fixed the behavior for questions without a default value when no answer is provided in the interactive questionnaire.

The new behavior is:

- `type: str` question: `""`
- `type: yaml` question: `None` (result of `yaml.safe_load("")`)
- `type: int`/`type: float`/`type: json` question: Validation error `Invalid input`
- `type: bool` question: `False` (implicit default)
- `choice: ...` question: The first choice (implicit default).

I think this new behavior is more intuitive and expected than the current one.

Fixes #355 #956.

Note: This PR does not change the behavior when the `--default` flag is passed because it is not yet clear how to deal with questions that lead to a validation error when no default value exists.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -31,6 +31,7 @@ from .subproject import Subproject
 from .template import Task, Template
 from .tools import Style, TemporaryDirectory, printf
 from .types import (
+    MISSING,
     AnyByStrDict,
     JSONSerializable,
     OptStr,
@@ -382,13 +383,14 @@ class Worker:
         for question in questions:
             # Display TUI and ask user interactively only without --defaults
             try:
-                new_answer = (
-                    question.get_default()
-                    if self.defaults
-                    else unsafe_prompt(
+                if self.defaults:
+                    new_answer = question.get_default()
+                    if new_answer is MISSING:
+                        raise ValueError(f'Question "{var_name}" is required')
+                else:
+                    new_answer = unsafe_prompt(
                         [question.get_questionary_structure()], answers=result.combined
                     )[question.var_name]
-                )
             except KeyboardInterrupt as err:
                 raise CopierAnswersInterrupt(result, question, self.template) from err
             previous_answer = result.combined.get(question.var_name)
--- a/copier/types.py
+++ b/copier/types.py
@@ -2,7 +2,17 @@
 
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, Mapping, Optional, Sequence, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Mapping,
+    NewType,
+    Optional,
+    Sequence,
+    TypeVar,
+    Union,
+)
 
 from pydantic.validators import path_validator
 
@@ -37,6 +47,8 @@ T = TypeVar("T")
 JSONSerializable = (dict, list, str, int, float, bool, type(None))
 VCSTypes = Literal["git"]
 Env = Mapping[str, str]
+MissingType = NewType("MissingType", object)
+MISSING = MissingType(object())
 
 
 class AllowArbitraryTypes:
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -452,9 +452,6 @@ def load_answersfile_data(
 
 def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
     """Cast answer to expected type."""
-    # Skip casting None into "None"
-    if type_fn is str and answer is None:
-        return answer
     try:
         return type_fn(answer)
     except (TypeError, AttributeError):
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -289,8 +289,6 @@ class Question:
 
     def filter_answer(self, answer) -> Any:
         """Cast the answer to the desired type."""
-        if answer == self.get_default_rendered():
-            return self.get_default()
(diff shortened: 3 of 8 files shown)


yajo:
For int and float, the default could be 0. And for Json, null. If we do it like that, we avoid the errors. What do you think?

sisp:
After giving this some more thought and reading https://github.com/copier-org/copier/issues/355#issuecomment-1019069511 again, I think whether the correct/consistent solution is this:

The `--defaults` flag (or equivalent argument via API) can only be used when the `default` field is set for *all* questions. Otherwise, an error is raised. Questions without a `default` field are required and must be answered either interactively or via the `--data` flag (or equivalent argument via API). In interactive mode, there are some implicit initial values/selections:

* `str` question: `""` because this is the best initial value for the input prompt.
* `bool` question: `False` because [`questionary`'s confirm question requires a default value](https://github.com/tmbo/questionary/blob/87a891c41a0199bb23b2b230e971fd15db32c17b/questionary/prompts/confirm.py#L22) and `False` seems better than `True` IMO.
* `choice: ...` question: "The first choice" because `questionary` doesn't support no initial choice and falls back to the first choice if no default value is provided.

For `int`/`float`/`yaml`/`json` questions, I'd initialize the input prompt empty and require a valid answer - also for `yaml` where I'd *not* treat an empty string or only spaces as `null` although `yaml.safe_load("")` behaves like that because it is inconsistent with `json` and not intuitive IMO.

That said, I don't agree with

> [...] but an empty string in the default field should not imply required=true since in some cases the default value may be an empty string.
>
> &mdash; https://github.com/copier-org/copier/issues/355#issuecomment-1019069511

because `default: ""` is a perfectly fine value or a `str` question and I see no reason why the question should ever imply `required: true` then. In fact, the presence of a `default` field implies an optional question and vice versa, so having both a `default` field and a `required` field is redundant.

WDYT?

yajo:
> The `--defaults` flag (or equivalent argument via API) can only be used when the `default` field is set for _all_ questions. Otherwise, an error is raised. Questions without a `default` field are required and must be answered either interactively or via the `--data` flag (or equivalent argument via API).

So, `--defaults` will work as long as all questions without defaults are answered via `--data`, right?

> because `default: ""` is a perfectly fine value or a `str` question and I see no reason why it should ever imply `required: true` then. In fact, the presence of a `default` field implies an optional question and the absence of a `default` field implies a required field, so having both a `default` field and a `required` field is redundant.

I agree. For a required str question, just add `validator: '{% if not answer %}Required{% endif %}'. It's not as ergonomic, but works and is simpler to maintain upstream.

yajo:
> for `yaml` where I'd _not_ treat an empty or spaces-only string as `null` although `yaml.safe_load("")` behaves like that because it is inconsistent with `json` and not intuitive IMO. If somebody wants to answer with `null`, then they should answer `null`.

This behavior is documented in [the YAML spec](https://yaml.org/spec/1.2.2/#72-empty-nodes), so I think we should respect the spec and let empty strings become `None` when parsed.

sisp:
> So, `--defaults` will work as long as all questions without defaults are answered via `--data`, right?

Correct.

> This behavior is documented in [the YAML spec](https://yaml.org/spec/1.2.2/#72-empty-nodes), so I think we should respect the spec and let empty strings become `None` when parsed.

I hadn't thought about that. Right, the spec should certainly be respected. :+1:

sisp:
I've made the changes as we discussed. Any question that has no `default` field is considered to be a required question and needs an explicit answer via interactive mode, CLI or API. When the `--defaults` flag is used, questions without a `default` field need to be answered via the `--data` flag. Note that `--defaults` implies non-interactive mode, i.e. it is not possible to interactively answer questions without a `default` field. This behavior seems consistent with the current behavior as `--defaults` for questions without a `default` field result in the answer `None`. If we want to distinguish between interactive and non-interactive mode, we might need to add another flag in a separate PR.

The implementation of this behavior requires a new value that has the semantics "not set" for the value of the `default` field when the `default` field is missing. `None` doesn't work because it might be a regular default value. Therefore, I introduced a sentinel type `MissingType` and a corresponding constant `MISSING` inspired by https://github.com/python/mypy/issues/5424#issuecomment-410866316.

A few tests require adjustment to the new behavior because they rely on the default answer to a question without a `default` field to be `None`. I've made those changes.

WDYT, @yajo?

yajo (approved):
Great!

yajo:
Would you please fix the conflics to merge?

sisp:
Done. And I've also added type hints made some minor stylistic improvements.

sisp:
I just pushed a commit that removes obsolete special treatment of `str` questions when the answer is `None` because this isn't valid anymore.

PR #964, opened 2023-02-13
refactor: simplify casting boolean question settings

sisp:
I've simplified casting boolean question settings to `bool` by using `cast_str_to_bool` directly instead of using `cast_answer_type(<value>, cast_str_to_bool)`, which (a) was semantically off because those values aren't answers to questions and (b) was unnecessary because `cast_answer_type(...)` contains additional logic for handling edge cases of answers that aren't relevant to those boolean-like settings values.

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -359,9 +359,7 @@ class Question:
 
     def get_multiline(self) -> bool:
         """Get the value for multiline."""
-        multiline = self.render_value(self.multiline)
-        multiline = cast_answer_type(multiline, cast_str_to_bool)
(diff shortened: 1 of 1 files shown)


sisp:
I think the failing CI job `CI /flake-check (macos-latest)` is caused by a (random) timeout.

yajo (approved):

yajo:
Yes, I have to increase the timeout in tests. At least, for mac.

Thanks for the patch, may you fix the conflicts please?

sisp:
Done. :rocket:

PR #965, opened 2023-02-13
refactor: remove useless code related to not asking a question

sisp:
I've removed some useless code related to not asking a question, i.e. not prompting, when (a) defaults are supposed be used (e.g. `--defaults` flag) or (b) the answer is provided via CLI (i.e. `--data` flag) or API (i.e. `data` argument passed to `run_auto(...)`/`copy(...)`). The removed code was about handling this case by setting `when` to `False`, but [it seems one must be careful with using `when` for purposes other than the intended one](https://github.com/copier-org/copier/discussions/678#discussioncomment-2822746), and [using defaults](https://github.com/copier-org/copier/blob/2201d0210a57cc4d88c1d5d2369b70ea4894e2a9/copier/main.py#L394-L395) and [using answers via CLI/API](https://github.com/copier-org/copier/blob/2201d0210a57cc4d88c1d5d2369b70ea4894e2a9/copier/main.py#L378-L380) is already handled elsewhere.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -381,7 +381,6 @@ class Worker:
             questions.append(
                 Question(
                     answers=result,
-                    ask_user=not self.defaults,
                     jinja_env=self.jinja_env,
                     var_name=var_name,
                     **details,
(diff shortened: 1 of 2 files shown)


yajo (approved):
Good catch!

PR #966, opened 2023-02-13
fix: --skip option is ignored

tgaugry:
Command line option "--skip" currently does not work and is just ignored, because its value is not passed to Worker class.


sisp:
A test that captures this bug would be great. 🙏

yajo:
Incredible that this passwd unnoticed for so long! Would you mind adding the test please? Thanks!

tgaugry:
Hello, could you relaunch the test workflow ? Thanks

yajo (approved):

PR #969, opened 2023-02-14
docs: add chapter on template hierarchies with meta-templates

### LINKED ISSUE #934: Updatable templates for template hierarchies and reuse

sisp:
I've written a first draft of a chapter on using Copier to create template hierarchies with meta-templates as discussed in #934.

I'd appreciate any kind of feedback, suggestions for improvement, criticism, typo fixes, ... :smile_cat: :pray:

Closes #934. Related to #79, #402, #416.

/cc @pykong @fdaguin


yajo:
Currently we have: 

* https://copier.readthedocs.io/en/latest/configuring/#include-other-yaml-files
* https://copier.readthedocs.io/en/latest/configuring/#applying-multiple-templates-to-the-same-subproject
* symlinks, being improved in #938 
* Jinja features like macros, import, extend...
* Git submodules.
* some creative use of tasks
* meta-templates
* Maybe something else?...

These subjects are a hot topic. We devs are DRY lovers! 🤪

I guess we should have a new file or directory in the docs called "advanced templating techniques" or similar. We should try to sort these techniques from simple to complex (IMHO meta-templates is the most complex). Then add for each of them:

- the problem it solves
- a description of the template and the rationale behind it 
- an example 

WDYT?

Regarding meta-templates, I will review it in depth after we settle the case in #934 👨‍⚖️

sisp:
You're right, it would be much better yo gather those techniques on a dedicated page with a standardized structure as you suggested.

I can work on a PR that migrates the already documented techniques first, and then when we come back to this PR after #934 I'll update these docs accordingly.

yajo:
OK, good idea.

yajo:
Possibly a new strategy to add to the collection: https://github.com/copier-org/copier/issues/1004#issuecomment-1446853228

yajo:
I was doing a quick scan on low-hanging fruits before doing a new release, now that all breaking changes are in place.

Is this ready to merge?

sisp:
As it turned out after our video call and https://github.com/copier-org/copier/issues/934#issuecomment-1515298048, I think template hierarchies wouldn't be built using meta-templates but rather by syncing a "fork" (child template) with the upstream project (the parent template) using Git. So I think this PR needs a rewrite, or we could close it and create a new one with the other approach once fully figured out. WDYT?

yajo:
Ok, let's close for now and let the future pr happen some day

PR #970, opened 2023-02-15
fix: skip tasks in pretend mode

### LINKED ISSUE #950: Do not run tasks in --pretend mode
My only task is to `git init` the project that was just created from the provided template. But when running in --pretend mode, there is no directory to `init`.

Make it optional to run tasks when running --pretend, or just don't run tasks at all.

**Additional context**

sisp:
Copier now skips tasks execution in pretend mode as discussed in #950. I've extended the fix to also skip _migration_ tasks in pretend mode although they weren't explicitly mentioned in #950, but I think that is consistent behavior.

Fixes #950.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -651,7 +651,8 @@ class Worker:
             if not self.quiet:
                 # TODO Unify printing tools
                 print("")  # padding space
-            self._execute_tasks(self.template.tasks)
+            if not self.pretend:
+                self._execute_tasks(self.template.tasks)
         except Exception:
             if not was_existing and self.cleanup_on_error:
                 rmtree(self.subproject.local_abspath)
@@ -743,9 +744,10 @@ class Worker:
                     )
                     diff = diff_cmd("--inter-hunk-context=0")
             # Run pre-migration tasks
-            self._execute_tasks(
-                self.template.migration_tasks("before", self.subproject.template)
-            )
+            if not self.pretend:
+                self._execute_tasks(
+                    self.template.migration_tasks("before", self.subproject.template)
+                )
             self._uncached_copy()
             # Try to apply cached diff into final destination
             with local.cwd(self.subproject.local_abspath):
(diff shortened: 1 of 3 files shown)


copier/main.py
```diff
@@ -651,7 +651,8 @@ def run_copy(self) -> None:
             if not self.quiet:
                 # TODO Unify printing tools
                 print("")  # padding space
-            self._execute_tasks(self.template.tasks)
+            if not self.pretend:
+                self._execute_tasks(self.template.tasks)
```
yajo, line 655: Maybe it makes more sense to add this logic directly into `_execute_tasks()`? So you just fix it in one place.

sisp: I had thought about that actually, but my reasoning against it was that `_execute_tasks()` is agnostic of the kind of task it executes. In fact, it's just a thin wrapper around `subprocess.run()` which runs in the commands in the subproject's root folder. I wasn't sure whether I could justify disabling *any* (future) kind of task in pretend mode, that's why I added the condition in the several places explicitly.

That said, do you think it's safe to disable any kind of task execution?

yajo: I think so. At some point we should let the user make an informed decision about whether executing each task or migration, to make copier safer. But that requires some refactoring. The shorter the code, the easier it will be to refactor when the time comes  IMHO.

sisp: Okay, makes sense. I've moved the pretend mode check to the `_execute_tasks()` method. 39717b5

copier/main.py
```diff
@@ -218,6 +218,8 @@ def _execute_tasks(self, tasks: Sequence[Task]) -> None:
         Arguments:
             tasks: The list of tasks to run.
         """
+        if self.pretend:
+            return
```
yajo, line 222: I think it should still print something, saying what you are pretending to do.

sisp: Good point. :+1: I've changed it accordingly.

yajo (approved):
The patch looks good to me, but I bet you want to add some types to these tests before merging, right? 😉

sisp:
Absolutely! :laughing: I've added the type hints and also improved the formatting a bit :wink:.

Issue #931, opened 2023-01-21
`poe lint` should be faster if user doesn't have nix installed


rafalkrupinski:
The contributors documentation asks contributors to use `poe lint`:
https://github.com/copier-org/copier/blob/f8ae8d6e8ed8ec5333abe4b20ad53bc87123bff0/CONTRIBUTING.md?plain=1#L102

but it appears to be misconfigured

```shell
 % poe lint
Poe => pre-commit run --all-files
An error has occurred: InvalidConfigError: 
=====> .pre-commit-config.yaml is not a file
Check the log at /Users/matte/.cache/pre-commit/pre-commit.log
% ls .pre-commit-config.yaml
ls: .pre-commit-config.yaml: No such file or directory
```

yajo:
Ah you were probably missing the previous steps: https://github.com/copier-org/copier/blob/f8ae8d6e8ed8ec5333abe4b20ad53bc87123bff0/CONTRIBUTING.md?plain=1#L59-L88

I switched to nix for the devshell not so long ago. Try it, it's quite nice!

rafalkrupinski:
OK,
but it seems rather intrusive. Are you trying to scare away potential contributors? ;)
I mean I could understand a requirement to install `make`...

rafalkrupinski:
Seriously there's no way I can start developing locally without **building** an operating system first?

yajo:
AFAIK every binary is prebuilt and pushed to our public cache, which in
turn is enabled automatically when you follow the instructions, so you
should be just one command away from the same dev env I have. It does
download a lot, but it shouldn't build almost anything.

FWIW everybody downloads full OS images these days with docker. In nix it's
not much different, except that you download it package by package instead
of layer by layer.

Also I'm extremely tired of unpredictabilities in build and CI systems.
Particularly in pre-commit. For example I collected a lot of related
problems in the past in
https://github.com/OCA/oca-addons-repo-template/issues/175#issue-1466067857.
Since I switched to Nix, all of that simply disappeared (including Docker
😂).

Gitpod is also supported. Just click the button and get a web IDE ready to
code. It's compatible with Nix and Direnv out of the box.

I would love to have a smarter CI that not only yells about style but also
pushes back the reformatted code. I hope some day I have time to do it.

rafalkrupinski:
> so you should be just one command away from the same dev env I have

Installing nix takes some time, building environment too. It took ca 1.5-2 hrs. But maybe I'm being picky - one command or three...

[...]

PR #980, opened 2023-02-15
build: easier containerized pre-commit execution for friends without nix

yajo:
`poe lint` now uses nix if available. Otherwise, it will try to use docker or podman, by downloading the nix image and running pre-commit nix-based from there.

Fixes https://github.com/copier-org/copier/issues/931.

### DIFF
--- a/devtasks.py
+++ b/devtasks.py
@@ -1,8 +1,12 @@
 """Development helper tasks."""
-import os
+import logging
 import shutil
 from pathlib import Path
-from subprocess import check_call
+
+from plumbum import TEE, CommandNotFound, ProcessExecutionError, local
+
+_logger = logging.getLogger(__name__)
+HERE = Path(__file__).parent
 
 
 def clean():
@@ -33,20 +37,53 @@ def clean():
 
 def dev_setup():
     """Set up a development environment."""
-    # Gitpod sets PIP_USER=yes, which breaks poetry
-    env = dict(os.environ, PIP_USER="no")
-    check_call(["poetry", "install", "--with", "docs"], env=env)
-    check_call(
-        [
-            "poetry",
-            "run",
-            "pre-commit",
-            "install",
-            "-t",
-            "pre-commit",
-            "-t",
-            "commit-msg",
-        ],
-        env=env,
-    )
-    check_call(["poetry", "run", "pre-commit", "install-hooks"], env=env)
+    with local.cwd(HERE):
+        local["direnv"]("allow")
+        local["poetry"]("install")
+
+
+def lint(recycle_container=False):
+    """Lint and format the project."""
+    args = [
+        "--extra-experimental-features",
+        "nix-command flakes",
+        "--extra-substituters",
+        "https://copier.cachix.org https://devenv.cachix.org",
+        "--extra-trusted-public-keys",
+        "copier.cachix.org-1:sVkdQyyNXrgc53qXPCH9zuS91zpt5eBYcg7JQSmTBG4= devenv.cachix.org-1:w1cLUi8dv3hnoSPGAuibQv+f9TZLr6cv/Hm9XgU50cw=",
+        "develop",
+        "--impure",
+        ".",
+        "--command",
+        "pre-commit",
+        "run",
+        "--color=always",
+        "--all-files",
+    ]
+    try:
+        local["nix"].with_cwd(HERE)[args] & TEE
+    except CommandNotFound:
+        _logger.warn("Nix not found; fallback to a container")
+        runner = local.get("docker", "podman")
+        try:
+            (
+                runner[
+                    "container",
+                    "create",
+                    "--name=copier-lint-v1",
+                    f"--volume={HERE}:{HERE}:rw,z",
+                    f"--workdir={HERE}",
+                    "docker.io/nixos/nix",
+                    "nix",
+                    args,
+                ]
+                & TEE
+            )
+        except ProcessExecutionError:
+            _logger.info(
+                "Couldn't create copier-lint-v1 container, probably because a previous one exists. "
+                "Remove it if you want to recycle it. Otherwise, this is OK."
+            )
+        runner["container", "start", "--attach", "copier-lint-v1"] & TEE
+    except ProcessExecutionError as error:
+        raise SystemExit(error.errno)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ cmd = "mkdocs serve"
 help = "start local docs server"
 
 [tool.poe.tasks.lint]
-cmd = "pre-commit run --all-files"
+script = "devtasks:lint"
 help = "check (and auto-fix) style with pre-commit"
 
 [tool.poe.tasks.test]


rafalkrupinski:
@yajo I'm sorry, I didn't communicate to well. By 👍 I meant that I will test it soon, so I don't leave you without an answer for a week.
I don't like docker desktop so I'm trying podman. I'll let you know how it worked once I figure out how to configure it.

rafalkrupinski:
I have docker client installed and I think it's interfering with poe.

```zsh
 % which docker
docker: aliased to podman
% docker ps
CONTAINER ID  IMAGE       COMMAND     CREATED     STATUS      PORTS       NAMES
% poe lint
Poe => lint
Nix not found; fallback to a container
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
...
```

yajo:
That's because aliases only work within bash, but we're executing a process directly. 

I guess it makes sense to priorize podman over docker.

yajo:
Please check with #1005.

rafalkrupinski:
After removing docker:

```zsh
% poe lint            
Poe => lint
Nix not found; fallback to a container
Error: creating container storage: the container name "copier-lint-v1" is already in use by a1093ee368e5ce414ac034fee3d2c0864a6f33e74947370b6b933c55b1dd9109. You have to remove that container to be able to reuse that name: that name is already in use
Error: unable to start container a1093ee368e5ce414ac034fee3d2c0864a6f33e74947370b6b933c55b1dd9109: preparing container a1093ee368e5ce414ac034fee3d2c0864a6f33e74947370b6b933c55b1dd9109 for attach: lsetxattr /Users/user/Documents/Projects/copier/.envrc: operation not supported
  File "/Users/user/Documents/Projects/copier/devtasks.py", line 64, in lint
    local["nix"].with_cwd(HERE)[args] & TEE
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/copier-U7tILSPN-py3.10/lib/python3.10/site-packages/plumbum/machines/local.py", line 242, in __getitem__
    return LocalCommand(self.which(cmd))
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/copier-U7tILSPN-py3.10/lib/python3.10/site-packages/plumbum/machines/local.py", line 202, in which
    raise CommandNotFound(progname, list(cls.env.path))
plumbum.commands.processes.CommandNotFound: (CommandNotFound(...), 'nix', [<LocalPath /Users/user/Library/Caches/pypoetry/virtualenvs/copier-U7tILSPN-py3.10/bin>, <LocalPath /Users/user/Library/Caches/pypoetry/virtualenvs/copier-U7tILSPN-py3.10/bin>, <LocalPath /Users/user/.pyenv/shims>, <LocalPath /Users/user/Library/pnpm>, <LocalPath /Users/user/.poetry/bin>, <LocalPath /Users/user/.nvm/versions/node/v16.13.1/bin>, <LocalPath /opt/local/bin>, <LocalPath /opt/local/sbin>, <LocalPath /Library/Frameworks/Python.framework/Versions/3.8/bin>, <LocalPath /usr/local/bin>, <LocalPath /System/Cryptexes/App/usr/bin>, <LocalPath /usr/bin>, <LocalPath /bin>, <LocalPath /usr/sbin>, <LocalPath /sbin>, <LocalPath /opt/X11/bin>, <LocalPath /Users/user/.local/bin>])
```

rafalkrupinski:
I guess that's the problem:
```lsetxattr /Users/user/Documents/Projects/copier/.envrc: operation not supported```

/Users is a directory mounted from the host system

rafalkrupinski:
I'm just a small contributor, so I don't thing I have any influence over this project, but I think using Nix is already an overkill for running a linter, and adding more components, like docker/podman is making the environment even more complex and less portable.
I know some people prefer using Nix or docker for development, but requiring it is IMO not a good practice, especially in a small open source project.
I'd rather KISS

rafalkrupinski:
I'm just a small contributor, so I don't thing I have any influence over this project, but I think using Nix is already an overkill for running a linter, and adding more components, like docker/podman is making the environment even more complex and less portable.
I know some people prefer using Nix or docker for development, but requiring it is IMO not a good practice, especially in a small open source project.
I'd rather _KISS_

yajo:
What I was trying to do with this enhancement was to make it simpler for those that don't have nix but have podman or docker.

Maintaining pre-commit configurations is a complex topic. It's so complex that my first approach to fix the problem was to create [Autopretty](https://github.com/copier-org/autopretty) which, as you see, was under the umbrella of the copier org. It fixed that problem for me for awhile.

Copier is a project I maintain in my free time. However, in my professional time I've stumbled _a huge lot_ of times with the problem that pre-commit hook installations fail randomly every now and then. Somebody will publish a new version of a dependency you use on your hooks that introduces some incompatibility and... 💣 I don't know why, but they usually do it in the day you most need your CI working like a charm because there's an emergency.

That's something that most devs accept because they don't know any other option, or because just yes. Recently I've been gathering some of these problems in https://github.com/OCA/oca-addons-repo-template/issues/175#issue-1466067857. As of today, it's counting 11 public issues linked there, but the total number of times I've fixed this is probably more than the triple. FWIW Autopretty is now archived because the last time I tried to use it, after some time without touching it, everything was broken yet again! 🤦🏼‍♂️ 

When I discovered nix (and https://github.com/cachix/pre-commit-hooks.nix in particular), it was mind-blowing to learn that I can write code in a way that will survive the pass of time. You will be able to develop Copier after 20 years if I don't touch it. Also you'll be able to build it and run it. If I have 30 minutes to dedicate to Copier, I don't want to waste them fighting with some random thing that suddenly crashed on CI.

So, although I'm the biggest fan of KISS, there are different definitions of Simple and Short. For me, something simple is also something that I know for certain it will keep working. Recently the https://devenv.sh/ project was born, as an attempt to make nix simpler for non-nix people. That's what I'm using in Copier, to KISS. IMHO, tools like devenv are going to make Nix more widespread. It just solves too many problems as to ignore it. Be ready to see more `flake.nix` and `devenv.nix` files here and there, each time in more projects.

Apart from all these reasons, Copier is also something I treat with lot of ❤️ . Sometimes I use it to experiment, and sometimes to try to share with the reader the wonderful things I've learned. Nix is one of the best things ever invented in IT, so I simply love to have it around. Makes development funnier!

I appreciate a lot the humility in your words and I hope you understand why nix is here to stay.

OTOH I dedicated some time to make sure all this new tooling works in gitpod. So if you want to do small contributions to Copier, probably the best way is to click the Gitpod badge on the README, wait for a minute until all dependencies are pulled, and give back your code. Also, you can always do `pipx run black .` and it should mostly work.

rafalkrupinski:
> I hope you understand why nix is here to stay.

Sure, it's your project and you can do with it whatever you want.

> So if you want to do small contributions to Copier on the README

I actually have a refactoring to push, and I'm not sure if it qualifies as small.

> Also, you can always do pipx run black . and it should mostly work.

Um, I did, and it changed files I didn't touch :D

Also running pytest outside of nix makes some tests fail. Now, how do you know copier will work on another users' machines, outside this carefully crafted environment?

What you say sounds to me like pre-commit itself is broken. If you compare it to poetry or node.js tools, they all have lock files that lock all runtime and dev dependencies versions. Seems pre-commit runs with whatever version is the last one, so if one of the hooks uses another packages internal code (black & click in one of the bugs you've linked), and it changes, than too bad for you. If that's the case, I see here a lot of effort to go around a single broken tool.

yajo:
Yes, that's why pre-commit is so broken.

sisp:
@yajo: Did you also experience pre-commit issues when most of the [tools were installed as dev-dependencies via Poetry](https://github.com/copier-org/copier/blob/v6.2.0/.pre-commit-config.yaml#L15-L54)? I know that [some pre-commit hooks were installed via pre-commit directly](https://github.com/copier-org/copier/blob/v6.2.0/.pre-commit-config.yaml#L56-L93), so that's certainly a problem. For Prettier, it would have been possible to manage it via `package.json` and a lock file of one of the common package managers, expect developers to install them using the corresponding package manager, and configure it in `.pre-commit-hooks.yaml` in the same way as the Poetry-installed Python pre-commit hooks. But then there were still the plenty of pre-commit hooks from https://github.com/pre-commit/pre-commit-hooks. Were they problematic? Just trying to understand your pain.

I've not used Nix or devenv yet, contributing to Copier with just Poetry and manually installed Flake8 + plugins has worked sufficiently, but you seem so convinced and fascinated by Nix and devenv that I'll actually look into it. :smile:

yajo:
I've had plenty of problems in the past with prettier installation. I actually get one or two installation random failures per week on my company's CI (in the projects that still don't use nix).

Poetry gave no problems back then, but its design tries to be kinda pure. However, in reality, it lacks important pieces. Example: https://github.com/python-poetry/poetry/issues/6154#issuecomment-1218493969

I think asking people to use npm to install dev dependencies is at least as bad as asking them to install nix. However it's much worse in many other aspects.

So, once I've found a solution that is good and covers any possible use case and is easy to use, it's just too hard to go back 😅

FWIW you don't need to "learn" nix to use it for copier.

PR #982, opened 2023-02-16
fix: delete conditionally created file when answer changes

### LINKED ISSUE #686: Conditionally created file is not deleted when answer changes
## **Describe the problem**

My template contains a conditionally-created file, as per [the docs](https://copier.readthedocs.io/en/stable/configuring/#conditional-files-and-directories).
When flipping my answer from _no_ to _yes_, the file is created. However, when changing the answer from _yes_ to _no_, the previously-created file is not deleted.

## **Template**

https://github.com/EBoisseauSierra/copier_test_template

## **To Reproduce**

```shell
cd "$(mktemp -d)"
git init
copier git@github.com:EBoisseauSierra/copier_test_template.git .
ls -lah
git add .
git commit -Ss -m "feat: Initialize project with copier"
copier update
ls -lah
git add .
git commit -Ss -m "fix: Identify as a teapot"
copier update
ls -lah
```

## **Logs**

<details>

```shell
ebosi:/tmp $ cd "$(mktemp -d)"
ebosi:/tmp/tmp.ZOZ617bw3W $ git init
copier git@github.com:EBoisseauSierra/copier_test_template.git .
ls -lah
git add .
git commit -Ss -m "feat: Initialize project with copier"
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /tmp/tmp.ZOZ617bw3W/.git/
🎤 What is your project name?
project_name? Format: str Foo bar
🎤 Are you a teapot?
is_a_teapot? Format: bool No

Copying from template version 0.1.0
 identical  .
    create  .copier-answers.yml
    create  README.md


total 8.0K
drwx------.  3 ebosi ebosi  100 May 27 20:23 .
drwxrwxrwt. 83 root  root  2.0K May 27 20:23 ..
-rw-r--r--.  1 ebosi ebosi  185 May 27 20:23 .copier-answers.yml
drwxr-xr-x.  7 ebosi ebosi  200 May 27 20:23 .git
-rw-r--r--.  1 ebosi ebosi   10 May 27 20:23 README.md
[master (root-commit) bae9686] feat: Initialize project with copier
 2 files changed, 8 insertions(+)
 create mode 100644 .copier-answers.yml
 create mode 100644 README.md
ebosi:/tmp/tmp.ZOZ617bw3W $ copier update
Updating to template version 0.1.0
🎤 What is your project name?
project_name? Format: str Foo bar
🎤 Are you a teapot?
is_a_teapot? Format: bool Yes

Copying from template version 0.1.0
 identical  .
  conflict  .copier-answers.yml
 overwrite  .copier-answers.yml
    create  error.txt
 identical  README.md


ebosi:/tmp/tmp.ZOZ617bw3W $ git add .
git commit -Ss -m "fix: Identify as a teapot"
[master deef2bd] fix: Identify as a teapot
 2 files changed, 2 insertions(+), 1 deletion(-)
 create mode 100644 error.txt
ebosi:/tmp/tmp.ZOZ617bw3W $ copier update
Updating to template version 0.1.0
🎤 What is your project name?
project_name? Format: str Foo bar
🎤 Are you a teapot?
is_a_teapot? Format: bool No

Copying from template version 0.1.0
 identical  .
  conflict  .copier-answers.yml
 overwrite  .copier-answers.yml
 identical  README.md


ebosi:/tmp/tmp.ZOZ617bw3W $ ls -lah
total 12K
drwx------.  3 ebosi ebosi  120 May 27 20:24 .
drwxrwxrwt. 91 root  root  2.1K May 27 20:24 ..
-rw-r--r--.  1 ebosi ebosi  185 May 27 20:24 .copier-answers.yml
-rw-r--r--.  1 ebosi ebosi    4 May 27 20:24 error.txt <---------------------- should have been deleted
drwxr-xr-x.  8 ebosi ebosi  260 May 27 20:24 .git
-rw-r--r--.  1 ebosi ebosi   10 May 27 20:24 README.md
ebosi:/tmp/tmp.ZOZ617bw3W $ cat .copier-answers.yml 
# DO NOT EDIT.
# Changes here will be overwritten by Copier.
_commit: 0.1.0
_src_path: git@github.com:EBoisseauSierra/copier_test_template.git
is_a_teapot: false
project_name: Foo bar

ebosi:/tmp/tmp.ZOZ617bw3W $ copier --version
copier 6.0.0
ebosi:/tmp/tmp.ZOZ617bw3W $ python --version
Python 3.10.4

```

</details>

## **Expected behavior**

The `error.txt` file is deleted when `is_a_teapot` is flipped from `true` to `false`.

## **Environment**
 - OS: Fedora 36
 - Copier version: `6.0.0` 
 - Python version: `3.10.4`
 - Installation method: pip

**Additional context**

Related: #461, #619 and #684 .

sisp:
I've fixed the `copier update` behavior to remove a conditionally created file when the answer changes such that the file would not be created anymore.

To achieve this behavior, I needed to change the update workflow a little.

**Before:**
1. The old template is copied into a temporary location with the answers from the answers file.
2. The new template is copied into a temporary location with the answers from the answers file. _By the way, I think this is not quite correct because the questionnaire might have changed in the new template version and there is no user interaction in this step that allows answering to new/changed questions._
3. A diff is created between the output of 1. and the real project.
4. Pre-migration tasks are executed on the real project.
5. The new template is copied into the real destination with possible user interaction where necessary unless `--data`/`--defaults` flags are passed accordingly.
6. The diff from 3. is applied to the real project.
7. Files that exist in 1. but not in 2. are removed from the real project.
8. Post-migration tasks are executed on the real project.

**After:**
1. The old template is copied into a temporary location with the answers from the answers file.
2. Pre-migration tasks are executed on the real project. _It's important to move this step up in the new workflow to support, e.g., answer file migrations. And it's also important to run this step after 1. because the old template must be copied before the pre-migration tasks have been executed in case of answer migrations._
3. The new template is copied into a temporary location. The possibly migrated answers from the real project are respected as "last answers" and otherwise _regular user interaction is possible_ unless `--data`/`--defaults` flags are passed accordingly, thus, changes in the questionnaire can be addressed by explicit new answers.
4. A diff is created between the output of 1. and the real project.
5. The new template is copied into the real destination with the answers given in 3. and without any user interaction because those answers match the template version.
6. The diff from 3. is applied to the real project.
7. Files that exist in 1. but not in 2. are removed from the real project.
8. Post-migration tasks are executed on the real project.

---

This workflow seems to work according to the test suite, but the (modified) implementation in `main.py::Worker._apply_update(...)` is a bit hacky and might need some refactoring. For instance, blocks like

```python
# Clear last answers cache to load possible answers migration
with suppress(AttributeError):
    del self.answers
with suppress(AttributeError):
    del self.subproject.last_answers
```

or

```python
# Copy the new template output into the actual destination with the
# answers from the temporary destination.
# TODO: Do this more elegantly.
old_data = self.data
old_defaults = self.defaults
old_quiet = self.quiet
self.data = new_worker.answers.combined
self.defaults = True
self.quiet = True
try:
    self.run_copy()
finally:
    self.data = old_data
    self.defaults = old_defaults
    self.quiet = old_quiet
```

don't feel right, but I think some more fundamental refactoring is needed to improve them. I'd postpone it to a follow-up PR to keep the scope of this PR small and ship improvements more quickly.

Fixes #686. Related to #619.

/cc @pawamoy

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -711,16 +711,26 @@ class Worker:
         ) as new_copy:
             old_worker = self._make_old_worker(old_copy)
             old_worker.run_copy()
-            recopy_worker = replace(
-                self,
-                dst_path=new_copy,
-                data=self.subproject.last_answers,
-                defaults=True,
-                quiet=True,
-                src_path=self.subproject.template.url,
+
+            # Run pre-migration tasks
+            self._execute_tasks(
+                self.template.migration_tasks("before", self.subproject.template)
             )
-            recopy_worker.run_copy()
-            compared = dircmp(old_copy, new_copy)
+
+            # Clear last answers cache to load possible answers migration
+            with suppress(AttributeError):
+                del self.answers
+            with suppress(AttributeError):
+                del self.subproject.last_answers
+            new_worker = replace(
+                self,
+                src_path=self.subproject.template.url,
+                dst_path=new_copy,
+            )
+            # Use last answers from the pre-migrated project
+            new_worker.subproject.last_answers = self.subproject.last_answers
+            new_worker.run_copy()
+
             # Extract diff between temporary destination and real destination
             with local.cwd(old_copy):
                 subproject_top = git(
@@ -742,11 +752,23 @@ class Worker:
                         file=sys.stderr,
                     )
                     diff = diff_cmd("--inter-hunk-context=0")
-            # Run pre-migration tasks
-            self._execute_tasks(
-                self.template.migration_tasks("before", self.subproject.template)
-            )
-            self._uncached_copy()
+
+            # Copy the new template output into the actual destination with the
+            # answers from the temporary destination.
+            # TODO: Do this more elegantly.
+            old_data = self.data
+            old_defaults = self.defaults
+            old_quiet = self.quiet
+            self.data = new_worker.answers.combined
+            self.defaults = True
+            self.quiet = True
+            try:
+                self.run_copy()
+            finally:
+                self.data = old_data
+                self.defaults = old_defaults
+                self.quiet = old_quiet
+
             # Try to apply cached diff into final destination
             with local.cwd(self.subproject.local_abspath):
                 apply_cmd = git["apply", "--reject", "--exclude", self.answers_relpath]
@@ -777,28 +799,18 @@ class Worker:
                             "after updating",
                             fname,
                             old_worker.subproject.local_abspath / fname,
-                            recopy_worker.subproject.local_abspath / fname,
+                            new_worker.subproject.local_abspath / fname,
                             retcode=None,
                         )
                         # Remove rejection witness
                         Path(f"{fname}.rej").unlink()
             # Trigger recursive removal of deleted files in last template version
-            _remove_old_files(self.subproject.local_abspath, compared)
+            _remove_old_files(self.subproject.local_abspath, dircmp(old_copy, new_copy))
 
         # Run post-migration tasks
         self._execute_tasks(
             self.template.migration_tasks("after", self.subproject.template)
         )
-
-    def _uncached_copy(self):
-        """Copy template to destination without using answer cache."""
-        # Clear last answers cache to load possible answers migration
-        with suppress(AttributeError):
-            del self.answers
-        with suppress(AttributeError):
-            del self.subproject.last_answers
-        # Do a normal update in final destination
-        self.run_copy()
 
     def _make_old_worker(self, old_copy):
         """Create a worker to copy the old template into a temporary destination."""
--- a/copier/main.py
+++ b/copier/main.py
@@ -700,23 +700,20 @@ class Worker:
         )
         subproject_subdir = self.subproject.local_abspath.relative_to(subproject_top)
 
-        # Copy old template into a temporary destination
         with TemporaryDirectory(
-            prefix=f"{__name__}.update_diff."
-        ) as old_copy, TemporaryDirectory(
-            prefix=f"{__name__}.recopy_diff."
-        ) as new_copy:
+            prefix=f"{__name__}.old_copy."
+        ) as old_copy, TemporaryDirectory(prefix=f"{__name__}.new_copy.") as new_copy:
+            # Copy old template into a temporary destination
             old_worker = replace(
                 self,
-                src_path=self.subproject.template.url,
                 dst_path=old_copy / subproject_subdir,
                 data=self.subproject.last_answers,
                 defaults=True,
                 quiet=True,
+                src_path=self.subproject.template.url,
                 vcs_ref=self.subproject.template.commit,
             )
             old_worker.run_copy()
-
             # Extract diff between temporary destination and real destination
             with local.cwd(old_copy):
                 self._git_initialize_repo()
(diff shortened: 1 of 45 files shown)


pawamoy:
Hello @sisp, thanks for the great work you're doing on this project :rocket: !

Your changes sound good to me. Would you also update the Mermaid diagram in the docs?

sisp:
Thanks for the praise, @pawamoy! :pray:

I've updated the diagram and added a step index (e.g. `(1)`, `(2)`, ...) because the order is important and not obvious otherwise.

While changing the diagram, I decided to move step 4 (computing the diff) up and perform it after step 1 because it's clearer that computing the diff doesn't depend on answer migration. Also, the diagram becomes easier to read because there are fewer jumps when following the order of the steps. This change is a no-op.

sisp:
I just resolved a merge conflict with `master` and added a missing type hint in the tests.

yajo:
Sorry I had to do an important fix in #1068. Could you solve conflicts again please?

copier/main.py
```diff
@@ -708,24 +708,15 @@ def _apply_update(self):
         ) as new_copy:
             old_worker = replace(
                 self,
+                src_path=self.subproject.template.url,
                 dst_path=old_copy / subproject_subdir,
                 data=self.subproject.last_answers,
                 defaults=True,
                 quiet=True,
-                src_path=self.subproject.template.url,
```
yajo, line 715: seems like you can revert this change.

tests/test_conditional_file_name.py
```diff
@@ -55,3 +57,34 @@ def test_dont_render_conditional_subdir(tmp_path_factory: TempPathFactory) -> No
...
+    copier.copy(str(src), dst, data={"condition": True})
+    assert (dst / "file.txt").exists()
+
+    with local.cwd(dst):
+        git("init")
+        git("add", ".")
+        git("commit", "-mv1")
+
+    copier.copy(dst_path=dst, data={"condition": False})
+    assert not (dst / "file.txt").exists()
```
yajo, line 90: I have tested this without the fix and it passes ✔️. Maybe some of the recent fixes fixed this already as a side effect?

If so, we could merge the test only and it should be enough. Or maybe the test isn't correct?

yajo (changes requested):

copier/main.py
sisp: Okay, will do.

tests/test_conditional_file_name.py
sisp: I think you're right that the recent fixes fixed also this problem as a side effect. I'll comment more below.

sisp:
That's great! I spent quite some time verifying that the recent fixes also fix this issue as a side effect. I suspected some missing edge case when answering interactively during the update process and extended the test case accordingly, but it looks like everything works fine now. Also, to the best of my knowledge the workflow diagram does not need to be changed anymore because there don't seem to be any changes in the workflow at the abstraction level of the diagram. So I reverted that change.

As it turns out, this PR now only adds a test case. You'll need to adjust the commit message accordingly. I followed your example in #1072 and parametrized the test case to cover both interactive and non-interactive answering instead of creating a second test case for interactive answering in `tests/test_prompt.py`.

yajo:
As an exception I kept the `fix` prefix in the commit to let it appear in the next changelog. After all, it got fixed in the release. Thanks!

PR #984, opened 2023-02-16
refactor: simplify question answering

sisp:
Just a minor cleanup: I've simplified the question answering logic a little by avoiding the unnecessary creation of a list of questions that shall be asked, i.e. for which no answers have been provided by the user via CLI/API.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -403,7 +403,6 @@ class Worker:
                 )
             except KeyboardInterrupt as err:
                 raise CopierAnswersInterrupt(result, question, self.template) from err
-
             previous_answer = result.combined.get(question.var_name)
             # If question was skipped and it's the 1st
             # run, you could be getting a raw templated value
--- a/copier/main.py
+++ b/copier/main.py
@@ -11,7 +11,7 @@ from functools import partial
 from itertools import chain
 from pathlib import Path
 from shutil import rmtree
-from typing import Callable, Iterable, List, Mapping, Optional, Sequence
+from typing import Callable, Iterable, Mapping, Optional, Sequence
 from unicodedata import normalize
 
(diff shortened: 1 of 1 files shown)


copier/main.py
```diff
@@ -406,6 +403,7 @@ def answers(self) -> AnswersMap:
                 )
             except KeyboardInterrupt as err:
                 raise CopierAnswersInterrupt(result, question, self.template) from err
+
```
yajo, line 406: There are people that add spaces, others remove them... let's keep only meaningful changes please 😊 

```suggestion
```

copier/main.py
```diff
@@ -415,6 +413,7 @@ def answers(self) -> AnswersMap:
                 new_answer = question.filter_answer(new_answer)
             if new_answer != previous_answer:
                 result.user[question.var_name] = new_answer
+
```
yajo, line 416:
```suggestion
```

yajo (changes requested):
Thanks! One little change please 😇

yajo:
Oh, BTW for you to know (because I see you're doing a lot of refactoring), I'm using commitizen to keep the changelog. It automatically adds entries to it when commits start with `refactor: `.

You probably understand that a change like this doesn't deserve an entry in the changelog, as it's just affecting the inner implementation of a function. For these changes, probably using commits with `style: ` prefix is better.

Also use `tests: ` when refactoring tests, as end users (those that read the changelog) shouldn't care about that.

For now don't worry, I will squash on merge and change it myself.

copier/main.py
sisp: Makes sense. 👍

sisp:
Ahhh, I hadn't read your comment about the `refactor:` prefix. Will remember it next time. 😄

PR #985, opened 2023-02-18
refactor(tests): add type hints and clean up

sisp:
I've added type hints to the tests, so they are type-checked by `mypy` as well, and cleaned up the tests a bit, e.g. more consistent setup of temporary template and subproject directories, formatting of template tree creation code for improved readability (IMO), consistent usage of the `pathlib` API instead of the `os`/`os.path` API, etc.

I know it's a quite big PR and probably not much fun to review :see_no_evil: - sorry about that :confused: -, but I felt the tests could use some love :smile_cat:.

### DIFF
--- a/tests/test_complex_questions.py
+++ b/tests/test_complex_questions.py
@@ -1,4 +1,5 @@
 import json
+from collections import deque
 from pathlib import Path
 from textwrap import dedent
 
@@ -191,9 +192,16 @@ def test_cli_interactive(template_path: str, tmp_path: Path, spawn: Spawn) -> No
     tui.sendline("- I'd love it")
     tui.send(Keyboard.Esc + Keyboard.Enter)
     expect_prompt(tui, "choose_list", "str", help="You see the value of the list items")
-    tui.expect_exact("first")
-    tui.expect_exact("second")
-    tui.expect_exact("third")
+    deque(
+        map(
+            tui.expect_exact,
+            [
+                "first",
+                "second",
+                "third",
+            ],
+        )
+    )
     tui.sendline()
     expect_prompt(
         tui,
@@ -201,21 +209,38 @@ def test_cli_interactive(template_path: str, tmp_path: Path, spawn: Spawn) -> No
         "str",
         help="You see the 1st tuple item, but I get the 2nd item as a result",
     )
-    tui.expect_exact("one")
-    tui.expect_exact("two")
-    tui.expect_exact("three")
+    deque(
+        map(
+            tui.expect_exact,
+            [
+                "one",
+                "two",
+                "three",
+            ],
+        )
+    )
     tui.sendline()
     expect_prompt(
         tui, "choose_dict", "str", help="You see the dict key, but I get the dict value"
     )
-    tui.expect_exact("one")
-    tui.expect_exact("two")
-    tui.expect_exact("three")
+    deque(
+        map(
+            tui.expect_exact,
+            [
+                "one",
+                "two",
+                "three",
+            ],
+        )
+    )
     tui.sendline()
     expect_prompt(tui, "choose_number", "float", help="This must be a number")
-    tui.expect_exact("-1.1")
-    tui.expect_exact("0")
-    tui.expect_exact("1")
+    deque(
+        map(
+            tui.expect_exact,
+            ["-1.1", "0", "1"],
+        )
+    )
     tui.sendline()
     expect_prompt(tui, "minutes_under_water", "int")
     tui.expect_exact("10")
@@ -397,14 +422,14 @@ def test_tui_inherited_default(
         "owner1": "example",
         "owner2": owner2,
     }
-    assert json.loads((dst / "answers.json").read_bytes()) == result
+    assert json.loads((dst / "answers.json").read_text()) == result
     with local.cwd(dst):
         git("init")
         git("add", "--all")
         git("commit", "--message", "init project")
     # After a forced update, answers stay the same
     run_update(dst, defaults=True, overwrite=True)
-    assert json.loads((dst / "answers.json").read_bytes()) == result
+    assert json.loads((dst / "answers.json").read_text()) == result
 
 
 def test_tui_typed_default(
@@ -434,7 +459,7 @@ def test_tui_typed_default(
     )
     tui = spawn(COPIER_PATH + ("copy", str(src), str(dst)), timeout=10)
     tui.expect_exact(pexpect.EOF)
-    assert json.loads((dst / "answers.json").read_bytes()) == {
+    assert json.loads((dst / "answers.json").read_text()) == {
         "_src_path": str(src),
         "test1": False,
         "test2": False,
--- a/tests/test_complex_questions.py
+++ b/tests/test_complex_questions.py
@@ -1,15 +1,16 @@
 import json
-from collections import deque
 from pathlib import Path
 from textwrap import dedent
 
 import pexpect
 import pytest
 import yaml
+from pexpect.popen_spawn import PopenSpawn
 from plumbum import local
 from plumbum.cmd import git
 
 from copier import copy, run_auto, run_update
+from copier.types import OptStr
 
 from .helpers import (
     BRACKET_ENVOPS,
@@ -17,6 +18,7 @@ from .helpers import (
     COPIER_PATH,
     SUFFIX_TMPL,
     Keyboard,
+    Spawn,
     build_file_tree,
     expect_prompt,
 )
@@ -26,12 +28,12 @@ BLK_END = BRACKET_ENVOPS["block_end_string"]
 
 
 @pytest.fixture(scope="module")
-def template_path(tmp_path_factory) -> str:
+def template_path(tmp_path_factory: pytest.TempPathFactory) -> str:
     root = tmp_path_factory.mktemp("template")
     build_file_tree(
         {
-            root
-            / "copier.yml": f"""\
+            (root / "copier.yml"): (
+                f"""\
                 _templates_suffix: {SUFFIX_TMPL}
                 _envops: {BRACKET_ENVOPS_JSON}
                 love_me:
@@ -91,9 +93,10 @@ def template_path(tmp_path_factory) -> str:
                 # Simplified format is still supported
                 minutes_under_water: 10
                 optional_value: null
-            """,
-            root
-            / "results.txt.tmpl": """\
+                """
+            ),
+            (root / "results.txt.tmpl"): (
+                """\
                 love_me: [[love_me|tojson]]
                 your_name: [[your_name|tojson]]
                 your_age: [[your_age|tojson]]
@@ -106,13 +109,29 @@ def template_path(tmp_path_factory) -> str:
                 choose_number: [[choose_number|tojson]]
                 minutes_under_water: [[minutes_under_water|tojson]]
                 optional_value: [[optional_value|tojson]]
-            """,
+            """
+            ),
         }
     )
     return str(root)
 
 
-def test_api(tmp_path, template_path):
+def check_invalid(
+    tui: PopenSpawn,
+    name: str,
+    format: str,
+    invalid_value: str,
+    help: OptStr = None,
+    err: str = "Invalid input",
+):
+    """Check that invalid input is reported correctly"""
+    expect_prompt(tui, name, format, help)
+    tui.sendline(invalid_value)
+    tui.expect_exact(invalid_value)
+    tui.expect_exact(err)
+
+
+def test_api(template_path: str, tmp_path: Path) -> None:
     """Test copier correctly processes advanced questions and answers through API."""
     copy(
         template_path,
@@ -128,8 +147,7 @@ def test_api(tmp_path, template_path):
         defaults=True,
         overwrite=True,
     )
-    results_file = tmp_path / "results.txt"
-    assert results_file.read_text() == dedent(
+    assert (tmp_path / "results.txt").read_text() == dedent(
         """\
             love_me: false
             your_name: "LeChuck"
@@ -147,16 +165,8 @@ def test_api(tmp_path, template_path):
     )
 
 
-def test_cli_interactive(tmp_path, spawn, template_path):
+def test_cli_interactive(template_path: str, tmp_path: Path, spawn: Spawn) -> None:
     """Test copier correctly processes advanced questions and answers through CLI."""
-
-    def check_invalid(tui, name, format, invalid_value, help=None, err="Invalid input"):
-        """Check that invalid input is reported correctly"""
-        expect_prompt(tui, name, format, help)
-        tui.sendline(invalid_value)
-        tui.expect_exact(invalid_value)
-        tui.expect_exact(err)
-
     tui = spawn(COPIER_PATH + ("copy", template_path, str(tmp_path)), timeout=10)
     expect_prompt(tui, "love_me", "bool", help="I need to know it. Do you love me?")
     tui.send("y")
@@ -181,16 +191,9 @@ def test_cli_interactive(tmp_path, spawn, template_path):
     tui.sendline("- I'd love it")
     tui.send(Keyboard.Esc + Keyboard.Enter)
     expect_prompt(tui, "choose_list", "str", help="You see the value of the list items")
-    deque(
-        map(
-            tui.expect_exact,
-            [
-                "first",
-                "second",
-                "third",
-            ],
-        )
-    )
+    tui.expect_exact("first")
+    tui.expect_exact("second")
+    tui.expect_exact("third")
     tui.sendline()
     expect_prompt(
         tui,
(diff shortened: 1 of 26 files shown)


yajo:
Thanks for your efforts! TBH I don't usually like touching tests unless there's an ❌ because there's always risk of breaking stuff. Sometimes they're testing non-obvious corner cases. Unlike with the rest of code, the more tests the better, and being repetitive is no big deal.

That said, the PR looks good. I can tell you know the depths of copier! 😀

I still didn't finish reviewing these 1500 boring lines but i know 1 thing: I wouldn't like to do this again. But I don't see any new mypy checks. Did you add them?

Thanks!

sisp:
I understand your hesitance to touch tests. It's really just a simple refactoring without semantic changes.

I've not added any typing tests, but the added type annotations to the test functions let mypy type-check them, too. mypy doesn't check the typing in bodies of untyped function defs (see [this example](https://mypy-play.net/?mypy=1.0.0&python=3.11&gist=8b2c2f7a307604bc47c581eecc46da92)). Adding `--check-untyped-defs` might help to some extent, too, but I think it's clearer and safer when also the tests are properly type-annotated.

Sorry for the utterly boring review ... 🫣

tests/test_migrations.py
```diff
@@ -72,109 +71,122 @@ def test_migrations_and_tasks(tmp_path: Path):
...
-        copy(src_path=str(template), defaults=True, overwrite=True)
-        answers = json.loads(Path(".copier-answers.yml").read_text())
-        assert answers["_commit"] == "v1"
-        assert answers["best_song"] == "la vie en rose"
-        assert json.loads(Path("songs.json").read_text()) == ["la vie en rose"]
+    copy(src_path=str(src), dst_path=dst, defaults=True, overwrite=True)
+    answers = json.loads((dst / ".copier-answers.yml").read_text())
+    assert answers["_commit"] == "v1"
+    assert answers["best_song"] == "la vie en rose"
+    assert json.loads((dst / "songs.json").read_text()) == ["la vie en rose"]
```
yajo, line 109: With this change, you're altering the test.

The previous implementation made sure that `dst_path` defaults to `.`, where the new one doesn't.

Small but important details. Please undo these kind of changes.

tests/test_empty_suffix.py
```diff
@@ -1,57 +1,60 @@
...
+    assert (dst / "pingu.txt").exists()
+    assert (dst / "pingu" / "render_me.txt").exists()
 
     expected = "Hello pingu!"
-    assert (dest / "render_me").read_text() == expected
-    assert (dest / "pingu.txt").read_text() == expected
-    assert (dest / "pingu" / "render_me.txt").read_text() == expected
+    assert (dst / "render_me").read_text() == expected
+    assert (dst / "pingu.txt").read_text() == expected
+    assert (dst / "pingu" / "render_me.txt").read_text() == expected
```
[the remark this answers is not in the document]
sisp: I repsect your position on this diff. But I really believe there is some value in harmonizing the tests a bit, also regarding the recurring creation code for the temporary template and subproject folders as well as the variable names for their paths. It removes variance in the tests where in fact none exists which makes it easier for new contributors to recognize recurring setup code. When several variations are used across the test suite, recognizing semantically identical code is more difficult.

If you really insist on keeping it as it is, I'll do it. But I genuinely think there is value in this harmonization.

tests/test_interrupts.py
```diff
@@ -10,75 +9,66 @@
...
+    ):
         with pytest.raises(CopierAnswersInterrupt) as err:
             worker.run_copy()
-        assert err.value.answers.user == answers
-        assert err.value.template == worker.template
+            assert err.value.answers.user == {
+                "question1": "foobar",
+                "question2": "yosemite",
+            }
+            assert err.value.template == worker.template
```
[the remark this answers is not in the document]
sisp: OMG, good catch! 🧐:bow: I understand your hesitance to touch tests unnecessarily. This bug is really subtle. I'll revert it of course.

tests/test_interrupts.py
```diff
@@ -10,75 +9,66 @@
...
-    ),
+    "side_effect",
+    [
+        # We override the prompt method from questionary to raise this
+        # exception and expect our surrounding machinery to re-raise
+        # it as a CopierAnswersInterrupt.
+        CopierAnswersInterrupt(Mock(), Mock(), Mock()),
+        KeyboardInterrupt,
+    ],
 )
```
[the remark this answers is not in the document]
sisp: Thanks! :pray::tada:

tests/test_migrations.py
sisp: Ouch, I see and hadn't thought about that! Good catch! :monocle_face::bow:

But one remark: I understand that my change did change the exact test scenario. But I think testing that `dst_path` defaults to `.` should be covered in a dedicated test (which I don't see at the moment) and not only implicitly in some unrelated test because then a change like mine, which was perfectly fine w.r.t. to the scenario under test, coincidentally removed another implicit test. WDYT, @yajo?

sisp:
> Would you get angry at me if I ask you to remove most diffs? :sweat_smile:

No worries, you're the authority of this project and I respect your opinion. No hard feelings. :crossed_fingers: I've tried extrapolating from your inline comments how you'd like this PR to change which has reduced the diff by quite a bit. I'll add some inline comments in a moment where I think there is sufficient value but I'm uncertain whether you'll the changes or where I have a question.

> I see lots of changes that are just stylistic or syntactical. I see some replacing `os.path` for `pathlib.Path`. Others replacing raw strings by `textwrap.dedent`. Others adding some parenthesis to `build_file_tree`... Most of these changes add little value to the code base (mostly when they're found on tests) and introduce big risks when changing just because of yes. OTOH they might fail to test some corner cases.

I completely agree on the cases you've shown that actually had negative impact on the scenarios under test. Regarding stylistic changes (I've also replied to one of your inline comments about this), I think there's some value in stylistic or syntactical harmonization even in the test suite because it reduces or removes variance where in fact none exists, so especially new contributors will find it easier to write better tests because the commonalities like the recurring temporary source/destination setup looks the same everywhere.

tests/test_normal_jinja2.py
```diff
@@ -24,45 +27,52 @@ def test_normal_jinja2(tmp_path_factory):
...
-    todo = (dst / "TODO.txt").read_text()
-    expected = "[[ Guybrush TODO LIST]]\n[# GROG #]\n    - Become a pirate\n"
-    assert todo == expected
+    assert (dst / "TODO.txt").read_text() == dedent(
+        """\
+        [[ Guybrush TODO LIST]]
+        [# GROG #]
+            - Become a pirate
+        """
+    )
```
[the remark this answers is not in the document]
sisp: I see. You're right, it was more explicit before. :+1: I'll revert it.

tests/test_config.py
```diff
@@ -427,45 +458,48 @@ def test_user_defaults(tmp_path_factory, user_defaults, data, expected):
...
+    expected_initial: str,
+    expected_updated: str,
+) -> None:
     src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
     with local.cwd(src):
         build_file_tree(
             {
-                src
-                / "copier.yaml": """\
+                (src / "copier.yaml"): (
```
sisp, line 485: Is it intended that the file path contains `src / ...` here while the current working directory is `src` already due to  `with local.cwd(src):`?

tests/test_complex_questions.py
```diff
@@ -181,55 +191,31 @@ def check_invalid(tui, name, format, invalid_value, help=None, err="Invalid inpu
...
-            [
-                "first",
-                "second",
-                "third",
-            ],
-        )
-    )
+    tui.expect_exact("first")
+    tui.expect_exact("second")
+    tui.expect_exact("third")
```
sisp, line 196: Do you agree with this change (and the same below)? I think it's better to reduce automation like loops etc. in tests and especially in assertions.

tests/test_exclude.py
```diff
@@ -26,26 +28,39 @@ def test_exclude_recursive_negate(tmp_path):
...
-    src, dst = tmp_path / "src", tmp_path / "dst"
-    build_file_tree({src / "copier.yml": "_exclude: ['*.txt']", src / "aaaa.txt": ""})
+def test_config_exclude(tmp_path_factory: pytest.TempPathFactory) -> None:
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    build_file_tree(
+        {
+            src / "copier.yml": "_exclude: ['*.txt']",
+            src / "aaaa.txt": "",
+        }
+    )
```
sisp, line 38: Do you agree with this change (and the other related ones in this file)? It's only stylistic, but I think it's much more readable to use a multiline dict when there is more than file to create.

tests/test_prompt.py
```diff
@@ -154,49 +160,54 @@ def test_copy_default_advertised(tmp_path_factory, spawn, name):
...
+                {
+                    "_envops": BRACKET_ENVOPS,
+                    "_templates_suffix": SUFFIX_TMPL,
+                    "question_1": question_1,
+                    "question_2": {
+                        "type": "yaml",
+                        "default": "something",
+                        "when": question_2_when,
+                    },
+                }
```
sisp, line 184: Do you agree with this change? I think that inlining the content of `copier.yml` is more concise and consistent with the rest of the test suite.

tests/test_subdirectory.py
```diff
@@ -155,165 +173,156 @@ def test_update_subdirectory_from_root_path(tmp_path_factory):
...
-            "<<<<<<< before updating\ndownstream version 1\n"
-            "=======\nupstream version 2\n>>>>>>> after updating\n",
+            dedent(
+                """\
+                <<<<<<< before updating
+                downstream version 1
+                =======
+                upstream version 2
+                >>>>>>> after updating
+                """
```
sisp, line 190: Do you agree with using a multiline string for the inline merge conflict? I think it's much more readable. The line breaks are implicit, yes, but it's much more obvious what the merge conflict syntax is.

tests/test_templated_prompt.py
```diff
@@ -236,14 +244,19 @@ def test_templated_prompt_builtins(tmp_path_factory):
...
 @pytest.mark.parametrize(
     "questions, raises, returns",
-    (
+    [
         ({"question": {"default": "{{ not_valid }}"}}, None, ""),
         ({"question": {"help": "{{ not_valid }}"}}, None, "None"),
         ({"question": {"type": "{{ not_valid }}"}}, InvalidTypeError, "None"),
         ({"question": {"choices": ["{{ not_valid }}"]}}, None, "None"),
-    ),
+    ],
```
sisp, line 252: There are a few of these minor changes in this PR. I think using a list of parameters is semantically better than using a tuple because in my understanding lists have the semantics of being collections while tuples have the semantics of ordered multi-value data. It's a minor distinction and both work fine technically. Also, [pytest's documentation](https://docs.pytest.org/en/stable/example/parametrize.html) uses a list if that's worth anything. :wink:

tests/test_config.py
```diff
@@ -36,7 +36,7 @@ def git_init(message="hello world") -> None:
 
 
 def test_config_data_is_loaded_from_file() -> None:
-    tpl = Template(str(Path("tests", "demo_data")))
+    tpl = Template("tests/demo_data")
```
sisp, line 39: I've reverted these kinds of changes because I think Copier handles POSIX paths fine both on Unix and Windows due to `pathlib.Path()` usage.

I tested this with `pathlib`:

```python
>>> pathlib.PureWindowsPath("a/b").parts
('a', 'b')
```

And `pathlib.Path()` seems to delegate the OS-specific implementation to the respective classes.

tests/test_config.py
yajo: probably just an overlook

tests/test_complex_questions.py
yajo: These are just 2 ways to express the same thing, so I'd keep it as it was.

tests/test_config.py
sisp: So I'll remove the `src / ` part, it shouldn't change the behavior of that test.

tests/test_empty_suffix.py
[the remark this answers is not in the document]
yajo: It's not an easy decision to take. On one hand, I obviously like prettier code. But OTOH, I already found 2 issues with the refactor, and the old motto "if something ain't broken, don't fix it" comes to my mind. I'm afraid that with the refactor we're introducing more non-obvious bugs, or at least skipping some other non-obvious corner cases that are being implicitly tested.

All my life I considered tests the only part of code that is "write and forget". I don't actually "maintain" tests. I just fix them when they're broken. And those breaks help me to know that the next release will be a major one, because something is broken. Also, writing tests is boring enough, so when a contributor contributes a test that is not diverging too much from the somehow modern way of testing copier, I just accept that. I mean: as long as the test is testing what it must test and goes green, it's mostly OK for me.

Putting all things in balance, sadly I think stability is weightier than beauty, so let me gladly ask you to please start reverting most of the diff. Let me do more comments on your specific questions below.

tests/test_prompt.py
yajo: Putting risk/value into balance, I think it's better to undo it.

tests/test_exclude.py
yajo: Let's undo it too.

tests/test_migrations.py
yajo: Of course, specific tests are better. But as long as we don't touch them, they won't add too much. Just undo the change; that'll be the most practical solution.

tests/test_subdirectory.py
yajo: OK, let's keep this one.

tests/test_templated_prompt.py
yajo: Let's undo this change too please.

tests/test_config.py
yajo: let's undo these too please.

tests/test_exclude.py
sisp: Oh well ... Done. :shrug:

tests/test_prompt.py
sisp: Oh well ... Done. :shrug:

tests/test_templated_prompt.py
sisp: Oh well ... Done. :shrug:

tests/test_empty_suffix.py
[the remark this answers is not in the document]
sisp: Oh well ... Done. :shrug:

sisp:
I've reverted some more changes. Would you like to take a look again, @yajo?

> On one hand, I obviously like prettier code. But OTOH, I already found 2 issues with the refactor, and the old motto "if something ain't broken, don't fix it" comes to my mind. I'm afraid that with the refactor we're introducing more non-obvious bugs, or at least skipping some other non-obvious corner cases that are being implicitly tested.
>
> All my life I considered tests the only part of code that is "write and forget". I don't actually "maintain" tests. I just fix them when they're broken. And those breaks help me to know that the next release will be a major one, because something is broken. Also, writing tests is boring enough, so when a contributor contributes a test that is not diverging too much from the somehow modern way of testing copier, I just accept that. I mean: as long as the test is testing what it must test and goes green, it's mostly OK for me.

I'm not sure how much further I want to take this discussion or just give in. :confused: Of course I understand the desire to keep the tests functional and the risk of losing that. But I disagree with tests code being "write and forget" because test code may evolve in similar ways as the main code does. And deliberately not maintaining test code contributes to technical debt. For instance, a project may have been started by a junior developer, which may have led to a bad tests organization, bad test scope, etc.; or the reviews in the early days of a project weren't as thorough as later; or certain testing patterns (setup/teardown code, naming conventions, etc.) became apparent only at a later point of the project. Those are all perfectly valid reasons to _maintain_, i.e. continuously refactor and clean up, a test suite because technical debt should be kept in check. While reading on this topic a bit, advice in several places was to never refactor the test suite _and_ change the main code which makes total sense and I haven't touched both here. Of course, there's no 100% guarantee, but the combination of leaving the main code untouched, using code quality metrics like code coverage as safeguards, using a linter, and thorough code review should minimize that risk. And especially purely stylistic changes, some of which I've even reverted by now, are obviously not problematic.

yajo (approved):
All those points have pros and cons. There's no silver bullet here.

For example, stylistic changes may seem more beautiful to some people and uglier to some others. They may make the code look better, but they also make `git blame` harder.

There's a difference between test _maintenance_ (which is important) and test _refactor_ (which is risky). For example, a test might have a bottleneck and make CI go slow. Or a library used for tests might introduce some breaking change and needs to update some tests. Or some app change involves a change in a test. That's just maintenance and is a good thing.

What I think is: what adds more value? To refactor a test, or to add a new one? And what is more risky?

In any case, let's merge this one as it seems to be OK now. Thanks for your help!

PR #986, opened 2023-02-19
feat: allow imports in inline templates

### LINKED ISSUE #718: Allow imports in inline templates

Copier users keep requesting variables support. There are good reasons to _avoid_ supporting variables:
- [You will want to use them to generate answer defaults](https://github.com/copier-org/copier/issues/716#issuecomment-1183326438). But you will want to use answers within them. 🐔 + 🥚 problem.
- Jinja already has many ways to DRY your code. Namely: [set](https://jinja.palletsprojects.com/en/3.0.x/templates/#assignments), [macro](https://jinja.palletsprojects.com/en/3.0.x/templates/#macros), [import](https://jinja.palletsprojects.com/en/3.0.x/templates/#import) and [include](https://jinja.palletsprojects.com/en/3.0.x/templates/#include).
- [YAML also supports anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/), which help you DRY your code within `copier.yaml`.
- There's already [the `ContextHook` extension for that](https://github.com/copier-org/copier-templates-extensions#context-hook-extension), and [it's on our docs](https://github.com/copier-org/copier/blob/e4f8956c30f8c4c75fa7e324ce0a435900da48be/docs/faq.md).

However, it's easy to understand that _people still need to keep their templates DRY_.

The real problem is not within the template files. As said, Jinja is DRY enough out of the box. The real problem lies within 2 other spaces:

1. Inside `copier.yaml`. Here, keys are not rendered but values are.
3. Templated file and directory names.

These are not files, so they get rendered in-line: https://github.com/copier-org/copier/blob/0f5257a0b01fde57ad69a2d51cb20eec1f9e6025/copier/user_data.py#L389

I really don't want to create a new system for declaring variables or importing code when Jinja already has them. I just want to be able to use them in in-line templates.

This way, I could declare a `copier.yaml` like this:

```yaml
_exclude:
  - name-slug

name:
  type: str
  default: A nice human-readable name

slug:
  type: str
  default: '{% include "name-slug.jinja" %}'
```

Also, I could have a template structure like this:

```
📄 copier.yaml
📄 name-slug.jinja
🐍  {% include "name-slug.jinja" %}.py
```

Apart from freeing Copier devs from reinventing some wheels, this would empower Copier templaters to *not only* add variables, but rather do a lot more stuff. Also it would make them more used to get reusable knowledge on the Jinja engine, rather than on a nic~~h~~e use case such as Copier.

**Describe how to implement it**

Quoting from [Jinja docs](https://jinja.palletsprojects.com/en/3.1.x/api/#basics):

> Using a template loader rather than passing strings to [Template](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.Template) or [Environment.from_string()](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.Environment.from_string) has multiple advantages. Besides being a lot easier to use _it also enables template inheritance_.

In other words, you can't import or include from an in-line template (or without a loader).

However, in our in-line templates, *we do have a filesystem location context*. So, the solution is relatively simple:

1. Write a [custom loader](https://jinja.palletsprojects.com/en/3.0.x/api/#loaders) that tells Jinja that context.
2. Let Copier know which loader to use, depending on where the rendered value comes from (is it a file? a path? or a `copier.yml` value?). This would mean possibly using one of [`PrefixLoader`](https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.PrefixLoader) or [`ChoiceLoader`](https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.ChoiceLoader).

See other user requests. All of them contain workarounds:

- #229
- #545 
- #629
- #678 
- #700
- #716 

**Additional context**

[Cookiecutter has private variables](https://cookiecutter.readthedocs.io/en/stable/advanced/private_variables.html).

sisp:
I've made an attempt at supporting Jinja2 [include](https://jinja.palletsprojects.com/en/3.1.x/templates/#include)/[import](https://jinja.palletsprojects.com/en/3.1.x/templates/#import) in inline templates in
1. templated values in `copier.yml`,
2. templated folder names,
3. templated file names, and
4. templated text file content.

Closes #718 (I hope :crossed_fingers:, or maybe some day).

---

To make this work, I've added a new option `_jinja_paths: Dict[str, str]` in `copier.yml` which is a mapping from identifiers to paths relative to the repository root from which additional Jinja2 templates are loaded, e.g.:

```yaml
_jinja_paths:
  $: includes
```

A template can be included like this:

```jinja
{% include "$:name-slug.jinja" %}
```

Under the hood, the Jinja2 `PrefixLoader` is used. My reasons for this approach are:

1. Templates to be included for, e.g., deriving values from answers are typically not copied to the project and IMO should be placed in a dedicated folder to avoid mixing them with other files. Instead of hardcoding the name for this folder, I think it's better to make it configurable.
6. I had thought about resolving the referenced template relative to where it is being used (e.g. an `{% include ... %}` in `copier.yml` would resolve relative to this file, an `{% include ... %}` in some templated folder name would resolve relative to this templated folder in the template, etc.), but this might easily require using `/` in the include path which isn't possible due to filesystem restrictions (i.e. a folder name like `{% include "includes/name-slug.jinja" %}` is not possible). IMO, the typical use case is to have a central collection of Jinja2 templates to be included or macros to be imported which suggests giving this "special" path an abstract identifier.

I consider this PR a basis for further discussion rather than ready to be merged (unless you think it's exactly what you had in mind :smile_cat:) since #718 doesn't fully specify the usage of this feature, so that I imagined how I would use it.

WDYT, @yajo?

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -461,7 +461,7 @@ class Worker:
         def _pathjoin(
             *path: str, mode: Literal["posix", "windows", "native"] = "posix"
         ) -> str:
-            return separators["mode"].join(path)
+            return separators[mode].join(path)
 
         env.filters["pathjoin"] = _pathjoin
         return env
--- a/copier/main.py
+++ b/copier/main.py
@@ -451,6 +451,9 @@ class Worker:
         env.filters["to_json"] = partial(
             env.filters["to_json"], default=pydantic_encoder
         )
+        # Add a helper function to join filesystem paths using the OS-specific
+        # path separator.
+        env.globals["pathjoin"] = os.path.join
         return env
 
     @cached_property
--- a/copier/main.py
+++ b/copier/main.py
@@ -451,9 +451,10 @@ class Worker:
         env.filters["to_json"] = partial(
             env.filters["to_json"], default=pydantic_encoder
         )
-        # Add a helper function to join filesystem paths using the OS-specific
-        # path separator.
-        env.globals["pathjoin"] = os.path.join
+        # Add a helper function to join filesystem paths using the path
+        # separator `/` as Jinja uses the POSIX and not OS-specific path
+        # separator for includes/imports.
+        env.globals["posixpath"] = lambda *p: "/".join(p)
         return env
 
     @cached_property
--- a/copier/main.py
+++ b/copier/main.py
@@ -33,6 +33,7 @@ from .tools import Style, TemporaryDirectory, printf
 from .types import (
     AnyByStrDict,
     JSONSerializable,
+    Literal,
     OptStr,
     RelativePath,
     StrOrPath,
@@ -449,10 +450,20 @@ class Worker:
         env.filters["to_json"] = partial(
             env.filters["to_json"], default=pydantic_encoder
         )
-        # Add a helper function to join filesystem paths using the path
-        # separator `/` as Jinja uses the POSIX and not OS-specific path
-        # separator for includes/imports.
-        env.globals["posixpath"] = lambda *p: "/".join(p)
+
+        # Add a filter to join filesystem paths.
+        separators = {
+            "posix": "/",
+            "windows": "\\",
+            "native": os.path.sep,
+        }
+
+        def _pathjoin(
+            *path: str, mode: Literal["posix", "windows", "native"] = "posix"
+        ) -> str:
+            return separators["mode"].join(path)
+
+        env.filters["pathjoin"] = _pathjoin
         return env
 
     @cached_property
--- a/copier/main.py
+++ b/copier/main.py
@@ -14,7 +14,7 @@ from shutil import rmtree
 from typing import Callable, Iterable, List, Mapping, Optional, Sequence
 from unicodedata import normalize
 
-from jinja2.loaders import ChoiceLoader, FileSystemLoader, PrefixLoader
+from jinja2.loaders import FileSystemLoader
 from jinja2.sandbox import SandboxedEnvironment
 from pathspec import PathSpec
 from plumbum import ProcessExecutionError, colors
@@ -425,19 +425,9 @@ class Worker:
 
         Respects template settings.
         """
-        loader = ChoiceLoader(
-            [
-                PrefixLoader(
-                    {
-                        prefix: FileSystemLoader(
-                            [str(self.template.local_abspath / path)]
-                        )
-                        for prefix, path in self.template.jinja_paths.items()
-                    },
-                    delimiter=":",
-                ),
-                FileSystemLoader([str(self.template.local_abspath)]),
-            ]
+        paths = [str(self.template.local_abspath)]
+        loader = FileSystemLoader(
+            paths,
         )
         default_extensions = [
             "jinja2_ansible_filters.AnsibleCoreFiltersExtension",
--- a/copier/main.py
+++ b/copier/main.py
@@ -451,7 +451,7 @@ class Worker:
             env.filters["to_json"], default=pydantic_encoder
         )
 
-        # Add a filter to join filesystem paths.
+        # Add a global function to join filesystem paths.
         separators = {
             "posix": "/",
             "windows": "\\",
@@ -463,7 +463,7 @@ class Worker:
         ) -> str:
             return separators[mode].join(path)
 
-        env.filters["pathjoin"] = _pathjoin
+        env.globals["pathjoin"] = _pathjoin
         return env
 
     @cached_property
--- a/copier/main.py
+++ b/copier/main.py
@@ -299,11 +299,7 @@ class Worker:
         return bool(ask(f" Overwrite {dst_relpath}?", default=True))
 
     def _render_allowed(
-        self,
-        dst_relpath: Path,
-        is_dir: bool = False,
-        expected_contents: bytes = b"",
-        expected_permissions=None,
+        self, dst_relpath: Path, is_dir: bool = False, expected_contents: bytes = b""
     ) -> bool:
         """Determine if a file or directory can be rendered.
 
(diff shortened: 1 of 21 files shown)


sisp:
Windows seems to choke on some character in the templated path name. I'll investigate this once I have some feedback on the general idea.

pawamoy:
Double quotes are indeed invalid characters in Windows paths, try using single quotes instead.

sisp:
Thanks, @pawamoy! I've changed the quotes, but CI on Windows is still failing. Might still be the `:` character.

About using `/` in paths: I just stumbled over the docs section on [generating a directory structure](https://copier.readthedocs.io/en/latest/configuring/#generating-a-directory-structure) which suggests it is actually possible to include path separators. Perhaps a Jinja2 include in a templated folder name might look like this:

```shell
📁 my_copier_template
├── 📄 copier.yml
├── 📁 includes
│   └── 📄 name-slug.jinja
└── 📁 template
    └── 📄 {% include '..' + _copier_conf.sep + 'includes' + _copier_conf.sep + 'name-slug.jinja' %}.py
```

Not super readable, but might work. Would this be preferable over using a prefixed/named load path?

Perhaps it's not actually the filesystem that doesn't support `/` in folder names, but I observed that `Path::iterdir()` in

https://github.com/copier-org/copier/blob/468b9ff78cf2331a70d8b99a6b1a812e3c40f1c4/copier/main.py#L542

splits at the `/` in `{% include '../includes/name-slug.jinja' %}`.

sisp:
> Not super readable, but might work.

The DX might be better if we used a Jinja2 filter, e.g.:

```jinja
{% include ['..', 'includes', 'name-slug.jinja']|join(_copier_conf.sep) %}.py
```

Or with a custom filter for even better readability which internally uses `_copier_conf.sep`:

```jinja
{% include joinpath('..', 'includes', 'name-slug.jinja') %}.py
```

pawamoy:
> Might still be the : character.

Ah, true, I remember now that I couldn't use `:` in tracks/movies file names on Windows.

sisp:
I have revised this PR and think it's ready for a review now.

I've reverted the initial approach with the `_jinja_paths` setting and Jinja's `PrefixLoader` because I dislike that API and don't think it's the right approach. I realized that [it is possible to use path separators in templated folder/file names](https://github.com/copier-org/copier/pull/986#issuecomment-1436003159), although it turned out that Jinja requires the POSIX path separator in include/import statements which means `_copier_conf.sep` cannot be used in templated folder or file names on Windows. So, I've come up with a more flexible and IMO intuitive solution:

Jinja includes/imports are now relative to the template root independent of the include/import target (`copier.yml`, templated folder name, templated file name, templated text file content) and I think this is the simplest, most flexible, most intuitive, and most robust approach. I was thinking about using relative paths (relative to the include/import target, e.g. relative to `copier.yml` when it's inside `copier.yml`, or relative to a folder when it's in a templated folder name, etc.), but I eventually didn't pursue this approach further for two reasons:

1. I think the main use case is to include/import from a central place rather than from a location relative to the target. Relative paths would make the include/import statements long and difficult to read when moving up several times in the filesystem.
2. I think the correct relative path might not be obvious when using a [generated directory structure](https://copier.readthedocs.io/en/latest/configuring/#generating-a-directory-structure) where the template contains only one folder that might render to a deeply nested folder, so it might be unclear whether the include/import path is relative to the templated folder name or the rendered (nested) folder.

I'm absolutely open for discussion though because it's an important design decision that will be difficult to change later without introducing a breaking change. WDYT?

copier/main.py
```diff
@@ -449,6 +449,10 @@ def jinja_env(self) -> SandboxedEnvironment:
         env.filters["to_json"] = partial(
             env.filters["to_json"], default=pydantic_encoder
         )
+        # Add a helper function to join filesystem paths using the path
+        # separator `/` as Jinja uses the POSIX and not OS-specific path
+        # separator for includes/imports.
+        env.globals["posixpath"] = lambda *p: "/".join(p)
```
yajo, line 455: WDYT about calling it `posixjoin` and using it as filter instead?

    "Base" | posixjoin("next", "next")| posixjoin ("woohoo more!")

docs/configuring.md
```diff
@@ -490,6 +490,132 @@ your_project
 You can either use any separator, like `.`, and replace it with `_copier_conf.sep`, like
 in the example above, or just use `/` in the answer (works on Windows too).
 
+## Importing Jinja templates and macros
+
+You can import templates and macros to reduce code duplication. A common scenario is the
```
yajo, line 495: I miss links to jinja docs

yajo (commented):
Brilliant. Amazing how such a small patch opens so many doors!

Non-surprisingly, the original creator of copier asked upstream about relative imports and the answer was similar to what you ended up reasoning: https://github.com/pallets/jinja/issues/1533

Some suggestions.

copier/main.py
yajo: Or even better: `pathjoin`, with a kwarg `mode` that accepts "native" , "windows" or "posix" (default).

sisp: Would you prefer a filter over a global function because the global function pollutes the context? E.g. a template creator would not be able to name a question `pathjoin` (after the renaminh you suggested)? Or is there a particular scenario you have in mind where chaining `pathjoin` filters comes in handy?

I'm not opposed to the filter approach, just trying to understand your thoughts.

sisp: I went ahead and made the change as you suggested. I'd still be curious about your thoughts.

sisp: Mhh, CI on Windows is failing. I'll take a look tomorrow.

sisp: @yajo I'm afraid using a filter in a folder of file name won't work because the [`|` character is not allowed there on Windows](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming-conventions). That's why the CI job in Windows is failing and I've manually verified it on a Windows machine as well. I suggest to register `pathjoin()` as a function (similar to before):

```jinja
{% include pathjoin('includes', 'name-slug.jinja') %}.py
```

I'd keep the optional kwarg `mode` of course. WDYT?

sisp: It's not critical for this particular feature, but I wonder whether maintaining Windows compatibility is worth the sacrifice of not being able to use Jinja filters in folder and file names in general. :thinking: My experience with developing software on Windows is awful. And since there is WSL, Windows users can escape into a sane development environment.

yajo: Yes, maintaining windows is terrible. However it's more or less done already, so let's stick with the function solution. The filter looked cooler to me, but given this technical limitation, I think it's the best.

sisp: Alright, I've reverted `pathjoin()` to a function.

sisp:
I think this PR is ready to be merged now. Two CI jobs are failing because of timeout issues.

PR #1006, opened 2023-02-26
fix: parse CLI data using question's answer parser

sisp:
I've fixed the parsing of answers provided via CLI using the `--data` flag (or implicitly also via API using the `data` argument when answers are provided as strings).

Before, answers passed via the `--data` flag (i.e. string values) were [parsed using `yaml.safe_load(...)`](https://github.com/copier-org/copier/blob/ab31906aba40bec33bffec99e9ea15fe717690e2/copier/cli.py#L194) which in general is not the same as parsing answers via the [question's answer parser](https://github.com/copier-org/copier/blob/ab31906aba40bec33bffec99e9ea15fe717690e2/copier/user_data.py#L401-L411). For instance, `--data=str_question=1.2` was parsed as a `float` value although it's a `str` question. I've removed the incorrect `yaml.safe_load(...)` parser and use the question's parser instead, so that answer are parsed according to the type of their corresponding question.

Note also that two tests in `tests/test_complex_questions.py` turned out to have incorrect expectations. I've fixed those, too.

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -50,7 +50,6 @@ from pathlib import Path
 from textwrap import dedent
 from unittest.mock import patch
 
-import yaml
 from plumbum import cli, colors
 
 from copier.tools import copier_version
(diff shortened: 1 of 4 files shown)


yajo (approved):
This one is great!

Issue #996, opened 2023-02-21
Validator support on prompt using choices


verdaatt:
**Is your feature request related to a problem? Please describe.**

Input validation is supported on prompts that do not use the `choices` option. There are however valid reasons for wanting to use the validator feature to validate if the chosen option from the choices list is compatible with inputs on other prompts. An example is these prompts for inputting Terraform and Terragrunt versions, where there's a Terraform version compatibility matrix for Terragrunt that we'd like to enforce using the validator:

```yaml
terraform_version:
  type: str
  default: "1.3.9"
  help: Which Terraform version do you want to use?
  choices:
    "1.3": "1.3.9"
    "1.2": "1.2.9"
    "1.1": "1.1.9"
    "1.0": "1.0.11"
    "0.15": "0.15.5"
    "0.14": "0.14.11"
    "0.13": "0.13.7"

terragrunt_version:
  type: str
  default: "0.43.2"
  help: |
    Which Terragrunt version you want to use?

    When choosing a previous version, please be mindful of Terraform version compatibility: https://terragrunt.gruntwork.io/docs/getting-started/supported-terraform-versions/
  choices:
    - "0.43.2"
    - "0.42.8"
    - "0.41.0"
    - "0.40.2"
    - "0.39.2"
    - "0.38.12"
    - "0.37.4"
    - "0.36.12"
    - "0.35.20"
    - "0.34.3"
    - "0.33.2"
    - "0.32.6"
    - "0.31.11"
    - "0.30.7"
    - "0.29.10"
    - "0.28.24"
    - "0.27.4"
    - "0.26.7"
    - "0.25.5"
  # Validator implements https://terragrunt.gruntwork.io/docs/getting-started/supported-terraform-versions/
  validator: >-
    {% if (terraform_version | regex_search('^1\.3\..*$')) %}
      {% if not (terragrunt_version | regex_search('^0\.(4\d|[5-9]\d).*$')) %}
      This version is not compatible with the Terraform 1.3.x, Please use version >= 0.40.0
      {% endif %}
    {% elif (terraform_version | regex_search('^1\.2\..*$')) %}
      {% if not (terragrunt_version | regex_search('^0\.(3[89]|[4-9]\d).*$')) %}
      This version is not compatible with the Terraform 1.2.x, Please use version >= 0.38.0
      {% endif %}
    {% elif (terraform_version | regex_search('^1\.1\..*$')) %}
[...]

PR #1010, opened 2023-02-27
feat: support choice validation via extended syntax 

### LINKED ISSUE #1001: Add support for conditional choices
Sometimes, there is a dependency between questions where the validity of an answer to a choice question depends on the answer to a previous question. Currently, Copier doesn't support enforcing constraints on choice questions.

I'd like to be able to make a choice item conditional, so that only valid choices are presented in interactive mode and an invalid answer would raise a validation error in CLI/API mode.

I believe conditional choice items could be implemented like this:

```yaml
# List syntax
question_with_list_choices:
  choices:
    - "{% if <condition-1> %}choice-1{% endif %}"
    - "{% if <condition-2> %}choice-2{% endif %}"
    - ...

# Dict syntax
question_with_dict_choices:
  choices:
    "{% if <condition-1> %}choice-1{% endif %}": null
    "{% if <condition-2> %}choice-2{% endif %}": null

# Tuple syntax
question_with_tuple_choices:
  choices:
    - ["{% if <condition-1> %}choice-1{% endif %}", null]
    - ["{% if <condition-2> %}choice-2{% endif %}": null]
```

Whenever (a) the list item is an empty string or (b) the dict key is an empty string or (c) the first element in the tuple is an empty string, the choice is ignored. Of course, a non-`null` value is possible for the dict and tuple syntax, but I wanted to keep the example consistent.

Copier currently allows an empty choice:

```yaml
question_with_list_choices:
  choices:
    - ""
```

```shell
$ copier $tpl $dst
🎤 question_with_list_choices
   (Use arrow keys)
 »
```

However, I don't think empty choices make much sense, so it should be fine to ignore them.

Instead of omitting invalid choices, it might be useful to show all choices and validate them, so that invalid choices show an error when selected. Regarding UX, I think omitting invalid choices is preferable to showing errors though.

**Additional context**
Related to #996.

sisp:
I've added support for conditional choices, i.e. choices whose validity depends on answers to previous questions.

In interactive mode, disabled choices are not shown and cannot be selected. In CLI/API mode, passing the value of a disabled choice will raise an error.

The templated prompt test case is based on the example from #996. If I understand the [compatibilities between Terraform and Terragrunt versions](https://terragrunt.gruntwork.io/docs/getting-started/supported-terraform-versions/) right, these versions should be compatible:

<table>
<tr>
<td colspan="2" rowspan="2"></td>
<th colspan="4">Terragrunt</th>
</tr>
<tr>
<td>0.38</td>
<td>0.39</td>
<td>0.40</td>
<td>0.41</td>
</tr>
<tr>
<th rowspan="2">Terraform</th>
<td>1.2</td>
<td>:heavy_check_mark:</td>
<td>:heavy_check_mark:</td>
<td>:heavy_check_mark:</td>
<td>:heavy_check_mark:</td>
</tr>
<tr>
<td>1.3</td>
<td></td>
<td></td>
<td>:heavy_check_mark:</td>
<td>:heavy_check_mark:</td>
</tr>
</table>

Closes #1001. Related to #996.

/cc @verdaatt


yajo (changes requested):
Please could you implement this design instead? https://github.com/copier-org/copier/issues/1001#issuecomment-1464842885

sisp:
I've changed the implementation to use the extended syntax we discussed. I've also updated the example and test case to a cloud & IaC tool questionnaire because I found it more intuitive and easier to grasp.

One test is failing now because this PR requires #1111 to be merged first.

sisp:
#1111 has been merged and I've pulled the latest state of `master`. The tests are passing now, I think this PR is ready for a review. :pray:

yajo (approved):
Code looks OK, thanks!

I still have to do the functional review.

PR #1022, opened 2023-03-05
introducing project mode

rafalkrupinski:
1. First commit extracts a Renderer class from Worker.
     Not only Worker is huge and does a lot, but the new class greatly simplifies rendering the template for each project-mode item in the next commit.
    The idea behind it was that it encapsulates the template and accepts answers to render the output.
    I decided to leave _allow_render in Worker and pass it as a function argument, as it uses data stored in Worker. Let me know if you can think of a better way to split responsibility between those two.
2. Add looping over items
  - accept `_items` in `copier.yaml` as either a string or a list of strings. Internally it's always a list.
  - use each key to read a list of items from answers, and render _its part_ of the template, stored under `{{template_root}}/{{key}}`
    So if you have _items: ['a'], copier expects an answer `a` to be a list of values, and uses each value to render a sub-template stored in 'a'.
  - all directories under root ,that are named like the keys under `_items` are excluded from normal rendering.
3. Not sure how to structure the documentation. Needs changes to both the template and configuration parts. Shell I create a new file for the project mode?
4. If nobody minds, I'd like to disable the "identical" message for directories. I'm getting it a hundred of times when run in a loop and it doesn't do anything useful for directories.

example template: https://github.com/python-lapidary/lapidary-template/tree/test2

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -27,6 +27,7 @@ from pydantic.json import pydantic_encoder
 from questionary import unsafe_prompt
 
 from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
+from .renderer import Renderer
 from .subproject import Subproject
 from .template import Task, Template
 from .tools import Style, TemporaryDirectory, printf
@@ -466,113 +467,6 @@ class Worker:
             )
         )
 
-    def _render_file(self, src_abspath: Path) -> None:
-        """Render one file.
-
-        Args:
-            src_abspath:
-                The absolute path to the file that will be rendered.
-        """
-        # TODO Get from main.render_file()
-        assert src_abspath.is_absolute()
-        src_relpath = src_abspath.relative_to(self.template.local_abspath).as_posix()
-        src_renderpath = src_abspath.relative_to(self.template_copy_root)
-        dst_relpath = self._render_path(src_renderpath)
-        if dst_relpath is None:
-            return
-        if src_abspath.name.endswith(self.template.templates_suffix):
-            try:
-                tpl = self.jinja_env.get_template(src_relpath)
-            except UnicodeDecodeError:
-                if self.template.templates_suffix:
-                    # suffix is not empty, re-raise
-                    raise
-                # suffix is empty, fallback to copy
-                new_content = src_abspath.read_bytes()
-            else:
-                new_content = tpl.render(**self._render_context()).encode()
-        else:
-            new_content = src_abspath.read_bytes()
-        dst_abspath = Path(self.subproject.local_abspath, dst_relpath)
-        src_mode = src_abspath.stat().st_mode
-        if not self._render_allowed(dst_relpath, expected_contents=new_content):
-            return
-        if not self.pretend:
-            dst_abspath.parent.mkdir(parents=True, exist_ok=True)
-            dst_abspath.write_bytes(new_content)
-            dst_abspath.chmod(src_mode)
-
-    def _render_folder(self, src_abspath: Path) -> None:
-        """Recursively render a folder.
-
-        Args:
-            src_path:
-                Folder to be rendered. It must be an absolute path within
-                the template.
-        """
-        assert src_abspath.is_absolute()
-        src_relpath = src_abspath.relative_to(self.template_copy_root)
-        dst_relpath = self._render_path(src_relpath)
-        if dst_relpath is None:
-            return
-        if not self._render_allowed(dst_relpath, is_dir=True):
-            return
-        dst_abspath = Path(self.subproject.local_abspath, dst_relpath)
-        if not self.pretend:
-            dst_abspath.mkdir(parents=True, exist_ok=True)
-        for file in src_abspath.iterdir():
-            if file.is_dir():
-                self._render_folder(file)
-            else:
-                self._render_file(file)
-
-    def _render_path(self, relpath: Path) -> Optional[Path]:
-        """Render one relative path.
-
-        Args:
-            relpath:
-                The relative path to be rendered. Obviously, it can be templated.
-        """
-        is_template = relpath.name.endswith(self.template.templates_suffix)
-        templated_sibling = (
-            self.template.local_abspath / f"{relpath}{self.template.templates_suffix}"
-        )
-        # With an empty suffix, the templated sibling always exists.
-        if templated_sibling.exists() and self.template.templates_suffix:
-            return None
-        if self.template.templates_suffix and is_template:
-            relpath = relpath.with_suffix("")
-        rendered_parts = []
-        for part in relpath.parts:
-            # Skip folder if any part is rendered as an empty string
-            part = self._render_string(part)
-            if not part:
-                return None
-            # {{ _copier_conf.answers_file }} becomes the full path; in that case,
-            # restore part to be just the end leaf
-            if str(self.answers_relpath) == part:
-                part = Path(part).name
-            rendered_parts.append(part)
-        result = Path(*rendered_parts)
-        if not is_template:
-            templated_sibling = (
-                self.template.local_abspath
-                / f"{result}{self.template.templates_suffix}"
-            )
-            if templated_sibling.exists():
-                return None
-        return result
-
-    def _render_string(self, string: str) -> str:
-        """Render one templated string.
-
-        Args:
-            string:
-                The template source string.
-        """
-        tpl = self.jinja_env.from_string(string)
-        return tpl.render(**self._render_context())
-
     @cached_property
     def subproject(self) -> Subproject:
         """Get related subproject."""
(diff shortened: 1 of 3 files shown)


rafalkrupinski:
For me these tests fail on master as well

sisp:
They pass for me on `master`, and CI seems happy, too. :thinking:

sisp (commented):
A couple of thoughts from my side:

1. If you think that the `main.Worker` class does too much and it would be better to factor out the rendering methods into `renderer.Renderer`, I suggest to do this refactoring in a dedicated PR to separate concerns and ease the review. In any case, @yajo will need to decide whether he likes the refactoring.
2. From a template creator's perspective, I find it a bit unexpected that the files in the folder marked via `_items` are merged in the target folder. For instance, in `test_render_items_list` the files in both `src/apples/` and `src/oranges/` are generated to `<dst>/` and not to `<dst>/apples/` and `<dst>/oranges/`, respectively. I imagine there are technical reasons for this behavior, but the point of declarative project templates is to make the them readable and close to the generated project. Arguably, [generating a directory structure](https://copier.readthedocs.io/en/latest/configuring/#generating-a-directory-structure) is also in violation of this goal, but in that case the benefits outweigh the loss in readability IMO, and at least there is no implicit magic unlike the implicit merging of the `src/apples/` and `src/oranges/` folders.
3. I'm a bit confused by `test_render_items_conditional`. The behavior tested there seems to equivalent to using a dynamic [subdirectory](https://copier.readthedocs.io/en/latest/configuring/#subdirectory). The only difference is the usage of the new special variable `_item`. Is the goal of this test case to test the rendering of dict items?
4. There was some minor discussion about nested loops in https://github.com/copier-org/copier/discussions/908. Would nested loops be possible with the current design? If yes, could you add a test case? I am vaguely worried that the current design might not be sufficiently flexible to address more advanced cases because the rendering context is only the main context plus the current item's data, but with nested loops the context from the parent loop seems to be unavailable in a child loop.

rafalkrupinski:
> A couple of thoughts from my side:
> 
>     1. If you think that the `main.Worker` class does too much and it would be better to factor out the rendering methods into `renderer.Renderer`, I suggest to do this refactoring in a dedicated PR to separate concerns and ease the review. In any case, @yajo will need to decide whether he likes the refactoring.

Yeah, I thought it's easier to change from a single PR to multiple, than the other way around

>     2. From a template creator's perspective, I find it a bit unexpected that the files in the folder marked via `_items` are merged in the target folder. For instance, in `test_render_items_list()` the files in both `src/apples/` and `src/oranges/` are generated to `<dst>/` and not to `<dst>/apples/` and `<dst>/oranges/`, respectively. I imagine there are technical reasons for this behavior, but the point of declarative project templates is to make the them readable and close to the generated project. Arguably, [generating a directory structure](https://copier.readthedocs.io/en/latest/configuring/#generating-a-directory-structure) is also in violation of this goal, but in that case the benefits outweigh the loss in readability IMO, and at least there is no implicit magic unlike the implicit merging of the `src/apples/` and `src/oranges/` folders.

The first version I did was with only one items list and without the directory. In result, all files and directories under the template root had to be named with a condition (`{% if _item %}` or `{% if not _item %}`). I found that hard to read.
With multiple item lists it gets slightly worse.

Those directories are already treated specially - their contents is rendered once for every item. I would find it weird if they were copied over to the output, but I guess it's just a matter of expectations of a new user of Copier vs a seasoned one.

I'd rather not add it as a flag, bc removing the directories IMO would be more reasonable default, but not what older users might expect.

How about those directories go to `_subtemplates` directory to signify their special treatment?
e.g. `src/_subtemplates/apples/* -> dst/*`

>     3. I'm a bit confused by `test_render_items_conditional()`. The behavior tested there seems to equivalent to using a dynamic [subdirectory](https://copier.readthedocs.io/en/latest/configuring/#subdirectory). The only difference is the usage of the new special variable `_item`. Is the goal of this test case to test the rendering of dict items?

Good catch. It's from the previous version, updated a bit thoughtlessly.

>     4. There was some minor discussion about nested loops in #908. Would nested loops be possible with the current design? If yes, could you add a test case? I am vaguely worried that the current design might not be sufficiently flexible to address more advanced cases because the rendering context is only the main context plus the current item's data, but with nested loops the context from the parent loop seems to be unavailable in a child loop.

You're referring to [this](https://github.com/copier-org/copier/discussions/908#discussioncomment-4803319). I never intended to implement it due to exponential complexity of such templates.

Considering the goal of this change (generating SDK or class model from a machine readable description), I'd say nested loops are out of scope. If you're generating a class model, you have a single list of table definitions or JSON schemas to generate classes for - that's a single list. In case of SDK for OpenAPI, you have a bunch of schemas and operations, that's two lists, processed in parallel. All that assuming your model is large enough to render it as multiple files, otherwise you don't even need project mode.

Do you have a use case in mind?

rafalkrupinski:
> For me these tests fail on master as well

The tests mostly pass on an identical environment in my fork.
https://github.com/python-lapidary/copier/actions/runs/4333218244

rafalkrupinski:
@yajo Any idea how to progress with this?

copier/renderer.py
```diff
@@ -0,0 +1,174 @@
+from collections import ChainMap
+from dataclasses import dataclass, field
```
yajo, line 2: Probably you should change this to a pydantic dataclass, just to be consistent with the rest of Copier.

copier/renderer.py
```diff
@@ -0,0 +1,174 @@
...
+            self.template,
+            self.subproject,
+            self._render_allowed,
+            self.pretend,
+            self.jinja_env,
+            self.answers_relpath,
+            ChainMap({"_item": items_value}, self._render_context),
+            items_key,
+            [],
+        )
```
yajo, line 174: You can do this probably:

```suggestion
        return replace(
            self,
            _render_context=ChainMap({"_item": items_value}, self._render_context),
            _items_key=items_key,
            _items_keys=[],
        )
```

copier/main.py
```diff
@@ -421,6 +425,13 @@ def all_exclusions(self) -> StrSeq:
 
     @cached_property
     def jinja_env(self) -> SandboxedEnvironment:
+        return cast(SandboxedEnvironment, self.abstract_jinja_env(SandboxedEnvironment))
+
+    @cached_property
+    def native_jinja_env(self) -> NativeEnvironment:
```
yajo, line 431: I don't see any uses of this...

copier/main.py
```diff
@@ -810,6 +711,33 @@ def _git_initialize_repo(self):
...
+
+    @cached_property
+    def project_mode_keys(self) -> Iterable[str]:
+        raw_items = self.template.config_data.get("items")
+        if not raw_items:
+            return ()
+        return (
+            [raw_items]
+            if isinstance(raw_items, str)
+            else cast(collections.abc.Iterable, raw_items)
```
yajo, line 738: If it is a string, then it should probably get rendered using the native jinja env, to allow for complex scenarios. Right?

copier/main.py
```diff
@@ -810,6 +711,33 @@ def _git_initialize_repo(self):
...
+            answers_relpath=self.answers_relpath,
+            _render_context=self._render_context(),
+            _items_keys=self.project_mode_keys,
+        )
+
+    def _render_string(self, string: str) -> str:
+        return self._default_renderer.render_string(string)
+
+    @cached_property
+    def project_mode_keys(self) -> Iterable[str]:
```
yajo, line 731: All have been projects since forever... let's give it a better name:

```suggestion
    def looped_mode_keys(self) -> Iterable[str]:
```

tests/test_project_mode.py
yajo, line 1: Call the file `test_looped_mode.py`

yajo (commented):
I find surprising also what @sisp said. I think that a looped template should behave as close as possible to a non-looped one.

My expectations are that it just gets rendered in loop under different context. Since the loop is sequential, any file that gets rendered more than once would get only the last result. It is up to the template designer to avoid that kind of situations; we just have to document that.

FWIW the normal thing would be that a file that gets rendered twice results in the same output if the context didn't change or `_item` isn't used within its content.

If it's needed to avoid that, then we can also add a key such as `_dynamic_files` with a list of gitignore-like patterns (just like `_exclude`) of files that should be rendered in loop. By default: all files. But that would be an extra thing to add later. I wouldn't include this in the MVP if possible.

So a simple design would be:

```yaml
# src/copier.yaml
_items: "{{ classes }}"

classes:
  type: yaml
  help: write a list of classes you want to generate
  default: [one, two]

user: your name
```

Then another file named `src/{{ _item }}.txt`:

```
{{ user }}
```

I'd run `copier -f src dst` and expect to get:

```
dst/one.txt (contents: "your name")
dst/two.txt (contents: "your name")
```

Do you think you would be able to implement such design?

yajo:
Oh and regarding the refactor... no problem with it.

sisp:
@yajo Have you seen my comment in the other discussion about this topic? https://github.com/copier-org/copier/discussions/908#discussioncomment-5466993 I think that approach would cover a broader range of scenarios with more intuitive API. WDYT?

yajo:
I like that design but i think it can be hard to implement. Gotta think
about this deeply.

Issue #1020, opened 2023-03-03
Add a 'check' CLI command / API to check for new template version


WernerWessely:
**Is your feature request related to a problem? Please describe.**

I would like for our developers or CI to check whether a new version of a project's template is available. This way they can know whether/when they should update.

I got the idea from the `cruft check` command: https://cruft.github.io/cruft/#checking-a-project. This command checks whether the project's template is up to date. Optionally the command can exit with some errorcode != 0 for use in scripts, e.g. in CI.

Does Copier maybe already offer this in some way that I overlooked?

**Describe the solution you'd like**

Perhaps a `copier check` command or a `copier update --check` option?

**Describe alternatives you've considered**
I am considering writing our own tool based on Copier until this becomes part of the CLI.  My solution as part of Click based CLI currently looks like this:
```Python
@broj.command(
    short_help="Check for a newer template version for an existing project.")
@click.argument("project_dir", default=".",
                type=click.Path(file_okay=False, resolve_path=True,
                                writable=True, path_type=Path)
                )
@click.option(
    "-e", "--error",
    help="Exit with code 128 if a newer version is available.",
    is_flag=True
)
def check(project_dir, error):
    """
    Check if a newer template version is available for an existing project.

    PROJECT_DIR must be a git repo, if it is not passed, the current
    directory is used. It must also contain a '.copier-answers.yml' file
    with the current template settings.

    This command will try to compute the version currently used by the project
    and the remote template and decide whether an update is necessary.
    """
    with project_dir:

        _check_answers_file()

        with Worker() as worker:
            proj_version = worker.subproject.template.version
            temp_version = worker.template.version
            temp = worker.template.url

            if proj_version is None:
                raise click.ClickException(
                    f"Failed to determine version for project {project_dir}!")
            click.echo(
                f"Project {project_dir} template version: {proj_version}")

            if temp_version is None:
                raise click.ClickException(
                    f"Failed to determine version for template {temp}!")
            click.echo(f"Template {temp} version: {temp_version}")

            if temp_version > proj_version:
                click.echo("Update with 'broj update'!")
                if error:
                    exit(128)
            else:
                click.echo("There is nothing to do!")
```

**Additional context**

Thanks for the great work you guys do on Copier! 

I would be willing to try my clumsy hand at a PR if we think such a feature makes sense.

sisp:
Copier doesn't have this feature yet, but I like it. 👍  from my side.

WernerWessely:
[...]

PR #1031, opened 2023-03-08
feat: add the 'copier check-update' command to check if template is up-to-date

WernerWessely:
Here is my proposal for a 'copier check' command.

Resolves: #1020 

I have not yet added any test or docs. If we think the idea and design is ok, I will get to work on those.

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -41,7 +41,7 @@ Basically, there are 4 different commands you can run:
         ```
 
 -   [`copier check`][copier.cli.CopierCheckSubApp] to check if a preexisting
-    project is using the latest version of its template
+    project is using the latest version of its template.
 
     !!! example
 
@@ -386,7 +386,7 @@ class CopierCheckSubApp(cli.Application):
         (it must be a key called `_src_path`).
 
         If that file contains also `_commit` and `destination_path` is a git
-        repository, this command will do its best determine whether a newer
+        repository, this command will do its best to determine whether a newer
         version is available, applying PEP 440 to the template's history.
         """
     )
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -1,7 +1,7 @@
 """
 Command line entrypoint. This module declares the Copier CLI applications.
 
-Basically, there are 3 different commands you can run:
+Basically, there are 4 different commands you can run:
 
 -   [`copier`][copier.cli.CopierApp], the main app, which is a shortcut for the
     `copy` and `update` subapps.
@@ -40,6 +40,15 @@ Basically, there are 3 different commands you can run:
         copier update
         ```
 
+-   [`copier check`][copier.cli.CopierCheckSubApp] to check if a preexisting
+    project is using the latest version of its template
+
+    !!! example
+
+        ```sh
+        copier check
+        ```
+
 Below are the docs of each one of those.
 """
 
@@ -49,12 +58,11 @@ from io import StringIO
 from pathlib import Path
 from textwrap import dedent
 from unittest.mock import patch
-
 from plumbum import cli, colors
 
-from .errors import UserMessageError
+from copier.tools import copier_version
+from .errors import UserMessageError, SubprojectOutdatedError
 from .main import Worker
-from .tools import copier_version
 from .types import AnyByStrDict, OptStr, StrSeq
 
 
@@ -113,6 +121,7 @@ class CopierApp(cli.Application):
         """\
         copier [MAIN_SWITCHES] [copy] [SUB_SWITCHES] template_src destination_path
         copier [MAIN_SWITCHES] [update] [SUB_SWITCHES] [destination_path]
+        copier [MAIN_SWITCHES] check [SUB_SWITCHES] [destination_path]
         """
     )
     VERSION = copier_version()
@@ -355,6 +364,59 @@ class CopierUpdateSubApp(cli.Application):
             conflict=self.conflict,
             context_lines=self.context_lines,
         ).run_update()
+        return 0
+
+
+@CopierApp.subcommand("check")
+class CopierCheckSubApp(cli.Application):
+    """The `copier check` subcommand.
+
+    Use this subcommand to check if an existing subproject is using the
+    latest version of a template.
+
+    Attributes:
+        exit-code: Set [exit-code][] option.
+    """
+
+    DESCRIPTION = "Check if a copy is using the latest version of its original template"
+    DESCRIPTION_MORE = dedent(
+        """\
+        The copy must have a valid answers file which contains info
+        from the last Copier execution, including the source template
+        (it must be a key called `_src_path`).
+
+        If that file contains also `_commit` and `destination_path` is a git
+        repository, this command will do its best determine whether a newer
+        version is available, applying PEP 440 to the template's history.
+        """
+    )
+
+    exit_code: cli.SwitchAttr = cli.SwitchAttr(
+        ["-e", "--exit-code"],
+        int,
+        default=0,
+        help=(
+            "Exit code for the command if a newer version is available, "
+            "for use in scripts."
+        ),
+    )
+
+    @handle_exceptions
+    def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
+        """Call [run_check][copier.main.Worker.run_check].
+
+        Parameters:
+            destination_path:
+                Only the destination path is needed to check, because the
+                `src_path` comes from [the answers file][the-copier-answersyml-file].
+
+                The subproject must exist. If not specified, the currently
+                working directory is used.
+        """
+        try:
+            self.parent._worker(dst_path=destination_path).run_check()
+        except SubprojectOutdatedError:
+            return self.exit_code
         return 0
 
 
--- a/copier/main.py
+++ b/copier/main.py
@@ -26,7 +26,7 @@ from pydantic.dataclasses import dataclass
 from pydantic.json import pydantic_encoder
 from questionary import unsafe_prompt
 
-from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError
+from .errors import CopierAnswersInterrupt, ExtensionNotFoundError, UserMessageError, SubprojectOutdatedError
 from .subproject import Subproject
 from .template import Task, Template
 from .tools import Style, TemporaryDirectory, printf, readlink
@@ -767,6 +767,43 @@ class Worker:
             )
         self._apply_update()
 
+    def run_check(self) -> None:
+        """Check if a subproject is using the latest version of it's template
+
+        See [checking a project][checking-a-project].
+        """
+        # Check all you need is there
+        if self.subproject.template is None or self.subproject.template.ref is None:
+            raise UserMessageError(
+                "Cannot check because cannot obtain old template references "
+                f"from `{self.subproject.answers_relpath}`."
+            )
+        if self.template.commit is None:
+            raise UserMessageError(
+                "Checking is only supported in git-tracked templates."
+            )
+        if not self.subproject.template.version:
+            raise UserMessageError(
+                "Cannot check: version from last update not detected."
+            )
+        if not self.template.version:
+            raise UserMessageError("Cannot check: version from template not detected.")
+
+        if not self.quiet:
+            # TODO Unify printing tools
+            print(
+                f"Currently using template version {self.subproject.template.version}, "
+                f"latest is {self.template.version}.",
+                file=sys.stderr)
+
+        if self.template.version > self.subproject.template.version:
+            if not self.quiet:
+                print("New template version available.", file=sys.stderr)
+            raise SubprojectOutdatedError("")
+        elif not self.quiet:
+            # TODO Unify printing tools
+            print("No newer template version available.", file=sys.stderr)
+
     def _apply_update(self):
         subproject_top = Path(
             git(
@@ -940,6 +977,22 @@ def run_auto(
     if src_path is None:
         return run_update(dst_path, data, **kwargs)
     return run_copy(src_path, dst_path, data, **kwargs)
+
+
+def run_check(
+    dst_path: StrOrPath = ".",
+    data: Optional[AnyByStrDict] = None,
+    **kwargs,
+) -> Worker:
+    """Check if a subproject is using the latest versio of its template.
+
+    See [Worker][copier.main.Worker] fields to understand this function's args.
+    """
+    if data is not None:
+        kwargs["data"] = data
+    with Worker(dst_path=Path(dst_path), **kwargs) as worker:
+        worker.run_check()
+    return worker
 
 
 def _remove_old_files(prefix: Path, cmp: dircmp, rm_common: bool = False):
--- a/copier/main.py
+++ b/copier/main.py
@@ -982,7 +982,7 @@ def run_check(
     data: Optional[AnyByStrDict] = None,
     **kwargs,
 ) -> Worker:
-    """Check if a subproject is using the latest versio of its template.
+    """Check if a subproject is using the latest version of its template.
 
     See [Worker][copier.main.Worker] fields to understand this function's args.
     """
--- a/copier/main.py
+++ b/copier/main.py
@@ -768,7 +768,7 @@ class Worker:
         self._apply_update()
 
     def run_check(self) -> None:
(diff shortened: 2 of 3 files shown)


copier/cli.py
```diff
@@ -40,6 +40,15 @@
         copier update
         ```
 
+-   [`copier check`][copier.cli.CopierCheckSubApp] to check if a preexisting
+    project is using the latest version of its template
```
sisp, line 44:
```suggestion
    project is using the latest version of its template.
```

copier/cli.py
```diff
@@ -351,6 +361,59 @@ def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
...
+
+    DESCRIPTION = "Check if a copy is using the latest version of its original template"
+    DESCRIPTION_MORE = dedent(
+        """\
+        The copy must have a valid answers file which contains info
+        from the last Copier execution, including the source template
+        (it must be a key called `_src_path`).
+
+        If that file contains also `_commit` and `destination_path` is a git
+        repository, this command will do its best determine whether a newer
```
sisp, line 383:
```suggestion
        repository, this command will do its best to determine whether a newer
```

copier/main.py
```diff
@@ -687,6 +687,43 @@ def run_update(self) -> None:
             )
         self._apply_update()
 
+    def run_check(self) -> None:
+        """Check if a subproject is using the latest version of it's template
```
sisp, line 691:
```suggestion
        """Check if a subproject is using the latest version of its template.
```

copier/main.py
```diff
@@ -687,6 +687,43 @@ def run_update(self) -> None:
...
+        if not self.quiet:
+            # TODO Unify printing tools
+            print(
+                f"Currently using template version {self.subproject.template.version}, "
+                f"latest is {self.template.version}.",
+                file=sys.stderr)
+
+        if self.template.version > self.subproject.template.version:
+            if not self.quiet:
+                print("New template version available.", file=sys.stderr)
```
sisp, line 721: I think it would be nice if the new version was printed, and the comment about unifying printing tools probably should be added here as well:

```suggestion
                # TODO Unify printing tools
                print(
                    f"New template version available: {self.template.version}",
                    file=sys.stderr,
                )
```

copier/main.py
```diff
@@ -687,6 +687,43 @@ def run_update(self) -> None:
...
+            raise SubprojectOutdatedError("")
```
sisp, line 722: No need for the empty string.

```suggestion
            raise SubprojectOutdatedError()
```

copier/main.py
```diff
@@ -687,6 +687,43 @@ def run_update(self) -> None:
...
+        elif not self.quiet:
+            # TODO Unify printing tools
+            print("No newer template version available.", file=sys.stderr)
```
sisp, line 725: For consistency with the other message that a new version is available:

```suggestion
            print("No new template version available.", file=sys.stderr)
```

Or perhaps the other way around?

```suggestion
            print("Project is up-to-date 🎉", file=sys.stderr)
```

copier/main.py
```diff
@@ -687,6 +687,43 @@ def run_update(self) -> None:
...
+        if not self.template.version:
+            raise UserMessageError("Cannot check: version from template not detected.")
+
+        if not self.quiet:
+            # TODO Unify printing tools
+            print(
+                f"Currently using template version {self.subproject.template.version}, "
+                f"latest is {self.template.version}.",
+                file=sys.stderr)
+
```
sisp, line 718: Wouldn't this message be a bit strange when no new template version is available?

```text
Currently using template version X.Y.Z, latest is X.Y.Z.
```

How about removing it and extending the message when a new template version is available to also that new version (see my suggestion below)?

```suggestion
```

copier/main.py
```diff
@@ -865,6 +902,22 @@ def run_auto(
     return run_copy(src_path, dst_path, data, **kwargs)
 
 
+def run_check(
+    dst_path: StrOrPath = ".",
+    data: Optional[AnyByStrDict] = None,
+    **kwargs,
+) -> Worker:
+    """Check if a subproject is using the latest versio of its template.
```
sisp, line 910:
```suggestion
    """Check if a subproject is using the latest version of its template.
```

copier/main.py
```diff
@@ -865,6 +902,22 @@ def run_auto(
...
+
+    See [Worker][copier.main.Worker] fields to understand this function's args.
+    """
+    if data is not None:
+        kwargs["data"] = data
```
sisp, line 915: The `check` command doesn't use data anyway, so I think this block can be removed.

```suggestion
```

copier/cli.py
```diff
@@ -114,6 +123,7 @@ class CopierApp(cli.Application):
         """\
         copier [MAIN_SWITCHES] [copy] [SUB_SWITCHES] template_src destination_path
         copier [MAIN_SWITCHES] [update] [SUB_SWITCHES] [destination_path]
+        copier [MAIN_SWITCHES] check [SUB_SWITCHES] [destination_path]
```
sisp, line 126: For consistency with the other subcommands:
```suggestion
        copier [MAIN_SWITCHES] [check] [SUB_SWITCHES] [destination_path]
```

sisp (changes requested):
Thanks for your PR, @WernerWessely! :pray:

I have a couple of inline comments/suggestions. Disclaimer: I'm not a maintainer, just a motivated contributor, so @yajo has the authority over this project.

In addition, I noticed that most global flags / main switches aren't used or even useful for the `check` command. Most flags have simply no effect, but e.g. the `--vcs-ref` flag is used to instantiate the `Template` object

https://github.com/copier-org/copier/blob/91966047d9741099b4925612c5c0c3a0dbe2490c/copier/main.py#L592

so setting, e.g., a non-latest version via this flag will compare the project's template version with the passed template version. This may be valid behavior, although I don't quite see the use case TBH.

The flags that are useful IMO are:

* `-a, --answers-file`
* `-g, --prereleases`
* `-q, --quiet`

Omitting the other flags from the `check` command would require some refactoring though, it might even involve a breaking change (I'm not sure off the top of my head).

WDYT, @yajo?

copier/cli.py
WernerWessely: Aren't the []'s there to indicate that this is optional, that you can just call copier and it will guess what you want? This is not the case for the check command...?

WernerWessely:
@sisp 
Thanks for taking a first look, I will fix all that stuff. I have been thinking about another thing:

What if the user is not really interested in whether a new version is available, but whether this new version would actually result in any changes for the subproject. Do you think this is something the check command should be able to do or should the user script this on his own by doing an update and seeing what happens in the git repo?

WernerWessely:
Mmm on second thought, this would involve dealing with all the complexities of an update such as new questions automatically, so probably not a great idea.

copier/cli.py
sisp: I think you're right. But `copier ...` without a subcommand does not automatically choose between generating a new project and updating it:

https://github.com/copier-org/copier/blob/91966047d9741099b4925612c5c0c3a0dbe2490c/copier/cli.py#L292-L296

Should the `copy` and `update` subcommands be documented with surrounding brackets, too, @yajo?

WernerWessely:
> Thanks for your PR, @WernerWessely! pray
> 
> I have a couple of inline comments/suggestions. Disclaimer: I'm not a maintainer, just a motivated contributor, so @yajo has the authority over this project.
> 
> In addition, I noticed that most global flags / main switches aren't used or even useful for the `check` command. Most flags have simply no effect, but e.g. the `--vcs-ref` flag is used to instantiate the `Template` object
> 
> https://github.com/copier-org/copier/blob/91966047d9741099b4925612c5c0c3a0dbe2490c/copier/main.py#L592
> 
> so setting, e.g., a non-latest version via this flag will compare the project's template version with the passed template version. This may be valid behavior, although I don't quite see the use case TBH.
> 
> The flags that are useful IMO are:
> 
>     * `-a, --answers-file`
> 
>     * `-g, --prereleases`
> 
>     * `-q, --quiet`
> 
> 
> Omitting the other flags from the `check` command would require some refactoring though, it might even involve a breaking change (I'm not sure off the top of my head).
> 
> WDYT, @yajo?

Yes, I guess sharing options between subcommands can complicate things when you start to add more subcommands.

lhupfeldt:
Great feature, but I think the name `check` is too generic for a subcommand. Check what? Maybe a flag `--new-template-check` (or the less descriptive `--check`) for the `update` command?

sisp:
If you add a flag to the `update` subcommand, then the problem remains that most flags will not apply when checking for a new template version, but there I think it's more difficult to solve and document than for an independent subcommand. Also, I think you could make the same argument for `copy` and `update`. In the context of Copier, I think `check` is sufficiently clear along with some CLI docs. What else could `check` check other than the project's template version?

sisp:
> What else could `check` check other than the project's template version?

Okay, on second thought, Copier might offer subcommand that checks/validates a template, e.g. validates `copier.yml`.

Hmmm 🤔

sisp:
For inspiration, Poetry has two check features:
* [`poetry check`](https://python-poetry.org/docs/cli/#check) for checking/validating `pyproject.toml`; no additional flags or subcommands.
* [`poetry lock --check`](https://python-poetry.org/docs/cli/#lock) for checking that `poetry.lock` and `pyproject.toml` are consistent; `--no-update` has no meaning when using `--check`.

Poetry also has [global flags](https://python-poetry.org/docs/cli/#global-options), most of which seem unrelated to the two check features. So the situation is somewhat similar to Copier.

So perhaps `copier update --check` as @lhupfeldt suggested? For my taste, `--new-template-check` is a bit too verbose.

And just dreaming about the template check/validation feature: Perhaps `copier check|validate|lint`?

WernerWessely:
> So perhaps copier update --check as @lhupfeldt suggested? For my taste, --new-template-check is a bit too verbose.

Yeah sounds fine.. so the global switches that make sense with `copier update --check` would be:

* `-a, --answers-file`
* `-g, --prereleases`
* `-q, --quiet`

What about `-r, --vcs-ref?`. It could make sense to use it, say if you just want to check whatever the current commit of your subproject is against some reference of the template repo?

WernerWessely:
And the `-e, --exit-code` option would be added to the update command right? And it would be documented that it only has an effect together with `-c, --check`

WernerWessely:
> In any case, I'd wait for some feedback from @yajo before making further changes.

Yeah will do :+1: thanks a lot for taking a look so far!

WernerWessely:
So what do you say @yajo ? Should I implement the `copier update --check / --exit-code` options in this PR?

yajo:
Let's triage #1020 first.

yajo:
I prefer a command because it does a different thing than update, and has different flags. But `check` is too generic. What about `copier is-updated`? That command name will also make clear that an exit code `0b10` means "no, it isn't".

yajo:
> sharing options between subcommands can complicate things when you start to add more subcommands

I opened #1081 to improve this situation.

WernerWessely:
`is-updated` ? Mmm I don't know sounds kinda ugly to me... is it even grammatically correct 😄 ? I prefer short one word commands, if its generic then so be it... but thats just my opinion. I will be happy to have `is-updated` as well. Guess I will sleep on it. Maybe we can still think of something a bit more punchy?

WernerWessely:
Some other options I was able to think of:

- `copier check-version`
- `copier is-latest`
- `copier query`
- `copier check-latest`
- `copier latest`
- `copier fetch`
- `copier uptodate`

Guess none of them are really better than `copier is-updated`...

yajo:
🤷‍♂️

WernerWessely:
What about you @sisp, which version do you prefer?

yajo:
> I kinda like `copier uptodate` actually...

It seems to me too similar to `update`. Might be confusing...

sisp:
I've done some checking on how a few other tools with somewhat similar functionality call such a (sub)command:

* `yum` used to have a command [`yum check-update`](https://linuxhint.com/centos-update/).
* `ncu` is the executable name of [npm-check-updates](https://github.com/raineorshine/npm-check-updates). It offers a [CLI switch `-e, --errorLevel`](https://github.com/raineorshine/npm-check-updates#options) which controls the exit code for when updates are available.
* `npm` has a subcommand [`npm outdated`](https://docs.npmjs.com/cli/v9/commands/npm-outdated). That's the opposite semantics of the other suggestions so far.

I like `copier outdated` or `copier check-update`. I would be okay with `copier is-updated` as well.

WernerWessely:
I also prefer either `copier outdated` or `copier check-update` to `copier is-updated` just a bit more. But you have the final word @yajo, which one should it be 😅 ?

yajo:
Go for check-update, please. Outdated would be a double-negative. I Always try to avoid those. De: WernerEnviado: martes, 18 de abril de 2023 6:30Para: copier-org/copierCC: Jairo Llopis; MentionAsunto: Re: [copier-org/copier] feat: add the 'copier update --check' option to check if template is up-to-date (PR #1031) I also prefer either copier outdated or copier check-update to copier is-updated just a bit more. But you have the final word @yajo, which one should it be 😅 ?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

yajo:
`copier check-update` seems good.

copier/main.py
```diff
@@ -767,6 +767,41 @@ def run_update(self) -> None:
...
+                # TODO Unify printing tools
+                print(
+                    f"NEW template version available. "
+                    f"Currently using {self.subproject.template.version}, "
+                    f"latest is {self.template.version}.",
+                    file=sys.stderr)
+            raise SubprojectOutdatedError()
+        elif not self.quiet:
+            # TODO Unify printing tools
+            print("Project is up-to-date 🎉", file=sys.stderr)
```
yajo, line 803: I'd like it better if the exception contained a couple of attributes to extract the current and latest versions from it.

Then I'd like it better too if the printing is done only in the CLI part.

Finally, I don't think we need to print anything when all is OK.

copier/cli.py
```diff
@@ -358,6 +367,59 @@ def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
...
+    exit_code: cli.SwitchAttr = cli.SwitchAttr(
+        ["-e", "--exit-code"],
+        int,
+        default=0,
+        help=(
+            "Exit code for the command if a newer version is available, "
+            "for use in scripts."
+        ),
+    )
+
```
yajo, line 403: I don't think this should be customizable. Instead, it should be documented and Copier should be the one keeping control over which exit codes it uses.

copier/cli.py
```diff
@@ -358,6 +367,59 @@ def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
...
+                Only the destination path is needed to check, because the
+                `src_path` comes from [the answers file][the-copier-answersyml-file].
+
+                The subproject must exist. If not specified, the currently
+                working directory is used.
+        """
+        try:
+            self.parent._worker(dst_path=destination_path).run_check()
+        except SubprojectOutdatedError:
+            return self.exit_code
```
yajo, line 419: Continuing from the comment above, It'd be even better if you add a constant at the beginning of this file:

```py
EXIT_CODE_OUTDATED = 0b10
```

That's the same as `2` in base 10, but for exit codes it's more comfortable to use binary base.

Then hardcode it here. We can use this one:

```suggestion
            return EXIT_CODE_OUTDATED
```

Finally, add it to the docstring, so it's clear for the user which exit code means the project is outdated.

yajo (changes requested):
Thanks! Code looks good in general, although there's a lot still to do 😄!

BTW there are no tests... we need them 💪🏼

yajo:
> I opened #1081 to improve this situation.

FWIW that's fixed already and it would help on the CLI UX for this PR.

pawamoy:
I'm willing to continue the work on this PR (starting next week probably) if @WernerWessely doesn't have the capacity anymore.

yajo:
Closing because this became stale.

Kaspiman:
@pawamoy @WernerWessely Hello! Do you have any plans to develop this feature?

pawamoy:
As you can see, no. Would you like to work on it? 🙂

Kaspiman:
Unfortunately, I don't program in Python :(

PR #1048, opened 2023-03-22
Fix local version guess using dunamai / PEP440 compliant.

cclecle:
fix: make dunamai.Version.from_git() PEP440 compliant (allow version tag without a leading 'v')


yajo (changes requested):
Thanks! This needs a test. Could you add one please?

cclecle:
> Thanks! This needs a test. Could you add one please?

Yes I will + fix wrong module call :).

cclecle:
pytest added.
That's still not perfect because if there is tag a tag like x.x.x.postY + dirt, it found x.x.x.postY.dev1 and not x.x.x.postY+1.
But that is still better (no more exception / template deployed)

tests/test_copy.py
```diff
@@ -68,6 +68,130 @@ def test_copy_with_non_version_tags(tmp_path_factory: pytest.TempPathFactory) ->
...
+
+    with local.cwd(src):
+        Path("v1.txt").write_text("Test content3")
+        git("add", ".")
+        git("commit", "-m1")
+
+    copier.run_auto(
+        str(src),
+        dst,
+        vcs_ref="HEAD",
```
yajo, line 107: If you pass this, you skip proper tag auto-detection! 

Instead, don't pass this arg and assert it picks the correct commit when copying.

tests/test_copy.py
```diff
@@ -68,6 +68,130 @@ def test_copy_with_non_version_tags(tmp_path_factory: pytest.TempPathFactory) ->
...
+        use_prereleases=True,
+        defaults=True,
+    )
+
+
+def test_copy_with_local_version_tags_PEP440_dev(
```
yajo, line 113: Please use pytest.mark.parametrize to have all 3 tests in one

yajo (changes requested):
Good one. It just needs a bit more polish please 😊

tests/test_copy.py
cclecle: That was the intented test behavior :-/.

If I remove vcs_ref="HEAD", copier will take the previous fully versionned tag.

sample:
commit n1 -> tag 1.1.1           
  -> copier from here ok, both with ou without HEAD
commit n2 -> tag 1.1.1.dev1 
  -> copier from here ok, without HEAD it will take the previous one, with HEAD this one.
commit n3 -> !! no tag           
 -> copier from here ok, but only if no HEAD

This pull request try to allow the last one to work too.
With this patch, the last one is recognized as tag 1.1.1.dev2+XXXXX, which is okay.
WIthout, this patch, copier raise an exception.

Only drawback is for "post" tags that would become: 1.1.1.pos1.dev1+XXXXX, would be better 1.1.1.post2+XXXXX, but it is still PEP440 compliant :)  So its another feature that can be for another later update... !

Is it ok for you if I keep this parameter, but add checking on copier stdout ?
I actually wanted to add that at previous push, but I do not know if relying on sdtdout was good ... as it doesnt aim to be a standardized interface I guess ?

copier/template.py
```diff
@@ -480,8 +480,12 @@ def version(self) -> Optional[Version]:
         try:
             with local.cwd(self.local_abspath):
                 # Leverage dunamai by default; usually it gets best results
+                # dunamai.Version.from_git needs Pattern.DefaultUnprefixed to be PEP440 compliant on version reading
                 return Version(
-                    dunamai.Version.from_git().serialize(style=dunamai.Style.Pep440)
+                    # dunamai.Version.from_git(pattern=dunamai.Pattern.DefaultUnprefixed).serialize(style=dunamai.Style.Pep440)
```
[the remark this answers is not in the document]
cclecle: of course, it is an ugly omission !!

tests/test_copy.py
yajo: AFAICS the main point of this test is the fact that the tag doesn't have a `v` prefix. Sorry, it wasn't very obvious. I ran the tests without the solution and indeed they fail. Thanks!

yajo (changes requested):
My doubts are cleared. This fix is indeed needed. Please could you attend the remaining comments before merge?

yajo (approved):
I fixed the missing parts myself. Thanks!

Issue #750, opened 2022-08-25
Add support for multiple prompts of the same type


sisp:
**Is your feature request related to a problem? Please describe.**
I'd like to prompt for zero or more additional dependencies (package names + optional version specifiers, e.g. `cowsay` or `cowsay@>=5.0`) to be included in the dependency specification of a generated Python project. Currently, there is no way of repeating a prompt and appending the answers to a list.

**Describe the solution you'd like**
I imagine a solution could look like this:
```yaml
dependencies:
  type: str
  help: Additional dependencies
  multiple:
    help: Add another dependency?
    min: 0
```

The `multiple` field could have the following variants:
```yaml
multiple: true            # uses defaults for all fields

multiple:
  help: <str>             # optional, defaults to "More?"
  min: <nonnegative int>  # optional, defaults to 0
  max: <positive int>     # optional, defaults to ∞
```

A `type: bool` question specified by `multiple.help` would be asked before accepting the input for any _optional_ item and deleted again from the terminal when answered. For instance, when `min: 2` then the confirm question would be asked only after 2 inputs, and when `max: 4` then the questionnaire would proceed to the next question after the fourth answer had been provided. The answers would be gathered in a list. The `validation` expression, when present, would be executed after every input with the current list of answers. For validating each item independently, only the last item of the list would be accessed in the `validation` expression. But complex validation logic that depends on previously entered answers or the number of answers so far would be possible, too.

Regarding UX, perhaps the nesting could be visualized using Unicode art similar to how the `tree` tool prints filesystem trees:

```
🎤 Additional dependencies
├─ cowsay@>=5.0
└─ requests
```

The implementation could take some inspiration from https://github.com/RhetTbull/questionary-superprompt#multiple-option.

**Describe alternatives you've considered**
Alternatively, a `type: yaml` or `type: json` prompt could be used, but the flow would not be guided, so the UX would be worse. Also, validation logic would be more complex.

**Additional context**
I'd be happy to work on a PR if this proposal gets accepted.

sisp:
The field `multiple` could also be named `many` instead, inspired by [`marshmallow`'s attribute for declaring collections](https://marshmallow.readthedocs.io/en/stable/quickstart.html#handling-collections-of-objects).

yajo:
Thanks for your contributions! It seems to me that #751 is a more complete proposal that would render this one unnecessary, so let me close here and continue there.

yajo:
[...]

PR #1049, opened 2023-03-22
Add multiple choice prompt (map list type to questionary checkbox)

### LINKED ISSUE #218: Add multiple choice prompt
#93/#107 implemented support for single choice questions. I'd love to see an option to use multiple choice prompts. Thinking about something like this:

```
ci_type:
  type: str
  help: Which CI configuration do you need?
  multiplechoice:
    - travis
    - circleci
    - gitlabci
    - jenkins
```

```
Which CI configuration do you need?
ci_type? Format: str
🎤
( ) travis
( ) circleci
( ) gitlabci
( ) jenkins
```

tlambert03:
Closes #218 by supporting `questionary.checkbox` when the question `type` is `list`.

```yaml
toppings:
    type: list
    choices:
        - Cheese
        - Tomatoes
        - Pepperoni
        - Sausage
```

This is a tiny PR at the moment... and it works fine for many cases.  But feel free to point me to edge cases that you want caught and tested.  (No need to guide my on the fixes yet, just state any concerns and I'll try to figure them out)

A potential elaboration would be to support `list[<type>]` with a more complicated `cast_fn`

For now, I am raising a validation error if `type == 'list'` but no `choices` are provided, which makes the validation logic much easier (since you'll always get a checkbox prompt from questionary).  In the future, or here if you prefer, we can add logic to cast more general strings to lists

### DIFF
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -491,6 +491,7 @@ def test_var_name_value_allowed(
     answers = yaml.safe_load((dst / ".copier-answers.yml").read_text())
     assert answers["value"] == "string"
 
+
 @pytest.mark.parametrize(
     "choices",
     (
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -532,11 +532,14 @@ def test_list_checkbox(
     # Copy
     tui = spawn(COPIER_PATH + (str(src), str(dst)), timeout=10)
     expect_prompt(tui, "pick_one", "list")
-    tui.send('i')  # invert selection
+    tui.send(Keyboard.Down)
+    tui.send(" ")  # select 2
+    tui.send(Keyboard.Down)
+    tui.send(" ")  # deselect 3
     tui.send(Keyboard.Enter)
     tui.expect_exact(pexpect.EOF)
     answers = yaml.safe_load((dst / ".copier-answers.yml").read_text())
-    assert answers["pick_one"] == [2]
+    assert answers["pick_one"] == [1, 2]
 
     with local.cwd(dst):
         git("init")
@@ -545,8 +548,8 @@ def test_list_checkbox(
     # Update
     tui = spawn(COPIER_PATH + (str(src), str(dst)), timeout=10)
     expect_prompt(tui, "pick_one", "list")
-    tui.send('i')  # invert selection
+    tui.send("i")  # invert selection
     tui.send(Keyboard.Enter)
     tui.expect_exact(pexpect.EOF)
     answers = yaml.safe_load((dst / ".copier-answers.yml").read_text())
-    assert answers["pick_one"] == [1, 3]
+    assert answers["pick_one"] == [3]
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -490,3 +490,63 @@ def test_var_name_value_allowed(
     # Check answers file
     answers = yaml.safe_load((dst / ".copier-answers.yml").read_text())
     assert answers["value"] == "string"
+
+@pytest.mark.parametrize(
+    "choices",
+    (
+        [1, 2, 3],
+        [["one", 1], ["two", 2], ["three", 3]],
+        {"one": 1, "two": 2, "three": 3},
+    ),
+)
+def test_list_checkbox(
+    tmp_path_factory: pytest.TempPathFactory,
+    spawn: Spawn,
+    choices: Union[
+        List[int],
+        List[List[Union[str, int]]],  # actually `List[Tuple[str, int]]`
+        Dict[str, int],
+    ],
+):
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    # Create template
+    build_file_tree(
+        {
+            (src / "copier.yml"): (
+                f"""\
+                _templates_suffix: {SUFFIX_TMPL}
+                _envops: {BRACKET_ENVOPS_JSON}
+                pick_one:
+                    type: list
+                    default: [1, 3]
+                    choices: {choices}
+                """
+            ),
+            (src / "[[ _copier_conf.answers_file ]].tmpl"): (
+                "[[ _copier_answers|to_nice_yaml ]]"
+            ),
+        }
+    )
+
+    # Copy
+    tui = spawn(COPIER_PATH + (str(src), str(dst)), timeout=10)
+    expect_prompt(tui, "pick_one", "list")
+    tui.send('i')  # invert selection
+    tui.send(Keyboard.Enter)
+    tui.expect_exact(pexpect.EOF)
+    answers = yaml.safe_load((dst / ".copier-answers.yml").read_text())
+    assert answers["pick_one"] == [2]
+
+    with local.cwd(dst):
+        git("init")
+        git("add", ".")
+        git("commit", "-m1")
+    # Update
+    tui = spawn(COPIER_PATH + (str(src), str(dst)), timeout=10)
+    expect_prompt(tui, "pick_one", "list")
+    tui.send('i')  # invert selection
+    tui.send(Keyboard.Enter)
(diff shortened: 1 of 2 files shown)


pawamoy:
Does it work if we define choices as a YAML map instead of a list?

```yaml
choices:
  key: value
  etc: etc
```

tlambert03:
> Does it work if we define choices as a YAML map instead of a list?

yes, test added.  Also work with `default`

tests/test_prompt.py
```diff
@@ -490,3 +490,67 @@ def test_var_name_value_allowed(
...
+        [["one", 1], ["two", 2], ["three", 3]],
+        {"one": 1, "two": 2, "three": 3},
+    ),
+)
+def test_list_checkbox(
+    tmp_path_factory: pytest.TempPathFactory,
+    spawn: Spawn,
+    choices: Union[
+        List[int],
+        List[List[Union[str, int]]],  # actually `List[Tuple[str, int]]`
```
yajo, line 508: What does that comment mean?

tests/test_prompt.py
```diff
@@ -490,3 +490,67 @@ def test_var_name_value_allowed(
...
+
+    # Create template
+    build_file_tree(
+        {
+            (src / "copier.yml"): (
+                f"""\
+                _templates_suffix: {SUFFIX_TMPL}
+                _envops: {BRACKET_ENVOPS_JSON}
+                pick_one:
+                    type: list
```
yajo, line 522: I'm not sure I like this interface.

Currently, when combining `type` and `choices`, type expresses the type of all choices. For example `type: str` tells Copier that no matter what choice is selected, it will be a string.

So I'd let that behavior as it is. For this example, we should have `type: int`.

Instead, I think it's better to add a new attribute `multiselect: true` which makes sense only when using `choices`. This way, the interface of this new option will be closer to current Copier behavior elsewhere.

WDYT?

tests/test_prompt.py
tlambert03: that was just copied from the fixture above, added in https://github.com/copier-org/copier/pull/985 ... 
I presume the intention was to state that it's always going to a list of length two, with a string and an int... but you'd have to ask @sisp why it was put that way

tests/test_prompt.py
tlambert03: > Currently, when combining type and choices, type expresses the type of all choices. For example type: str tells Copier that no matter what choice is selected, it will be a string.  So I'd let that behavior as it is. For this example, we should have type: int.

I'm not sure I agree here.  Currently, with choices, you *always* get a single item back.  So it makes sense to use the type (or union of types) that you will get back.  Here, however, questionary *will* return you a `list`.  So saying `type: list` isn't a hack, it's actually describing exactly what the type of the value will be.  If you wanted to specify the type of each item, then I'd suggest using something like `list[<type>] with a more complicated cast_fn` as I mentioned in my original post. 

I'm not opposed to the more explicit `multi-select: true` flag in addition though.  But I do think you'd be setting things up for a bit of confusion if you leave `type` as the contained type, and users end up getting lists back

tests/test_prompt.py
sisp: A tuple-style choice doesn't actually use a tuple but a list with two elements where the first element is a string and the type of the second element must match the question's type. So semantically the type hint is the type hint for a choice is `Tuple[str, int]`, but since it's a list and not a tuple the actual type is `List[Union[str, int]]` because there is no way to express a tuple that uses a list container.

tests/test_prompt.py
yajo: Hmm makes sense. Ok, let me play a bit with it 😁

sisp: I think this design decision is related to #750 and should take that feature proposal into consideration as well.

sisp: With #750 in mind, I disagree with

> [...] it makes sense to use the type (or union of types) that you will get back.

because following that logic, a repeated question would need `type: list[<type>]` (in fact `type: list[list[<type>]]`) as well which would be ambiguous with multiple choice prompts.

Copier's questions syntax isn't a generic schema language like JSON Schema. It's rather a questions configuration language. I think it's important to not confuse the two. So `type` isn't the type of the variable that contains the answer to that question but rather one config setting of a question among several whose meaning depends on the kind of question defined in conjunction with the other settings.

tlambert03: > Copier's questions syntax isn't a generic schema language like JSON Schema.

correct, and it looks like #750 would actually require something more like that no?  Here we're just allowing that the return value be a list of values, selected from some set of possibilities.  from my brief look at it, i think that #750, while definitely powerful and interesting, would be the one that would require the more powerful configuration language, no?

tlambert03: in any case, it's more important that the maintainers here (you) like the syntax and typing plan :joy: so after you all give it some more thought, if you'd like to propose a clear alternative to the current PR, I'm happy to implement it, or feel free to close and reimplement in a follow up

sisp: I think #750 would work perfectly fine as suggested there without a more schema-like language.

As I said, I think Copier's question configuration is not a schema language for describing an answer value or data structure. For instance, configuring

```
q:
  type: list[str]
  multiple: true
```

is redundant because `multiple: true` implies a list of answers. The same goes for multiple choice questions.

But ultimately @yajo needs to decide as he is the maintainer. I'm just trying to share my thoughts. 😉

tlambert03: i agree, it's redundant, and as mentioned, i do also like the explicit `multiple:` or @yajo's `multi-select`.  The only thing that I think would be confusing is the use of `multiple` *and* the use of a non-container `type`:

```yaml
q:
  type: str
  multiple: true
```

yajo: We currently have `yaml` and `json` types which result in getting any kind of object/list/str/float/int.. anything supported by those formats.

I think it's confusing that you have `type: list` because Copier users have used lists since long ago by using those other 2 types.

Let's just make it clear that we're letting the user do a multi-select of choices, while still letting the `type` reflect the value contained within each of those choices. Just like it happens with single-select choices. Considering all options, let's use this implementation: https://github.com/copier-org/copier/pull/1049#discussion_r1147742760

FWIW the default type is `yaml` IIRC. So we could have a `{type: yaml, multiselect: true}` question and that'd be just fine and readable. The docs should clarify usage anyways.

brpaz: @yajo I was experimenting a bit with this, before seeing this PR.

I was implementing like you said, to have a `multiple` option and keep the original prompt type - https://github.com/copier-org/copier/compare/master...brpaz:copier:multiselect?expand=1

The only issue I faced with this approach, is when rendering the jinja template, the value is passed as a string and not as a list, making any kind of operation like looping over all the choices difficult.

I managed to get a workaround, by setting the type as `json`.  And maybe `yaml` also works as you said. I think it´s not very intuitive but I can live with that. :)

So, @tlambert03, any chances that you will do these changes on this PR? Otherwise, I can open a PR from my branch, if @yajo thinks it´s in the right direction

tlambert03: Go for it :)

yajo: The truth is that the type is a bit redundant when using select (or, now, multiselect) questions.

The only purpose of a type is to convert any user input to the expected type. But when you're only selecting options, it's the template designer who wrote the possible answers. Thus, it's fair to assume that it's their responsibility to be consistent on the chosen types.

Of course, this didn't suppose a problem when it was a single answer. But, if now, it's a problem, maybe a very simple approach would be to raise an error if you use `multiple` with a type that is not `yaml` or `json`. As usual, just add docs and go for the simplest solution :)

tlambert03: Thanks for the input all!  Gonna close this and let @brpaz follow up

PR #1059, opened 2023-03-27
feat: add support for yaml data files

Ian2012:
This PR adds support for yaml files as a data input

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -189,16 +189,8 @@ class CopierApp(cli.Application):
                 Each value in the list is of the following form: `NAME=VALUE`.
         """
         for arg in values:
-            if "=" in arg:
-                key, value = arg.split("=", 1)
-                self.data[key] = value
-            else:
-                if not Path(arg).is_file():
-                    raise UserMessageError(f"Data file {arg} not found")
-                if not Path(arg).suffix in {".yaml", ".yml"}:
-                    raise UserMessageError(f"Data file {arg} must be a YAML file")
-                with open(arg) as f:
-                    self.data.update(yaml.safe_load(f))
+            key, value = arg.split("=", 1)
+            self.data[key] = value
 
     def _worker(self, src_path: OptStr = None, dst_path: str = ".", **kwargs) -> Worker:
         """
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -180,6 +180,7 @@ class CopierApp(cli.Application):
         "VARIABLE=VALUE",
         list=True,
         help="Make VARIABLE available as VALUE when rendering the template",
+        excludes=["--data-file"],
     )
     def data_switch(self, values: StrSeq) -> None:
         """Update [data][] with provided values.
@@ -196,6 +197,7 @@ class CopierApp(cli.Application):
         ["--data-file"],
         str,
         help="Load data from a YAML file",
+        excludes=["--data"],
     )
     def data_file_switch(self, path: str) -> None:
         """Update [data][] with provided values.
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -192,6 +192,24 @@ class CopierApp(cli.Application):
             key, value = arg.split("=", 1)
             self.data[key] = value
 
+    @cli.switch(
+        ["--data-file"],
+        str,
+        help="Load data from a YAML file",
+    )
+    def data_file_switch(self, path: str) -> None:
+        """Update [data][] with provided values.
+
+        Arguments:
+            path: The path to the YAML file to load.
+        """
+        if not Path(path).exists():
+            raise UserMessageError(f"Data file {path} not found")
+        if not Path(path).suffix not in [".yaml", ".yml"]:
+            raise UserMessageError(f"Data file {path} must be a YAML file")
+        with open(path) as f:
+            self.data.update(yaml.safe_load(f))
+
     def _worker(self, src_path: OptStr = None, dst_path: str = ".", **kwargs) -> Worker:
         """
         Run Copier's internal API using CLI switches.
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -195,7 +195,7 @@ class CopierApp(cli.Application):
             else:
                 if not Path(arg).exists():
                     raise UserMessageError(f"Data file {arg} not found")
-                if not Path(arg).suffix not in [".yaml", ".yml"]:
+                if not Path(arg).suffix in {".yaml", ".yml"}:
                     raise UserMessageError(f"Data file {arg} must be a YAML file")
                 with open(arg) as f:
                     self.data.update(yaml.safe_load(f))
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -43,6 +43,7 @@ Basically, there are 3 different commands you can run:
 Below are the docs of each one of those.
 """
 
+import yaml
 import sys
 from functools import wraps
 from io import StringIO
@@ -188,8 +189,16 @@ class CopierApp(cli.Application):
                 Each value in the list is of the following form: `NAME=VALUE`.
         """
         for arg in values:
-            key, value = arg.split("=", 1)
-            self.data[key] = value
+            if "=" in arg:
+                key, value = arg.split("=", 1)
+                self.data[key] = value
+            else:
+                if not Path(arg).exists():
+                    raise UserMessageError(f"Data file {arg} not found")
+                if not Path(arg).suffix not in [".yaml", ".yml"]:
+                    raise UserMessageError(f"Data file {arg} must be a YAML file")
+                with open(arg) as f:
+                    self.data.update(yaml.safe_load(f))
 
     def _worker(self, src_path: OptStr = None, dst_path: str = ".", **kwargs) -> Worker:
         """
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -195,18 +195,16 @@ class CopierApp(cli.Application):
 
     @cli.switch(
         ["--data-file"],
-        str,
+        cli.ExistingPath,
         help="Load data from a YAML file",
         excludes=["--data"],
     )
-    def data_file_switch(self, path: str) -> None:
+    def data_file_switch(self, path: plumbum.LocalPath) -> None:
         """Update [data][] with provided values.
 
         Arguments:
             path: The path to the YAML file to load.
         """
-        if not Path(path).exists():
-            raise UserMessageError(f"Data file {path} not found")
         if not Path(path).suffix not in [".yaml", ".yml"]:
             raise UserMessageError(f"Data file {path} must be a YAML file")
         with open(path) as f:
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -205,8 +205,6 @@ class CopierApp(cli.Application):
         Arguments:
             path: The path to the YAML file to load.
         """
-        if not Path(path).suffix not in [".yaml", ".yml"]:
-            raise UserMessageError(f"Data file {path} must be a YAML file")
         with open(path) as f:
             self.data.update(yaml.safe_load(f))
 
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -43,7 +43,6 @@ Basically, there are 3 different commands you can run:
 Below are the docs of each one of those.
 """
 
-import yaml
 import sys
 from functools import wraps
 from io import StringIO
@@ -51,6 +50,7 @@ from pathlib import Path
 from textwrap import dedent
 from unittest.mock import patch
 
+import yaml
 from plumbum import cli, colors
 
 from .errors import UserMessageError
@@ -193,7 +193,7 @@ class CopierApp(cli.Application):
                 key, value = arg.split("=", 1)
                 self.data[key] = value
             else:
-                if not Path(arg).exists():
+                if not Path(arg).is_file():
                     raise UserMessageError(f"Data file {arg} not found")
                 if not Path(arg).suffix in {".yaml", ".yml"}:
                     raise UserMessageError(f"Data file {arg} must be a YAML file")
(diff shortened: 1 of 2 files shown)


copier/cli.py
sisp, line 1: Currently, the switch uses `list=True`, so the switch can be passed repeatedly, which makes sense for passing key-value pairs. But typically only one data file would be passed. `plumbum.cli` doesn't support overloading a switch with multiple specs though, so I see two options:

1. The `list=False` behavior when passing a data file needs to be implemented manually, i.e. repeatedly providing that switch with a data file raises an error.
2. A second switch, e.g. `--data-file`, could be added which is [mutually exclusive](https://plumbum.readthedocs.io/en/latest/cli.html#mutual-exclusion) with `-d, --data`.

That said, it might actually make sense to allow passing multiple files whose contents would be merged. And to take this thought even further, it might even make sense to allow mixing key-value pairs with data files and all data would be merged. This way, `-d, --data` could be reused and, in fact, the current implementation would already handle this scenario. WDYT, @yajo?

When reusing the `-d, --data` switch for providing a data file, the `argname` (`VARIABLE=VALUE`), `help` string, and method docstring need to be updated to reflect this behavior.

copier/cli.py
```diff
@@ -189,8 +190,16 @@ def data_switch(self, values: StrSeq) -> None:
                 Each value in the list is of the following form: `NAME=VALUE`.
         """
         for arg in values:
-            key, value = arg.split("=", 1)
-            self.data[key] = value
+            if "=" in arg:
+                key, value = arg.split("=", 1)
+                self.data[key] = value
+            else:
+                if not Path(arg).exists():
```
sisp, line 197: Should we use [`Path(arg).isfile()`](https://docs.python.org/3/library/os.path.html#os.path.isfile) instead to be more precise?

docs/configuring.md
```diff
@@ -573,6 +573,29 @@ questions with default answers.
     copier -fd 'user_name=Manuel Calavera' copy template destination
     ```
 
+As alternative you can also pass the path to yaml file that contains your data. This
+cannot be used in the API.
```
sisp, line 577: A few minor linguistic corrections.

And how about a little rewording in the second sentence?

```suggestion
As an alternative you can also pass the path to a YAML file that contains your data.
Passing a data file is only supported through the CLI.
```

docs/configuring.md
```diff
@@ -573,6 +573,29 @@ questions with default answers.
...
+
+!!! example
+
+    Example CLI usage to take a yaml file as data:
```
sisp, line 581:
```suggestion
    Example CLI usage with a YAML file containing data:
```

docs/configuring.md
```diff
@@ -573,6 +573,29 @@ questions with default answers.
...
+
+    ```yml
```
sisp, line 583: For consistency with the rest of the docs.

```suggestion
    ```yaml
```

docs/configuring.md
```diff
@@ -573,6 +573,29 @@ questions with default answers.
...
+    user_name: Alice Bob
```
sisp, line 584: For consistency with the key-value pair example above:

```suggestion
    user_name: Manuel Calavera
```

docs/configuring.md
```diff
@@ -573,6 +573,29 @@ questions with default answers.
...
+    height: 1.83
+    ```
+
+    Calling this:
+    ```sh
+    copier -d input.yml copy template destination
+    ```
+
+    does the same thing as:
+    ```sh
```
sisp, line 595: For consistency with the rest of the docs.

```suggestion
    ```shell
```

docs/configuring.md
```diff
@@ -573,6 +573,29 @@ questions with default answers.
...
+    copier -d "user_name=Alice Bob" -d "age=7" -d "height=1.83" copy template destination
```
sisp, line 596: Same as above:

```suggestion
    copier -d 'user_name=Manuel Calavera' -d 'age=7' -d 'height=1.83' copy template destination
```

docs/configuring.md
```diff
@@ -573,6 +573,29 @@ questions with default answers.
...
+
+    Example CLI usage to take a yaml file as data:
+
+    ```yml
+    user_name: Alice Bob
+    age: 7
+    height: 1.83
+    ```
+
+    Calling this:
```
sisp, line 589: Maybe a bit more explicit:

```suggestion
    Passing a data file

```

Also note the empty line. Prettier doesn't format Markdown markup within the admonition block because it's non-standard syntax and Prettier interprets it as an indented code block.

sisp:
Follow-up of #485. Resolves #540.

sisp (changes requested):
Thanks for your contribution, @Ian2012!

I have a couple of minor suggestions and raised a topic for discussion around repeatedly passing the `-d, --data` switch with data files.

Please also add tests for both the success and failure scenarios once the conceptual discussion has been resolved.

As always, @yajo needs to perform the ultimate review, I'm only trying to support.

copier/cli.py
Ian2012: At our team, we designed a copier wrapper that covers this specific scenario where we need to handle those cases, where the `-d, --data` can be both (data file or key value pair) and in fact, the key value pairs arguments are prioritized.

Ian2012:
@sisp Thanks for your review. I've already handle all your comments, if you agree, could you resolve all the conversations so we can focus on the conceptual discussion?

copier/cli.py
Ian2012: I agree

sisp: LGTM

sisp:
Thanks, @Ian2012. There are still a few open suggestions of mine in the docs though. And I can't seem to mark the other comments as resolved. Can you resolve them?

copier/cli.py
sisp: That's great. It means you have an actual use case for mixing a data file and key-value pairs.

I think there are two valid behavior options for mixing them:

1. Data files and key-value pairs are merged/overridden in the order of their occurrence in the CLI. The current implementation works like this (with shallow merging).
2. Key-value pairs always override data from data files independent of the switch order. The current implementation does not work like this.

To me, 1. is a bit more intuitive, but 2. also follows a sensible logic. Would 1. also work for your use case?

Another aspect to discuss is merge behavior. Should answers to `json`/`yaml` questions be merged via shallow or deep merging?

Ian2012: I think case 1 is a more general case, so we can adopt that change for our use case. It's a good solution for us.

Also, I just remember that we are using copier programmatically, should I change something on another part of the code?

Regarding the merging part, I think shallow merge is the way to go, but any of those cases will support our use case and we are open to any modification

Ian2012: @yajo what do you think?

Ian2012:
Is there anything I can do to speed up the review process?

copier/cli.py
yajo: IMHO I prefer solution 2 from https://github.com/copier-org/copier/pull/1059#discussion_r1150702483. It is the MVP for the feature, and it seems much simpler and straightforward. No merging, no multiple args.

Ian2012: @yajo Done, there is a new switch `--data-file` that is mutually exclusive with `--data`

docs/configuring.md
```diff
@@ -703,6 +703,39 @@ questions with default answers.
...
+    ```yaml
+    user_name: Manuel Calavera
+    age: 7
+    height: 1.83
+    ```
+
+    Passing a data file
+
+    ```shell
+    copier -d input.yml copy template destination
```
sisp, line 731: The flag needs to be updated:

```suggestion
    copier --data-file input.yml copy template destination
```

docs/configuring.md
```diff
@@ -703,6 +703,39 @@ questions with default answers.
...
+Passing a data file is only supported through the CLI.
+
+This cannot be defined in `copier.yml`, where its equivalent would be just normal
+questions with default answers.
+
+!!! example
+
+    Example CLI usage with a YAML file containing data:
+
+    ```yaml
```
sisp, line 722: How about adding the file name `input.yml` as the title, so there is a better connect with the CLI call below?

```suggestion
    ```yaml
```

docs/configuring.md
Ian2012: Addressed

docs/configuring.md
Ian2012: Addressed

sisp:
@Ian2012 I think it would be good to mention in the docs for both `--data` and `--data-file` (with anchor links to each other) that they are mutually exclusive.

Ian2012:
I totally agree, here is an updated version https://copier--1059.org.readthedocs.build/en/1059/configuring/#data

docs/configuring.md
```diff
@@ -703,6 +703,47 @@ questions with default answers.
     copier -fd 'user_name=Manuel Calavera' copy template destination
     ```
 
+!!! info
+
+    The `--data` option is mutually exclusive with [`--data-file`](#data_file) option.
```
sisp, line 708: Juat a minor suggestion to retain the wording of the docs:

```suggestion
    The `-d, --data` flags are mutually exclusive with the [`--data-file`](#data_file) flag.
```

WDYT, @Ian2012?

docs/configuring.md
```diff
@@ -703,6 +703,47 @@ questions with default answers.
...
+    ```
+
+    is equivalent to passing its content as key-value pairs:
+
+    ```shell
+    copier -d 'user_name=Manuel Calavera' -d 'age=7' -d 'height=1.83' copy template destination
+
+!!! info
+
+    The `--data-file` option is mutually exclusive with [`--data`](#data) option.
```
sisp, line 745: Juat a minor suggestion to retain the wording of the docs:

```suggestion
    The `--data-file` flag is mutually exclusive with the [`-d, --data`](#data) flags.
```

WDYT, @Ian2012?

copier/cli.py
```diff
@@ -191,6 +193,25 @@ def data_switch(self, values: StrSeq) -> None:
...
+        excludes=["--data"],
+    )
+    def data_file_switch(self, path: str) -> None:
+        """Update [data][] with provided values.
+
+        Arguments:
+            path: The path to the YAML file to load.
+        """
+        if not Path(path).exists():
+            raise UserMessageError(f"Data file {path} not found")
```
yajo, line 209: You can use some plumbum helpers:

```suggestion
    @cli.switch(
        ["--data-file"],
        cli.ExistingPath,
        help="Load data from a YAML file",
        excludes=["--data"],
    )
    def data_file_switch(self, path: plumbum.LocalPath) -> None:
        """Update [data][] with provided values.

        Arguments:
            path: The path to the YAML file to load.
        """
```

docs/configuring.md
```diff
@@ -703,6 +703,47 @@ questions with default answers.
...
+
+    The `--data` option is mutually exclusive with [`--data-file`](#data_file) option.
+
+### `data_file`
+
+-   Format: `str`
+-   CLI flags: `--data-file`
+-   Default value: N/A
+
+As an alternative you can also pass the path to a YAML file that contains your data.
```
yajo, line 716: Just in case some reader lands here directly without reading the previous chapter before:

```suggestion
As an alternative to [`--data`][data] you can also pass the path to a YAML file that contains your data.
```

docs/configuring.md
```diff
@@ -703,6 +703,47 @@ questions with default answers.
...
+Passing a data file is only supported through the CLI.
+
+This cannot be defined in `copier.yml`, where its equivalent would be just normal
+questions with default answers.
```
yajo, line 720: Just following the standard from other similar options:

```suggestion
!!! info
    Not supported in `copier.yml` or API calls. Only supported through the CLI.
```

yajo (changes requested):
Some rewording on docs and simplification on code.

Thanks for the feature! Really helpful and a small patch. I like it a lot 😊 

This needs a test though. Could you add it too please?

docs/configuring.md
```diff
@@ -703,6 +703,45 @@ questions with default answers.
...
+
+    The `-d, --data` flags are mutually exclusive with the [`--data-file`](#data_file) flag.
+
+### `data_file`
+
+-   Format: `str`
+-   CLI flags: `--data-file`
+-   Default value: N/A
+
+As an alternative to [`--data`][data] you can also pass the path to a YAML file that contains your data.
```
sisp, line 716: A newline is missing. Also, how a out adding the short flag here as well? And shouldn't the link be an anchor link?

```suggestion
As an alternative to [`-d, --data`](#data) you can also pass the path to a YAML file that contains your data.

```

docs/configuring.md
```diff
@@ -703,6 +703,45 @@ questions with default answers.
...
+!!! info
```
sisp, line 717: A newline is missing:

```suggestion
!!! info

```

sisp (changes requested):

Ian2012:
I would like to add tests but I'm unable to setup a development environment using local env, do you guys have the manual instructions?

When I run `direnv allow` it does nothing, I don't know how to install `poe` and the developments dependencies

yajo:
There's a small guide in https://github.com/copier-org/copier/blob/master/CONTRIBUTING.md#dev-environment-setup

The fastest way is just to use gitpod.

Ian2012:
@yajo I'm really struggling with the test, could you give me a hand there?

I'm not used to test

sisp:
@Ian2012 If you're still struggling with setting up the development environment, it's typically sufficient to run

```shell
$ poetry install --with dev,docs
$ poetry run pip install black isort  # optionally but useful to get the formatters
$ poetry run poe test
```

if you have Git and a sufficiently recent Python interpreter installed on your system. Nix provides a complete and deterministic setup including Git, Python, pre-commit, code formatters and linters, but you should be able to develop also without it.

yajo:
> I'm not used to test

I think it's more a matter of not knowing how to do a unit test, right?

I'll try to help when I have some free time. Or maybe you can help with that @sisp?

yajo:
Superseded by #1325.

Issue #312, opened 2020-11-28
Allow to execute when git is not installed


yajo:
Copier has lots of git-related stuff, but it also has features that work without it. It should be a lazy requirement.

astrojuanlu:
Can confirm:

```
root@a620fae20704:/# which git || echo "git is not available"
git is not available
root@a620fae20704:/# copier --version
Traceback (most recent call last):
  File "/usr/local/bin/copier", line 5, in <module>
    from copier.cli import CopierApp
  File "/usr/local/lib/python3.10/site-packages/copier/__init__.py", line 6, in <module>
    from .main import *  # noqa: F401,F403
  File "/usr/local/lib/python3.10/site-packages/copier/main.py", line 23, in <module>
    from plumbum.cmd import git
ImportError: cannot import name 'git' from 'plumbum.cmd' (/usr/local/lib/python3.10/site-packages/plumbum/cmd.py)
```

(For the same reason, `copier /path/to/template /output` doesn't work without `git` installed either)

astrojuanlu:
Gave this a try in gh-1074.

Issue #1045, opened 2023-03-21
Replace `git` subprocess by `pygit2`


astrojuanlu:
**Is your feature request related to a problem? Please describe.**
Now `copier` relies on `git` being installed in the system:

https://github.com/copier-org/copier/blob/47f3d59c27094e6f8ef0bc71acc74d88cbf6bb78/README.md?plain=1#L25

However, it would be nice if all the dependencies could be installed with pip without having to reach for the system package manager. This would be useful for folks that don't have admin access to their systems, but can `pip install` to their user locations.

**Describe the solution you'd like**
Replace subprocess calls to `git` through `plumbum`:

https://github.com/copier-org/copier/blob/47f3d59c27094e6f8ef0bc71acc74d88cbf6bb78/copier/main.py#L801-L811

with `pygit2`, the Python bindings for `libgit2`, which ship precompiled binary wheels for plenty of platforms:

https://pypi.org/project/pygit2/1.11.1/#files

As a side effect, this will also probably be faster (but I don't have data to back this up).

**Describe alternatives you've considered**
Could not think of other alternatives.

**Additional context**
This was rejected for cookiecutter as early as 2016: https://github.com/cookiecutter/cookiecutter/issues/846#issuecomment-258049634

In Kedro we are trying to remove our dependency with system `git` https://github.com/kedro-org/kedro/issues/2051 which prompted me to open this issue.

sisp:
I like the idea of removing the prerequisite of having Git installed. But I stumbled over one caveat that is a blocker for me:

> Binary wheels for Windows are available, but they don’t have support for ssh.
>
> &mdash; https://www.pygit2.org/install.html#quick-install (under "Caveats")

I don't use Windows, but I know many who do, and supporting SSH is quite important in our environment.

astrojuanlu:
I havw no experience with either, but @pawamoy suggestion looks sensible enough!

astrojuanlu:
Just to be sure: is this issue open for pull requests then? Or is there more discussion needed?

yajo:
I think this can be split in 2.

The 1st thing is to not require git. Copier can work with gitless templates. They have less features, but they work. Git should only be required when using a gitful template. 

If you want a gitful template, I can't imagine why would you not have git installed, so that could be the real fix here.

If there are reasons why that can happen and still dulwitch works for us (copier uses git deeply and intensely), then replacing git is the 2nd part of the story.

[...]

PR #1074, opened 2023-04-02
Do not immediately fail if git is not available

astrojuanlu:
Fix gh-312.

There are several ways of doing this, I picked one that I considered simple enough. The idea is to always import `git` from a centralized place, and in that place do a soft fail if `git` fails to be imported from `plumbum.cmd`.

This is how `copier --version` looks like with these changes:

```
root@885f691c5e79:/# which git || echo "git is not available"
git is not available
root@885f691c5e79:/# copier --version
/usr/local/lib/python3.10/site-packages/copier/vcs.py:25: UserWarning: git command not found, some functionality might not work
  warnings.warn(
copier 7.1.0a0.post98.dev0+5d600ff
```

On the other hand, I made `copier.vcs.is_git_bundle` more consistent with `is_git_shallow_repo` and `is_in_git_repo`, and now it returns `False` if there is an `OSError` or `ProcessExecutionError` error. Thanks to that, now copier can work with gitless templates as well, as mentioned in https://github.com/copier-org/copier/issues/1045#issuecomment-1492956700:

```
root@885f691c5e79:/# copier /home/copier-pylib-main/ /tmp/
/usr/local/lib/python3.10/site-packages/copier/vcs.py:25: UserWarning: git command not found, some functionality might not work
  warnings.warn(
🎤 Your account or organization
   astrojuanlu
...

 identical  .
    create  .gitignore
...
```

I have two questions:

1. First, I have no idea how to test this. Initially, I tried some form of mocking or monkeypatching, but since `plumbum.cmd.git` is not an actual attribute, things are more complicated:

```
    @pytest.fixture
    def no_plumbum_git(monkeypatch):
>       monkeypatch.delattr("plumbum.cmd.git")
E       AttributeError: git

tests/test_main.py:6: AttributeError
```

I could do a very deep patching of how `plumbum.local` retrieves the command, but wanted to ask for advice first.

2. As I explained in the second commit, one could actually consider that `copier.vcs.is_git_bundle`, `is_git_shallow_repo`, and `is_in_git_repo` should return None or something else than `True` or `False`, if the git status could not be determined because the git command is missing. In this scheme, returning `None` would mean "I don't know if this directory is actually a git repo". For this reason, I preferred to wait before proceeding any further adding tests.

Let me know what you folks think.

For the record, I too was a bit puzzled by the heavyweight requirements to run the linter https://github.com/copier-org/copier/issues/931 so going forward I think I'll use Gitpod to continue this PR (I have only used Nix a bit when experimenting with Replit earlier this year, not sure if I already want to commit to installing it on my home laptop).

### DIFF
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -14,21 +14,25 @@ from plumbum import TF, ProcessExecutionError, colors, local
 from .errors import DirtyLocalWarning, ShallowCloneWarning
 from .types import OptBool, OptStr, StrOrPath
 
-try:
-    from plumbum.cmd import git
 
-    GIT_VERSION = Version(re.findall(r"\d+\.\d+\.\d+", git("version"))[0])
-except ImportError:
+def get_git():
+    """Gets `git` command, or fails if it's not available"""
+    try:
+        from plumbum.cmd import git
+        return git
+    except ImportError as exc:
+        raise OSError("git is not available") from exc
+
+try:
+    GIT_VERSION = Version(re.findall(r"\d+\.\d+\.\d+", get_git()("version"))[0])
+except OSError:
     import warnings
 
     warnings.warn(
         "git command not found, some functionality might not work", UserWarning
     )
 
-    def git(*args, **kwargs):
-        raise OSError("git is not available")
-
-    GIT_VERSION = None
+    GIT_VERSION = None  # type: ignore
 
 
 GIT_PREFIX = ("git@", "git://", "git+", "https://github.com/", "https://gitlab.com/")
@@ -45,7 +49,7 @@ def is_git_repo_root(path: StrOrPath) -> bool:
     """Indicate if a given path is a git repo root directory."""
     try:
         with local.cwd(Path(path, ".git")):
-            return git("rev-parse", "--is-inside-git-dir").strip() == "true"
+            return get_git()("rev-parse", "--is-inside-git-dir").strip() == "true"
     except OSError:
         return False
 
@@ -53,7 +57,7 @@ def is_git_repo_root(path: StrOrPath) -> bool:
 def is_in_git_repo(path: StrOrPath) -> bool:
     """Indicate if a given path is in a git repo directory."""
     try:
-        git("-C", path, "rev-parse", "--show-toplevel")
+        get_git()("-C", path, "rev-parse", "--show-toplevel")
         return True
     except (OSError, ProcessExecutionError):
         return False
@@ -62,7 +66,7 @@ def is_in_git_repo(path: StrOrPath) -> bool:
 def is_git_shallow_repo(path: StrOrPath) -> bool:
     """Indicate if a given path is a git shallow repo directory."""
     try:
-        return git("-C", path, "rev-parse", "--is-shallow-repository").strip() == "true"
+        return get_git()("-C", path, "rev-parse", "--is-shallow-repository").strip() == "true"
     except (OSError, ProcessExecutionError):
         return False
 
@@ -73,8 +77,8 @@ def is_git_bundle(path: Path) -> bool:
         path = path.resolve()
     with TemporaryDirectory(prefix=f"{__name__}.is_git_bundle.") as dirname:
         with local.cwd(dirname):
-            git("init")
-            return bool(git["bundle", "verify", path] & TF)
+            get_git()("init")
+            return bool(get_git()["bundle", "verify", path] & TF)
 
 
 def get_repo(url: str) -> OptStr:
@@ -122,6 +126,7 @@ def checkout_latest_tag(local_repo: StrOrPath, use_prereleases: OptBool = False)
         use_prereleases:
             If `False`, skip prerelease git tags.
     """
+    git = get_git()
     with local.cwd(local_repo):
         all_tags = filter(valid_version, git("tag").split())
         if not use_prereleases:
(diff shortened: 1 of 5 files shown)


copier/vcs.py
```diff
@@ -10,15 +10,30 @@
...
+
+    GIT_VERSION = Version(re.findall(r"\d+\.\d+\.\d+", git("version"))[0])
+except ImportError:
+    import warnings
+
+    warnings.warn(
+        "git command not found, some functionality might not work", UserWarning
+    )
+
+    def git(*args, **kwargs):
```
yajo, line 29: I don't think this will work as expected because plumbum commands can get subscripted, and we actually do that quite often.

Example:

```py
_git = git["-C", dst_path]
_git("clone", src_path)
```

You can do all this inside a method and decorate it with `@proxy_tools.module_property`, so it's lazy loaded. See https://github.com/jtushman/proxy_tools.

Another way would be to also put it inside a method, but don't use that dependency. In that case, every usage of `vcs.git` should become refactored `vcs.git()` (because that method would return the git command). This way we'd save that dependency.

Since you'd be doing import within a lazy method, you could let the `ImportError` raise in case something actually requires git.

yajo (changes requested):
About testing, you can obtain the git interpreter through `plumbum.local["git"]` instead. That'll be easier to mock.

https://plumbum.readthedocs.io/en/latest/local_commands.html#guide-local-commands

copier/vcs.py
astrojuanlu: Thanks @yajo, it's clear that this needs more work. Will give all this some thought!

yajo: No problem! I marked this as draft in the mean time.

astrojuanlu: I ended up going for your second option (called it `get_git` instead of `git` to minimize confusion, although it makes the diff larger). Let me know what you think.

astrojuanlu:
This is ready for a second pass again. I think this now will be easier to test (haven't started doing it yet).

On the other hand, my other question still stands:

> As I explained in the second commit, one could actually consider that copier.vcs.is_git_bundle, is_git_shallow_repo, and is_in_git_repo should return None or something else than True or False, if the git status could not be determined because the git command is missing. In this scheme, returning None would mean "I don't know if this directory is actually a git repo". For this reason, I preferred to wait before proceeding any further adding tests.

yajo:
> As I explained in the second commit, one could actually consider that copier.vcs.is_git_bundle, is_git_shallow_repo, and is_in_git_repo should return None or something else than True or False, if the git status could not be determined because the git command is missing. In this scheme, returning None would mean "I don't know if this directory is actually a git repo". For this reason, I preferred to wait before proceeding any further adding tests.

I don't think this is a good solution.

Instead, those functions should be able to return a boolean without relying in the presence of git. That shouldn't be too hard (just check if there's a `.git` file/dir in the folder or any of its parents, usually).

This way, _if it really is a git repo and git is not available_, then copier should fail.

copier/vcs.py
```diff
@@ -10,14 +10,33 @@
...
 
 from .errors import DirtyLocalWarning, ShallowCloneWarning
 from .types import OptBool, OptStr, StrOrPath
 
+
+def get_git():
+    """Gets `git` command, or fails if it's not available"""
+    try:
+        from plumbum.cmd import git
+        return git
```
yajo, line 22: I think this is better, if you don't mind:

```suggestion
        return local["git"]
```

copier/vcs.py
```diff
@@ -10,14 +10,33 @@
...
+    except ImportError as exc:
+        raise OSError("git is not available") from exc
```
yajo, line 24: If you change the above, you can probably leave the original exception.

copier/vcs.py
```diff
@@ -10,14 +10,33 @@
...
+try:
+    GIT_VERSION = Version(re.findall(r"\d+\.\d+\.\d+", get_git()("version"))[0])
+except OSError:
+    import warnings
+
+    warnings.warn(
+        "git command not found, some functionality might not work", UserWarning
+    )
+
+    GIT_VERSION = None  # type: ignore
```
yajo, line 35: I think this is better done in a function. `def git_version(): ...`

This way it's lazy and you actually get an exception if you really need a git version for whatever reason.

yajo (changes requested):

astrojuanlu:
Applied changes and ran `poe lint` (doing so from Gitpod was more or less straightforward 👍🏽)

astrojuanlu:
> > As I explained in the second commit, one could actually consider that copier.vcs.is_git_bundle, is_git_shallow_repo, and is_in_git_repo should return None or something else than True or False, if the git status could not be determined because the git command is missing. In this scheme, returning None would mean "I don't know if this directory is actually a git repo". For this reason, I preferred to wait before proceeding any further adding tests.
> 
> I don't think this is a good solution.
> 
> Instead, those functions should be able to return a boolean without relying in the presence of git. That shouldn't be too hard (just check if there's a `.git` file/dir in the folder or any of its parents, usually).
> 
> This way, _if it really is a git repo and git is not available_, then copier should fail.

I'm reverting this commit for now

yajo (approved):

sharpencrag:
Apologies if this is not the right place to bring it up, but after this merge, I am still seeing a failure when attempting to run `copier` without git installed.

`is_git_bundle` is still calling `get_git` without handling the exception when it fails

astrojuanlu:
See https://github.com/copier-org/copier/pull/1074#issuecomment-1694355508, left it out of this PR

PR #1075, opened 2023-04-03
feat: validate default values

sisp:
I've added validation of default values which revealed an error in one of the test cases and some general weakness in the answer parsing/casting. In particular, a `type: str/int/float/bool` question could have `None` as its default value and Copier didn't complain for several reasons:

* `str` had a [special treatment](https://github.com/copier-org/copier/blob/17135c5f346bb2daad32c40ae43f8790ac50c270/copier/user_data.py#L441-L442) which [seems obsolete](https://github.com/copier-org/copier/pull/958/commits/bdf84ad346ffe3f2484e0213607ab0e209fc9031) to me after #958 has been merged.
* `bool` silently handles `None` where `bool(None) is False`.
* `int` and `float` both raise a `TypeError` for `int(None)`/`float(None)` which got swallowed by a [too broad `try`/`except` block](https://github.com/copier-org/copier/blob/17135c5f346bb2daad32c40ae43f8790ac50c270/copier/user_data.py#L443-L447) actually intended only for allowing `type: json/yaml` questions to parse answers with native (i.e. not stringified) Python values.

**Note**: This PR requires #958 to be merged first. The test suite fails right now but passes for me after merging #958 locally.


copier/user_data.py
```diff
@@ -450,13 +445,25 @@ def load_answersfile_data(
         return {}
 
 
-def cast_answer_type(answer: Any, type_fn: Callable) -> Any:
+def cast_answer_type(answer: Any, type_name: str) -> Any:
     """Cast answer to expected type."""
+    try:
+        type_fn = CAST_STR_TO_NATIVE[type_name]
+    except KeyError:
+        raise InvalidTypeError("Invalid answer type")
```
sisp, line 453: I'm not sure how to cover this error by a test as it doesn't occur in practice because the question type is checked to be one of the keys of `CAST_STR_TO_NATIVE` before. Nevertheless, this function could theoretically be called in a different context, so I think it makes sense to keep the check here.

sisp (commented):
@yajo I think this PR is ready for review now. There is one uncovered line, please see my inline comment.

yajo:
Could you please solve the conflicts?

sisp:
Done.

Issue #987, opened 2023-02-20
Autoselecting version tags


FlorianLudwig:
**Describe the problem**

Given the following git tags in a repo:
```
v1
v1.0
v1.0.1
v1.0.0
```

**To Reproduce**

1. Run `copier copy` on a repo with the above mentioned tags
2. copier will use `v1`

**Expected behavior**
copier will use `v1.0.1`

**Environment**
 - Copier version: 7.0.1

FlorianLudwig:
I can still reproduce this:

```
$ copier --version                                                              
copier 7.0.1

$ copier copy git@github.com:FlorianLudwig/copier-987.git test
$ cat test/.copier-answers.yml                                
# Changes here will be overwritten by Copier
_commit: v1
_src_path: git@github.com:FlorianLudwig/copier-987.git
```

sisp:
The problem is that

https://github.com/copier-org/copier/blob/17135c5f346bb2daad32c40ae43f8790ac50c270/copier/template.py#L264

returns `v1`, probably because multiple tags point to that same commit.

yajo:
That's not a bug then

FlorianLudwig:
@yajo would you accept a new feature to support repos that contain multiple tags on the same commit?

FlorianLudwig:
suggested algorithm:

- get all tags
- filter valid versions
- select tag with longest string
```
tags = git("tag", "--points-at", "HEAD")
tags = filter(valid_version, tags)
tags.sort(key=lambda x: len(x))
```

sisp:
> * filter valid versions
> 
> * select tag with longest string

I think the algorithm should rather be similar to this one:

https://github.com/copier-org/copier/blob/c96ff3cf5ebdea6bbd89f6204c3f1490bb30020c/copier/vcs.py#L103-L129

FlorianLudwig:
@sisp the issue for me is that it does not solve my use-case.

Having tags for branches (e.g. `v1`, `v1.1` and `v1.1.0`) at the same is not too uncommon, I think.

sorting those with by version results in `['v1.1', 'v1.1.0', 'v1']`. So it selects "highest" version first and "least specific" next. Since it's tags on the same commit sorting by "highest" version should not make any difference as all tags are actually the "same" version.

yajo:
> Since it's tags on the same commit sorting by "highest" version should not make any difference as all tags are actually the "same" version.

[...]

PR #1080, opened 2023-04-06
Draft: add tag selecting algorithm

FlorianLudwig:
In a VCS tracked template where the selected commit has multiple tags pointing to it select a tag by:

- Ensure only a tag that is actually a version tag might be selected
- If multiple version tags exist, select the most specific (longest string)

See #987

### DIFF
--- a/copier/template.py
+++ b/copier/template.py
@@ -28,7 +28,7 @@ from .errors import (
 )
 from .tools import copier_version, handle_remove_readonly
 from .types import AnyByStrDict, Env, OptStr, StrSeq, Union, VCSTypes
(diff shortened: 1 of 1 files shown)


yajo:
The algorithm is wrong because length doesn't have a meaning.

We use PEP440 as a tag indicator, so `v2` is newer than `v1.0.1`, and `v1.1.alpha4` is lower than `v1.0.1` unless you're in prerelease mode.

Since this PR is wrong, let me close this while we triage the issue in #987. If we see it's needed, you'll be able to open a new one later. 😉 

Thanks for your help!

PR #1101, opened 2023-04-14
feat: add support for validating answers with JSON Schema

### LINKED ISSUE #1065: Add support for validating JSON/YAML questions with JSON Schema
Copier supports `type: json` and `type: yaml` questions which means their answers can be complex structured data. Copier also supports [answer validation via the `validator` field](https://copier.readthedocs.io/en/v7.0.1/configuring/#advanced-prompt-formatting) using a Jinja expression to render an error message or an empty string when validation succeeds. This validation approach is fine for simple answers but not well suited for complex structured data.

JSON/YAML data can be validated very nicely using the widely adopted [JSON Schema](https://json-schema.org/) language. I suggest to add support for validating `type: json` and `type: yaml` questions using a JSON Schema document in addition to the Jinja based validation approach. The existing `validator` field could be reused to also accept a YAML object which contains the JSON Schema document. And the popular Python library [`jsonschema`](https://github.com/python-jsonschema/jsonschema) could be used to perform the validation of the data against the schema.

For example:

```yaml
complex_question:
  type: yaml  # or `json`
  validator:
    # Here comes the JSON Schema document that
    # validates a simple object in this example
    type: object
    required:
      - name
      - age
    additionalProperties: false
    properties:
      name:
        type: string
      age:
        type: integer
        minimum: 0
```

The existing validator approach using complex Jinja if-else rules could be used, but it would be very verbose and difficult to read.

**Additional context**
n/a

sisp:
I've added support for validating answers with JSON Schema using a new Jinja filter `jsonschema(schema_uri: str)` which returns a [`jsonschema.ValidationError`](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/exceptions/#jsonschema.exceptions.ValidationError) object when validation fails and `None` otherwise. For example:

```yaml
question:
  type: yaml
  validator: "{{ question | jsonschema('schema.yml') | default('', true) }}"
```

The Jinja filter supports schemas written in JSON and YAML, numerous dialects supported by the installed Python `jsonschema` library, local schemas (relative to the local template root), remote schemas (HTTP and HTTPS protocols), schema references (including references with mixed formats, i.e. a YAML schema can reference a JSON schema and vice versa), and JSON Pointers. See the test cases for an overview of these features.

Local schemas and YAML support currently require a custom `RefResolver`. Starting with the next minor release v4.18.0 of the Python `jsonschema` library, a [new approach based on the `referencing` library](https://python-jsonschema.readthedocs.io/en/latest/referencing/#migrating-from-refresolver) will replace the `RefResolver` approach. I'll send a PR to migrate to the new approach then.

I considered writing a custom [Jinja extension](https://jinja.palletsprojects.com/en/3.1.x/extensions/#extension-api), but decided against it in the end because the filter needs contextual information about the local template root path to ensure that a local schema or a schema reference won't access a file above the local template root. According to the Jinja docs, this is not easily achieved with Jinja extensions:

> As extensions are created by the environment they cannot accept any arguments for configuration. One may want to work around that by using a factory function, but that is not possible as extensions are identified by their import name. The correct way to configure the extension is storing the configuration values on the environment. Because this way the environment ends up acting as central configuration storage the attributes may clash which is why extensions have to ensure that the names they choose for configuration are not too generic.

Closes #1065.


sisp:
@yajo This PR is ready for review now. :pray:

yajo:
To solve the problem of the lack of context in the jinja extension, couldn't you do this? 

```jinja
{% set spec | from_yaml %}
{% include 'schema.yml' %}
{% endset %}
{{ question | jsonschema(spec) }}
```

sisp:
That would still not solve the Jinja extension issue because I need to know the local template root path for resolving local references and checking that they don't resolve to outside the local template root path.

This might be a third option for providing a schema though, either via a Jinja include or by defining the schema in the Jinja set block directly. Then, a local reference in the root schema would always resolve to a path relative to the local template root path.

sisp:
But it might be possible to solve thr Jinja extension issue if we changed the current working directory to the local template root path processing the questionnaire. Then, the Jinja extension could simply check that local references don't resolve to above the current working directory.

sisp:
I believe I've found a better solution for registering the filter via a proper Jinja extension. Instead of passing the local template path to the filter/extension as a configuration and implementing the loading of the schema from the filesystem, I now use the Jinja loader (`FileSystemLoader`) that is already configured in the Jinja environment to load the schema. This way, there's no need to provide the local template path explicitly because the Jinja loader is already configured to load files from a location relative to the local template path, it also prevents accessing files above the configured root path, it requires POSIX-style paths as that's simply how Jinja works, and absolute paths (relative to the local template path) are now supported for both the root schema and in references well. As a result, the implementation has become quite a bit simpler.

In addition, I've also added support for loading inline schemas, e.g. by using a Jinja block assignment. Any local references in an inline schema are relative to the local template path. Using a Jinja block assignment in combination with a Jinja include isn't a replacement for passing a local file path to the filter though because it doesn't allow using local references with relative paths to the root schema file.

WDYT, @yajo?

yajo:
Thanks! That's perfect. 😊 

However, as you can see, now this extension has no relationship with Copier itself. It's a very useful extension that should work with any Jinja template. It could easily be expanded to work even with other loaders, not only the filesystem one (if it isn't already).

So, may you please package it separately in another project? We can point to it in our docs as a "featured" one, [as we have with others](https://copier.readthedocs.io/en/latest/configuring/#jinja_extensions).

yajo:
If you want, I can create the repo and host it in copier org and put you as
admin. Just tell me the project name you want 😃

sisp:
Nice! I think `jinja2-jsonschema` sounds appropriate, and [`jinja2-jsonschema` seems to be available on pypi.org](https://pypi.org/project/jinja2-jsonschema/).

yajo:
I thought Github would tell you, but you're the admin of https://github.com/copier-org/jinja2-jsonschema now.

Thus I close it here, and you can reopen a new one for the docs when the extension is published 🚀

yajo:
BTW I'm thinking that there's a bad design here. I think it should be a [test](https://jinja.palletsprojects.com/en/3.0.x/templates/#tests) and not a [filter](https://jinja.palletsprojects.com/en/3.0.x/templates/#filters). IIUC, we're just asserting a value, not converting it.

sisp:
> I thought Github would tell you, but you're the admin of https://github.com/copier-org/jinja2-jsonschema now.

It did, I just haven't found the time to turn this PR into an independent package. Hopefully next week. I've been working on `copier doctor`. :laughing:

> Thus I close it here, and you can reopen a new one for the docs when the extension is published :rocket:

:+1:

> BTW I'm thinking that there's a bad design here. I think it should be a [test](https://jinja.palletsprojects.com/en/3.0.x/templates/#tests) and not a [filter](https://jinja.palletsprojects.com/en/3.0.x/templates/#filters). IIUC, we're just asserting a value, not converting it.

A Jinja test returns a boolean, but this `jsonschema()` filter converts data into an error object or `None` in case the validation is successful. That's why I think it's a filter and not a test.

PR #1110, opened 2023-04-18
fix: fix answer validation for question with complex choices

sisp:
I've fixed a bug related to answer validation of questions with complex choices. Previously, choice values were collected in a set which requires the choice values to be hashable. But complex choice values such as `dict` or `list` values aren't hashable. Now, they are collected in a list.

Fixes #1108.

Issue #1108, opened 2023-04-18
Cannot pass yaml as --data cli argument


svsk417:
**Describe the problem**
I am trying to pass a list as a `--data` cli argument while generating the project like:
```
copier --force --data 'layers=[data, domain]' --vcs-ref HEAD copy <my template dir> <my target dir>
```

Within the copier.yaml file the layers question is defined as follows: 

```
layers: 
  type: yaml
  help: Which layers do you need? 
  default: [data, domain]
  choices:
    data & domain: [data, domain]
    data: [data]
```

**To Reproduce**

1. Run `copier --force --data 'layers=[data, domain]' --vcs-ref HEAD copy <my template dir> <my target dir>`
2. See error

**Logs**

<details>
```
Traceback (most recent call last):
  File "/usr/local/bin/copier", line 8, in <module>
    sys.exit(copier_app_run())
  File "/usr/local/lib/python3.10/site-packages/plumbum/cli/application.py", line 639, in run
    inst, retcode = subapp.run(argv, exit=False)
  File "/usr/local/lib/python3.10/site-packages/plumbum/cli/application.py", line 634, in run
    retcode = inst.main(*tailargs)
  File "/usr/local/lib/python3.10/site-packages/copier/cli.py", line 68, in _wrapper
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/copier/cli.py", line 295, in main
    ).run_copy()
  File "/usr/local/lib/python3.10/site-packages/copier/main.py", line 697, in run_copy
    src_abspath = self.template_copy_root
  File "/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/usr/local/lib/python3.10/site-packages/copier/main.py", line 670, in template_copy_root
    subdir = self._render_string(self.template.subdirectory) or ""
  File "/usr/local/lib/python3.10/site-packages/copier/main.py", line 644, in _render_string
    return tpl.render(**self._render_context())
  File "/usr/local/lib/python3.10/site-packages/copier/main.py", line 249, in _render_context
    "answers_file": self.answers_relpath,
  File "/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
[...]

PR #1111, opened 2023-04-19
fix: work around Pydantic bug when parsing choices

sisp:
I've fixed a bug when parsing complex-valued choices caused by an [upstream bug in Pydantic v1](https://github.com/pydantic/pydantic/issues/5523).

The workaround is to change the order of the type union so that list-style choices are parsed using the `Sequence[Any]` type because it matches before `Dict[Any, Any]` (which shouldn't match in that case at all but does due to the bug).

Related to #1108 & #1110.

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -202,7 +202,7 @@ class Question:
     var_name: str
     answers: AnswersMap
     jinja_env: SandboxedEnvironment
-    choices: Union[Dict[Any, Any], Sequence[Any]] = field(default_factory=list)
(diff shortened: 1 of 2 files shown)


sisp:
Sometimes, I wonder why Codecov thinks there's a coverage drop. :thinking: I can't image any reason why coverage should drop in this PR. :confused:

yajo:
It is usually because some jobs fail to upload because there's no way to
read the secret token for external prs. All of them aggregate results from
every os. Thus, some jobs failing means some lines not covered. Gotta fix
that some day 😅

PR #1117, opened 2023-04-20
feat(cli): add runtime type to `--answers-file` switch

sisp:
Just a minor improvement of the CLI switch `--answers-file`.


sisp:
Ah, never mind, we can't use `cli.ExistingFile` here because

1. the file cannot exist yet when the project is generated, and
2. the file must exist _in the project folder_ when the project is updated, but `cli.ExistingFile` searches relative to the current working directory and not relative to the project folder.

PR #1126, opened 2023-04-26
fix: skip question when skip condition is met

### LINKED ISSUE #1125: Error with --force and `when` condition with copier >= 7.1.0
**Describe the problem**
Hello, I have a copier.yaml like the following in which I first asks for the kind of config and than ask a more specific question. So when I select kind=foo, the question for key testbar will be skipped.
But when I execute the command with `--force` it raises this Error: `ValueError: Question "testbar" is required`

**Template**

``` yaml
kind:
  type: str
  help: What kind do you need?
  choices:
    foo: foo
    bar: bar

testbar:
  when: "{{ kind == 'bar' }}"
  type: str
  help: any string bar?

testfoo:
  when: "{{ kind == 'foo' }}"
  type: str
  help: any string foo?
```

**To Reproduce**

`copier https://github.com/myproject.git --data "kind=foo" --data "testfoo=helloworld" --force .`
ValueError: Question "testbar" is required

**Logs**

<details>

```
  File "/root/.local/bin/copier", line 8, in <module>
    sys.exit(copier_app_run())
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/plumbum/cli/application.py", line 639, in run
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/plumbum/cli/application.py", line 634, in run
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/cli.py", line 68, in _wrapper
    return method(*args, **kwargs)
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/cli.py", line 291, in main
    self.parent._worker(
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/main.py", line 697, in run_copy
    src_abspath = self.template_copy_root
  File "/usr/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/main.py", line 670, in template_copy_root
    subdir = self._render_string(self.template.subdirectory) or ""
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/main.py", line 644, in _render_string
    return tpl.render(**self._render_context())
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/main.py", line 249, in _render_context
    "answers_file": self.answers_relpath,
  File "/usr/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/main.py", line 430, in answers_relpath
    return Path(template.render(**self.answers.combined))
  File "/usr/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/root/.local/pipx/venvs/copier/lib/python3.8/site-packages/copier/main.py", line 399, in answers
    raise ValueError(f'Question "{var_name}" is required')
ValueError: Question "testbar" is required
```

</details>

**Expected behavior**

Executing the command without `--force` works, but I need it with `--force` for using it in Github Actions and so on. With the version `7.0.1` the command worked without Error

**Environment**
 - OS: ubuntu 20.04
 - Copier version: 7.1.0
 - Python version: 3.8.10
 - Installation method: pipx

sisp:
I've indirectly fixed the validation of answers to questions for which the skip condition is met. Before, a question with a truthy skip condition was not properly skipped such that (a) its default value (if present) was still validated and could cause a validation error or (b) an error was raised that the question was required although it should have been skipped. As a side effect of the previous implementation (which didn't properly skip questions), answers of skipped questions were recorded in the answers file which IMO makes no sense.

I've changed the way skipped questions are handled such that the skip condition is evaluated early and the question is immediately skipped when the skip condition is truthy. To prevent records of such answers in the answers file, e.g. when they originate from previous runs, when they are provided by the user via the `--data` flag, etc., any answers in the internal answers map are removed (or rather marked as removed). As a result, I needed to adapt a few tests and even removed one that specifically tested for the records in the answers file.

This new behavior should be backwards compatible and is more logical in my opinion. As a side effect, some slightly confusing code block could be removed which handled some special case of using a templated default answer.

Fixes #1125.

### DIFF
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -512,6 +512,42 @@ def test_validate_init_data(
         copier.copy(str(src), dst, data={"q": value})
 
 
+@pytest.mark.parametrize("defaults", [False, True])
+def test_validate_init_data_with_skipped_question(
+    tmp_path_factory: pytest.TempPathFactory, defaults: bool
+) -> None:
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    build_file_tree(
+        {
+            (src / "copier.yml"): (
+                f"""\
+                _envops: {BRACKET_ENVOPS_JSON}
+
+                kind:
+                    type: str
+                    help: What kind do you need?
+                    choices:
+                        foo: foo
+                        bar: bar
+
+                testbar:
+                    when: "[[ kind == 'bar' ]]"
+                    type: str
+                    help: any string bar?
+
+                testfoo:
+                    when: "[[ kind == 'foo' ]]"
+                    type: str
+                    help: any string foo?
+                """
+            ),
+        }
+    )
+    copier.copy(
+        str(src), dst, defaults=defaults, data={"kind": "foo", "testfoo": "helloworld"}
+    )
+
+
 @pytest.mark.parametrize(
     "type_name",
     ["str", "int", "float", "bool", "json", "yaml"],
--- a/copier/main.py
+++ b/copier/main.py
@@ -390,6 +390,12 @@ class Worker:
                 var_name=var_name,
                 **details,
             )
+            # Skip a question when the skip condition is met and remove any data
+            # from the answers map, so no answer for this question is recorded
+            # in the answers file.
+            if not question.get_when():
+                result.remove(var_name)
+                continue
             if var_name in result.init:
                 # Try to parse the answer value.
                 answer = question.parse_answer(result.init[var_name])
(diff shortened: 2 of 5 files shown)


tests/test_copy.py
```diff
@@ -512,6 +512,42 @@ def test_validate_init_data(
...
+                    when: "[[ kind == 'foo' ]]"
+                    type: str
+                    help: any string foo?
+                """
+            ),
+        }
+    )
+    copier.copy(
+        str(src), dst, defaults=defaults, data={"kind": "foo", "testfoo": "helloworld"}
+    )
```
yajo, line 548: add some file to be rendered here and assert its contents, please.

yajo (changes requested):

tests/test_copy.py
sisp: Done. Is this okay for you? a2c1103

PR #1136, opened 2023-05-14
Fix filesystem view to match later graphic (and homepage)

mitchnegus:
This is a really minor change, but it looks like the 'Minimal Example' filesystem graphic on the [Creating a template](https://copier.readthedocs.io/en/stable/creating/#minimal-example) page is slightly off.

According to the graphic a bit further down on the page, `.copier-answers.yml` should be found under `generated_project`, and so `{{_copier_conf.answers_file}}.jinja` should be directly inside the `my_copier_template` (not `{{project_name}}`.


yajo (approved):
Ah, good catch! Thanks. 😊

PR #1145, opened 2023-05-17
docs: replace `run_auto()` by `run_copy()`

sisp:
Fantastic changes in #1143! :bow: :star_struck: :rocket:

I just updated the docs to fix a few leftovers from the past regarding usage of `copier.run_auto()`.


yajo (approved):
Thanks!

PR #1151, opened 2023-05-22
chore: pull issue templates from github.com:copier-org/.github

sisp:
I've pulled the issue templates from https://github.com/copier-org/.github/tree/main/.github/ISSUE_TEMPLATE as we discussed (https://github.com/copier-org/.github/issues/3). I've preserved the history and intend the commits to be squashed, just in case you're wondering about the number of commits in this PR, @yajo.


yajo:
Thanks! I'll squash them because otherwise these commits would pollute the changelog.

sisp:
Yes, that's what I intended you to do. :+1:

Issue #943, opened 2023-02-03
copier not overwriting file on update, .rej file is wrong


lunik1:
**Describe the problem**
I'll admit I'm trying to do something a little strange! I am trying to create a template that will start off substituting the current year:
```
2023
```
and will, in future years, automatically expand this to a range of years. So if I were to update a project with this template in 2 years time, I would get:
```
2023-2025
```
Specifically, I would like to use this in `LICENSE` files.

I have tried to achieve this by creating a `start_year` question, which is set to the current year when the project is initialised, and saved to the answers file. I also create `current_year` which is set to the current year but not saved to the answers file, via (ab)use of secrets. I then compare `start_year` and `end_year` to know whether to print a range of years or not (the full template and MWE is linked in the next section).

Unfortunately, this does not work. When updating in a future year from the `start_year` copier reports a conflict, but the file is unchanged. Oddly, if changes are rejected, the reject file shows
```patch
diff a/dates.txt b/dates.txt    (rejected hunks)
@@ -1 +1 @@
-2023-2025
+2023
```

**Template**
https://gist.github.com/lunik1/d41aea8a0b643a34d2f47b97957c2bd7

**To Reproduce**

1. Run `copier copy {{ my_template }} dir`
2. Run `cd dir && git init && git add . && git commit -m "Initial commit"`
3. Run `faketime -f '+2y' copier update .`*
4. See issue outlined above

*requires libfaketime to be installed. Alternatively, you can wait.

**Logs**

<details>

```
 identical  .
  conflict  dates.txt
 Overwrite dates.txt? [Y/n] y
 identical  .copier-answers.yml
```

</details>

**Expected behavior**

The `date.txt` file to contain `{{ start_year }}-{{ current _year }}`.

...well to be honest I appreciate this is a bit of a hack so I _expected_ something to go subtly wrong somewhere. Is there a better way of achieving this, or could support be added for e.g. variables that are created on copy and then can be used by update?

**Environment**
 - OS: Debian GNU/Linux 11 (bullseye) (WSL)
 - Copier version: 7.0.1
 - Python version: 3.10.9
 - Installation method: nixpkgs

yajo:
Have you tried installing from Master and using the --conflict inline
option? That can give you a better result I think.

PR #1156, opened 2023-05-23
fix: don't lie about updated files

yajo:
Updating involves some low-level git-fu. This can cause unpredicted files to get updated, and the report to the user sometimes is fake.

It'd be great to tell the truth, but for now at least let's just not lie.

Fix https://github.com/copier-org/copier/issues/943.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -823,6 +823,7 @@ class Worker:
                 quiet=True,
             )
             current_worker.run_copy()
+            self.answers = current_worker.answers
             # Render with the same answers in an empty dir to avoid pollution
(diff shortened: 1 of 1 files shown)


yajo:
Well, it seemed more simpler but it seems to break too many things... I'll publish a better fix when I have time.

sisp:
@yajo You nearly had it working, only the answers need recorded in the temporary (quiet) worker need to be assigned back to the main worker. I took the liberty to reopen this PR and add the missing piece.

yajo (commented):
Oh cool!

Issue #1137, opened 2023-05-14
Reduce template dangers


yajo:
> The problem is, templates just need these features. Writing good migrations
takes time, but they provide smooth updates; but what if users are able to
turn them off? Then you can't predict the output. Tasks also are needed to
replay last render when updating. And jinja would die if you use it without
the required extensions.

Good point!

> So, although it's OK to warn the user, the features are needed for copier
to do its work. Basically, running Copier without --UNSAFE would only work
with templates without these features. At least, we shouldn't require that
flag when the operation won't involve any of those. Also we should provide
a different exit code for this exit path.

:ok_hand:

>  think it'd be OK to present a question before starting, which tells all
dangers at once, so user allows them all or exits. In the case of updates,
we need to include replay dangers and migrations if there are any between
the two versions involved.

:+1:

> Copier is a development tool, so it will usually need access to other
development tools expected in that environment.

Also good point! I hadn't considered that.

> I don't recommend running malicious templates in a container. I just
recommend not running them :laughing:.
>
> However, it would be good to have a page in the docs dedicated to our
threat model and possible workarounds. And that could happen even without
implementing any other fix. It could even be "the fix". :thinking:

Yes, I think docs plus the additional question when tasks/migrations/extensions are present or the `--UNSAFE` flag in case of non-interactive usage, as you described above.

I think this sounds like a good actions plan to raise awareness of the dangers and let users make conscious decisions. :+1:

_Originally posted by @sisp in https://github.com/copier-org/copier/discussions/1132#discussioncomment-5881590_

yajo:
@sisp would you like to tackle this issue?

sisp:
Yes, feel free to assign me to this issue. :+1:

fcollonval:
Hey, I understand all the security concerns raised here. But I feel like the `--UNSAFE` flag is a bit scary for newbies that don't understand all the good reasons for it. I'm wondering if `copier` could store a list of trusted template source in config file in the user home directory. The idea is to avoid having to set UNSAFE every time you use a trusted template.

We could then add a flag `--trust` to `copier copy` that implies `--UNSAFE` and add the template to the trusted list;

```sh
# First time use
copier copy --trust https://github.com/org/trusted-template.git .

# Update does not require the flag
copier update

# Next time use, does not raise
copier copy https://github.com/org/trusted-template.git .
```

pawamoy:
[...]

PR #1171, opened 2023-06-01
feat: disable unsafe features by default and add `--UNSAFE` switch

sisp:
I've disabled the use of unsafe features (Jinja extensions, migrations, and tasks) by default and added a new CLI switch `--UNSAFE` which enables them. Templates that don't use unsafe features are unaffected by this change. But Copier will raise an error for templates that do use unsafe features unless the `--UNSAFE` flag is passed.

I've not added an interactive prompt that asks for consent for using unsafe features because I think it's not clear how to distinguish between interactive prompting and raising an error when `--UNSAFE` is not passed. For this, I think Copier would need a switch that clearly states whether interactive or non-interactive mode is desired. Currently, `--defaults` implies this for questions.

We could add an interactive prompt in a follow-up PR ones the path forward is clear. I'd also add a dedicated page to the docs regarding Copier's security architecture and threat model in a separate PR.

Resolves partially #1137.

### DIFF
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1319,10 +1319,11 @@ Copier templates can use dangerous features that allow arbitrary code execution:
 -   [Migrations](#migrations)
 -   [Tasks](#tasks)
 
-Therefore, these features are disabled by default and Copier will raise an error when
-they are found in a template. In this case, please verify that no malicious code gets
-executed by any of the used features. When you're sufficiently confident or willing to
-take the risk, set `unsafe=True` or pass the CLI switch `--UNSAFE`.
+Therefore, these features are disabled by default and Copier will raise an error (and
+exit from the CLI with code `2`) when they are found in a template. In this case, please
+verify that no malicious code gets executed by any of the used features. When you're
+sufficiently confident or willing to take the risk, set `unsafe=True` or pass the CLI
+switch `--UNSAFE`.
 
 !!! danger
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1307,6 +1307,31 @@ templates suffix is _not_ empty, Copier will abort and print an error message.
 
     Copier 7+ no longer uses the old default independent of [min_copier_version][].
 
+### `unsafe`
+
+-   Format: `bool`
+-   CLI flags: `--UNSAFE`
+-   Default value: `False`
+
+Copier templates can use dangerous features that allow arbitrary code execution:
+
+-   [Jinja extensions](#jinja_extensions)
+-   [Migrations](#migrations)
+-   [Tasks](#tasks)
+
+Therefore, these features are disabled by default and Copier will raise an error when
+they are found in a template. In this case, please verify that no malicious code gets
+executed by any of the used features. When you're sufficiently confident or willing to
+take the risk, set `unsafe=True` or pass the CLI switch `--UNSAFE`.
+
+!!! danger
+
+    Please be sure you understand the risks when allowing unsafe features!
+
+!!! info
+
+    Not supported in `copier.yml`.
+
 ### `use_prereleases`
 
 -   Format: `bool`
--- a/tests/test_unsafe.py
+++ b/tests/test_unsafe.py
@@ -8,7 +8,7 @@ from plumbum import local
 from plumbum.cmd import git
 
 from copier.cli import CopierApp
-from copier.errors import UserMessageError
+from copier.errors import UnsafeTemplateError
 from copier.main import run_copy, run_update
 from copier.types import AnyByStrDict
 
@@ -33,7 +33,7 @@ class JinjaExtension(Extension):
         (
             {"_tasks": ["touch task.txt"]},
             pytest.raises(
-                UserMessageError, match="Template uses unsafe feature: tasks"
+                UnsafeTemplateError, match="Template uses unsafe feature: tasks"
             ),
         ),
         (
@@ -59,7 +59,7 @@ class JinjaExtension(Extension):
         (
             {"_jinja_extensions": ["tests.test_unsafe.JinjaExtension"]},
             pytest.raises(
-                UserMessageError,
+                UnsafeTemplateError,
                 match="Template uses unsafe feature: jinja_extensions",
             ),
         ),
@@ -69,7 +69,7 @@ class JinjaExtension(Extension):
                 "_jinja_extensions": ["tests.test_unsafe.JinjaExtension"],
             },
             pytest.raises(
-                UserMessageError,
+                UnsafeTemplateError,
                 match="Template uses unsafe features: jinja_extensions, tasks",
             ),
         ),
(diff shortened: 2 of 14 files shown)


tests/test_unsafe.py
```diff
@@ -0,0 +1,374 @@
...
+                "_migrations": [
+                    {
+                        "version": "v1",
+                        "before": ["touch v1-before.txt"],
+                        "after": ["touch v1-after.txt"],
+                    }
+                ]
+            },
+            does_not_raise(),
+        ),
```
yajo, line 54: Why this case doesn't raise? 🤔

tests/test_unsafe.py
```diff
@@ -0,0 +1,374 @@
...
+        {(src / "copier.yaml"): yaml.safe_dump({"_tasks": ["touch task.txt"]})}
+    )
+    _, retcode = CopierApp.run(
+        ["copier", "copy", *(["--UNSAFE"] if unsafe else []), str(src), str(dst)],
+        exit=False,
+    )
+    if unsafe:
+        assert retcode == 0
+    else:
+        assert retcode == 1
```
yajo, line 108: Could we have a special return code for this kind of failure?

yajo (changes requested):
Neat! Thanks, looks very good. Actually it looks so good that I'd gladly forget the idea of interactive prompting for unsafety. You're right in that it could be confused with the usual questionary.

Some questions below, though.

BTW would you mind amending the commit and adding a `BREAKING CHANGE:` trailer so it helps the autogeneration of changelog when I launch the next release? See [the example](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer).

tests/test_unsafe.py
sisp: Because `run_copy()` doesn't execute migration tasks, so I think they should be ignored by the check.

tests/test_unsafe.py
sisp: Good idea. I've added a new error class and catch it in the CLI code to exit with code 2 in this case.

sisp:
> BTW would you mind amending the commit and adding a BREAKING CHANGE: trailer so it helps the autogeneration of changelog when I launch the next release? See [the example](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer).

Sure, no problem. :slightly_smiling_face: I've amended the commit and added a `BREAKING CHANGE:` trailer.

yajo (approved):

docs/configuring.md
```diff
@@ -1307,6 +1307,31 @@ templates suffix is _not_ empty, Copier will abort and print an error message.
...
+Copier templates can use dangerous features that allow arbitrary code execution:
+
+-   [Jinja extensions](#jinja_extensions)
+-   [Migrations](#migrations)
+-   [Tasks](#tasks)
+
+Therefore, these features are disabled by default and Copier will raise an error when
+they are found in a template. In this case, please verify that no malicious code gets
+executed by any of the used features. When you're sufficiently confident or willing to
+take the risk, set `unsafe=True` or pass the CLI switch `--UNSAFE`.
```
yajo, line 1325: I think it would be useful to explain that these kind of failures give an exit code 2 on CLI.

yajo (changes requested):
Please just one little extra request.

Besides, please mind the commits. There's one `fix: support python 3.7`. The _fix_ commits appear in the changelog, and that wouldn't make sense because python 3.7 was already supported before this change.

Well, if you prefer, just forget about commits and I'll just squash on merge, rewording it. I wanted to tell just in case you were wondering why I usually squash PRs... 😆 I know you're the kind of person that likes to understand even these kind of details. 😉

docs/configuring.md
sisp: Like this? 0c3e859f0bdd5ebec5b19d6792a5feca27a70fc7

sisp:
> Besides, please mind the commits. There's one `fix: support python 3.7`. The _fix_ commits appear in the changelog, and that wouldn't make sense because python 3.7 was already supported before this change.
> 
> Well, if you prefer, just forget about commits and I'll just squash on merge, rewording it. I wanted to tell just in case you were wondering why I usually squash PRs... laughing I know you're the kind of person that likes to understand even these kind of details. wink

Haha, yes, indeed I do. But my logic has been this: The first commit should have a proper change type (here, `feat:`). All subsequent commits in this PR will get change types according to their change relative to the first commit in the PR. So because I messed up Python 3.7 support _in this PR_, I fixed it. I always assume such PRs to be squashed, so the first commit message (or rather PR title) is used for the final commit message in the `master` branch. Is this wrong? :thinking:

yajo:
No, it's not wrong, don't worry!

It's just simpler for me to hit "Rebase and merge" than to rewrite the squash message.

But it's really the most irrelevant thing, so don't worry about that. Thanks for the PR!

sisp:
I'm still curious to understand the change type ting better. How would would you do it?
1. Amend the first commit and force-push?
1. Always use the change type of the first commit for any subsequent commit in the PR independent of its actual change type?
1. Something else?

:thinking:

yajo:
For me, the rule is: does the commit type appear in the release notes and
is it meaningful there? Then add a commit. Otherwise, squash.

sisp:
It's still not clear to me ... :confused: What would you have done, e.g., for the [`fix: support Python 3.7`](https://github.com/copier-org/copier/pull/1171/commits/19fe0c8f76e6bb1e5e3defd4defbc865e869dcc2) commit? Would you have amended the first commit? Or would you have used a different change type (`feat:`)?

yajo:
Does the commit type (fix) appear in the changelog? Yes.

Does it make sense there? No, because python 3.7 was already supported
before.

Then, I'd squash it.

sisp:
Sorry to be so persistent. With squashing, there's no problem then. But it seems you'd prefer no squashing. Then, the only alternative is to amend the first commit after a review that requests changes. Somehow I don't think that's a good workflow because it's less clear how a PR has evolved until it got merged. And committing review suggestions via the GitHub UI also creates new commits. I think squash-merging is the way to go for almost all PRs, and then subsequent commits don't contribute to the changelog no matter what their change types are.

yajo:
Haha yes indeed.

I do that most of the time.

In the case of this PR, it made sense to have both feat commits in the
changelog, but the fix commit didn't.

In any case, this really isn't an extremely important problem. Don't worry
about it too much and contribute the way it's easiest for you. 😊

pawamoy:
Personally I use a fixup commit instead of amending: `git commit --fixup SHA_OF_COMMIT_TO_FIX`.
This way we still have the commit history in the branch / PR, and I squash the commits when merging. It also makes it very easy to squash them locally: `git rebase -i --autosquash master` (or something like `HEAD~N` instead of master, where N is a number of commits).

PR #1179, opened 2023-06-10
feat: add `--trust` as a less scary alternative to `--UNSAFE`

yajo:
See rationale in https://github.com/copier-org/copier/issues/1137#issuecomment-1579304713.

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -144,7 +144,7 @@ class _Subcommand(cli.Application):
         help="Use prereleases to compare template VCS tags.",
     )
     unsafe = cli.Flag(
-        ["--UNSAFE"],
+        ["--UNSAFE", "--trust"],
         help=(
             "Allow templates with unsafe features (Jinja extensions, migrations, tasks)"
         ),
--- a/tests/test_unsafe.py
+++ b/tests/test_unsafe.py
@@ -1,5 +1,5 @@
 from contextlib import nullcontext as does_not_raise
-from typing import ContextManager
+from typing import ContextManager, Union
 
 import pytest
 import yaml
--- a/tests/test_unsafe.py
+++ b/tests/test_unsafe.py
@@ -318,7 +318,7 @@ def test_update(
 def test_update_cli(
     tmp_path_factory: pytest.TempPathFactory,
     capsys: pytest.CaptureFixture[str],
-    unsafe: bool,
+    unsafe: Union[bool, str],
 ) -> None:
     src, dst = map(tmp_path_factory.mktemp, ["src", "dst"])
     unsafe_args = [unsafe] if unsafe else []
--- a/tests/test_unsafe.py
+++ b/tests/test_unsafe.py
@@ -314,13 +314,14 @@ def test_update(
         run_update(dst, overwrite=True, unsafe=unsafe)
 
 
-@pytest.mark.parametrize("unsafe", [False, True])
+@pytest.mark.parametrize("unsafe", [False, "--trust", "--UNSAFE"])
 def test_update_cli(
     tmp_path_factory: pytest.TempPathFactory,
     capsys: pytest.CaptureFixture[str],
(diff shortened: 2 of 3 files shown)


tests/test_unsafe.py
```diff
@@ -335,7 +335,7 @@ def test_update_cli(
         git("tag", "v1")
 
     _, retcode = CopierApp.run(
-        ["copier", "copy", "--UNSAFE", str(src), str(dst)],
+        ["copier", "copy", "--trusted", str(src), str(dst)],
```
sisp, line 338: How about parametrizing this test to check that both `--UNSAFE` and `--trusted` both work?

copier/cli.py
```diff
@@ -144,7 +144,7 @@ class _Subcommand(cli.Application):
         help="Use prereleases to compare template VCS tags.",
     )
     unsafe = cli.Flag(
-        ["--UNSAFE"],
+        ["--UNSAFE", "--trusted"],
```
sisp, line 147: Or `--trust` instead of `--trusted`? :thinking: "Unsafe" is an adjective, "trusted" as well, but many other switches/flags are in imperative mood. Not sure. WDYT, @yajo @pawamoy?

sisp (commented):
Some minor suggestions or ideas from my end, see inline comments.

copier/cli.py
pawamoy: +1 for `trust`. Shorter, easier to write and pronounce. It better conveys the fact that you want to trust the template.

tests/test_unsafe.py
yajo: I don't think it's needed, because the same test uses `--UNSAFE` a few lines later.

copier/cli.py
yajo: Thanks! Done.

tests/test_unsafe.py
sisp: Hm, I think it would make the test clearer (one might wonder, is there a particular reason why it's first `--trusted`/`--trust` and later `--UNSAFE`) and possibly add a tiny bit of safety.

yajo: Done.

tests/test_unsafe.py
```diff
@@ -314,13 +314,14 @@ def test_update(
         run_update(dst, overwrite=True, unsafe=unsafe)
 
 
-@pytest.mark.parametrize("unsafe", [False, True])
+@pytest.mark.parametrize("unsafe", [False, "--trust", "--UNSAFE"])
 def test_update_cli(
     tmp_path_factory: pytest.TempPathFactory,
     capsys: pytest.CaptureFixture[str],
     unsafe: bool,
```
sisp, line 321: This type hint needs an update now:

```suggestion
    unsafe: Union[bool, str],
```

And `Union` needs to be imported.

sisp (changes requested):
Just a minor request, looks great otherwise. 👌

tests/test_unsafe.py
pawamoy: How do you feel about postponed annotations @sisp @yajo? Allowing to do `bool | str`.

sisp: I find `bool | str` more readable, and it's shorter. Is there any disadvantage because of the `__future__` import?

pawamoy: As long as you don't declare Pydantic models inside of functions, there will be no issues (both with or without `__future__` annotations).

sisp: I'd vote +1 for converting the code base to annotations, we should do a batch conversion in a separate PR though.

WDYT, @yajo?

sisp (changes requested):
The only remaining problem is the missing import of the `Union` type now.

sisp (approved):
Nice :ok_hand:

PR #1181, opened 2023-06-12
docs: add commit message and PR merge guidelines

yajo:
Now that we have more maintainers, I think this will be helpful.

cc @copier-org/maintainers


pawamoy (approved):
:+1:

PR #1188, opened 2023-06-14
feat: add OS identifier to render context

sisp:
I've added a new property `os` to the `_copier_conf` variable that is provided to the render context which holds operating system identifiers for Linux (`linux`), macOS (`macos`) and Windows (`windows`). Any other operating system type is represented by `None`.

Knowing the operating system is useful, e.g., for rendering OS-specific tasks.

That said, knowing the operating system might not be enough, e.g. it might be necessary to even differentiate between Windows CMD and Windows PowerShell. We might consider providing also shell information, e.g. using [`shellingham`](https://github.com/sarugaku/shellingham). But I'd leave this to a potential follow-up PR.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -34,7 +34,7 @@ from .errors import (
 )
 from .subproject import Subproject
 from .template import Task, Template
-from .tools import Style, TemporaryDirectory, printf, readlink
+from .tools import OS, Style, TemporaryDirectory, printf, readlink
 from .types import (
     MISSING,
     AnyByStrDict,
@@ -298,6 +298,7 @@ class Worker:
                 "src_path": self.template.local_abspath,
                 "vcs_ref_hash": self.template.commit_hash,
                 "sep": os.sep,
+                "os": OS,
             }
         )
 
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -2,6 +2,7 @@
 
 import errno
 import os
+import platform
 import shutil
 import stat
 import sys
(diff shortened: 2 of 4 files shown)


sisp:
I'm a bit puzzled why the tests are failing on Windows. :confused:

```text
FAILED tests/test_tasks.py::test_os_specific_tasks[Linux-linux.txt] - PermissionError: [Errno 13] Permission denied: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\popen-gw0\\dst53'
FAILED tests/test_tasks.py::test_os_specific_tasks[Darwin-macos.txt] - PermissionError: [Errno 13] Permission denied: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\popen-gw0\\dst54'
FAILED tests/test_tasks.py::test_os_specific_tasks[Java-unsupported.txt] - PermissionError: [Errno 13] Permission denied: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\popen-gw0\\dst55'
```

Surprisingly, the `("Windows", "windows.txt")` case isn't failing. This feels like the monkeypatching isn't working on Windows. But I can't seem to figure out what's wrong. Does anybody have an idea?

pawamoy:
Maybe it's because you're trying to patch an object within an object of the main module? What about patching `platform` directly, with a fake class that only has a `system` method to which you assign the lambda?

sisp:
> What about patching `platform` directly, with a fake class that only has a `system` method to which you assign the lambda?

Wouldn't this be dangerous if other functions from `platform` were used elsewhere in Copier?

sisp:
I found the problem:

https://github.com/copier-org/copier/blob/cadf32687e225e7d5d3a7d1ce5e17d3cd3d13b6c/copier/main.py#L395

This hack doesn't work when I monkeypatch `platform.system()`.

yajo (approved):
Thanks, looks good. It has conflicts though.

Issue #368, opened 2021-03-11
Unify logging


Yajo:
There are many logging system around the codebase. Makes it confusing:

- Old one https://github.com/copier-org/copier/blob/76d6c2f502db149e500bd91dc93f1871d6a9fc38/copier/tools.py#L60-L88
- plumbum one https://github.com/copier-org/copier/blob/b3190c2abffb04d5f4b21fa71a51bf4036baad5a/copier/main.py#L605-L609
- Native warnings https://github.com/copier-org/copier/blob/d86c88d051a8e9127499a7d8eae1bb84572f4780/copier/template.py#L114-L117
- Plumbum's ask, as opposed to Questionary's https://github.com/copier-org/copier/blob/d86c88d051a8e9127499a7d8eae1bb84572f4780/copier/main.py#L251

Gotta pick one and use only it. Possibly Questionary for interactive output, and https://docs.python.org/3/library/logging.html with `captureWarnings` for informational output.

karolzlot:
@Yajo 

It should be written which file is processed before processing it, not after.

If I have error then I have to guess which file is next.

example error which I am now debugging:

```
    create  .
    create  .dependabot
    create  .dependabot/config.yml
    create  .en v copy
    create  .env
    create  .gitignore
    create  LICENSE
    create  README.md
    create  Banana
    create  Banana/.prettierignore
    create  Banana/backend
    create  Banana/backend/alembic.ini
    create  Banana/backend/app
    create  Banana/backend/app/alembic
    create  Banana/backend/app/alembic/env.py
    create  Banana/backend/app/alembic/README
    create  Banana/backend/app/alembic/script.py.mako
    create  Banana/backend/app/alembic/versions
    create  Banana/backend/app/alembic/versions/91979b40eb38_create_users_table.py
    create  Banana/backend/app/alembic/__init__.py
    create  Banana/backend/app/alembic.ini
    create  Banana/backend/app/api
    create  Banana/backend/app/api/api_v1
    create  Banana/backend/app/api/api_v1/routers
    create  Banana/backend/app/api/api_v1/routers/auth.py
    create  Banana/backend/app/api/api_v1/routers/tests
    create  Banana/backend/app/api/api_v1/routers/tests/test_auth.py
    create  Banana/backend/app/api/api_v1/routers/tests/test_users.py
    create  Banana/backend/app/api/api_v1/routers/tests/__init__.py
    create  Banana/backend/app/api/api_v1/routers/users.py
[...]

PR #1194, opened 2023-06-15
feat: add support for pre-copy and post-copy messages

### LINKED ISSUE #723: Add intro message
I would like a clean way to display an intro message before the input questions.
Right now I'm doing this by adding it to the first question.
There are a couple og potentional problems with this:
- An argument gets added before the first argument, causing the intro message to be display after the first question
- It is not possible to programmatically use the intro message, e.g. for a UI.

I suggest adding an `intro_message` configuration item.
Or maybe `pre_message`, `post_success_message`, `post_fail_message`?

sisp:
I've added support for pre-copy and post-copy messages that can be written in plain text or markdown. Messages can be written inline or included via Jinja includes, and messages are rendered.

I'll work on pre-update and post-update messages in a separate PR once we've clarified whether the implementation in this PR is good.

Closes #723.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -243,6 +243,10 @@ class Worker:
         if features:
             raise UnsafeTemplateError(sorted(features))
 
+    def _print_message(self, message: str) -> None:
+        if message and not self.quiet:
+            print(self._render_string(message), file=sys.stderr)
+
     def _answers_to_remember(self) -> Mapping:
         """Get only answers that will be remembered in the copier answers file."""
         # All internal values must appear first
@@ -727,6 +731,7 @@ class Worker:
         See [generating a project][generating-a-project].
         """
         self._check_unsafe("copy")
+        self._print_message(self.template.message_before_copy)
         self._ask()
         was_existing = self.subproject.local_abspath.exists()
         src_abspath = self.template_copy_root
@@ -746,6 +751,7 @@ class Worker:
             if not was_existing and self.cleanup_on_error:
                 rmtree(self.subproject.local_abspath)
             raise
+        self._print_message(self.template.message_after_copy)
         if not self.quiet:
             # TODO Unify printing tools
             print("")  # padding space
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -996,6 +996,64 @@ on them, so they are always installed when Copier is installed.
     [other Jinja2 topics](https://github.com/search?q=jinja&type=topics), or
     [on PyPI using the jinja + extension keywords](https://pypi.org/search/?q=jinja+extension).
 
+### `message_after_copy`
+
+-   Format: `str`
+-   CLI flags: N/A
+-   Default value: `""`
+
+A message to be printed after [generating](../generating) or
+[regenerating](../generating#regenerating-a-project) a project _successfully_.
+
+If the message contains Jinja code, it will be rendered with the same context as the
+rest of the template. A [Jinja include](#importing-jinja-templates-and-macros)
+expression may be used to import a message from a file.
+
+The message is suppressed when Copier is run in [quiet mode](#quiet).
+
+!!! example
+
+    ```yaml title="copier.yml"
+    project_name:
+        type: str
+        help: An awesome project needs an awesome name. Tell me yours.
+
+    _message_after_copy: |
+        Your project "{{ project_name }}" has been created successfully!
+
+        Next steps:
+
+        1. Change directory to the project root:
+
+           $ cd {{ _copier_conf.dst_path }}
+
+        2. Read "CONTRIBUING.md" and start coding.
+    ```
+
+### `message_before_copy`
+
+-   Format: `str`
+-   CLI flags: N/A
+-   Default value: `""`
+
+Like [`message_after_copy`](#message_after_copy) but printed _before_
+[generating](../generating) or [regenerating](../generating#regenerating-a-project) a
+project.
+
+!!! example
+
+    ```yaml title="copier.yml"
+    project_name:
+        type: str
+        help: An awesome project needs an awesome name. Tell me yours.
+
+    _message_before_copy: |
+        Thanks for generating a project using our template.
+
+        You'll be asked a series of questions whose answers will be used to
+        generate a tailored project for you.
+    ```
+
 ### `migrations`
 
 -   Format: `List[dict]`
--- a/copier/template.py
+++ b/copier/template.py
@@ -322,6 +322,16 @@ class Template:
         return tuple(self.config_data.get("jinja_extensions", ()))
 
     @cached_property
+    def message_after_copy(self) -> str:
+        """Get message to print after copy action specified in the template."""
+        return self.config_data.get("message_after_copy", "")
+
+    @cached_property
+    def message_before_copy(self) -> str:
+        """Get message to print before copy action specified in the template."""
+        return self.config_data.get("message_before_copy", "")
+
+    @cached_property
     def metadata(self) -> AnyByStrDict:
         """Get template metadata.
 
--- a/tests/test_output.py
+++ b/tests/test_output.py
@@ -1,12 +1,13 @@
 import re
 from pathlib import Path
 
+import pexpect
 import pytest
 
 from copier.errors import InvalidTypeError
-from copier.main import run_copy
+from copier.main import run_copy, run_recopy
 
-from .helpers import build_file_tree, render
+from .helpers import COPIER_PATH, Spawn, build_file_tree, expect_prompt, render
 
 
 def test_output(capsys: pytest.CaptureFixture[str], tmp_path: Path) -> None:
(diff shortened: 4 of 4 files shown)


pawamoy (approved):
Nice!

If we print these messages through Rich, I think we should do it for questions description/help too, for consistency. Users might be tempted to use Markdown in questions description/help and would be surprised to see that it does not get rendered the same way as pre/post messages. This can of course be done in another PR.

sisp:
Is that possible? The help text is printed through Questionary. It would be great though!

pawamoy:
I don't know :smile:

yajo (changes requested):
There's something wrong here. I've tested the test interactively and this is what happens:

```sh
➤ copier copy '/tmp/pytest-of-yajo/pytest-0/popen-gw0/src0' /tmp/b
🎤 project_name
   a
Thank you for using our template                                                                                                                                                                           

 • You're on linux                                                                                                                                                                                         
 • Let's generate your project                                                                                                                                                                             

    create  .

Project a successfully created                                                                                                                                                                             

Next steps:                                                                                                                                                                                                

 • cd /tmp/b                                                                                                                                                                                               
 • Start coding                                                                                                                                                                                            
```

The welcome message appears *after* the questionary. It should be *before*, right?

yajo:
> If we print these messages through Rich, I think we should do it for questions description/help too, for consistency.

Yikes, yet another method of printing stuff... 🤦🏼‍♂️ #368 gets bigger.

sisp:
Oh, good catch about the printing after the question. It should be before the question, of course. Strange, I'll look into it. It seems I'll need to change the tests to use a tui, was hoping to avoid that.

copier/main.py
```diff
@@ -241,6 +243,12 @@ def _check_unsafe(self, mode: Literal["copy", "update"]) -> None:
         if features:
             raise UnsafeTemplateError(sorted(features))
 
+    def _print_message(self, message: str) -> None:
+        if message and not self.quiet:
+            Console(stderr=True).print(
+                Markdown(self._render_string(message), hyperlinks=True)
```
sisp, line 249: The call to `self._render_string(...)` starts the questionnaire, so the `message_before_copy` is rendered _after_ answering the questions. :thinking:

sisp: It's a side effect of using an [impure property for prompting](https://github.com/copier-org/copier/blob/1b351bc8560d19d70e927cdb2fafb4f7ce41092d/copier/main.py#L411-L463) instead of a method that records the answers in the state of the `Worker` class.

sisp: I've submitted #1198 to trigger the questionnaire imperatively and, thus, get explicit control, so that the render context can be accessed prior to the questionnaire, of course without containing any answers then.

sisp:
I've extended the tests to also check interactive prompting and make sure the message before copy is shown before the prompt and the message after copy is shown after the prompt.

I'd appreciate another critical review. :pray:

copier/main.py
```diff
@@ -25,6 +25,8 @@
 from pydantic.dataclasses import dataclass
 from pydantic.json import pydantic_encoder
 from questionary import unsafe_prompt
+from rich.console import Console
+from rich.markdown import Markdown
```
yajo, line 29: I'm double-thinking about this.

I really like it that we can render markdown in the terminal. However, I'm worried about the lack of consistency.

I can imagine somebody writing an intro message in markdown and a question message also in markdown and wondering... why one works and the other doesn't?

Do you think it would be good to defer markdown support until we can add it consistently?

Other than that, the PR is great.

sisp: I've looked into supporting Markdown in `questionary` help text through `rich`, but it seems this is currently not possible because `questionary` applies its own styles through `prompt-toolkit` to the message. So when I render it first with `rich` and pass it to `questionary`, then the formatting remains but the styling gets overwritten by `questionary`/`prompt-toolkit`. For instance, currently the help text is printed in bold font, and then also the rendered Markdown becomes bold font. Disabling `questionary`/`prompt-toolkit` styles for help text seems to be impossible at the moment.

I also thought about replacing `rich` by `prompt-toolkit` to avoid introducing yet another printing library, but `prompt-toolkit` doesn't support Markdown rendering. `questionary` is heavily based on `prompt-toolkit`, so converging on a common toolstack would mean we shouldn't use `rich`.

Perhaps we need to answer a few more fundamental questions:

1. How important is Markdown rendering for pre/post copy messages?
2. How important is Markdown rendering for help text?
3. How important is having consistent rendering capabilities? I.e., would we only want Markdown rendering at all when all relevant texts support it?
4. Is it possible to achieve a unified look and feel when using `prompt-toolkit` and `rich` in different places?

We could defer answering those questions by simply removing Markdown support for pre/post copy messages now (and, thus, remove `rich` again), so we can merge this PR more quickly.

WDYT?

lhupfeldt:
I think this could be made more general by replacing the `message_before_copy` with just `message` and allowing messages mixed with questions (and `vars` when we get `when: false` back). Of cause a message may only refer context already resolved. That way grouping of questions with message will be possible.
Is `message_after_copy` only for the `copy` operation or will be printed after all operations? I think the default should be always, and if necessary a `when: operation=update/copy/...` could be added (later). In that case a more generic name, like `post_message` would be better.

sisp:
I'm afraid I don't understand your proposal. Could you provide an example of how you think `copier.yml` would look?

lhupfeldt:
> I'm afraid I don't understand your proposal. Could you provide an example of how you think `copier.yml` would look?

Something like:
```
_tasks:
  - exec: "./doit.sh"
    message: "We did it!"

_message: "Overall template intro."

_message: "Your Personal info"

name: "Your full name"
email: "Your email"

_message: "Family info"

mother: "Your mother's name"
father: "Your father's name"

_message:
  text: "Ouch, your name is 'dummy'?"
  when: "{{ name == 'dummy' }}"

_message:
  text: "Before tasks first task"
  when: has_tasks

# Here we run tasks, so there may be task message output, or output printed from tasks.
# A `message` method could also be made available to take advantage of standardized formatting of messages printed from task.

_message:
  text: "Yeah"
  when: "{{ copier_finished and copier_success copier_action == 'update' }}"

_message:
  text: "Congratulation"
  when: "{{ copier_finished and copier_success }}"

_message:
  text: "Oops!"
  when: "{{ copier_finished and copier_error }}"

```

I think that a single `_message` with a `when` condition is both more flexible and more user friendly. 
The example is just some ideas, don't take it literally.
Task messages are just an extra feature.

sisp:
This isn't valid YAML according to the spec (https://yaml.org/spec/1.2.2/#mapping):

> The content of a mapping node is an unordered set of key/value node pairs, with the restriction that each of the keys is [unique](https://yaml.org/spec/1.2.2/#node-comparison).

PyYAML overwrites the result with the last key.

lhupfeldt:
> This isn't valid YAML according to the spec (https://yaml.org/spec/1.2.2/#mapping):

You are right.
I guess the only way to make a generic messaging mechanism similar to my suggestion would be to handle `_message_X` where `X` can be anything.

copier/main.py
sisp: I've decided to remove `rich` and Markdown support again until we have a proper solution to render Markdown consistently in other places as well.

yajo: I think it's the best solution for the moment.

yajo:
Apart from being invalid yaml, you already have the question help text to print whatever you want, so that proposal doesn't add much value IMHO.

docs/configuring.md
```diff
@@ -996,6 +996,68 @@ on them, so they are always installed when Copier is installed.
...
+-   Default value: `""`
+
+A message to be printed before [project creation](../generating).
+
+If the message contains Jinja code, it will be rendered with the initial context which
+does not yet contain any answers to questions. A
+[Jinja include](#importing-jinja-templates-and-macros) expression may be used to import
+a message from a file.
+
+The message is suppressed when Copier is run in [quiet mode](#quiet).
```
yajo, line 1045: I think you could refer to one section from the other, to DRY.

yajo (approved):
Really good, thanks! 

Just one minor suggestion for the docs part. Non blocking.

lhupfeldt:
It is still not clear to me if this only before "copy", or also before
"update", ...

ons. 19. jul. 2023 06.05 skrev Jairo Llopis ***@***.***>:

docs/configuring.md
sisp: Done.

sisp:
@lhupfeldt Only before _copy_, not before update. But you got me thinking and it turns out this message is printed also for _re_copy. I'll extend the docs and add some tests.

sisp:
I've extended the tests to check the message printing also for recopying and edited the docs accordingly.

Since those changes don't alter the PR in general, I'll take the liberty of merging now without further review. I hope that's okay, @copier-org/maintainers.

lhupfeldt:
I know this already merged, but I still think that it would be better to
have a 'when' condition to limit which action you want it printed before.
It seems more natural to me the the default should be before all actions,
including 'update'.

ons. 19. jul. 2023 12.13 skrev Sigurd Spieckermann ***@***.***

sisp:
To avoid repetition, you could rely on YAML anchors and aliases to print the same message before copy and update (assuming `_message_before_update` was already available):

```yaml
_message_before_copy: &message_before |-
  Hello, thanks for using our template!

_message_before_update: *message_before
```

This way, it's explicit when the message will be printed.

lhupfeldt:
Correct, but it would still be more verbose, and if new actions are added
it will become even more verbose. Having a 'when' clause with the
possibility to ask which action was requested seems to fit well with the
existing design.
The possibility to query the action is something I would really like for
questions as well.

tor. 20. jul. 2023 13.47 skrev Sigurd Spieckermann ***@***.***

yajo:
The benefit of having dedicated keys is that it's easier to compose with
!include

PR #1198, opened 2023-06-16
refactor: request answers imperatively instead of implicitly via impure property

sisp:
I've refactored the implementation of how the answers of the questionnaire are requested and recorded. Before, the questionnaire was implicitly triggered via the impure cached property `Worker(...).answers` which made it difficult to understand when exactly the prompting takes place. For instance, this results in [printing a welcome message _after completing_ instead of _before beginning_ the questionnaire](https://github.com/copier-org/copier/pull/1194#pullrequestreview-1481962789) as this message gets rendered with Jinja where the [render context includes the answers retrieved from the impure property](https://github.com/copier-org/copier/blob/1b351bc8560d19d70e927cdb2fafb4f7ce41092d/copier/main.py#L307) which triggers the prompting.

I've changed the implementation by triggering the prompting imperatively and recording the answers in an instance attribute of the `Worker` class which gives explicit control over triggering the prompting and will allow rendering the welcome message with a render context that does not yet contain the answers from the questionnaire.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -202,6 +202,8 @@ class Worker:
     context_lines: PositiveInt = 3
     unsafe: bool = False
 
+    answers: AnswersMap = field(default_factory=AnswersMap, init=False)
+
     def __enter__(self):
         """Allow using worker as a context manager."""
         return self
@@ -408,12 +410,8 @@ class Worker:
             return True
         return self._solve_render_conflict(dst_relpath)
 
-    @cached_property
-    def answers(self) -> AnswersMap:
-        """Container of all answers to the questionary.
-
-        It asks the user the 1st time it is called, if running interactively.
-        """
+    def _ask(self) -> None:
+        """Ask the questions of the questionary and record their answers."""
         result = AnswersMap(
             default=self.template.default_answers,
             user_defaults=self.user_defaults,
@@ -460,7 +458,7 @@ class Worker:
                 raise CopierAnswersInterrupt(result, question, self.template) from err
             result.user[var_name] = new_answer
 
-        return result
+        self.answers = result
 
     @cached_property
     def answers_relpath(self) -> Path:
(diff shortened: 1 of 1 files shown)


pawamoy (approved):
LGTM :+1:

I'm fighting with this in my own projects: cached properties are very convenient but a great way to shoot yourself in the foot.

copier/main.py
```diff
@@ -202,6 +202,8 @@ class Worker:
     context_lines: PositiveInt = 3
     unsafe: bool = False
 
+    answers: AnswersMap = field(default_factory=AnswersMap, init=False)
```
yajo, line 205: I'd rename it now, to avoid pydantic making it a parameter: 

```suggestion
    _answers: AnswersMap = field(default_factory=AnswersMap, init=False)
```

yajo (changes requested):

copier/main.py
sisp: It's not a Pydantic parameter because Pydantic respects `field(init=False)`:

```python
from dataclasses import field


@dataclass
class A:
    x: int = field(default=1, init=False)


A(x=2)
```

raises an error:

```
  File "<stdin>", line 1, in <module>
  File ".venv/lib/python3.9/site-packages/pydantic/dataclasses.py", line 322, in new_init
    handle_extra_init(self, *args, **kwargs)
  File ".venv/lib/python3.9/site-packages/pydantic/dataclasses.py", line 286, in handle_extra_init
    init(self, *args, **{k: v for k, v in kwargs.items() if k in self.__dataclass_fields__})
TypeError: __init__() got an unexpected keyword argument 'x'
```

yajo: Ah! I overlooked that. Thanks

yajo (approved):

copier/main.py
sisp: No problem 🙂

PR #1199, opened 2023-06-16
refactor: assign immutable default value without dataclass `field()` function

sisp:
Just a minor refactoring which simplifies setting an immutable default value for a dataclass attribute without calling the `field()` function.


pawamoy (approved):

sisp:
Rebased to get `install-nix-action` v22 which should let `CI / flake-check (macos-latest)` pass now.

PR #1200, opened 2023-06-16
refactor(typing): use `Literal` type hint for `conflict` attribute

sisp:
I've made the type hint for the `conflict` attribute more precise by using `Literal["rej", "inline"]` instead of `str`.

@yajo Is `refactor(typing)` the correct change type in your opinion? It seemed the most appropriate to me, but I couldn't find a guideline for committing only typing changes.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -198,7 +198,7 @@ class Worker:
     overwrite: bool = False
     pretend: bool = False
     quiet: bool = False
-    conflict: str = "inline"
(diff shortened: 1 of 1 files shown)


pawamoy (approved):

yajo:
I relabeled the commit to the build category. Refactor commits appear in the changelog, and I don't think this one deserves such a honor. 😜

Issue #1205, opened 2023-06-18
Support hidden variables without user prompt/input


lhupfeldt:
### Actual Situation

In version 7 you could use `when: false` to set temporary variables without prompting the user. This feature is not available i version 8, see  #1175.

### Desired Situation

Functionality similar to `when: false` is restored

### Proposed solution

Introduce a new flag e.g. `hidden: true`
There must be a default value set in config.
Values will not be stored so will always be calculated based on static or Jinja template defaults.

sisp:
What do you think about #1203?

PR #1206, opened 2023-06-19
fix: restore behavior for skipped question with default value

sisp:
I've partially reverted #1126 and restored the behavior for a skipped question with a default value such that its answer is recorded in the answers file. This behavior is well-documented

> If a question is skipped, its answer will be:
>
> - The default value, if you're generating the project for the 1st time.
> - The last answer recorded, if you're updating the project.

and was incompatibly changed in #1126.

Resolves #1204. See https://github.com/copier-org/copier/discussions/1204#discussioncomment-6211104 for a detailed explanation of the history that led to this regression.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -415,7 +415,6 @@ class Worker:
         It asks the user the 1st time it is called, if running interactively.
         """
         result = AnswersMap(
-            default=self.template.default_answers,
             user_defaults=self.user_defaults,
             init=self.data,
             last=self.subproject.last_answers,
@@ -429,10 +428,10 @@ class Worker:
                 var_name=var_name,
                 **details,
             )
-            # Skip a question when the skip condition is met and remove any data
-            # from the answers map, so no answer for this question is recorded
-            # in the answers file.
-            if not question.get_when():
+            # Skip a question when the skip condition is met and it has no
+            # default value, and remove any data from the answers map, so no
+            # answer for this question is recorded in the answers file.
+            if not question.get_when() and question.default is MISSING:
                 result.remove(var_name)
                 continue
             if var_name in result.init:
@@ -454,7 +453,8 @@ class Worker:
                         raise ValueError(f'Question "{var_name}" is required')
                 else:
                     new_answer = unsafe_prompt(
-                        [question.get_questionary_structure()], answers=result.combined
+                        [question.get_questionary_structure()],
+                        answers={question.var_name: question.get_default()},
                     )[question.var_name]
             except KeyboardInterrupt as err:
                 raise CopierAnswersInterrupt(result, question, self.template) from err
--- a/copier/template.py
+++ b/copier/template.py
@@ -282,11 +282,6 @@ class Template:
         return result
 
     @cached_property
-    def default_answers(self) -> AnyByStrDict:
-        """Get default answers for template's questions."""
-        return {key: value.get("default") for key, value in self.questions_data.items()}
-
-    @cached_property
     def envops(self) -> Mapping:
         """Get the Jinja configuration specified in the template, or default values.
 
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -103,11 +103,6 @@ class AnswersMap:
         last:
             Data from [the answers file][the-copier-answersyml-file].
 
-        default:
-            Default data from the template.
-
-            See [copier.template.Template.default_answers][].
-
         user_defaults:
             Default data from the user e.g. previously completed and restored data.
 
@@ -123,7 +118,6 @@ class AnswersMap:
     metadata: AnyByStrDict = field(default_factory=dict)
     last: AnyByStrDict = field(default_factory=dict)
     user_defaults: AnyByStrDict = field(default_factory=dict)
-    default: AnyByStrDict = field(default_factory=dict)
 
     @property
     def combined(self) -> Mapping[str, Any]:
@@ -135,7 +129,6 @@ class AnswersMap:
                 self.metadata,
                 self.last,
                 self.user_defaults,
-                self.default,
                 DEFAULT_DATA,
             )
         )
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -182,10 +182,12 @@ Supported keys:
 
     This is most useful when [templated](#prompt-templating).
 
-    If a question is skipped, its answer will be:
+    If a question _with_ a default value is skipped, its answer will be:
 
     -   The default value, if you're generating the project for the 1st time.
     -   The last answer recorded, if you're updating the project.
+
+    If a question _without_ a default value is skipped, no answer will be recorded.
 
     !!! example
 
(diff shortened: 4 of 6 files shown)


pawamoy:
In all transparency, I think @yajo would simply prefer declaring that `when` and `default` are incompatible.

With this change, when a template changes the default value, but `when` keeps evaluating to false, users won't be able to get the new default, not even by running Copier interactively, answering all questions again, since the question will be skipped. They'll have to update it in the answers file directly, which is a bad idea. Or they'll have to recopy the project (which will generate noise in the git diff). Or pass the new default as command line data (which requires them to *know* the default before hand).

Don't get me wrong, I've been wanting unprompted/computed values since the beginning. But I'm wondering if and how the tradeoff will impact me. Honestly, I can handle myself and my own templates. But when the template's writer and its users are not the same person, it's a bit more problematic. WDYT @sisp?

sisp:
I don't think `when` and `default` are incompatible because you might want to provide a default value when `when` evaluates to `true`. But when `when` evaluates to `false`, then the answer/variable would not exist; this is the current behavior since #1126 has been merged.

@yajo's comment https://github.com/copier-org/copier/discussions/678#discussioncomment-2822746 sounds to me like he approves the behavior prior to #1126 _when `when` is not hardcoded to `false`. That's why I thought we should restore it. Omitting an answer to a skipped question _without a default value_ when `when` evaluates to `false` is definitely correct because `None` carrying the semantics of "missing" is simply wrong.

If we don't restore the old behavior for skipped questions _with default values_, then #1126 should be marked as a breaking change in the changelog and we'd need to update the docs to explain the new behavior. I'm fine with this as well because I agree with what you said about not being able to change the implicit answer even when the (evaluated) default value changes. This would break @tlambert03's use case though, assuming he wants the implicit answer to be locked &ndash; but I doubt it TBH. Then, he'd need truly computed variables which can be done through [Jinja import/include](https://copier.readthedocs.io/en/v8.0.0/configuring/#importing-jinja-templates-and-macros) for computing string variables. Alternatively, I suggested #1203, but there has been no feedback yet.

So, in summary, personally I'm fine with closing this PR if we decide that #1126 was in fact a breaking change and update the docs and changelog.

tlambert03:
> In all transparency, I think @yajo would simply prefer declaring that when and default are incompatible.

With great appreciation for all that you do... to be honest, this feature was pretty much the main reason I switched to copier from cookiecutter :joy:, and though I do like copier, its removal would make me very much consider returning to cookiecutter for its stability. (or fork the project for my users). 

I understand that it makes it easier to maintain to reduce the feature set, but that would be a real loss in my opinion.

> With this change, when a template changes the default value, but when keeps evaluating to false, users won't be able to get the new default.  They'll have to update it in the answers file directly

Personally, I'm 100% fine with this compromise, since I think my target audience will use copier *far* more often for project creation, and less so for project updating.  I do recognize that updating is a "killer app" for copier (the "cruft" of cookiecutter) and having a condition with surprising behavior is never desirable.  However, if it's a choice between not offering the feature at all, and offering it with a big "**warning using `when` and `default` may lead to unexpected behavior during updates!**", I think a well documented warning would be way better.

> So, in summary, personally I'm fine with closing this PR if we decide that https://github.com/copier-org/copier/pull/1126 was in fact a breaking change and update the docs and changelog.

That would be 😭

pawamoy:
I'm personally in favor of merging this, I was just trying to be the most transparent as possible since I'm biased towards the feature :smile:

pawamoy:
By the way, an alternative path would be to make this configurable: `copier update --use-new-defaults-for-skipped-questions` (I'm sure I can find an even longer option name).

sisp:
I wouldn't go with a new flag, this needs even extra code for the different behaviors. I think it's okay to restore the old behavior for questions with a default value and add a big warning to the docs that this should not be used for deriving values.

yajo:
I'm evaluating the implications of going back this path, so please don't get angry at me. 😀 I just want to make sure we're going the correct path.

Thanks for your kind words and for keeping this discussion so nice!

The fact that a question that is explicitly skipped is missing in the context and the answers file is a good thing IMHO. It makes things predictable. Of course it's a breaking change, but it was something that didn't make much sense after all.

You have plenty of ways to solve this in jinja. You can declare a variable, use the default filter, import a macro or namespace... a lot!

Besides, there's an extension to declare variables.

So... isn't this just a consequence of using a feature for an unsupported purpose? 

> Personally, I'm 100% fine with this compromise, since I think my target audience will use copier _far_ more often for project creation, and less so for project updating.

Ok, but what would you do when this really happens?

tlambert03:
> Ok, but what would you do when this really happens?

yep, definitely a fair/good question. 
If moving all the logic to a jinja template is the _only_ way you can see to support this use case both in template creation and in updating, then I can accept that it's probably the best way forward.  but I do think it's a shame that we once had the feature.

to me it was much more elegant to have context variable that has a default value, gated on a question, than it is to have a project file with a big ol' jinja conditional in the file name (or to separate concerns to place macros and or post-generation scripts to handle the case)

> but it was something that didn't make much sense after all.

I _do_ actually think it made a lot of sense,  it was essentially saying `value = 'something' if some_condition else 'something else'` ... which is a perfectly sensical programmatic thing to do.  I agree perhaps that the hack of using `when: false` didn't make sense, but i think my use case (described in #1204) was a very sensical thing to do :joy:

ultimately, I will of course be able to figure out a workaround.  So definitely feel free to close this if the thought of having to maintain it gives you deep angst :)
but if we just step back and ask whether this looks like a sane design pattern:

```yaml
some_variable:
  when: "{{ some_other_variable == 'some value' }}"
  type: bool
  default: "{{ some_other_variable != 'some other value' }}"
```

then I'd say yep! 👍

tlambert03:
> The fact that a question that is explicitly skipped is missing in the context and the answers file is a good thing IMHO.

I'm also not sure I agree with that either.  I think having a variable that is sometimes present and sometimes not is probably less predictable than one that is always present and evaluated as described in the config file

yajo:
What if we don't record the answer but still provide it in the context if it has a value? No matter if `when` is hard-coded or not.

That would make it work essentially as a variable, and not as a question, which is what everybody wants. Since it wouldn't be recorded (just like a secret question), every time its `default` would be evaluated, until at some update its `when` becomes true and _then_ we'd record it.

sisp:
Sounds reasonable to me for the use case where semantically those variables are answers to questions (e.g. first a question like "Advanced questions? [y/N]" and if "Y" then a series of detailed questions gets asked). But for truly derived variables, I think something like #1203 is still needed because question types are very limited and a question has more settings than what's needed for a derived variable (e.g. `choices`).

lhupfeldt:
I think `when=false` will do the trick if properly documented as a way to set variables.
I think that a more explicit way of way of setting variables, e.g. #1205 or `prompt=false` could be better though, as like @sisp said, there are things which only makes sense for real questions.
But this is a minor detail which should not delay the feature.

sisp:
I've submitted #1220 which adds computed values via skipped questions (`when: false`) as suggested in https://github.com/copier-org/copier/pull/1206#issuecomment-1605892551. If #1220 gets merged, it will supersede this PR.

sisp:
Superseded by #1220.

Issue #1177, opened 2023-06-09
Copier update crashes when template includes secret questions


staffanlinnaeus:
### Describe the problem

When an answers file is rendered using `{{ _copier_answers|to_nice_yaml -}}`, no questions with `secret: true` will be included.

This causes problems when running `copier update`. If a template has secret questions, they won't be included in the answers file and copier will crash. 

I tried running `copier update --vcs-ref=0.2.0 --data password=somevalue` but the `--data` flag seems to be ignored.

It's possible to work around this by setting the secret question manually in the answers file, but this is obviously not something that you should do.

Tested with Copier 7.2.0 and 8.0.0.

### Example template

To reproduce the problem, create a new template including the following files:

```yaml
# Filename: copier.yml
username:
    type: str
    help: Enter username

password:
    type: str
    secret: true
    help: Enter password
```
```yaml
# Filename: {{ _copier_conf.answers_file }}.jinja
{{ _copier_answers|to_nice_yaml -}}
```
```yaml
# Filename: somefile.yaml.jinja
---
version: 0.1.0 # Example value that can be changed when setting new git tags
username: {{ username }}
password: {{ password }}
```

### To Reproduce

1. Create a project from a template which has secret questions and a rendered answers file (or use the example above).
2. Run `copier update --vcs-ref=<version>`
   OR
   run `copier update --vcs-ref=<version> --data password=somevalue`

### Logs

```console
Updating to template version 0.2.0
Traceback (most recent call last):
  File "/home/username/.local/bin/copier", line 8, in <module>
    sys.exit(copier_app_run())
  File "/home/username/.local/pipx/venvs/copier/lib/python3.10/site-packages/plumbum/cli/application.py", line 638, in run
    inst, retcode = subapp.run(argv, exit=False)
  File "/home/username/.local/pipx/venvs/copier/lib/python3.10/site-packages/plumbum/cli/application.py", line 633, in run
    retcode = inst.main(*tailargs)
  File "/home/username/.local/pipx/venvs/copier/lib/python3.10/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
  File "/home/username/.local/pipx/venvs/copier/lib/python3.10/site-packages/copier/cli.py", line 66, in handle_exceptions
    return method(*args, **kwargs)
  File "/home/username/.local/pipx/venvs/copier/lib/python3.10/site-packages/copier/cli.py", line 369, in main
    ).run_update()
[...]

PR #1209, opened 2023-06-20
refactor: clean up secret questions management

sisp:
I've cleaned up the internal bookkeeping logic for secret questions. The new structure will simplify raising an error when a secret question has no default value, which we discussed in https://github.com/copier-org/copier/issues/1177#issuecomment-1585556997.

There are two ways to declare a question as secret:

1. By setting `secret: true` in the config of a question.
2. By adding the variable name of a question to the `_secret_questions` list. This option is needed when a question that uses the `<name>: <default value>` syntax shall be secret.

Internally, [Copier uses the a secret questions list to omit answers to secret questions from the answers file](https://github.com/copier-org/copier/blob/1b351bc8560d19d70e927cdb2fafb4f7ce41092d/copier/main.py#L258). This list contains the variable names of all secret questions independent of how they are declared as secret, so questions with `secret: true` are added to this list internally.

Before, adding questions with the `secret: true` setting to the `_secret_questions` list took place both in [`template.py:filter_config()`](https://github.com/copier-org/copier/blob/1b351bc8560d19d70e927cdb2fafb4f7ce41092d/copier/template.py#L70-L71) and in [`template.py:Template.secret_questions`](https://github.com/copier-org/copier/blob/1b351bc8560d19d70e927cdb2fafb4f7ce41092d/copier/template.py#L399-L400) which was redundant and also a bit error-prone because there was no clear separation of concerns where this information shall be synced.

I've decided to let `template.py:filter_config()` only handle separating config and questions data as its docstring suggests, and sync the secret information in [`template.py:Template.secret_questions`](https://github.com/copier-org/copier/blob/1b351bc8560d19d70e927cdb2fafb4f7ce41092d/copier/template.py#L399-L400) (as it has been done already, so no change) and [`template.py:Template.questions_data`](https://github.com/sisp/copier/blob/708eac9828242a53bddad8ed0bf23db4ef9205f0/copier/template.py#L386-L388).

In addition, I've fixed some minor oddities:

- A literal dict `{}` was used as a default value for getting the `_secret_questions` list, which should be a `list`.
- The empty config data was `{"secret_questions": set()}` which makes not much sense (or is at least unnecessary).

Those changes are purely internal though.

### DIFF
--- a/copier/template.py
+++ b/copier/template.py
@@ -55,21 +55,17 @@ DEFAULT_TEMPLATES_SUFFIX = ".jinja"
 
 def filter_config(data: AnyByStrDict) -> Tuple[AnyByStrDict, AnyByStrDict]:
     """Separates config and questions data."""
-    conf_data: AnyByStrDict = {"secret_questions": set()}
+    config_data: AnyByStrDict = {}
     questions_data = {}
     for k, v in data.items():
-        if k == "_secret_questions":
-            conf_data["secret_questions"].update(v)
-        elif k.startswith("_"):
-            conf_data[k[1:]] = v
+        if k.startswith("_"):
+            config_data[k[1:]] = v
         else:
             # Transform simplified questions format into complex
             if not isinstance(v, dict):
                 v = {"default": v}
             questions_data[k] = v
-            if v.get("secret"):
-                conf_data["secret_questions"].add(k)
-    return conf_data, questions_data
+    return config_data, questions_data
 
 
 def load_template_config(conf_path: Path, quiet: bool = False) -> AnyByStrDict:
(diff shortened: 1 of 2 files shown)


yajo (approved):

sisp:
Just FYI, @yajo: Auto-merge is a bit useless now because a second approval is needed. And I can't approve my own PR. So you or @pawamoy still need to take action.

yajo:
Thanks. I downgraded that to 1 required approval.

sisp:
@yajo The CI check `CI / flake--check (macos-latest)` keeps failing with this error:

```
  Could not set environment: 150: Operation not permitted while System Integrity Protection is engaged
  Error: Process completed with exit code 150.
```

Any idea what's wrong?

sisp:
I think it has been fixed in [`install-nix-action` v22](https://github.com/cachix/install-nix-action/releases/tag/v22).

sisp:
I just noticed that `install-nix-action` had already been bumped to v22 by #1208, so I just rebased this PR.

PR #1220, opened 2023-06-29
feat: add support for computed values via skipped questions

sisp:
I've added support for a long awaited feature: computed values via skipped questions (`when: false`). :tada: See https://github.com/copier-org/copier/pull/1206#issuecomment-1605892551 for more details.

Supersedes #1206. Resolves #1204 #1205. Resolves (I guess ...) #1203. :stuck_out_tongue_winking_eye:

### DIFF
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -181,8 +181,8 @@ Supported keys:
     If it is a string, it is converted to boolean using a parser similar to YAML, but
     only for boolean values. The string can be [templated][prompt-templating].
 
-    If a question is skipped, its answer is not recorded, but its default value is available
-    in the render context.
+    If a question is skipped, its answer is not recorded, but its default value is
+    available in the render context.
 
     !!! example
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -179,9 +179,9 @@ Supported keys:
     creating a computed value.
 
     If it is a string, it is converted to boolean using a parser similar to YAML, but
-    only for boolean values. The string can be [templated](#prompt-templating).
+    only for boolean values. The string can be [templated][prompt-templating].
 
-    If a question is skipped, no answer is recorded, but its default value is available
+    If a question is skipped, its answer is not recorded, but its default value is available
     in the render context.
 
     !!! example
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -175,17 +175,14 @@ Supported keys:
 
 -   **when**: Condition that, if `false`, skips the question.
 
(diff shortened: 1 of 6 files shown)


docs/configuring.md
```diff
@@ -175,17 +175,14 @@ Supported keys:
...
-    only for boolean values.
+    only for boolean values. The string can be [templated](#prompt-templating).
 
-    This is most useful when [templated](#prompt-templating).
-
-    If a question is skipped, its answer will be:
-
-    -   The default value, if you're generating the project for the 1st time.
-    -   The last answer recorded, if you're updating the project.
+    If a question is skipped, no answer is recorded, but its default value is available
```
yajo, line 184:
```suggestion
    If a question is skipped, its answer is not recorded, but its default value is available
```

yajo (changes requested):
Little improvement for docs. The rest looks great!

yajo:
I'm thinking...  We're just adding a feature, but what if a template relied on the fact that a skipped question was always missing? Should we label this as a breaking change? 🤔

sisp:
Missing in the context? For this to be breaking, an injected variable through an extension like the context hooks one would need to have the same name as the skipped question. I guess, it can be breaking in rare cases. :thinking:

I forgot to edit the inline comments to describe the new behavior. I'll push a commit with corrected comments in a moment.

yajo:
Well, the feature was never supported, so after all any breakages would come from unsupported use. Thus we can consider this a normal improvement.

sisp:
I think this PR isn't quite right yet. We should make sure that a skipped question's answer can only ever be its default value but not a value passed via `--data` or read from an edited answers file. I'll look into it.

yajo:
> not a value passed via `--data` or read from an edited answers file

I think you don't need to care so deeply for this. Those are either weird or unsupported use cases anyways. If you do them, you know you're gonna get something wrong, so it should be OK for us IMHO.

sisp:
Alright, I guess we can keep it like it is now to ship it more quickly, possibly add more rigor later for those anyway weird cases.

yajo:
It seems you have style violations. Could you run pre-commit please?

sisp:
Ooops, yes, should be fixed now.

pawamoy:
Best PR :smiling_face_with_tear: Many users will be super happy about this!

tlambert03:
thanks for your work on this @sisp!

I can confirm that with this PR, my template at https://github.com/pydev-guide/pyrepo-copier works as it did before v8

sisp:
More important is @yajo's careful assessment about the best implementation of this feature.

PR #1229, opened 2023-07-02
fix: Support Pydantic v2

### LINKED ISSUE #1225: Pydantic 2.0 (complete rewrite) breaks copier
### Describe the problem

After over a year and two PyCons, the Rust rewrite of Pydandic with 17x+ speedup (last I heard maybe it was 23x?) is out (four hours ago)! Unfortunately, it breaks copier. As much as I [hate suggesting it](https://iscinumpy.dev/post/bound-version-constraints/), you might want to do a quick patch release with a version cap, then work on making it work (Unless it's really easy to fix, of course, which it might be).

Also, it's a really bad idea to use things from dependencies that start with underscores. Just saying. :P

(They did an incredible amount of work to try to make sure they didn't break users. Users importing underscore objects doesn't count!)

### Template

https://github.com/scientific-python/cookie

(If you want one, that is - you can't even start copier on a fresh install)

### To Reproduce

Run copier.

That's all.

### Logs

```pytb
  File "/home/runner/work/cookie/cookie/.nox/compare_copier/bin/copier", line 5, in <module>
    from copier.__main__ import copier_app_run
  File "/home/runner/work/cookie/cookie/.nox/compare_copier/lib/python3.8/site-packages/copier/__init__.py", line 6, in <module>
    from .main import *  # noqa: F401,F403
  File "/home/runner/work/cookie/cookie/.nox/compare_copier/lib/python3.8/site-packages/copier/main.py", line 29, in <module>
    from .errors import (
  File "/home/runner/work/cookie/cookie/.nox/compare_copier/lib/python3.8/site-packages/copier/errors.py", line 6, in <module>
    from pydantic.errors import _PathValueError
ImportError: cannot import name '_PathValueError' from 'pydantic.errors' (/home/runner/work/cookie/cookie/.nox/compare_copier/lib/python3.8/site-packages/pydantic/errors.py)
```

### Expected behavior

Not to break when pydantic is installed, or not to install pedantic 2.0.

_No response_

### Operating system

macOS

Any

### Copier version

8.0.0

### Python version

All

### Installation method

pipx+pypi

### Additional context

Broken logs at https://github.com/scientific-python/cookie/actions/runs/5426746583/jobs/9869183393

pawamoy:
Closes #1225.


sisp (approved):
LGTM! :tada: The code coverage reduction seems irrelevant to me.

henryiii:
Could you bump the pydantic version in the CI too? Looks like it’s pinned and/or cached too:
```
 Installing pydantic (1.10.9): Skipped for the following reason: Already installed
```

lhupfeldt:
Please release this a a quick fix for both 7.x an 8.x unless #1220 is release.

pawamoy:
@henryiii, good catch. I'm personally not used to track lock files for libraries, even when they provide a CLI :smile: @sisp should I update the lock file in this PR or another?

@lhupfeldt yes, don't worry, we'll surely merge and release this. We can't go faster than what our free time allows. Waiting for either:

- a second approval by @yajo
- a fix in the merge permissions, so that @sisp's approval is enough and allows me to merge

lhupfeldt:
> @henryiii, good catch. I'm personally not used to track lock files for libraries, even when they provide a CLI 😄 @sisp should I update the lock file in this PR or another?
> 
> @lhupfeldt yes, don't worry, we'll surely merge and release this. We can't go faster than what our free time allows. Waiting for either:
> 
>     * a second approval by @yajo
> 
>     * a fix in the merge permissions, so that @sisp's approval is enough and allows me to merge

Thank you, I'm aware it is a volunteer effort and it is much appreciated.
Note that I'm really hoping for a backport of the fix to 7.x, because I now have two issues :)

pawamoy:
> Note that I'm really hoping for a backport of the fix to 7.x, because I now have two issues :)

The fix is simple, so I think that's reasonable :slightly_smiling_face:

sisp:
> We use copier as part of our CI pipeline. Implementing the suggested workarounds is not feasible for us.

@lhupfeldt Why not? You should be able to [cap Pydantic as a temporary workaround] (https://github.com/copier-org/copier/issues/1225#issuecomment-1616656418) on your end.

> I'm personally not used to track lock files for libraries, even when they provide a CLI 😄

Slightly off-topic, but how do you ensure deterministic dev and test environments then? 🤔 

> @sisp should I update the lock file in this PR or another?

I think in this PR to check that the fix actually works for Pydantic v2, although this means we'll be developing against Pydantic v2 then since that'll be the installed version; meaning we can't test against v1 anymore, but the same is true vice versa. I assume Pydantic won't keep supporting v1 anyway. To test against both v1 and v2 we'd need something like Tox which doesn't work well with lock files though. 😵‍💫

I think @henryiii has been thinking about dependencies, constraints, upper bounds, etc. a lot (fantastic [article](https://iscinumpy.dev/post/bound-version-constraints/) by the way 👏). Any advice from your side?

pawamoy:
> Slightly off-topic, but how do you ensure deterministic dev and test environments then?

I don't. Each dev has their own lock file. I let things break as new versions of dependencies come out. It sometimes break my own CI (for PRs on my projects), but I'm fine with that as it warns me early that something is broken. Locally, I can update my lock file to test on latest dependencies versions, without worrying about reverting the lock file before commits.

Also, allowing each dev to have their own lock file integrates better with projects offering insiders versions (Material for MkDocs, my own projects :wink:), as these versions can be fetched from local, self-hosted PyPI servers. With a tracked lock file, devs using insiders versions have to re-lock, test/check then revert each time. It's also easier for devs that do not have access to the same index as the one that was used to lock (users whose government blocks pypi.org for example, and who have to use mirrors). 

> we'd need something like Tox

Not sure about Poetry, but PDM can be used with both [Nox](https://pdm.fming.dev/latest/usage/advanced/#use-nox-as-the-runner) and [Tox](https://pdm.fming.dev/latest/usage/advanced/#use-tox-as-the-runner).

henryiii:
>  I'm fine with that as it warns me early that something is broken

Yes, it would have been nice for this breakage to have been caught here, rather than in all the downstream projects that weren't using lock files. Same reasoning, I'd much rather get early warning of a breakage than have stable CI that doesn't tell me that my users are currently broken. :) (For a library, not a deployment of something, obviously)

However, the caching is nice to have. I wonder if enabling dev versions would generally help? There were a lot of dev releases of Poetry. Not all packages do that, but locking + periodic automated updates + dev versions would be really a good way to go if packages did that often enough.

On important projects, I usually test against minimum supported versions too - I have a test/constraints.txt file with all my specified minimums, and have a job or two dedicated to testing that.

> Not sure about Poetry, but PDM can be used with both [Nox](https://pdm.fming.dev/latest/usage/advanced/#use-nox-as-the-runner) and [Tox](https://pdm.fming.dev/latest/usage/advanced/#use-tox-as-the-runner).

Huge fan of PDM. It supports _all_ PEP 621 backends, so you can use hatchling, flit-core, setuptools, not just pdm-backend. And because you are using standards instead of custom Poetry config, Black and Ruff can pick up your minimum version of Python directly from your config, no need to set it for each tool! :)

Hatch is nice too, and has multiple environment support, but doesn't support locking yet, which is the main reason I would use a tool like that.

henryiii:
Yes, looks like someone has mentioned it at https://github.com/nix-community/poetry2nix/issues/1183. Not familiar at all with nix. The GitHub dependency graph supports PEP 621 now. Not sure about dependabot, but it's easy to write a job that runs pdm update (though making CI run on that job requires an extra step).

sisp:
This is what Poetry's docs say about comitting a lock file for a library: https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control

lhupfeldt:
> @lhupfeldt Why not? You should be able to [cap Pydantic as a temporary workaround] ([#1225 (comment)](https://github.com/copier-org/copier/issues/1225#issuecomment-1616656418)) on your end.

Feasible was probably not the right word, just some pinning, committing, PR review, build, and then undoing it all which I would rather not have to do:) We are generally setup for CI. Pinning stuff that breaks should be easier than it is, but that is work in progress.

majidaldo:
> Yes, looks like someone has mentioned it at [nix-community/poetry2nix#1183](https://github.com/nix-community/poetry2nix/issues/1183). Not familiar at all with nix. The GitHub dependency graph supports PEP 621 now. Not sure about dependabot, but it's easy to write a job that runs pdm update (though making CI run on that job requires an extra step).

👋

Have you guys seen [rye](https://rye-up.com/) ? it creates two lockfiles, one for running and another for dev'ing. 

From there, the path to nix I think would be [pip2nix](https://github.com/nix-community/pip2nix).

pawamoy:
PDM is also able to create/use multiple lockfiles with specific groups of dependencies :slightly_smiling_face:

lhupfeldt:
Now that #1220 has been merged, it is probably not necessary to backport this fix to 7.x.

sisp:
@pawamoy Could you update the locked Pydantic version to the latest v2, so that also CI can verify the correctness of the fix?

```sh
poetry update pydantic
```

pawamoy:
Oh yes, thanks for the reminder! Done.

sisp (changes requested):
@pawamoy Could you please update your Poetry version to the most recent? The lock file nkw contains many entries `category = "..." ` again which got dropped in Poetry v1.5.

sisp:
Oh, it seems there are more incompatibilities with Pydantic v2.

pawamoy:
Done. Let me know if I should rework the history to keep the lock file commits apart from the fix. Otherwise feel free to squash them all.

pawamoy:
Also I'll take a look at the other errors :slightly_smiling_face:

pawamoy:
Ha, too bad, bump-pydantic didn't update anything :thinking: 
I pushed other fixes, but it's not finished yet. I converted the PR back to draft.
Also, do we want to keep support Pydantic v1? I'm afraid I don't have enough time to work on this in a serious way...

yajo:
Hi folks! Many things popped up here, sorry to be late to the party 😆 

I see a bunch of off-topic discussions and I think they are interesting, but let me kindly suggest you to move them to the forum. Regarding the PR, I think the important part here is:



> do we want to keep support for Pydantic v1?

IMHO no. They broke compatibility. We should just update the dependency and forget Pydantic v1.

sisp:
@pawamoy Is it okay for you if I push commits to this PR? I have finished the Pydantic v2-only adaptations and would like to (a) retain your credit for your work on this and (b) avoid creating another PR.

pawamoy:
Of course, no need to ask :)

sisp (commented):
Pydantic v2 is working now, only the `flake-check` jobs are failing with some error seemingly related to Nix that I don't understand yet.

sisp:
I suspect that `annotated-types` needs to be added to https://github.com/nix-community/poetry2nix/blob/master/overrides/build-systems.json, but I'm not familiar enough with Nix and `poetry2nix`. Do.you know, @yajo?

yajo:
Yes, that's it. I'll see if I can do that later, it's just a routine task.

yajo (approved):
The rest of the code looks great. I'll push later the Nix build fix so we can merge.

sisp (approved):

PR #1244, opened 2023-07-10
fix(pydantic): add upper dependency bound to fix unlocked installations

yajo:
Fix https://github.com/copier-org/copier/issues/1225.


yajo:
@copier-org/maintainers do you see this as a good temporary fix while we settle #1229?

These months I'll have reduced availability to work on Copier, so I think going for the MVP is a good idea.

sisp (approved):
I'm fine with it. We can move on to Pydantic v2 shortly afterwards. If @pawamoy doesn't have the capacity to finish #1229, I can take over.

PR #1252, opened 2023-07-17
build: drop support for Python 3.7

sisp:
I've dropped support for Python 3.7. Resolves #1251.

### DIFF
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -3,27 +3,19 @@
 import errno
 import os
 import platform
-import shutil
 import stat
 import sys
-import tempfile
-import warnings
 from contextlib import suppress
+from importlib.metadata import version
 from pathlib import Path
 from types import TracebackType
-from typing import Any, Callable, Optional, TextIO, Tuple, Union, cast
+from typing import Any, Callable, Literal, Optional, TextIO, Tuple, Union, cast
 
 import colorama
 from packaging.version import Version
 from pydantic import StrictBool
 
-from .types import IntSeq, Literal
-
-# TODO Remove condition when dropping python 3.8 support
-if sys.version_info < (3, 8):
-    from importlib_metadata import version
-else:
-    from importlib.metadata import version
+from .types import IntSeq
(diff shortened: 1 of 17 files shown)


copier/tools.py
```diff
@@ -147,7 +140,8 @@ def handle_remove_readonly(
 ) -> None:
     """Handle errors when trying to remove read-only files through `shutil.rmtree`.
 
-    This handler makes sure the given file is writable, then re-execute the given removal function.
+    On Windows, `shutil.rmtree` does not handle read-only files very well. This handler
+    makes sure the given file is writable, then re-execute the given removal function.
```
sisp, line 144: It seems that`shutil.rmtree` doesn't only have this issue under Python <3.8 on Windows. See the Windows CI checks when I had this handler removed: https://github.com/copier-org/copier/actions/runs/5573531416

It seems okay to remove the custom `TemporaryDirectory` class though. I hope that's actually correct and not just uncovered by tests.

sisp:
I think this PR is ready for a review now.

I'm not sure why there are 3 `build (<OS>-latest, 3.7)` jobs, they shouldn't exist because I removed `"3.7"` from the CI matrix. :thinking:

copier/tools.py
pawamoy: Is the `handle_remove_readonly` utility still used somewhere else in the code?

yajo:
That was due to repo settings. Fixed now.

copier/tools.py
sisp: No, only in [`template.Template._cleanup`](https://github.com/copier-org/copier/blob/a27366bc72927919914fbe09987ffd16518f0ad8/copier/template.py#L213).

yajo (approved):

sisp:
@yajo Haha, looks like I'm still struggling with the commit change types. :laughing: I was actually wondering whether `build:` was a bad choice because it wouldn't appear in the changelog.

PR #1253, opened 2023-07-17
style: prefer imports from modules

sisp:
Just some minor consistency improvement regarding imports. I think it's preferable to import things like exceptions and utilities from their corresponding modules and not from the root module, and only import `run_copy`/`run_recopy`/`run_update` (and perhaps `Worker`, although I think that should be hidden long-term) from the root module.

This is a minor initial effort towards #1250.

### DIFF
--- a/tests/test_symlinks.py
+++ b/tests/test_symlinks.py
@@ -5,8 +5,9 @@ import pytest
 from plumbum import local
 from plumbum.cmd import git
 
-from copier import readlink, run_copy, run_update
+from copier import run_copy, run_update
 from copier.errors import DirtyLocalWarning
+from copier.tools import readlink
 
 from .helpers import build_file_tree
 
--- a/tests/test_vcs.py
+++ b/tests/test_vcs.py
@@ -9,11 +9,13 @@ from packaging.version import Version
 from plumbum import local
 from plumbum.cmd import git
 
-from copier import Worker, errors, run_copy, run_update, vcs
+from copier import Worker, run_copy, run_update
+from copier.errors import ShallowCloneWarning
+from copier.vcs import GIT_VERSION, checkout_latest_tag, clone, get_repo
 
 
 def test_get_repo() -> None:
-    get = vcs.get_repo
+    get = get_repo
 
     assert get("git@git.myproject.org:MyProject") == "git@git.myproject.org:MyProject"
     assert (
@@ -65,7 +67,7 @@ def test_get_repo() -> None:
 
 @pytest.mark.impure
 def test_clone() -> None:
-    tmp = vcs.clone("https://github.com/copier-org/copier.git")
+    tmp = clone("https://github.com/copier-org/copier.git")
     assert tmp
     assert Path(tmp, "README.md").exists()
     shutil.rmtree(tmp, ignore_errors=True)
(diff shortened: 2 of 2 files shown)


pawamoy:
It's in my personal "best-practices" list: always use canonical imports within your own code base, even if importing from private modules. Use public imports when importing from other packages, and never add an indirection level (like importing external objects from an internal module).

Importing from canonical locations reduces the risk of getting cyclic imports. I don't think any linter offer such checks though. Avoiding indirect imports and wildcard imports helps static analysis tools.

sisp:
@pawamoy Just clarifying: You second importing from the specific modules and not from the root module, right? :laughing:

pawamoy:
> Just clarifying: You second importing from the specific modules and not from the root module, right?

Ah, not exactly! Thanks for asking. To me, the public API is for users, and internal code should use the internal API, so importing from specific modules. But here you modify the imports in the tests suite: it is good to actually test the public API!

<details><summary>My second point is somehow a rant. Collapsing to reduce noise as it's off-topic here.</summary>

Linters usually forbid you to import from protected modules (starting with `_`), or importing protected objects from any module. But they forbid that without regards to where the object comes from: it should be tolerated to import [from] protected objects [modules] as long as they exist within your own code base.

```python
from external_pkg import _protected_thing  # rightly forbidden
from external_pkg._protected_module import thing  # rightly forbidden

from mypackage import _protected_thing  # should be tolerated
from mypackage._protected_module import thing  # should be tolerated
```

Without such tolerance, you get noise in linter warnings, and end up either ignoring them inconsistently (which puts you at risk of unannounced breaking changes) or stopping to protect objects (which is terrible for maintaining backward-compatibility).

Sorry, I'm just being an old man yelling at the linters.
</details>

sisp:
Ah, I see. Thanks for the detailed clarification! :pray:

In that case, how would you do

* `from copier.tools import readlink` and
* `from copier.vcs import GIT_VERSION, checkout_latest_tag, clone, get_repo`

in the tests instead? Those are internal utilities, but the they happen to be needed also in the tests. If they were protected my renaming `tools.py` to `_tools.py` and `vcs.py` to `_vcs.py`, are you suggesting to do

* `from copier._tools import readlink` and
* `from copier._vcs import GIT_VERSION, checkout_latest_tag, clone, get_repo`

in the tests?

The current situation is suboptimal anyway because [Copier's public API isn't well defined](https://github.com/orgs/copier-org/discussions/1250) IMO. That's why I thought to import those utilities from the modules directly as I _consider_ them internal although they aren't yet marked as such by a leading underscore. Cleanly exposing only public identifiers would be another PR for me.

pawamoy:
> are you suggesting to do `from copier._tools import readlink` and...

If this is the only way to import them, yes. If they have public locations too, use the public location in tests (but keep using the canonical/protected location within the code, unless linters are too annoying).

> The current situation is suboptimal anyway because [Copier's public API isn't well defined](https://github.com/orgs/copier-org/discussions/1250) IMO. That's why I thought to import those utilities from the modules directly as I consider them internal although they aren't yet marked as such by a leading underscore. Cleanly exposing only public identifiers would be another PR for me.

Yup, we can discuss about this in a later PR :slightly_smiling_face:

pawamoy (approved):

PR #1258, opened 2023-07-19
docs: use relative path to Markdown file for page linking

sisp:
I've unified the linking to pages to [use relative paths to Markdown files](https://www.mkdocs.org/user-guide/writing-your-docs/#linking-to-pages) instead of relative URL paths like `../<page name>` which is quite unintuitive because of the leading `../` that is needed as MkDocs page URLs have a trailing `/`.

### DIFF
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -724,7 +724,7 @@ code hunk in the file itself, similar to the behavior of `git merge`.
 
 During a project update, Copier needs to compare the template evolution with the
 subproject evolution. This way, it can detect what changed, where and how to merge those
-changes. [Refer here for more details on this process](./updating.md).
+changes. [Refer here for more details on this process](updating.md).
 
 The more lines you use, the more accurate Copier will be when detecting conflicts. But
 you will also have more conflicts to solve by yourself. FWIW, Git uses 3 lines by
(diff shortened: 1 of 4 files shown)


pawamoy (approved):

sisp:
Thanks, @pawamoy! :pray:

pawamoy:
Thank you @sisp :smile:

yajo:
Wouldn't be better to use the auto-generated shortcuts?

sisp:
What are auto-generated shortcuts? :thinking:

yajo:
I mean this: https://github.com/mkdocstrings/autorefs

pawamoy:
That's a possibility indeed, and could make things more consistent.

@sisp for example, instead of

```md
see [using a context hook](faq.md#how-can-i-alter-the-context-before-rendering-the-project)
```

it would be like

sisp:
Ah, I see. But this only works for anchors, not for plain page links, right?

sisp:
And what would be the advantage over using links to Markdown files plus anchors when referencing headings?

pawamoy:
True, it only works for linking to actual headings, not plain pages.
The advantage is that you don't have to remember on which page it is, or find out how to link to the page using relative paths (sometimes a headache), and this will continue to work even if the sections move from pages to pages.

sisp:
I don't think I've experienced any of these issues, so I don't think I'm qualified to judge whether autorefs would be better than regular page links and anchors. Please feel free to decide whether you'd prefer autorefs and I'll simply trust your judgment.

The only problem I've encountered a few times is quite the opposite &ndash; duplicate subsections under different parent sections on the same page which lead to anchor collisions, and then I needed custom anchors via non-standard Markdown syntax.

pawamoy:
Headings with the same id *in the same page* will automatically get suffixed with -1, -2, etc.
Headings with the same id *in different pages* won't be suffixed though, so autorefs falls short on this.

Also, just in case it wasn't clear: autorefs is already used in a few (most?) places in the docs, so it's not like we're adding a new dependency.

I don't have a strong opinion on this, just commenting on autorefs. I believe both regular and autorefs links can co-exist without issues. Just a matter of preference :slightly_smiling_face:

yajo:
> I believe both regular and autorefs links can co-exist without issues. Just a matter of preference slightly_smiling_face

Indeed. It just surprised me that, given most docs use autorefs, this change didn't go that path.

FWIW I find useful autorefs when linking from docs to code docstrings and viceversa.

PR #1270, opened 2023-07-25
fix: clean up tmp dir

yajo:
Until now, Copier always left a lot of garbage in the tmp dir.

When running in batch, this could become a big problem, as templates would fill up memory (`/tmp/` is usually mounted as tmpfs).

@moduon MT-3411

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -287,6 +287,7 @@ class Worker:
         # Backwards compatibility
         # FIXME Remove it?
         conf = asdict(self)
+        conf.pop("_cleanup_hooks")
         conf.update(
             {
                 "answers_file": self.answers_relpath,
--- a/copier/main.py
+++ b/copier/main.py
@@ -15,6 +15,7 @@ from tempfile import TemporaryDirectory
 from typing import (
     Callable,
     Iterable,
+    List,
     Literal,
     Mapping,
     Optional,
@@ -187,6 +188,7 @@ class Worker:
     unsafe: bool = False
 
     answers: AnswersMap = field(default_factory=AnswersMap, init=False)
+    _cleanup_hooks: List[Callable] = field(default_factory=list, init=False)
 
     def __enter__(self):
         """Allow using worker as a context manager."""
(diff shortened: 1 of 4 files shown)


sisp:
Pydantic can't validate the bound cleanup method. I've been wondering anyway why the `Worker` and `Subproject` classes should be Pydantic models as they don't parse external data, which is Pydantic's primary use case.

yajo:
My pydantism is very rusty now... Do you have any clue on how to fix the test? 🤔

copier/main.py
```diff
@@ -187,6 +188,7 @@ class Worker:
     unsafe: bool = False
 
     answers: AnswersMap = field(default_factory=AnswersMap, init=False)
+    _cleanup_hooks: List[Callable] = field(default_factory=list, init=False)
```
sisp, line 191: I _think_ you can wrap the `Callable` type hint with Pydantic's `SkipValidation` container (https://docs.pydantic.dev/latest/usage/validators/#special-types):

```suggestion
    _cleanup_hooks: List[SkipValidation[Callable]] = field(default_factory=list, init=False)
```

Don't forget to import it from the `pydantic` package.

I haven't tried it though.

sisp: Oh, wait a second. It's about serialization, not validation. 🤔

sisp:
There is an occurrence of `{{ _copier_conf|to_json }}`, and it seems `_copier_conf` now contains this cleanup callable. I believe

https://github.com/copier-org/copier/blob/a9e7b6ba3c9ae3fca6606eb8ae22e7da52b2d69d/copier/main.py#L285

adds `_cleanup_hooks` to the dict which can't be serialized to JSON. I guess it would work if you explicitly deleted it from the dict.

copier/main.py
sisp: I think this was wrong, so I'm resolving this conversation as invalid.

sisp:
I took the liberty to fix the problem here, but for some reason GitHub doesn't show the commit 34d429fffda4e539e37e8c1478742a5fb51a01fc in https://github.com/copier-org/copier/tree/cleanup in this PR. :thinking:

sisp:
Okay, not sure what went wrong, but I amended the commit without change to get a new commit hash and force-pushed again, now GitHub picked it up.

yajo (commented):
Great, thanks!

sisp (approved):
LGTM! :tada:

The only minor hiccup I have is accessing the `Subproject._cleanup` method in the cached property `Worker.subproject`, but I think it's fine for now and would probably require an architectural change to avoid it.

I've pulled the latest changes in `master`, especially the changes from #1156, and adapted the `current_worker.run_copy()` call to use the context manager as well.

Issue #1178, opened 2023-06-09
Don't require users to review all questions when running copier update


drewoldag:
### Actual Situation

When someone runs `copier update`, they have to review all of the responses that they have provided in their answers file. People get kind of blind to these responses and just return through the list as fast as possible. 

### Desired Situation

It would be nice if there was an argument that would take all the previous responses found in the answers file and only present the user with new template questions that were added after the hydrated project was created. 

### Proposed solution

I have no technical solution, but an initial questions like this would be amazing. (Actual wording should be improved.) 
```
>> Would you like an interactive update?
          Full interaction - review previous answers
    >> Some interaction - use previous answers, show me new questions
          No interaction - show no questions, just take previous answers and default values for new questions
```

sisp:
I like the idea. :+1: I'd use a switch though, e.g. `copier update --keep-answers` (just an idea, to be discussed) and `--defaults` might be added to use default answers for new questions.

yajo:
Using `--defaults`, any question without a default value will be asked. Does that solve your problem?

drewoldag:
`--defaults` would be fine, but I like the clarity of the `--keep-answers` flag. 

At the end of the day, it would just be nice not to be forced to review all the prior responses when updating :)

yajo:
Maybe, but `--keep-answers` is a bit absurd as a flag name. Answers are always kept on updates and recopies. Something like `--skip-answered` or `--ask-only-new-questions` makes more sense.

I don't have plans to implement this, but any PR would be welcome.

PR #1276, opened 2023-07-28
feat: Adding '--skip-answered' flag to enhance user interaction during copier update

ljossha:
This Pull Request addresses a usability issue troubling users when they run 'copier update'. Previously, each time 'copier update' was run, the user had to review all responses given in the answers file. This was a repetitive and laborious process, particularly for users dealing with extensive templates.

This PR introduces a new flag- --skip-answered to improve the user experience and reduce friction during updates. The new flag allows users to bypass the review of questions already answered in the past. It automatically pulls responses from the answers file and presents only the newly added template questions to the user for their review.

`skip-answered` will skip all answered questions and ask user input only for new inputs if they are present.

Issue: #1178


copier/main.py
```diff
@@ -847,11 +851,12 @@ def _apply_update(self):
...
             )
-            # Clear last answers cache to load possible answers migration
-            with suppress(AttributeError):
-                self.answers = AnswersMap()
-            with suppress(AttributeError):
-                del self.subproject.last_answers
+            # Clear last answers cache to load possible answers migration, if skip_answered flag is not set
+            if self.skip_answered is False:
+                with suppress(AttributeError):
+                    self.answers = AnswersMap()
```
yajo, line 857: This suppress looks like unnecessary, isn't it?

```suggestion
                self.answers = AnswersMap()
```

yajo (approved):
The code looks perfect to me. Just a minor non-blocking comment. Thanks for this contribution!

It seems you still have a little style issue... if you can run pre-commit it'll fix it for you.

ljossha:
> The code looks perfect to me. Just a minor non-blocking comment. Thanks for this contribution!
> 
> It seems you still have a little style issue... if you can run pre-commit it'll fix it for you.

I've fixed your suggested changes. 
But when I try to run pre-commit, I receive:
```
An error has occurred: InvalidConfigError: 
=====> .pre-commit-config.yaml is not a file
```
Any ideas?

yajo:
You will have to run `poe lint` (as you did), and commit and push the changes. 

Next release will still take some weeks. I don't want to release while on holidays. 🏖

ljossha:
> You will have to run `poe lint` (as you did), and commit and push the changes.
> 
> Next release will still take some weeks. I don't want to release while on holidays. 🏖

Yeah, I see. Pushed new changes.

ljossha:
@yajo Hey man!
It seems that everything is fine. 
Where will you be able to merge it? 
Do you need anything from me at this point?

yajo:
Since this is a new feature, I'd want to hand-test it before merge, so this will take a bit more than usual due to holidays. Maybe 1 or 2 weeks. But for now you can relax. According to the code, it seems perfect, so I'll just merge when I can give it a try. Otherwise I'll post what to change. Thanks!

tests/test_migrations.py
```diff
@@ -151,6 +151,86 @@ def test_pre_migration_modifies_answers(
...
+        git("tag", "v2")
+    # User updates subproject to v2 template
+    with local.cwd(dst):
+        run_update(defaults=True, overwrite=True, unsafe=True, skip_answered=True)
+        answers = json.loads(Path(".copier-answers.yml").read_text())
+        assert answers["_commit"] == "v2"
+        assert "best_song" not in answers
+        assert answers["best_song_list"] == ["paranoid android"]
+        assert json.loads(Path("songs.json").read_text()) == ["paranoid android"]
+        assert answers["bad_song"] == ["creep"]
```
yajo, line 231: I already hand-tested this feature. The feature is working fine, but the test isn't appropriate.

What this test is testing is that the migration is correct. However, this feature is not about migrations, but about interactive questions.

In v1 of template you create in the test, you add the question `best_song`. In v2, you replace that one with `bad_song` and `best_song_list`. Since the old question disappeared in v2, it wouldn't be answered, even without this feature.

Your feature is only noticed when using Copier interactively, but you're adding `defaults=True, overwrite=True` and disabling interactiveness.

So, summarizing, you can remove this test and add a new one that:
- Is interactive.
- Checks that the old question isn't asked again when updating.
- Checks that new questions are asked.

Interactive (TUI) tests are a bit more complex, so you can inspire yourself by this other test: https://github.com/copier-org/copier/blob/d6c04b478d5a4126dc10181e884ecd838d79f436/tests/test_prompt.py#L49-L117

yajo (changes requested):

tests/test_migrations.py
ljossha: Hey, yeah, it makes sense. 
I've removed my old test and added 2 more. 

Can you check it?

yajo: Much better now, thanks!

yajo (approved):

PR #1282, opened 2023-08-04
docs: add light/dark scheme support for star history chart

sisp:
star-history.com charts now support dark mode. :sunglasses:


sisp:
Me too. 🤓

PR #1288, opened 2023-08-12
feat: add support for pre-update and post-update messages

sisp:
I've added support for pre-update and post-update messages that can be written in plain text. Messages can be written inline or included via Jinja includes, and messages are rendered.

This (simple) implementation became possible due to #1156 since there's no output of an internal copy call of the update algorithm anymore. Without this copy call output suppression, the internal copy call would print the pre-copy/post-copy messages in case they are configured which we wouldn't want to see when running an update.

Related to #723. Follow-up of #1194.

### DIFF
--- a/tests/test_output.py
+++ b/tests/test_output.py
@@ -144,15 +144,6 @@ def test_messages_with_inline_text(
     # clear capture output log
     capsys.readouterr()
 
-    copy_pattern = (
-        r"^"
-        r"Thank you for using our template on (linux|macos|windows)"
-        r".+"
-        r"Project {project_name} successfully created"
-        r"\s*"
-        r"$"
-    )
-
     # copy
     if interactive:
         tui = spawn(COPIER_PATH + ("copy", "-r", "v1", str(src), str(dst)), timeout=10)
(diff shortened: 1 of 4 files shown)


tests/test_output.py
```diff
@@ -131,47 +155,84 @@ def test_message_copy_with_inline_text(
...
+    assert (dst / "version.txt").read_text() == "v2"
+    _, err = capsys.readouterr()
+    assert re.search(
+        (
+            r"^"
+            r"Updating on (linux|macos|windows)"
+            r".+"
+            r"Project {project_name} successfully updated"
+            r"\s*"
+            r"$"
```
yajo, line 208: https://docs.python.org/3/library/re.html#re.VERBOSE

It's beautier to use a verbose regexp IMHO.

tests/test_output.py
```diff
@@ -204,37 +294,72 @@ def test_message_copy_with_included_text(
...
     _, err = capsys.readouterr()
-    assert re.search(pattern.format(project_name="myproj_new"), err, flags=re.S)
+    assert re.search(
+        (
+            r"^"
+            r"Updating on (linux|macos|windows)"
+            r".+"
+            r"Project {project_name} successfully updated"
+            r"\s*"
+            r"$"
```
yajo, line 347: Same here

yajo (approved):
Looks very nice. Just a suggestion for those regex

tests/test_output.py
yajo: I'd do it slightly different. I think it makes the code clearer TBH:

```py
assert re.search(
    fr"""
    ^Updating\ on\ (linux|macos|windows).+
    Project\ {project_name}\ successfully\ updated\s*$
    """,
    err,
    flags=re.S | re.X,
)
```

sisp: Thanks, that looks more concise. :+1: I've added a commit that updates the regexes using verbose syntax also for copy/recopy/update messages. So this PR should probably be rebase-merged, not squash-merged.

Issue #857, opened 2022-11-02
Copier update unknown switch --vcs_ref


szebenyib:
**Describe the problem**
Using copier update 7.0.1

**Template**
Any template

**To Reproduce**
1. Run `copier update --vcs_ref=HEAD` as specified in the docs
2. Observe error: Error: Unkown switch --vcs_ref

**Expected behavior**
Ref to HEAD working

**Environment**
 - OS:  Arch Linux
 - Copier version: 7.0.1
 - Python version: 3.9.5
 - Installation method: pip in virtualenv
 
**Additional context**
As a side note just after generating from template the update does not work, and refers to a non-existing commit.

pawamoy:
Which part of the docs mention `vcs_ref`?

This part correctly mentions `vcs-ref` with a dash, though it's missing a leading one: `-vcs-ref` instead of `--vcs-ref` https://copier.readthedocs.io/en/stable/configuring/#vcs_ref

szebenyib:
Here: "to update to the latest commit..."
https://copier.readthedocs.io/en/latest/updating/

pawamoy:
The docs seem to be correct: `--vcs-ref=HEAD`.

szebenyib:
Yet this command is not recognised, this is the output I receive when running this:

```
someone@somewhere (git)-[master] % copier update --vcs-ref=HEAD
Error: Unknown switch --vcs-ref
------
copier update 7.0.1

Update a copy from its original template

The copy must have a valid answers file which contains info from the last Copier execution, including the source template (it must be a key called `_src_path`).

If that file contains also `_commit` and `destination_path` is a git repository, this command will do its best to respect the diff that you have generated since the last
`copier` execution. To avoid that, use `copier copy` instead.

Usage:
    copier update [SWITCHES] args...

Hidden-switches:
    -h, --help         Prints this help message and quits
    --help-all         Prints help messages of all sub-commands and quits
    -v, --version      Prints the program's version and quits
```

mshafer-NI:
Options get passed to copier rather than the sub-command.

Try `copier --vcs-ref=HEAD update`

szebenyib:
Ouch, that is true. Thank you, this somehow did not occur to me.

PR #1323, opened 2023-09-10
use correct vcs-ref syntax

12rambau:
I wanted to use a specific version of my template but I was not able to select it using vcs-ref option. Looking at the doc: https://copier.readthedocs.io/en/latest/generating/#copying-dirty-changes, I was trying to run: 

```
copier copy --vcs-ref master https://github.com/foo/copier-template.git ./path/to/destination
```

leading me to the following error: 
```
Error: Unknown switch --vcs-ref
```

According to this thread: https://github.com/copier-org/copier/issues/857, it seems the option is not for `copy` but `copier`. I updated the code in the doc to reflect that.

### DIFF
--- a/docs/generating.md
+++ b/docs/generating.md
@@ -54,14 +54,14 @@ to use.
 For example to use the latest master branch from a public repository:
 
 ```shell
-copier copy --vcs-ref master https://github.com/foo/copier-template.git ./path/to/destination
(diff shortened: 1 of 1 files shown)


pawamoy:
What version of Copier are you using?

12rambau:
v7.2.0, agreed I'm a bit late to the party ;-)

pawamoy:
OK, I'll close then, let us know if there is any other issue :slightly_smiling_face:

PR #1283, opened 2023-08-05
docs: use autorefs for section references

sisp:
I've updated the docs to use autorefs for section references instead of (page links with) anchors. See the preceding discussion on this topic starting at https://github.com/copier-org/copier/pull/1258#issuecomment-1644936923.

Follow-up of #1258.

PR #1325, opened 2023-09-11
feat: implement reading template data from yaml file

mspiegel31:
## Description

re-implements #1059, but adds a test case and uses the current state of `master`.

I'm excited for this feature;  feedback welcome

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -49,8 +49,10 @@ from pathlib import Path
 from textwrap import dedent
 from unittest.mock import patch
 
+import yaml
 from decorator import decorator
 from plumbum import cli, colors
+from plumbum.path import LocalPath
 
 from .errors import UnsafeTemplateError, UserMessageError
 from .main import Worker
@@ -156,6 +158,7 @@ class _Subcommand(cli.Application):
         "VARIABLE=VALUE",
         list=True,
         help="Make VARIABLE available as VALUE when rendering the template",
+        excludes=["--data-file"],
     )
     def data_switch(self, values: StrSeq) -> None:
         """Update [data][] with provided values.
@@ -167,6 +170,20 @@ class _Subcommand(cli.Application):
         for arg in values:
             key, value = arg.split("=", 1)
             self.data[key] = value
+
+    @cli.switch(
+        ["--data-file"],
+        cli.ExistingFile,
+        help="Load data from a YAML file",
+        excludes=["--data"],
+    )
+    def data_file_switch(self, path: LocalPath) -> None:
+        """Update [data][] with provided values.
+        Arguments:
+            path: The path to the YAML file to load.
+        """
+        with open(path) as f:
+            self.data.update(yaml.safe_load(f))
 
     def _worker(self, src_path: OptStr = None, dst_path: str = ".", **kwargs) -> Worker:
         """
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -179,6 +179,7 @@ class _Subcommand(cli.Application):
     )
     def data_file_switch(self, path: LocalPath) -> None:
         """Update [data][] with provided values.
+
         Arguments:
             path: The path to the YAML file to load.
         """
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -52,7 +52,6 @@ from unittest.mock import patch
 import yaml
 from decorator import decorator
 from plumbum import cli, colors
-from plumbum.path import LocalPath
 
 from .errors import UnsafeTemplateError, UserMessageError
 from .main import Worker
@@ -177,7 +176,7 @@ class _Subcommand(cli.Application):
         help="Load data from a YAML file",
         excludes=["--data"],
     )
-    def data_file_switch(self, path: LocalPath) -> None:
+    def data_file_switch(self, path: cli.ExistingFile) -> None:
         """Update [data][] with provided values.
 
         Arguments:
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -760,7 +760,7 @@ questions with default answers.
 
 !!! info
 
-    The `-d, --data` flags are mutually exclusive with the [`--data-file`](#data_file) flag.
+    The `-d, --data` flags are mutually exclusive with the [`--data-file`][data_file] flag.
 
 ### `data_file`
 
@@ -768,10 +768,12 @@ questions with default answers.
 -   CLI flags: `--data-file`
 -   Default value: N/A
 
-As an alternative to [`-d, --data`](#data) you can also pass the path to a YAML file
+As an alternative to [`-d, --data`][data] you can also pass the path to a YAML file
 that contains your data.
 
-!!! info Not supported in `copier.yml` or API calls. Only supported through the CLI.
+!!! info
+
+    Not supported in `copier.yml` or API calls. Only supported through the CLI.
 
 !!! example
 
@@ -786,17 +788,18 @@ that contains your data.
     Passing a data file
 
     ```shell
-    copier --data-file input.yml copy template destination
+    copier copy --data-file input.yml template destination
     ```
 
     is equivalent to passing its content as key-value pairs:
 
     ```shell
-    copier -d 'user_name=Manuel Calavera' -d 'age=7' -d 'height=1.83' copy template destination
+    copier copy -d 'user_name=Manuel Calavera' -d 'age=7' -d 'height=1.83' template destination
+    ```
 
 !!! info
 
-    The `--data-file` flag is mutually exclusive with the [`-d, --data`](#data) flags.
+    The `--data-file` flag is mutually exclusive with the [`-d, --data`][data] flags.
 
 ### `envops`
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -768,8 +768,8 @@ questions with default answers.
 -   CLI flags: `--data-file`
 -   Default value: N/A
 
-As an alternative to [`-d, --data`][data] you can also pass the path to a YAML file
-that contains your data.
+As an alternative to [`-d, --data`][data] you can also pass the path to a YAML file that
+contains your data.
 
 !!! info
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -758,6 +758,46 @@ questions with default answers.
     copier copy -fd 'user_name=Manuel Calavera' template destination
     ```
 
+!!! info
+
+    The `-d, --data` flags are mutually exclusive with the [`--data-file`](#data_file) flag.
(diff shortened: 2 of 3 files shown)


copier/cli.py
```diff
@@ -168,6 +171,20 @@ def data_switch(self, values: StrSeq) -> None:
...
 
+    @cli.switch(
+        ["--data-file"],
+        cli.ExistingFile,
+        help="Load data from a YAML file",
+        excludes=["--data"],
+    )
+    def data_file_switch(self, path: LocalPath) -> None:
+        """Update [data][] with provided values.
+        Arguments:
```
sisp, line 182: Could you please add an empty line here?

```suggestion

        Arguments:
```

docs/configuring.md
```diff
@@ -758,6 +758,46 @@ questions with default answers.
...
+
+    The `-d, --data` flags are mutually exclusive with the [`--data-file`](#data_file) flag.
+
+### `data_file`
+
+-   Format: `str`
+-   CLI flags: `--data-file`
+-   Default value: N/A
+
+As an alternative to [`-d, --data`](#data) you can also pass the path to a YAML file
```
sisp, line 771: Prefer autorefs (see above):

```suggestion
As an alternative to [`-d, --data`][data] you can also pass the path to a YAML file
```

docs/configuring.md
```diff
@@ -758,6 +758,46 @@ questions with default answers.
...
+that contains your data.
+
+!!! info Not supported in `copier.yml` or API calls. Only supported through the CLI.
```
sisp, line 774: The markup for this admonition isn't correct.

```suggestion
!!! info

    Not supported in `copier.yml` or API calls. Only supported through the CLI.
```

docs/configuring.md
```diff
@@ -758,6 +758,46 @@ questions with default answers.
...
+    Passing a data file
+
+    ```shell
+    copier --data-file input.yml copy template destination
+    ```
+
+    is equivalent to passing its content as key-value pairs:
+
+    ```shell
+    copier -d 'user_name=Manuel Calavera' -d 'age=7' -d 'height=1.83' copy template destination
```
sisp, line 795: Closing fence is missing and flags are passed after the subcommand since Copier v8.0.0:

```suggestion
    copier copy -d 'user_name=Manuel Calavera' -d 'age=7' -d 'height=1.83' template destination
    ```
```

docs/configuring.md
```diff
@@ -758,6 +758,46 @@ questions with default answers.
...
+    ```yaml title="input.yml"
+    user_name: Manuel Calavera
+    age: 7
+    height: 1.83
+    ```
+
+    Passing a data file
+
+    ```shell
+    copier --data-file input.yml copy template destination
```
sisp, line 789: Flags are passed after the subcommand since Copier v8.0.0:

```suggestion
    copier copy --data-file input.yml template destination
```

docs/configuring.md
```diff
@@ -758,6 +758,46 @@ questions with default answers.
...
+    ```
+
+    is equivalent to passing its content as key-value pairs:
+
+    ```shell
+    copier -d 'user_name=Manuel Calavera' -d 'age=7' -d 'height=1.83' copy template destination
+
+!!! info
+
+    The `--data-file` flag is mutually exclusive with the [`-d, --data`](#data) flags.
```
sisp, line 799: Prefer autorefs (see above):

```suggestion
    The `--data-file` flag is mutually exclusive with the [`-d, --data`][data] flags.
```

docs/configuring.md
```diff
@@ -758,6 +758,46 @@ questions with default answers.
     copier copy -fd 'user_name=Manuel Calavera' template destination
     ```
 
+!!! info
+
+    The `-d, --data` flags are mutually exclusive with the [`--data-file`](#data_file) flag.
```
sisp, line 763: In #1283, we migrated to consistently using autorefs.

```suggestion
    The `-d, --data` flags are mutually exclusive with the [`--data-file`][data_file] flag.
```

tests/test_cli.py
```diff
@@ -131,6 +131,75 @@ def test_cli_data_parsed_by_question_type(
...
+                """
+            ),
+        }
+    )
+
+    run_result = CopierApp.run(
+        [
+            "copier",
+            "copy",
+            f"--data-file={existing_data_file.absolute()}",
```
sisp, line 190: * Simply use the `build_file_tree()` helper.
* `local` is an absolute path already, no need to convert the file path to absolute anymore.
* Just a minor cosmetic and subjective thing: I'd not introduce a local variable for the data file for simplicity.

```suggestion
    build_file_tree(
        {
            (src / "copier.yml"): (
                f"""\
                question:
                    type: {type_}
                """
            ),
            (src / "{{ _copier_conf.answers_file }}.jinja"): (
                """\
                # Changes here will be overwritten by Copier
                {{ _copier_answers|to_nice_yaml }}
                """
            ),
            (local / "data.yml"): yaml.dump({"question": answer}),
        }
    )

    run_result = CopierApp.run(
        [
            "copier",
            "copy",
            f"--data-file={local / 'data.yml'}",
```

sisp (changes requested):
Thanks for picking up this feature implementation, @mspiegel31! :pray:

Looks good overall :+1:, I just have a couple of remarks.

mspiegel31:
Thanks for the feedback (and the suggestion markers).  I've accepted all feedback verbatim, should be ready to look at again.

copier/cli.py
```diff
@@ -168,6 +171,21 @@ def data_switch(self, values: StrSeq) -> None:
             key, value = arg.split("=", 1)
             self.data[key] = value
 
+    @cli.switch(
+        ["--data-file"],
+        cli.ExistingFile,
+        help="Load data from a YAML file",
+        excludes=["--data"],
+    )
+    def data_file_switch(self, path: LocalPath) -> None:
```
sisp, line 180: Wouldn't the type hint `cli.ExistingFile` be more adequate?

```suggestion
    def data_file_switch(self, path: cli.ExistingFile) -> None:
```

I missed this in the previous review. In case you agree, don't forget to remove the corresponding import.

sisp (changes requested):
Thanks for updating the PR, @mspiegel31! :pray:

Just a minor comment/question about a type hint that I missed in the previous review. And I think it would be great testing that `-d, --data` and `--data-file` are mutually exclusive, i.e. testing that using both simultaneously raises an error.

copier/cli.py
mspiegel31: addressed in https://github.com/copier-org/copier/pull/1325/commits/f841be471d7dd4d229480c446b42820a82663517

tests/test_cli.py
```diff
@@ -131,6 +131,114 @@ def test_cli_data_parsed_by_question_type(
...
+            '--data=question="also does not matter"',
+            str(src),
+            str(dst),
+            "--quiet",
+        ],
+        exit=False,
+    )
+    assert run_result[1] == 2  # cli failure
+    out, _ = capsys.readouterr()
+    assert "Error: Given --data-file, the following are invalid ['--data']" in out
```
sisp, line 238: It seems the order of the error for mutually exclusive flags isn't deterministic. Sometimes, this order of flags is printed, sometimes is swapped, i.e.:

```
assert "Error: Given --data, the following are invalid ['--data-file']" in out
```

Perhaps we can simply check for both cases:

```suggestion
    assert any(
        [
            "Error: Given --data-file, the following are invalid ['--data']" in out,
            "Error: Given --data, the following are invalid ['--data-file']" in out,
        ]
    )

```

sisp (changes requested):
The error related to mutually exclusive flags apparently isn't deterministic, so we need to be more robust in asserting the message. And there's a minor Prettier formatting error.

tests/test_cli.py
mspiegel31: Happy to add this; it does seem like asserting an error condition of 2 and no data project created Shou be sufficient though?

sisp: True, this seems to be a [`plumbum.cli.switches.SwitchCombinationError` which results in a return code 2](https://github.com/tomerfiliba/plumbum/blob/40fcd2708863d36fb570be4ca19c14d2457ab7b7/plumbum/cli/application.py#L626), but [Copier also uses return code 2 for `UnsafeTemplateError`](https://github.com/copier-org/copier/blob/ddafdee3ce162da7b889c44e11aad8b19c910136/copier/cli.py#L72-L74), so it's ambiguous. Resolving this return code collision is a [separate topic for discussion](https://github.com/copier-org/copier/issues/1328) by the way. So, I'd say it's safer to assert also the error message.

tests/test_cli.py
```diff
@@ -131,6 +131,114 @@ def test_cli_data_parsed_by_question_type(
...
+        ("yaml", "1.2", 1.2),
+        ("yaml", "true", True),
+        ("yaml", "null", None),
+        ("yaml", '["string", 1, 1.2, true, null]', ["string", 1, 1.2, True, None]),
+        ("json", '"string"', "string"),
+        ("json", "1", 1),
+        ("json", "1.2", 1.2),
+        ("json", "true", True),
+        ("json", "null", None),
+        ("json", '["string", 1, 1.2, true, null]', ["string", 1, 1.2, True, None]),
```
yajo, line 158: I think we don't need 30 tests here.

Just pass a `data.yml` that contains a few types and you solve all with a single test.

yajo (approved):
All looks good, once the CI goes ✔️

tests/test_cli.py
```diff
@@ -131,6 +131,120 @@ def test_cli_data_parsed_by_question_type(
...
+    )
+    assert run_result[1] == 2  # cli failure
+    out, _ = capsys.readouterr()
+    assert any(
+        [
+            "Error: Given --data-file, the following are invalid ['--data']" in out,
+            "Error: Given --data, the following are invalid ['--data-file']" in out,
+        ]
+    )
+
```
sisp, line 244: No need to introduce an empty line here:

```suggestion
```

sisp (changes requested):
Just a minor new change request, the formatting error is still unresolved, and [@yajo's change request](https://github.com/copier-org/copier/pull/1325/files#r1328389674) is still pending.

yajo (approved):
Looks perfect, thanks!

sisp (approved):
Perfect, @mspiegel31! :ok_hand: Thanks for your contribution and patience. :pray:

mspiegel31:
Thank you for the awesome DX tools in the repo @sisp @yajo!  contribution is a breeze.

PR #1330, opened 2023-09-19
fix(cli): use return code 4 for unsafe template error

### LINKED ISSUE #1328: Ambiguous exit code 2
### Describe the problem

[Copier uses exit code 2 for `UnsafeTemplateError`](https://github.com/copier-org/copier/blob/ddafdee3ce162da7b889c44e11aad8b19c910136/copier/cli.py#L72-L74) while Plumbum uses the same exit code for `plumbum.cli.switches.SwitchError` type errors, e.g. mutually exclusive switches that are passed simultaneously.

### Template

There is none yet that exhibits this problem, but it would occur after merging #1325.

### To Reproduce

_No response_

### Logs

_No response_

### Expected behavior

Copier should use different exit codes for semantically different error types.

_No response_

### Operating system

Linux

any

### Copier version

next feature release that includes #1325

### Python version

any

### Installation method

pipx+pypi

### Additional context

_No response_

sisp:
I've updated the return code for `UnsafeTemplateError` errors to use 4 (`1 << 2` / `0b100`) instead of 2 which is used by Plumbum already for `SwitchError` type errors.

Technically, this is a breaking change :thinking:, but I anticipate hardly anybody relying on the return code. WDYT, @copier-org/maintainers?

Fixes #1328.

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -71,7 +71,8 @@ def handle_exceptions(method, *args, **kwargs):
         return 1
     except UnsafeTemplateError as error:
         print(colors.red | "\n".join(error.args), file=sys.stderr)
-        return 2
+        # DOCS https://github.com/copier-org/copier/issues/1328#issuecomment-1723214165
+        return 0b100
 
 
 class CopierApp(cli.Application):
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1439,7 +1439,7 @@ Copier templates can use dangerous features that allow arbitrary code execution:
 -   [Tasks][tasks]
 
 Therefore, these features are disabled by default and Copier will raise an error (and
-exit from the CLI with code `2`) when they are found in a template. In this case, please
+exit from the CLI with code `4`) when they are found in a template. In this case, please
 verify that no malicious code gets executed by any of the used features. When you're
 sufficiently confident or willing to take the risk, set `unsafe=True` or pass the CLI
 switch `--UNSAFE` or `--trust`.
--- a/tests/test_unsafe.py
+++ b/tests/test_unsafe.py
@@ -106,7 +106,7 @@ def test_copy_cli(
     if unsafe:
         assert retcode == 0
     else:
-        assert retcode == 2
+        assert retcode == 4
         _, err = capsys.readouterr()
         assert "Template uses potentially unsafe feature: tasks." in err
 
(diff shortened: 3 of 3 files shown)


copier/cli.py
```diff
@@ -71,7 +71,7 @@ def handle_exceptions(method, *args, **kwargs):
         return 1
     except UnsafeTemplateError as error:
         print(colors.red | "\n".join(error.args), file=sys.stderr)
-        return 2
+        return 1 << 2
```
pawamoy, line 74: I so rarely shift bits that without knowing it's 4, I would have had to look up on the web. Is that how the other exit codes are written within the code base? If not, any reason why not using `0b100` which, to me at least, is more readable? In comparison, `1 << 2` adds an indirection: `1` shifted by 2, so `100`, then in decimal gives `4`.

Also, maybe dumb question, but why not using `4` directly, with a short comment about using only multiples of 2?

pawamoy (commented):
I have no idea if someone is relying on the current exit code :shrug: 
I'd say nobody does, but you never know :shrug: In any case the code was conflicting with plumbum so we could say it was a buggy code?

copier/cli.py
sisp: Fair point. I've replaced the bitwise shift by a literal `4` with a comment and in the tests just with a literal `4`.

copier/cli.py
```diff
@@ -71,7 +71,7 @@ def handle_exceptions(method, *args, **kwargs):
         return 1
     except UnsafeTemplateError as error:
         print(colors.red | "\n".join(error.args), file=sys.stderr)
-        return 2
+        return 4  # 0b100
```
yajo, line 74: This might be a bit nitpicking, but IMHO this would be more helpful:

```suggestion
        # DOCS https://github.com/copier-org/copier/issues/1328#issuecomment-1723214165
        return 0b100
```

I'm based on the fact that somebody reading this code will want to understand why in the world is it written in bytes, and if they follow the link they'll maybe learn something interesting.

yajo (changes requested):

copier/cli.py
sisp: Absolutely fine with me. :ok_hand:

sisp: Done: f95585f

yajo (approved):

yajo:
> the code was conflicting with plumbum so we could say it was a buggy code?

Technically this wasn't buggy because we didn't have yet any excluding flags.

IMHO we can label this as a breaking change. We still have lots of major versions to release until we reach chrome! 😂

Versions are free, after all...

Do you agree?

sisp:
Agreed, I've amended the commit message with a `BREAKING CHANGE` footer.

yajo:
Ok

PR #1332, opened 2023-09-20
feat: allow overriding data file with CLI arguments

mspiegel31:
## Description

Users should be able to pass CLI data arguments and use a `--data-file` at the same time.  Let's just take inspiration from the [helm values file](https://helm.sh/docs/chart_template_guide/values_files/) and give highest priority to data values passed at the command line.

from the helm doc site:
>The list above is in order of specificity: values.yaml is the default, which can be overridden by a parent chart's values.yaml, which can in turn be overridden by a user-supplied values file, which can in turn be overridden by --set parameters.

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -47,7 +47,6 @@ import sys
 from io import StringIO
 from pathlib import Path
 from textwrap import dedent
-from typing import Set
 from unittest.mock import patch
 
 import yaml
@@ -106,10 +105,8 @@ class _Subcommand(cli.Application):
     """Base class for Copier subcommands."""
 
     def __init__(self, executable):
-        self.cli_data_args: Set[str] = set()
+        self.data: AnyByStrDict = {}
         super().__init__(executable)
-
-    data: AnyByStrDict = {}
 
     answers_file = cli.SwitchAttr(
         ["-a", "--answers-file"],
@@ -174,7 +171,6 @@ class _Subcommand(cli.Application):
         for arg in values:
             key, value = arg.split("=", 1)
             self.data[key] = value
-            self.cli_data_args.add(key)
 
     @cli.switch(
         ["--data-file"],
@@ -191,7 +187,7 @@ class _Subcommand(cli.Application):
             file_updates: AnyByStrDict = yaml.safe_load(f)
 
         updates_without_cli_overrides = {
-            k: v for k, v in file_updates.items() if k not in self.cli_data_args
+            k: v for k, v in file_updates.items() if k not in self.data
         }
         self.data.update(updates_without_cli_overrides)
 
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -104,7 +104,9 @@ class CopierApp(cli.Application):
 class _Subcommand(cli.Application):
     """Base class for Copier subcommands."""
 
-    data: AnyByStrDict = {}
+    def __init__(self, executable):
+        self.data: AnyByStrDict = {}
+        super().__init__(executable)
 
     answers_file = cli.SwitchAttr(
         ["-a", "--answers-file"],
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -199,22 +198,17 @@ def test_data_and_data_file_mutually_exclusive(
...
-        [
-            "Error: Given --data-file, the following are invalid ['--data']" in out,
-            "Error: Given --data, the following are invalid ['--data-file']" in out,
-        ]
-    )
-    assert not (dst / ".copier-answers.yml").exists()
+    assert run_result[1] == 0
+    answers = yaml.safe_load((dst / ".copier-answers.yml").read_text())
+    assert answers["_src_path"] == str(src)
+    assert answers["question"] == expected
```
yajo, line 211: Could you please also assert that some other question that got answered only by yaml still keeps that value? You'll have to modify a bit more the copier.yml and input.yml files

yajo (changes requested):
Good enhancement, thanks!

tests/test_cli.py
mspiegel31: addressed in https://github.com/copier-org/copier/pull/1332/commits/69be4d11e47009e35112ba31067b73dc38010565

mspiegel31:
@yajo I'm a little confused about the coverage failure, it looks like `codecov` is stating this PR reduces test coverage on other test files ([ref](https://app.codecov.io/gh/copier-org/copier/pull/1332/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=copier-org))?  Would love some clarification or help here.

copier/cli.py
```diff
@@ -103,7 +103,9 @@ class CopierApp(cli.Application):
 class _Subcommand(cli.Application):
     """Base class for Copier subcommands."""
 
-    data: AnyByStrDict = {}
+    def __init__(self, executable):
+        self.data: AnyByStrDict = {}
```
yajo, line 107: Why this change?

yajo (commented):
Looking good now! 

Just one more doubt...

yajo:
https://github.com/copier-org/copier/pull/1332#issuecomment-1728329806: Some upload error on codecov: https://github.com/copier-org/copier/actions/runs/6251262443/job/16972001045?pr=1332#step:12:51

Gotta check that.

yajo:
I updated the token, hoping (not much) it fixes codecov uploads forever.

copier/cli.py
mspiegel31: when `data` is a _class_ member, test data from previous test runs is unintentionally shared between all tests in the `test_cli.py` test suite.  I've pushed https://github.com/copier-org/copier/pull/1332/commits/ed656c5aabedd414dfa1458d2961c099dde2a21e to demo the CI failure that occurs.  

this failure won't show up if you run just run the failing test on its own (i.e. `pytest tests/test_cli.py -k 'test_data_file_parsed_by_question_type' `); it only shows up when tests before it are ran.

making `data` an _instance_ member solves this issue.  I'm guessing that this only matters for the test suites, since every time you invoke `copier` at the command line it spins up all state "fresh".

mspiegel31:
@yajo/@sisp still not quite sure what to do about these codecov errors;  can you provide any guidance? Given that the "dropped" coverage seems to occur other test files, would you be OK admin merging this one without the `codecov/project` task passing?

copier/cli.py
yajo: Ah, mutability... I wish I had the time to rewrite copier in haskell 😂

Thanks!

yajo (commented):
Please could you rebase? to see if that fixes codecov.

mspiegel31:
@yajo the rebase fixed this;  should be all set for a final review!

yajo (approved):

PR #1333, opened 2023-09-21
build(typing): add missing type hints

sisp:
Just some polishing :wink:: I've added a couple of missing type hints.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -1002,7 +1002,7 @@ def run_update(
     return worker
 
 
-def _remove_old_files(prefix: Path, cmp: dircmp, rm_common: bool = False):
+def _remove_old_files(prefix: Path, cmp: dircmp, rm_common: bool = False) -> None:
     """Remove files and directories only found in "old" template.
 
     This is an internal helper method used to process a comparison of 2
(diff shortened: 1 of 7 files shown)


pawamoy (approved):
Nice!

Maybe you know that already: if you add `from __future__ import annotations` at the top of modules (to stringify annotations, lots to say on the subject, I invite anyone to read PEP 649 and related), and use Ruff to auto-fix the code, Ruff will automatically place imports that are only used for annotations under an `if TYPE_CHECKING:` condition.

sisp:
We were briefly discussing about this topic a while ago without conclusion: https://github.com/copier-org/copier/pull/1179#discussion_r1230194072 What was [your point](https://github.com/copier-org/copier/pull/1179#discussion_r1231255519) about at the end, @yajo?

pawamoy:
Ha, I followed my own advice and read PEP 649 more carefully (it improved a lot over time). Cyclic imports and `if TYPE_CHECKING` are still an ongoing problem in cases where annotations need to be resolved at runtime (Pydantic models, dataclasses). So it's not always possible to put imports under `if TYPE_CHECKING`, and I'm not sure how Ruff handles that.

yajo (approved):

yajo:
> What was [your point](https://github.com/copier-org/copier/pull/1179#discussion_r1231255519)

I meant that, according to the docs, adding `from __future__ import annotations` will enable [PEP 563](https://peps.python.org/pep-0563/), but you were talking about [PEP 604](https://peps.python.org/pep-0604/) in that thread, which is only py3.10+.



> use Ruff to auto-fix the code

TIL about Ruff and looks promising. I'd be glad to use that in Copier!

PR #1335, opened 2023-09-23
fix: fix answer validation for `type: str` questions

sisp:
I've fixed the answer validation for `type: str` questions. Copier does loose parsing/validation, i.e. non-string values that can be interpreted as strings are valid, but the loose parsing/validation has been too loose. For instance, values such as `object()` or `{}` or `[]` were considered valid.

I've created a custom validator/parser inspired by [Pydantic's `str_validator`](https://github.com/pydantic/pydantic/blob/3b297adf6132636646f36e77a5cd2aacd6b4f03b/pydantic/v1/validators.py#L59-L77) which makes more sense IMO.

WDYT?

### DIFF
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -6,6 +6,8 @@ import platform
 import stat
 import sys
 from contextlib import suppress
+from decimal import Decimal
+from enum import Enum
 from importlib.metadata import version
 from pathlib import Path
 from types import TracebackType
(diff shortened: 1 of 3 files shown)


copier/tools.py
```diff
@@ -90,7 +92,23 @@ def printf_exception(
...
+        value:
+            Anything to be casted to a str.
+    """
+    if isinstance(value, str):
+        return value.value if isinstance(value, Enum) else value
+    if isinstance(value, (float, int, Decimal)):
+        return str(value)
+    if isinstance(value, (bytes, bytearray)):
+        return value.decode()
+    raise ValueError(f"Could not convert {value} to string")
```
yajo, line 108: How about this?

```suggestion
    return repr(value)
```

sisp: This would let non-string-like things like `object()` pass validation as `"<object object at 0xfa2598>"` which I think should fail instead because `object()` is not a proper answer to a `type: str` question. IMO, if somebody really wants to pass the `__repr__` value of an object, then they should pass it explicitly by calling `repr(...) explicitly/consciously. WDYT, @yajo?

yajo: Of course. Thanks for the explanation!

yajo (approved):

PR #1349, opened 2023-10-01
docs: fix generating `--help-all` output in CLI reference

sisp:
I've removed dead code that was meant to generate `--help-all` output in the API reference. It was shown in the [docs until v6.0.0](https://copier.readthedocs.io/en/v6.0.0/reference/cli/#copier.cli) but then disappeared. Do you happen to know why, @pawamoy?

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -41,13 +41,17 @@ Basically, there are 3 different commands you can run:
         ```
 
 Below are the docs of each one of those.
+
+CLI help generated from `copier --help-all`:
+
+```sh exec="on" result="text"
+copier --help-all
+```
 """
 
 import sys
-from io import StringIO
 from pathlib import Path
 from textwrap import dedent
-from unittest.mock import patch
 
 import yaml
 from decorator import decorator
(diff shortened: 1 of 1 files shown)


pawamoy:
Oh, good catch! Here's the most probable explanation:

- before, mkdocstrings used pytkdocs to collect data from sources, and pytkdocs uses dynamic analysis, so the updated value of the `__doc__` attribute was properly collected
- now mkdocstrings uses Griffe, which is still able to use dynamic analysis, but prefers static analysis if possible. Without a custom Griffe extension, modifying the `__doc__` attribute value at runtime doesn't have any effect anymore on the data collected by Griffe, and therefore on the output of mkdocstrings.

If we want to keep adding the output of `--help-all` in the docs, I'd recommend using markdown-exec instead:

````md
```bash exec="on"
copier --help-all 2>&1
```
````

(not sure the stderr-to-stdout redirection is needed)

Happy to provide a PR!

**Comment by pawamoy** [comment]
Also, we can merge this one as clean up first, and then I send the PR using markdown-exec :slightly_smiling_face:

**Review by pawamoy — APPROVED** [review]

**Comment by sisp** [comment]
Thanks for the prompt investigation and suggestion to use `markdown-exec`, @pawamoy! :ok_hand:

WDYT, @yajo?

**Comment by pawamoy** [comment]
Or really you can simply update this PR and add this at the end of the module docstring:

````md
CLI help generated from `copier --help-all`:

```bash exec="on" result="text"
copier --help-all
```
````

yajo:
Looks good, but if the real fix is so easy, I think it's worth it keeping all in a commit. Could you please add that suggested change from @pawamoy?

sisp:
Done, @yajo. That's a perfect fix. :ok_hand: Thanks for the idea, @pawamoy! :+1:

sisp:
I'll just merge without further review as the changes have indirectly been approved by the aforegoing discussion.

Issue #1224, opened 2023-06-30
Copier update fails if git gpg commit signing is enabled globally


scmmmh:
### Describe the problem

If signing git commits is enabled globally for a user using `git config --global commit.gpgsign true`, then running `copier update` fails with the error "error: gpg failed to sign data". This is caused by `copier/main.py` in lines 917-928, where user name and email are set and then a commit is attempted. Since there is no GPG key for the given e-mail address GPG cannot sign the commit and the update process fails in line 926.

### Template

Error is template-agnostic.

### To Reproduce

1. Run `git config --local commit.gpgsign true`
2. In a project created using `copier copy` run `copier update`
3. Update fails with `error: gpg failed to sign the data`

### Expected behavior

The update is applied

### Operating system

Linux

### Operating system distribution and version

Ubuntu 23.04

### Copier version

8.0.0

### Python version

CPython 3.11

### Installation method

pipx+pypi

sisp:
I believe #856 is related but abandoned. It would be great to finish that work.

scmmmh:
Looking at #856, that definitely looks as if that would fix my bug, as that is essentially the workaround I'm currently applying.

sisp:
Great! Would you be interested in working on a PR that finishes what was started in #856?

scmmmh:
What is there that remains to be done? As far as I can see there is just one failing check, but I can't see why it failed anymore.

sisp:
It looks like it was complete but didn't get merged for some reason. Since then, the code base has evolved, there are merge conflicts and probably new tests that need this extra Git config command as well. It should be quite straightforward to replicate #856 for the current state of the project.

scmmmh:
I am happy to pitch in, but I currently can't make any promises as to when I will find time for it.

sisp:
I've been thinking about this issue and #856 a little more and dislike the many added Git config lines in the tests of #856. So why were they added? Because the test suite fails for developers with GPG commit signing enabled, and that's because the repo/global/system Git config affects Copier's _internal_ usage of `git`, e.g. in the update algorithm. I believe we should differentiate between two Git usage scenarios in Copier:

1. Programmatic usage of `git` respecting user's configs. This is needed for cloning a template, e.g. when the Git project is private and the user has configured credentials or an SSH key.
2. Internal usage of `git`, e.g. the update algorithm.

[...]

Issue #616, opened 2022-03-20
Tests fail if gpg-sign is true


jmgilman:
**Describe the problem**
If the user has a git configuration with `gpg-sign` enabled, several tests will fail with:

```
error: gpg failed to sign the data
```

**Template**
N/A

**To Reproduce**
1. Clone the repo
2. In your git config, set `gpgsign = true` under the `[commit]` header.
3. Run `poetry install && poetry run poe test`

**Logs**

Essentially, any test that calls `git commit` fails with something like below:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
copier/main.py:758: in run_update
    worker.run_update()
copier/main.py:685: in run_update
    git("commit", "--allow-empty", "-am", "dumb commit 2")
../../Library/Caches/pypoetry/virtualenvs/copier-Vmj8QV6h-py3.9/lib/python3.9/site-packages/plumbum/commands/base.py:99: in __call__
    return self.run(args, **kwargs)[1]
../../Library/Caches/pypoetry/virtualenvs/copier-Vmj8QV6h-py3.9/lib/python3.9/site-packages/plumbum/commands/base.py:240: in run
    return p.run()
../../Library/Caches/pypoetry/virtualenvs/copier-Vmj8QV6h-py3.9/lib/python3.9/site-packages/plumbum/commands/base.py:201: in runner
    return run_proc(p, retcode, timeout)
../../Library/Caches/pypoetry/virtualenvs/copier-Vmj8QV6h-py3.9/lib/python3.9/site-packages/plumbum/commands/processes.py:322: in run_proc
    return _check_process(proc, retcode, timeout, stdout, stderr)
../../Library/Caches/pypoetry/virtualenvs/copier-Vmj8QV6h-py3.9/lib/python3.9/site-packages/plumbum/commands/processes.py:24: in _check_process
    proc.verify(retcode, timeout, stdout, stderr)
[...]

PR #1360, opened 2023-10-05
fix(vcs): prevent local clone from being stuck by gpg prompts

noirbizarre:
A quick and simple PR preventing the `copy` command to get stuck if you have `commit.gpgSign` and try to run the `copy` command on WIP local project for trial.


sisp (commented):
Thanks, @noirbizarre! :pray:

I think you're aiming for a subset of #856, which unfortunately seems abandoned. So, I anticipate updating a generated project with GPG commit signing enabled would be a problem for you, too. Would you be interest in finishing #856 here? I.e., add the `git config` line in `main.py` and adapt the tests accordingly? This would be in line with @yajo's pragmatic suggestion:

> Another option is forget about that and just add the `--no-gpg-sign` flag wherever needed. I don't think this would be too complicated honestly.
>
> &mdash; https://github.com/copier-org/copier/issues/1224#issuecomment-1646140385

Related to #1224.

noirbizarre:
Thanks, I didn saw it (my bad). I'll resubmit the PR integrating #1224 and #856 👍🏼

yajo (changes requested):
There are still more git commits issued. Maybe it's the time to fix all?

https://github.com/copier-org/copier/blob/1196644aaf928d84201a893b0c769cef48c0cdeb/copier/main.py#L946-L947

https://github.com/copier-org/copier/blob/1196644aaf928d84201a893b0c769cef48c0cdeb/tests/helpers.py#L149

There are still more in the tests suite. See https://github.com/copier-org/copier/issues/616.

noirbizarre:
OK, there is in fact 2 distinct issues in this:
- tests runs with the user config which might conflict/interfere witht the test
- copier also commit with the user config but this is legit and should be properly handled

PR updated with the following:
- tests use a sandboxed `.gitconfig` using [`pytest-gitconfig`](https://pypi.org/project/pytest-gitconfig/)
- a test specific explicit test case has been added for the `user.gpgsigninkey`/`commit.gpgsign` to ensure copier is properly handling this case
- `--no-gpg-sign` has been added to copier commit calls only (it is faster and easier to maintain than trying to monkeypatch plumbum or gitconfigm there was only 3 calls)

If there are some new or untested case which might conflict with `user.gpgsigninkey`/`commit.gpgsign`, am explicit test case for this case should be added on the same model than `tests/test_dirty_local.py::test_copy_dirty_head_with_gpg`

yajo (approved):
Awesome!

yajo:
I made a push including https://github.com/nix-community/poetry2nix/pull/1342 to see if that fixes the build.

noirbizarre:
I have no expertise on Nix and I really can't help on this part.

TBH I was surprise with the `.envrc` automatically building for 30min and I wonder if this part should be optin and excluded from CI ? It feels like artificially raising the complexity with an hard system dependency. NiX is meant for reproducibility but I can't reproduce this error, build pass as well as pre-commit checks 🤷‍♂️

However I am the `pytest-gitconfig` author so if something is needed for NiX compatibility (given the module it would be surprizing, but again, I am not knowledgeable on NiX), tell me, I'll do the change (this is the first time I encounter an error on this one)

yajo:
Don't worry, it's just a common routine for projects like this one that use poetry2nix. There's nothing special for you to do in pytest-gitconfig to add that compatibility.

Using nix is just because it's great! Most concerns about it is "it's complex", but in reality, once you learn it, it removes lots of complexities. It's obsoleting containers. I just can't live without it. 🤷🏼‍♂️ 

Regarding test failures, I see mypy failures (probably typing errors in some test) and also gpg signinig failures. It seems like, when you wanted to skip gpg, you actually enabled it everywhere! But you can see that all versions of CI (even those that don't use Nix) are broken, so it doesn't seem like a Nix problem at first sight.

I wouldn't like to go too off-topic about nix here, but if you have any questions please feel free to open thread(s) in the forum. 😊

noirbizarre:
Last failures were due to:
- a `pytest-gitconfig` compatibility issue with 3.8
- `plumbum.local.env` being snapshotted once at start and requiring its own monkeypatching has changes to `os.environ` are ignored

PR (and `pytest-gitconfig`) updated and should be fixed 👍🏼

pyproject.toml
sisp, line 1: IIRC from own experience, these formatting changes are caused by a VS Code TOML extension. Would you mind undoing them and saving the file without formatting, @noirbizarre?

sisp (changes requested):

pyproject.toml
noirbizarre: Done 👍🏼 

However, I can't explain windows errors:
- doesn't seem directly related to this change
- fails only on windows (which I can't reproduce, I only have Linux)
But I'll try to find out why.

Is it possible to remove the approval for the workflow (if you can do it on user or PR based)? 
Because this is rather hard to push and wait days to see if CI is green or still needs a fix while I could have probably fixed it in the same time (The windows part, I can't have seen it before has windows and Mac tests are passing). 
(I am curious, this is not the only open source project with a workflow approval to run on PRs, what is justifying that ?)

noirbizarre:
Sorry, comment posted in the wrong place.

Note: @yajo for nix it was just bad timing, latest update was broken for 3 days on `archlinux`. It's now fixed and it works 👍🏼 

EDIT: I think I also found the Windows issue: some git parameter were set from outside in CI (and so ignored by `pytest-gitconfig` doing it's job). Given `user.name` and `user.email` are already automatically set by `pytest-gitconfig`, I think windows was just missing the `core.autocrlf=input`. Ci should pass (unless there are other Windows specific issues)

yajo (approved):

PR #1366, opened 2023-10-08
docs: mention Cruft for Cookiecutter in comparison on template updates

sisp:
I've added a footnote in the comparison on template updates for Cookiecutter that refers to [Cruft](https://github.com/cruft/cruft) as a tool that uses Cookiecutter to enable updates. I think Cruft should be mentioned for a fair comparison.


pawamoy (approved):
Yes, definitely agree!

PR #1386, opened 2023-10-27
feat(choices): support questionary checkbox for multiple choices using `multiple: true`

noirbizarre:
This PR is a sequel to https://github.com/copier-org/copier/pull/1049, taking into account review comments and https://github.com/copier-org/copier/issues/750 design proposal.

This PR only include the `multiple: true` form as the advanced form needs more specification and design I think, and it mostly relies on https://github.com/RhetTbull/questionary-superprompt#multiple-option which is not part of questionary.
However, it remains compatible as it is possible to add the advanced part later.

WDYT ?

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -127,6 +127,10 @@ class Question:
             Selections available for the user if the question requires them.
             Can be templated.
 
+        multiselect:
+            Indicates if the question supports multiple answers.
+            Only supported by choices type.
+
         default:
             Default value presented to the user to make it easier to respond.
             Can be templated.
@@ -173,6 +177,7 @@ class Question:
     answers: AnswersMap
     jinja_env: SandboxedEnvironment
     choices: Union[Sequence[Any], Dict[Any, Any]] = field(default_factory=list)
+    multiselect: bool = False
     default: Any = MISSING
     help: str = ""
     multiline: Union[str, bool] = False
@@ -215,6 +220,8 @@ class Question:
                 f'to question "{self.var_name}" of type "{type_name}"'
             )
         try:
+            if self.multiselect and isinstance(answer, list):
+                return [type_fn(item) for item in answer]
             return type_fn(answer)
         except (TypeError, AttributeError) as error:
             # JSON or YAML failed because it wasn't a string; no need to convert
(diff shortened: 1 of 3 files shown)


noirbizarre:
I don't get why coverage on `tests/test_tools.py` changed:
- it's a test
- not related to the feature
- it has not been skipped

BTW, [POE `coverage` commands](https://github.com/copier-org/copier/blob/master/pyproject.toml#L84) differ from [the CI one](https://github.com/copier-org/copier/blob/master/.github/workflows/ci.yml#L69).

Why not this in CI ?:
```shell
poetry run poe coverage --cov-report=xml
```

This way it would be the same as locally with poe with only the xml coverage report in addition (the `-ra` is already in the `pyproject.toml` `tool.pytest.ini_options.addopts` setting) and tests would not be included in coverage by tests.

noirbizarre:
https://github.com/copier-org/copier/pull/1388 should fix the unexpected coverage change

copier/user_data.py
```diff
@@ -173,6 +177,7 @@ class Question:
     answers: AnswersMap
     jinja_env: SandboxedEnvironment
     choices: Union[Sequence[Any], Dict[Any, Any]] = field(default_factory=list)
+    multiple: bool = False
```
yajo, line 180: **issue:** the `multiple` word might seem OK as of today, but when https://github.com/copier-org/copier/issues/750 is implemented some day, it could become confusing.

This PR is about having the possibility to select multiple choices in a choices question. However, https://github.com/copier-org/copier/issues/750 is about asking the same question multiple times. In other words, the issue you're fixing is https://github.com/copier-org/copier/issues/218 instead.

Looking at https://github.com/RhetTbull/questionary-superprompt#multiple-option, it's possible that we use that solution when we implement https://github.com/copier-org/copier/issues/750. Thus, I'm afraid that using `multiple` now could be a shot in our own foot for the future.

So I think we should be able to rename this now, to help our future self.

WDYT about renaming this to `multiselect`?


```suggestion
    multiselect: bool = False
```

The same for other occurrences of this option.

FWIW, I can imagine a questionary where somebody wants to be able to make a multiselect choice multiple times (in the future, after this and #750 are implemented).

yajo (changes requested):

copier/user_data.py
noirbizarre: Oh, I am OK with this, I chose `multiple` as this was the term chosen in #750 (#750 define 2 behaviors, `multiple: true` and `multiple: object`) and #1049.

In fact, I thought a lot about it (that's what I was implying when saying "the advanced form needs more specification and design I think") because I think part of questionnary-superprompt might also be useful for multiselect: min/max make sense in multiselect as much as it make sense in multiple prompts. Poitn being that a possible evolution of `multiselect: true` is the same pattern than `multiple: true`: allowing to defined expected boundaries.

So `multiselect` it is, I'll update the PR accordingly 👍🏼 (thanks for pointing #218 because I missed it)

yajo (approved):
Excellent, thanks!

PR #1387, opened 2023-10-27
build: test and advertise Python 3.12 support

noirbizarre:
A basic pull request adding Python 3.12 to tested versions as well as proper PyPI classifiers for Python 3.12 support advertisement.

No functional changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ${{ matrix.os }}
     steps:
       # HACK https://github.com/actions/cache/issues/315
--- a/poetry.lock
+++ b/poetry.lock
@@ -251,13 +251,13 @@ files = [
 
 [[package]]
 name = "distlib"
-version = "0.3.6"
+version = "0.3.7"
 description = "Distribution utilities"
 optional = false
 python-versions = "*"
 files = [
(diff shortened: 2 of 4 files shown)


sisp (changes requested):
Thanks for caring about this maintenance work, @noirbizarre! 🙏 

I believe you just ran `poetry lock` or `poetry update`, right? It seems only `pywin32` needed an update though. I'd prefer not mixing a complete lock file update with adding support for Python 3.12. Would you mind reverting `poetry.lock` and trying to update only `pywin32` through `poetry update pywin32`?

sisp:
By the way, I think the tests are failing now because of the most recent mypy version. See also https://github.com/copier-org/copier/pull/1286#issuecomment-1673306751.

noirbizarre:
Yep, exactly. PR updated with minimal lock change.
Sorry for the multiple trials, I don't have other mean to test against Windows and MacOS.

yajo (approved):
Thanks!

sisp:
One CI job is failing though, I'm not quite sure why. It seems to be related to the pip vendored by Python 3.12. https://github.com/pypa/pip/issues/11501 looks like the same problem. I wonder why only the Ubuntu job is failing.

noirbizarre:
I think `virtualenv` needed an update, so I just did `poetry update virtualenv`.
For reference, the idea comes from https://github.com/pre-commit/pre-commit/issues/3041

sisp (approved):
Perfect, thanks for investigating that problem! :ok_hand:

LGTM! :tada: Thanks again for engaging also in this maintenance work, much appreciated! :pray:

noirbizarre:
I try to contribute as much as possible to tools I use, even for small tasks.
Given I contribute to those tools, I benefit too from housekeeping and maintenance.
Anyway, I'm pleased to help !

PR #1388, opened 2023-10-28
test(coverage): make coverage consistent and reproducible

noirbizarre:
While working on #1386, I noticed coverage change when it shouldn't on some module. This is the consequence of 2 things:
- tests are included in coverage, meaning that both `coverage` and `pytest-cov`  should be loaded early to avoid depending on plugins and tests load order
- using `pytest-xdist` without `--cov-append` leading to some reports parts being erased/overwritten

For reference:
- https://pytest-cov.readthedocs.io/en/latest/plugins.html *says about pytest plugins but apply in fact as soon as you want to measure coverage on test material (cover tests, provided fixtures...)
- https://github.com/pytest-dev/pytest-cov/issues/277

So this PR:
- use both `coverage run` to start pytest and `-p pytest_cov` to ensure pytest-cov is loaded before other plugins
- move all possible parameters into `pyproject.toml` to ensure consistent parameters whether tests/coverage are running in CI or locally
- add the `--cov-append` parameter to coverage for better `pytest-xdist` handling
- make the CI use `poe coverage` to ensure developers can run the same test suite locally

Note: test coverage has been kept, but I think codebase coverage should be enough


pyproject.toml
```diff
@@ -81,7 +81,7 @@ script = "devtasks:clean"
 help = "remove build/python artifacts"
 
 [tool.poe.tasks.coverage]
-cmd = "pytest --cov-report html --cov copier copier tests"
+cmd = "coverage run -m pytest -p pytest_cov --cov-report html --cov-report xml --cov-append"
```
yajo, line 84: Ran this, but I don't get any xml or html files produced. Am I missing something? Maybe some docs will help.

noirbizarre:
My bad, I have been through multiple iteration to find one working, and I didn't submit the right one. This should be fixed.

But honestly, I would simply remove coverage on `tests/` as this is totally unreliable (as you can see, you already have coverage reports [for jinja files](https://app.codecov.io/gh/copier-org/copier/tree/master/tests%2Fdemo_extensions_default), or [other templates](https://app.codecov.io/gh/copier-org/copier/tree/master/tests%2Fdemo), I tried to exclude them with the `omit` setting). You can try locally, even with this configuration (which will only remove some false-negative case but not all), you might notice coverage changes between 2 consecutive runs without a single code change as well as part marked as uncovered while this is a test which has run).
I also think it dilute the real project coverage (because most of the time you will have more tests than code to test and test coverage will always be 100% as soon as the test is executed, and the overall coverage will be higher than it should).

In the case of #1386 I think this is only a codecov misreport as the file flag as loosing coverage was in fact already not covered: https://app.codecov.io/gh/copier-org/copier/blob/master/tests%2Ftest_tools.py

pyproject.toml
noirbizarre: Sorry, bad version submitted, this should be fixed: it was missing the `--cov` but I also haven't kept the `coverage run -m` as it was not bringing so much benefit (on some pytest plugin I wrote, I wasn't able to have proper coverage with `pytest-cov` while it was working jsut fine with `coverage run` this is why I tried this prefix)

yajo:
I tested it and now I get the report I expected.

I noticed this generates hundreds of `.coverage.$hostname.*` files. Can't that be avoided?



> I would simply remove coverage on `tests/` as this is totally unreliable

I think it's useful in the sense that some parts of the test you write could be not really exercised for some reason. Maybe you're missing some parametrization.

Keep in mind that codecov is configured to aggregate all reports. Some tests run only on some OS, but we push all of them and then they get summed. Maybe that affects?

yajo:
Setting as draft until attended. @noirbizarre do you have plans to complete this?

noirbizarre:
OMG, sorry, I completly forgot this one.
I'm going to complete it soon

PR #1379, opened 2023-10-22
fix: Only fix Windows console with Colorama

pawamoy:
Closes #1378 

Not sure how to test this.

PR #1403, opened 2023-11-11
[FIX] update dependencies to lasted colorama

celm1990:
After this PR https://github.com/copier-org/copier/pull/1379  method `just_fix_windows_console` is used instead of `init` So is need install `colorama>=0.4.6` because this method is add in commit https://github.com/tartley/colorama/commit/5f1794d1aac9f2f5be4893bcc8ba8a061b9aeb08
otherwise raise error
`AttributeError: module 'colorama' has no attribute 'just_fix_windows_console'`


sisp (commented):
Good catch, @celm1990, and thanks for the PR! :pray:

Could you please run `poetry lock --no-update` to update the lock file (the content hash of `pyproject.toml` in there)? It currently warns:

```
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
```

sisp (approved):
Perfect, thanks! LGTM! :tada:

Issue #1414, opened 2023-11-17
make the context-hook extension a "safe" extension


mspiegel31:
### Actual Situation

For many of our templates, we would like to derive intermediate values from user input.  The FAQ [states](https://copier.readthedocs.io/en/stable/faq/#how-can-i-alter-the-context-before-rendering-the-project) the best way to do this is with a third-party extension, the [context-hook](https://github.com/copier-org/copier-templates-extensions#context-hook-extension) extension.

Using this extension requires that users leverage the `--unsafe` ([docs](https://copier.readthedocs.io/en/stable/configuring/#unsafe)) flag, as this requires loading a jinja extension.  This makes template adoption much trickier in scenarios where templates are being rendered in CI environments.  It's even scarier if users need to pass a command that has `--unsafe` in the input

### Desired Situation

Since the context-hook extension is recommneded by the docs (and is maintained by the `copier-org` organization), it would be useful to be able to use this extension without needing to supply it as a custom `jinja_extension` in `copier.yaml`.

### Proposed solution

I think this extension could be auto-loaded, similarly to how `jinja2_ansible_filters.AnsibleCoreFiltersExtension` is always loaded

sisp:
I'm afraid the situation is more complex. The context hook extension cannot be compared with the Ansible core filters extension because Ansible core filters is a finite set of well-defined Jinja filters whereas the context hook extension enables arbitrary code execution of code written by template authors. Thus, auto-loading the context hook extension and excluding it from the need for the `--UNSAFE, --trust` flag would be unsafe. The idea behind the flag is to guarantee that a template which doesn't require it will only rely on the Jinja filters/tests that are available through Copier, which are not harmful. Tasks, migrations, and additional Jinja extensions – or context hooks registered via the the context hook extension – allow arbitrary code execution, so Copier can't make any guarantees about the safety of a template using these features.

If your templates support updating, which is among the biggest benefits of Copier IMO, you'll sooner or later need migrations to provide a smooth update experience, so `--UNSAFE, --trust` will become necessary anyway. This flag is not there to scare users but to raise awareness for the possible risks of using a template with these features, so a conscious decision about using the template can be made.

pawamoy:
Author of the context-hook extension here: completely agree with @sisp here. It wouldn't make sense to trust this extension by default as it allows arbitrary code execution.

yajo:
[...]

PR #1420, opened 2023-11-20
docs: explain how to use computed variables, more clearly

yajo:
Fixes https://github.com/copier-org/copier/issues/1414 by documenting a clearer and safe alternative that covers most cases.


sisp (approved):
LGTM! :ok_hand:

PR #1421, opened 2023-11-21
fix: support older versions of colorama

em230418:
just_fix_windows_console was introduces in colorama 0.4.6

Some stable GNU/Linux OS use packages with older version of colorama. Examples:
https://packages.altlinux.org/en/p10/binary/python3-module-colorama/noarch/
https://packages.debian.org/bullseye/python3-colorama

In order to use OS specific packages with latest copier (which is required for example here https://github.com/OCA/oca-addons-repo-template/blob/master/copier.yml#L4), this patch is introduced.


pawamoy:
Might be already fixed by #1403

pawamoy:
If we are to accept this fix, #1403 should also be reverted.

em230418:
> If we are to accept this fix, #1403 should also be reverted

Done.

yajo (changes requested):
IMHO we shouldn't merge this.

Don't get me wrong. The patch looks quite OK. It's just that older distros will happen always.

Our supported installation methods include pipx and nix, which are both self-contained. You can use also pip and any venv implementation you want.

Thus, I think it's better that our users use self-contained installations, than having to deal with older distros.

At least, that's my opinion right now. But I could be wrong...

em230418:
Removed modification in poetry.lock and added pragma: nocover

> Our supported installation methods include pipx and nix, which are both self-contained. You can use also pip and any venv implementation you want.

In copier among these installation methods, there is installation method that just makes sure, that there is minimal versions of dependecies that can be used (those dependencies are described in pyproject.toml). I use this installation method and this PR is for those people who use the same method.

> It's just that older distros will happen always.

Listed distros in my first message are not ancient and are still supported.

sisp:
> In order to use OS specific packages with latest copier [...]

Just trying to understand the motivation for this PR better, @em230418: Why do you need to rely on a Colorama version from an OS distro? Thr `oca-addons-repo-template` README instructs installing Copier via `pipx`, so you should be able to get a recent version of Colorama. Am I missing something? :thinking_face:

em230418:
> Why do you need to rely on a Colorama version from an OS distro?

Because I trust packages from OS distro more, than packages from pypi in general. That thing is not specific for colorama.
As an exclusion, if is no package in OS distro, I use from ones from pypi or github.

pawamoy:
I suppose @em230418 installs Copier like `apt install python-copier` (or similar OS-specific command), where `python-copier` is packaged by volunteers.

pawamoy:
I'd vote to entirely remove the call to Colorama. I am myself considering the removal of Colorama from my projects:

- it's easy enough to hardcode/copy-paste a few ANSI sequences in a codebase (or use Plumbum colors in our case)
- `just_fix_windows_console` is not needed in recent tools like Microsoft's [Terminal](https://github.com/microsoft/terminal), or shells running in WSL, where ANSI sequences are natively supported
- higher-level libraries like Rich handle this transparently for us.

Obviously, there are still people using powershell, or cmd.exe through Cmder/else. Modern Windows (10) actually supports ANSI, but it's disabled by default. [This SuperUser post shows how to enable it permanently](https://superuser.com/a/1300251).

yajo:
Ok for dropping it

yajo:
@em230418 do you have intentions to go ahead with this PR and remove colorama?

Marking as draft while this isn't ready.

PR #1431, opened 2023-11-25
fix: mark package as typed

### LINKED ISSUE #1381: Add py.typed file
### Actual Situation

Mypy is unable to find type information for the `copier` package, which requires me to manually ignore type errors for `copier` whenever I import it.

### Desired Situation

Mypy doesn't complain about missing type information, and the type safety of my own code is increased as a result. I also have to invest less mental energy worrying about ignoring types for a package I'm using.

### Proposed solution

Add a `py.typed` file to the `copier` package.

sisp:
I've added the file `py.typed` to mark the `copier` package as typed when imported in other projects.

Fixes #1381.


sisp:
Or would the change type `build(typing)` be better? 🤔 It wouldn't appear in the changelog though ...

bswck:
Not strictly related to the PR. This project needs some modernization of type hints (don't use `Union[...]`, `typing.Callable`, etc.), a bit like here https://github.com/CERT-Polska/malduck/pull/111. I would wait with merging this before this passes on `mypy --strict`, which I can see now it doesn't, because of non-parametrized generics.

yajo (approved):
@bswck of course you're welcome to help improving type hints! You could even add the strict mode to our mypy checks, so we make sure there are no regressions on that.

OTOH, I see no reason for blocking this PR to wait for your improvement.

PR #1444, opened 2023-12-01
build(lint): switch over to ruff; add taplo

yajo:
Ruff is faster, better integrated, easier to configure... Cool!

Taplo also helps with toml maintenance.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ copier = "copier.__main__:copier_app_run"
 "Bug Tracker" = "https://github.com/copier-org/copier/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<4.0" # HACK https://github.com/PyCQA/isort/issues/1945
+python = ">=3.8"
 colorama = ">=0.4.6"
 decorator = ">=5.1.1"
 dunamai = ">=1.7.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,15 +5,15 @@ version = "0.0.0"
 description = "A library for rendering project templates."
 license = "MIT"
 classifiers = [
-    "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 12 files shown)


sisp:
There's some Nix error, but I still don't know enough about Nix to say what's wrong. 🫣

And I wonder whether the Ruff pre-commit hook also checks/applies formatting because the [hook spec](https://github.com/cachix/pre-commit-hooks.nix/blob/e5ee5c5f3844550c01d2131096c7271cec5e9b78/modules/hooks.nix#L1040-L1046) looks a bit outdated. With a recent Ruff version, it's `ruff check` for linting and `ruff format` for formatting. We'd want to make sure the formatting is also checked through CI. Do you know, @yajo?

pyproject.toml
```diff
@@ -27,7 +27,7 @@ copier = "copier.__main__:copier_app_run"
 "Bug Tracker" = "https://github.com/copier-org/copier/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<4.0" # HACK https://github.com/PyCQA/isort/issues/1945
+python = ">=3.8,<4.0"                                               # HACK https://github.com/PyCQA/isort/issues/1945
```
pawamoy, line 30: Wow, what's this extra space :joy:? Does it try to align comment on the far right or something?

yajo: Wow amazing, I didn't see this. Let me check

yajo: Fun, it turns out that the issue itself was closed long ago, so I removed the hack.

yajo:
Well, nix configs make the `.pre-commit-config.yaml` nix-generated file use taplo. It's just a TOML formatter.

You have a brief explanation of how it works [here](https://devenv.sh/pre-commit-hooks/#managing-the-pre-commit-configyaml-file).

PR #1505, opened 2024-02-06
add 'pygrep' lint group

danieleades:
what is says on the tin

- reduces the scope of mypy errors
- prevents accidentally 'hiding' new errors due to blanket lint suppression

### DIFF
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -76,7 +76,7 @@ def printf(
     if not style:
         return action + _msg
 
-    out = style + [action] + Style.RESET + [INDENT, _msg]  # type: ignore
+    out = style + [action] + Style.RESET + [INDENT, _msg]  # type: ignore[operator]
     print(*out, sep="", file=file_)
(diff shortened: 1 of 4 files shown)


danieleades:
some of the formatting changes here surprise me. probably used to using a different version of ruff? i don't see anywhere where the version of ruff is pinned. Am i missing something?

sisp (changes requested):
There's a minor Ruff linting error:

```
devtasks.py:63:9: PGH002 `warn` is deprecated in favor of `warning`
```

danieleades:
> There's a minor Ruff linting error:
> 
> ```
> devtasks.py:63:9: PGH002 `warn` is deprecated in favor of `warning`
> ```

should be resolved now.

Oddly, i couldn't reproduce this error locally using `poetry run ruff check .`

sisp (approved):
LGTM! :tada: Thanks, @danieleades! :pray:

PR #1506, opened 2024-02-06
add 'flake-future-annotations' lint group

danieleades:
allows using more modern typing syntax without sacrificing compatibility

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ pyyaml = ">=5.3.1"
 pyyaml-include = ">=1.2"
 questionary = ">=1.8.1"
 typing-extensions = { version = ">=3.7.4,<5.0.0", python = "<3.9" }
+eval-type-backport = { version = "^0.1.3", python = "<3.10" }
 
 [tool.poetry.group.dev]
 optional = true
@@ -109,7 +110,7 @@ style = "pep440"
 vcs = "git"
 
 [tool.ruff.lint]
-extend-select = ["B", "D", "E", "F", "I", "PGH", "UP"]
+extend-select = ["B", "D", "E", "F", "FA", "I", "PGH", "UP"]
 extend-ignore = ['B028', "B904", "D105", "D107", "E501"]
 
 [tool.ruff.lint.per-file-ignores]
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -84,21 +84,18 @@ class CopierApp(cli.Application):
     """The Copier CLI application."""
 
     DESCRIPTION = "Create a new project from a template."
-    DESCRIPTION_MORE = (
-        dedent(
-            """\
+    DESCRIPTION_MORE = dedent(
+        """\
             Docs in https://copier.readthedocs.io/
 
             """
-        )
-        + (
-            colors.yellow
-            | dedent(
-                """\
+    ) + (
+        colors.yellow
+        | dedent(
+            """\
                 WARNING! Use only trusted project templates, as they might
                 execute code with the same level of access as your user.\n
                 """
-            )
         )
     )
     VERSION = copier_version()
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -84,18 +84,21 @@ class CopierApp(cli.Application):
     """The Copier CLI application."""
 
     DESCRIPTION = "Create a new project from a template."
-    DESCRIPTION_MORE = dedent(
-        """\
+    DESCRIPTION_MORE = (
+        dedent(
+            """\
             Docs in https://copier.readthedocs.io/
 
             """
-    ) + (
-        colors.yellow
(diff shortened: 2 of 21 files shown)


danieleades:
i've removed some type aliases on the basis that (especially with the modern syntax allowed by the `from __future__ annotations` import) they are:

- needless indirection
- no shorter or more readable than the explicit types
- not used consistently throughout the codebase anyway

personally i'm inclined to remove the 'sequence' type aliases as well but thought that might be ever so slightly more controversial

danieleades:
i've gone ahead and removed the `Sequence` aliases, but it's in a separate commit for ease of review. That can be thrown away if it's not of interest

yajo (approved):
Lovely, thanks! ❤️

yajo:
It seems there are test failures, could you fix them please?

danieleades:
> It seems there are test failures, could you fix them please?

hmm this is a bug i've not run into before. The `from __future__ import annotations` line in `copier.cli` causes the errors. I suspect that the `@decorator` decorator isn't playing nicely with it

danieleades:
@yajo could you rerun the CI for me?
I *think* i've resolved the issues by rebasing on #1513

sisp (commented):
@danieleades Would you mind addressing https://github.com/copier-org/copier/pull/1513#pullrequestreview-1874341496 and then rebasing this PR first?

danieleades:
> @danieleades Would you mind addressing [#1513 (review)](https://github.com/copier-org/copier/pull/1513#pullrequestreview-1874341496) and then rebasing this PR first?

i've addressed that comment. I'll rebase this PR once that's merged

danieleades:
note for reviewers- I haven't been able to get the Nix environment working, so i'm not running the same `ruff` version locally and have avoided formatting, however i suspect that this PR *should* be formatted. I'm at an impasse.
Can a reviewer check this out and apply formatting using the Nix environment?

I take it the CI isn't checking the formatting? That would solve this problem for me- I could manually apply the changes if the CI showed a diff

pyproject.toml
sisp, line 1: How about adding

```toml
[tool.ruff.lint.isort]
required-imports = ["from __future__ import annotations"]
```

to ensure that all Python files use the future annotations import?

https://docs.astral.sh/ruff/settings/#lint_isort_required-imports

danieleades: i'd personally prefer not to, though I don't feel strongly about it

- the ruff `UP` lint group will automatically remove the `from __future__ import annotations` imports once python 3.9 is dropped (i believe) which will conflict with this configuration setting
- the `FA` lint group will already ping a user to add the import in any file that needs it, and that's not a large burden
- it can *occasionally* cause unexpected behaviour (as per my last PR!)

sisp: Agreed, thanks for sharing your reasoning. :ok_hand:

sisp: @danieleades I just remembered this discussion and your comment about Ruff's `UP` lint group removing `from __future__ import annotations` at some point, and then I checked [`pyupgrade`'s imports plugin](https://github.com/asottile/pyupgrade/blob/main/pyupgrade/_plugins/imports.py) and couldn't find any such removal rule there. Based on this and after skimming over https://docs.python.org/3/howto/annotations.html, I'm not sure whether there's any timeline yet for dropping `from __future__ import annotations`. But the situation is a bit confusing to me. Do you have more clarity?

/cc @copier-org/maintainers

sisp: I've read some more about this situation, and I believe it's this:

In Python 3.7's release notes, [postponed evaluation of annotations][python-37.pep563] ([PEP 563][python.pep563]) was introduced with the following statement:

> Since this change breaks compatibility, the new behavior needs to be enabled on a per-module basis in Python 3.7 using a `__future__` import:
>
> ```python
> from __future__ import annotations
> ```
>
> It will become the default in Python 3.10.

However, the [decision was retracted][python-310.pep563-retraction], and [PEP 649][python.pep649] will [supersede PEP 563 in Python 3.13][python-313.pep649-supersedes-pep563]. So prior to supporting Python 3.13, `from __future__ import annotations` should be added. I'm not sure yet how compatibility across Python versions prior to 3.13 and 3.13+ will work though.

[python.pep563]: https://peps.python.org/pep-0563/
[python.pep649]: https://peps.python.org/pep-0649/
[python-37.pep563]: https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations
[python-310.pep563-retraction]: https://mail.python.org/archives/list/python-dev@python.org/thread/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/
[python-313.pep649-supersedes-pep563]: https://github.com/python/steering-council/issues/63#issuecomment-1538909692

PR #1510, opened 2024-02-08
Simplify migrations

### LINKED ISSUE #326: Simplify migrations
The current migrations workflow is quite difficult to maintain.

The goal was to give full flexibility, but after some usage, it seems to make things more difficult.

Instead of having a migrations definition such as this:

```yaml

_migrations:
  - version: v1.0.0
    before:
      - &mig
        - "[[ _copier_conf.src_path / 'migrations.py' ]]"
        - "[[ _copier_conf|tojson ]]"
    after:
      - *mig
  - version: v2
    before: [*mig]
    after:
      - *mig
      - "rm delete-in-migration-$VERSION_CURRENT.txt"
```

We should be having something like this:

```yaml
_migrations:
  working_directory: "[[ _copier_conf.src_path ]]" # default value
  command: [poetry, run, migrations.py]
```

Then, all data needed for migrations will be passed as env variables, which that script can load and use to handle all of them comfortably.

At the end of the day, that code is easier to maintain in a migrations script, and this way we avoid the need to repeat the same definition once and again for each version (as seen in the 1st example using `*mig`).

AdrianFreundQC:
Resolves #326 

Here is a suggestion for how to simplify migrations. Some changes made to migrations were also applied to tasks for consistency.
The task configuration is 100% backwards compatible. The migration configuration is not backwards compatible. The old style should be seen as deprecated. This PR support both old and new style and shows a warning when the old style is used.

## Tasks

Tasks can still be configured like before:
```yaml
_tasks:
  - git init
```
  
  Additionally Tasks can be a dictionary, which allows passing one or more of the optional arguments `when` and `working_directory`. `working_directory` defaults to the destination directory and can be either relative to it, or absolute. Both arguments can be templated.
  
```yaml
_tasks:
  - command: git init
    working_directory: "submodule"
    when: "{{ enable_submodules }}"
```

## Migrations

It is no longer necessary to duplicate the same command line for every version to run a migration script.
```yaml
_migrations:
  - "{{ _copier_python }} {{ _copier_conf.src_path }}/migrate.py"
```
This runs twice (before and after) on every update and still passes environment variables to the script.
Note that it can't pass `$VERSION_(PEP440_)CURRENT` because there is no _current_ version. Only a from and to version.

It is still possible to only run migrations on a specific version:
```yaml
_migrations:
  - version: v1.4
    command: rm old_file.txt || true
```
In this case `$VERSION_(PEP440_)CURRENT` is still passed to the command. 

This still runs both before and after.
`working_directory` and `when` are also supported for migrations. In addition all the environment variables passed to the command are also available for use in templating:

```yaml
_migrations:
  - version: v1.4
    command: rm old_file.txt
```

## Current State

All the features described here work.
I still want to do a bit of code cleanup and improve the documentation in the next few days.
I would also like some feedback on this design. Is there anything you don't agree with? Something I should change?


sisp (changes requested):
Thanks for taking the initiative and working on this long-standing feature proposal, @AdrianFreundQC! :pray:

I've done a first review iteration over the main code and docs (not the tests yet). It looks great, just a couple of minor comments so far.

copier/template.py
```diff
@@ -372,17 +387,44 @@ def migration_tasks(
             "VERSION_PEP440_FROM": str(from_template.version),
             "VERSION_PEP440_TO": str(self.version),
         }
+        extra_context = { "_" + key.lower(): value for key, value in extra_env.items()}
```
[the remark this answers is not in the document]
AdrianFreundQC: The one in the new format branch is only executed if a `version` is specified.
However, this is part of the code I wanted to clean up anyways.

docs/configuring.md
```diff
@@ -1146,26 +1146,29 @@ Like [`message_before_copy`][message_after_copy] but printed before
...
-    trigger this migration. It is evaluated using [PEP 440][].
--   **before** (optional): Commands to execute before performing the update. The answers
-    file is reloaded after running migrations in this stage, to let you migrate answer
-    values.
--   **after** (optional): Commands to execute after performing the update.
+-   **command**: The migration command to run
+-   **version** (optional): Indicates the version that the template update has to go
+    through to trigger this migration. It is evaluated using [PEP 440][]. If not version
+    is specified the migration will run on every update.
+-   **when** (optional): Specifies a condition that needs to hold for the task to run.
```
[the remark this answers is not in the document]
AdrianFreundQC: Yes. I added the jinja variables after writing the documentation. Will document them later.

docs/configuring.md
```diff
@@ -1191,13 +1194,12 @@ Migration processes will receive these environment variables:
...
-        - version: v1.0.0
-          before:
-              - rm ./old-folder
-          after:
-              # {{ _copier_conf.src_path }} points to the path where the template was
-              # cloned, so it can be helpful to run migration scripts stored there.
-              - invoke -r {{ _copier_conf.src_path }} -c migrations migrate $VERSION_CURRENT
+      # {{ _copier_conf.src_path }} points to the path where the template was
+      # cloned, so it can be helpful to run migration scripts stored there.
+      - invoke -r {{ _copier_conf.src_path }} -c migrations migrate $VERSION_FROM $VERSION_TO
```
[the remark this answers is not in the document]
AdrianFreundQC: I know that it's different, but it's just an arbitrary example.
I just wanted to go with something that clearly shows how it works.

I guess I can add $STAGE, but in reality none of those would have to be passed as arguments, as they are available as env variables to the task.

AdrianFreundQC:
I pushed the fixes as new commits to simplify review.
I can rebase them back into three commits (feat, test, docs) or you can just squash merge if you prefer that.

AdrianFreundQC:
CI failure looks unrelated. Should be fixed with a re-run

docs/configuring.md
```diff
@@ -1191,15 +1199,29 @@ Migration processes will receive these environment variables:
...
-              - invoke -r {{ _copier_conf.src_path }} -c migrations migrate $VERSION_CURRENT
+      # {{ _copier_conf.src_path }} points to the path where the template was
+      # cloned, so it can be helpful to run migration scripts stored there.
+      - invoke -r {{ _copier_conf.src_path }} -c migrations migrate $STAGE $VERSION_FROM $VERSION_TO
+      - version: v1.0.0
+        command: rm ./old-folder
+        when: "{{ _stage == 'before' }}"
     ```
 
+In copier versions before v10.0.0 a different configuration format had to be used. This
```
AdrianFreundQC, line 1210: Or v9.2.0?

freundTech:
One more thought I had: the migration command running twice on every update might be confusing and unexpected for new users, however it is required for more complex migrations.

Changing the default value of `when` from `true` to `{{ _stage == "after" }}` might solve this issue. Migrations without `when` only run once, while experienced users can still set `when` to `true` to have them run in the before and after stage.

AdrianFreundQC:
One more thought I had: the migration command running twice on every update might be confusing and unexpected for new users, however it is required for more complex migrations.

Changing the default value for `when` from `true` to `{{ _state == "after" }}` might solve this issue. Migrations without `when` only run once, while experience users can still set `when` to `true` to have them run in the before and after stage.

sisp:
I like this idea, it sounds like a better default behavior than running the command twice. 👌

By the way, I haven't forgotten about reviewing this PR, just haven't had enough time yet. 🙁 I'll try to do it on Monday. 🤞

tests/test_migrations.py
sisp, line 1: I wonder whether we should rename the original version of this file to `test_migrations_legacy.py` and create a copy named `test_migrations.py` which contains the same tests using the new format _plus_ additional tests covering new aspects like:

* The availability of Jinja variables equivalent to the environment variables.
* The default `when` behavior (e.g., `when: "{{ _stage == 'after' }}"` [as you suggested](https://github.com/copier-org/copier/pull/1510#issuecomment-1962313028).
* The behavior of, e.g., `when: true` and possibly other typical cases.
* The default working directory.
* The behavior of a custom working directory.
* More?

This way, we can ensure:

* The legacy format remains functional as tested by the original tests.
* The original tests adapted to the new format also pass for the new format.
* The additional capabilities of the new format are tested.
* When we decide to drop support for the legacy format, we only need to delete `test_migrations_legacy.py` and the demo fixture instead of surgically removing code associated with the legacy format.

WDYT, @AdrianFreundQC and @copier-org/maintainers?

tests/test_tasks.py
sisp, line 1: Same thought as for `test_migrations.py`.

sisp (changes requested):
I've done a full review this time. The implementation looks good to me. The legacy format test does not ensure identical behavior as before anymore, and I think the new format needs a bit more systematic testing. I think keeping the legacy tests unchanged and adapting and extending a copy thereof for the new format would be better than surgically adding some testing of the new format and mixing it with the old format (see my inline comment for more details).

docs/configuring.md
```diff
@@ -1146,43 +1146,51 @@ Like [`message_before_copy`][message_after_copy] but printed before
...
--   **version**: Indicates the version that the template update has to go through to
-    trigger this migration. It is evaluated using [PEP 440][].
--   **before** (optional): Commands to execute before performing the update. The answers
-    file is reloaded after running migrations in this stage, to let you migrate answer
-    values.
--   **after** (optional): Commands to execute after performing the update.
+-   **command**: The migration command to run
+-   **version** (optional): Indicates the version that the template update has to go
+    through to trigger this migration. It is evaluated using [PEP 440][]. If not version
+    is specified the migration will run on every update.
```
yajo, line 1158: suggestion(docs): I think it's worth to mention that using ˋversionˋ can still save you from being marked as an unsafe template for every update, because the unsafe part will only happen in *that* specific update, and not in *any* update.

yajo (changes requested):
This definitely looks like the path forward.  Thanks for taking care of all this!

docs/configuring.md
```diff
@@ -1146,43 +1146,51 @@ Like [`message_before_copy`][message_after_copy] but printed before
...
-    migration tasks.
--   `$VERSION_PEP440_FROM`, `$VERSION_PEP440_TO`, `$VERSION_PEP440_CURRENT`: Same as the
-    above, but normalized into a standard [PEP 440][] version string indicator. If your
-    scripts use these environment variables to perform migrations, you probably will
-    prefer to use these variables.
+There are a number of additional variables available for templating of migrations. Those
+variables are also passed to the migration process as environment variables. Migration
+processes will receive a number of these environment variables:
+
+-   `_stage`/`$STAGE`: Either `before` or `after`.
```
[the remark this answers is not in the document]
sisp: Or simply `_task` dict? "env" carries some semantics of environment variables IMO.

copier/main.py
```diff
@@ -277,7 +282,15 @@ def _execute_tasks(self, tasks: Sequence[Task]) -> None:
                 )
             if self.pretend:
                 continue
-            with local.cwd(self.subproject.local_abspath), local.env(**task.extra_env):
+
+            working_directory = (
+                    # We can't use _render_path here, as that function has special handling for files in the template
+                    self.subproject.local_abspath / Path(self._render_string(str(task.working_directory), extra_context))
```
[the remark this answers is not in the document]
AdrianFreundQC: I don't think that's needed, but it could make the code a bit clearer.
The current solution also allows using absolute paths:
```python
>>> from pathlib import Path
>>> p = Path("/user/test")
>>> p
PosixPath('/user/test')
>>> p / "/user/other"
PosixPath('/user/other')
```

docs/configuring.md
```diff
@@ -1191,15 +1199,29 @@ Migration processes will receive these environment variables:
...
+
+-   **version**: Indicates the version that the template update has to go through to
+    trigger this migration. It is evaluated using [PEP 440][].
+-   **before** (optional): Commands to execute before performing the update. The answers
+    file is reloaded after running migrations in this stage, to let you migrate answer
+    values.
+-   **after** (optional): Commands to execute after performing the update.
+
+The migration variables mentioned above are available as environment variables, but
+can't be used in jinja templates.
```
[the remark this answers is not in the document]
AdrianFreundQC: They have once this is merged, but this part talks about older versions.
The only case where you can use the Jinja variables is if you already have the new version, but are using the old configuration format.

docs/configuring.md
pawamoy: Also "if not version" -> "if no version".

docs/configuring.md
yajo: Yes

copier/main.py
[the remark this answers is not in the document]
yajo: Ah I didn't know that, thanks.

docs/configuring.md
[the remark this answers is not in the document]
yajo: Well, both options have pros and cons, so I think it's fair to respect your strong preference, since you're the one contributing this :)

docs/configuring.md
```diff
@@ -1392,14 +1414,24 @@ This allows you to keep separate the template metadata and the template code.
...
 -   Default value: `[]`
 
 Commands to execute after generating or updating a project from your template.
 
 They run ordered, and with the `$STAGE=task` variable in their environment.
 
+If a `dict` is given it can contain the following items:
+
+-   **command**: The task command to run.
+-   **when** (optional): Specifies a condition that needs to hold for the task to run.
```
hparfr, line 1428: thanks for this work

>Commands to execute after generating or updating a project from your template.

Because tasks are run after generating AND updating, is it possible to test this in the "when" clause ?

```yaml
- command:  git clone something in_directory/
  when : `copier copy`

- command: cd in_directory && git pull
  when:  `copier update`
```

?

sisp: Work on supporting pre/post copy/update tasks has started in #1511 and will probably build upon this PR.

docs/configuring.md
```diff
@@ -1414,12 +1446,10 @@ They run ordered, and with the `$STAGE=task` variable in their environment.
...
         # OS-specific task (supported values are "linux", "macos", "windows" and `None`)
-        - >-
-          {% if _copier_conf.os in ['linux', 'macos'] %}
-          rm {{ name_of_the_project }}/README.md
-          {% elif _copier_conf.os == 'windows' %}
-          Remove-Item {{ name_of_the_project }}/README.md
-          {% endif %}
+        - command: rm {{ name_of_the_project }}/README.md
+          when: "{{ _copier_conf.os in  ['linux', 'macos'] }}"
+        - command: Remove-Item {{ name_of_the_project }}/README.md
```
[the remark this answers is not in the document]
AdrianFreundQC: I just copied the command from the old version, where it already used a `/`.
I don't know if PowerShell can also deal with `/`, but I can just change it to `\`

AdrianFreundQC:
Sorry for the delay. I've been quite busy recently.
I'm still working on the tests for the new migrations implementation.
I'll have some time in the next two weeks and if that's not enough I'll probably be able to finish it in the first half of May.

sisp:
Thanks for the update and your continuous work on this PR, @AdrianFreundQC! :pray:

AdrianFreundQC:
I have no idea why the MacOS flake test is suddenly failing now. Maybe a broken cache?

> /Users/runner/work/_temp/be5ceacc-d596-4a1f-b513-54880e12e09d.sh: /Users/runner/work/copier/copier/.venv/bin/copier: /Users/runner/work/copier/copier/.venv/bin/python: bad interpreter: No such file or directory

The flake check runs fine on my machine.

sisp:
@AdrianFreundQC Sorry for the inconvenience. It's because `macos-latest` on GitHub-hosted runners got bumped to macOS 14 (ARM) for which there are no binaries for Python 3.8 and 3.9. Before, `macos-latest` was referring to macOS 12 (x64) for which those binaries existed. See https://github.com/copier-org/copier/pull/1602#pullrequestreview-2019446447 for more details.

We're waiting for @yajo to resolve #1604 (because of a project-level settings change only he can make). Then, after rebasing your PR, CI checks should work again.

docs/configuring.md
AdrianFreundQC: Done.

AdrianFreundQC:
I think all discussions should have been addressed now and the CI seems to be working (assuming my last change didn't break anything).

tests/test_migrations.py
AdrianFreundQC: I kept the legacy test as is, but started from scratch for the new tests, to produce more maintainable and better structured tests. I think they should cover everything.

AdrianFreundQC:
I think all discussions should have been addressed now and the CI seems to be working (assuming my last change didn't break anything).

docs/configuring.md
```diff
@@ -1146,43 +1146,53 @@ Like [`message_before_copy`][message_after_copy] but printed before
...
+    template as it was before updating.
+-   `_version_to`/`$VERSION_TO`: [Git commit description][git describe] of the template
+    as it will be after updating.
+-   `_version_current`/`$VERSION_CURRENT`: The `version` detector as you indicated it
+    when describing migration tasks (only when `version` is given).
+-   `_version_pep440_from`/`$VERSION_PEP440_FROM`,
+    `_version_pep440_to`/`$VERSION_PEP440_TO`,
+    `_version_pep440_current`/`$VERSION_PEP440_CURRENT`: Same as the above, but
+    normalized into a standard [PEP 440][] version. In Jinja templates these are
+    represented as `packaging.version.Version` objects and allow access to their
```
yajo, line 1193: suggestion: it'd be cool to link to docs of this object if possible.

```suggestion
    represented as [packaging.version.Version](https://packaging.pypa.io/en/stable/version.html#packaging.version.Version) objects and allow access to their
```

yajo (approved):
Very amazing work, Thanks!

There's just one minor suggestion to enhance docs. Would you like to push it before merging?

docs/configuring.md
AdrianFreundQC: Done!

sisp (changes requested):
Looks great, @AdrianFreundQC! :ok_hand: Just some totally minor remarks from my side. Plus in general, some lines seem very long. Would you mind running Ruff's formatter once?

About code formatting: IIRC, there is no Nix-based pre-commit hook for Ruff's formatter, which would explain why the `flake-check` checks pass although I'm sure Ruff's formatter would disagree. @yajo Is there a way to enable Ruff's formatter to check code formatting? And does the Nix-locked version of Ruff even have the formatter already? We might need to update the lock file.

AdrianFreundQC:
I ran addressed the two comments above and ran `ruff format` (nix pinned version) on `main.py` and `template.py`. I didn't run it on all files to not clutter this PR.

I also rebased on master again.

tests/test_migrations.py
```diff
@@ -1,311 +1,516 @@
...
+
+    assert not (dst / "foo").exists()  # Migrations don't run on initial copy
+
+    with local.cwd(src):
+        git("tag", "v2")
+    with local.cwd(dst):
+        run_update(defaults=True, overwrite=True, unsafe=True, pretend=True)
+
+    assert not (dst / "foo").exists()  # In pretend mode the command shouldn't run
+
```
sisp, line 285: One more empty line here:

```suggestion


```

tests/test_migrations.py
```diff
@@ -1,311 +1,516 @@
...
+    assert not (dst / "foo").exists()  # Migrations don't run on initial copy
+
+    with local.cwd(src):
+        git("tag", "v2")
+    with local.cwd(dst):
+        run_update(defaults=True, overwrite=True, unsafe=True, skip_tasks=True)
+
+    assert (dst / "foo").exists()  # Migrations are not skipped by skip_tasks
+
```
sisp, line 317: One more empty line here:

```suggestion


```

tests/test_migrations.py
```diff
@@ -1,311 +1,516 @@
...
+    with local.cwd(src):
+        build_file_tree({"foo": ""})
+        git_save(tag="v2")
+    with local.cwd(dst):
+        run_update(defaults=True, overwrite=True, unsafe=True)
+
+    assert not (dst / "foo").exists()
+    assert (dst / "bar").is_file()
+
```
sisp, line 402: One more empty line here:

```suggestion


```

tests/test_migrations.py
```diff
@@ -1,311 +1,516 @@
...
+@pytest.mark.parametrize("with_version", [True, False])
+def test_migration_env_variables(tmp_path_factory: pytest.TempPathFactory, with_version: bool) -> None:
```
sisp, line 404: This line is too long. Could you run Ruff's formatter on this file?

tests/test_migrations.py
```diff
@@ -1,311 +1,516 @@
...
+    assert (dst / "env.txt").is_file()
+    env = (dst / "env.txt").read_text().split("\n")
+    for variable, value in variables.items():
+        assert f"{variable}={value}" in env
+
+    for variable, value in current_only_variables.items():
+        assert (f"{variable}={value}" in env) == with_version
+
+@pytest.mark.parametrize("with_version", [True, False])
+def test_migration_jinja_variables(tmp_path_factory: pytest.TempPathFactory, with_version: bool) -> None:
```
sisp, line 458: This line is too long. Could you run Ruff's formatter on this file?

tests/test_tasks.py
```diff
@@ -38,7 +44,7 @@ def template_path(tmp_path_factory: pytest.TempPathFactory) -> str:
 
 
 def test_render_tasks(template_path: str, tmp_path: Path) -> None:
-    copier.run_copy(template_path, tmp_path, data={"other_file": "custom"}, unsafe=True)
+    copier.run_copy(template_path, tmp_path, data={"other_file": "custom", "condition": "true"}, unsafe=True)
```
sisp, line 47: This line is too long. Could you run Ruff's formatter on this file?

tests/test_migrations.py
```diff
@@ -1,311 +1,516 @@
...
+        "_version_pep440_from": "1",
+        "_version_pep440_to": "3",
+    }
+    current_only_variables = {
+        "_version_current": "v2",
+        "_version_pep440_current": "2",
+    }
+    all_variables = {**variables, **current_only_variables}
+
+    command = "&&".join(f"echo {var}={{{{ {var} }}}} >> vars.txt" for var in all_variables)
```
sisp, line 475: This line is too long. Could you run Ruff's formatter on this file?

sisp (changes requested):
Sorry for the delay. 🫣

Content-wise it looks good to me. I still see some minor formatting issues. Would you mind trying to fix them?

AdrianFreundQC:
No problem.
I reformatted the files and rebased on master again.

sisp (approved):
Looks good now! :tada: Thank you very much for the hard work, implementing this great feature with proper tests, and patience with the reviews even on pedantic formatting fixes! :pray: Fantastic job! :1st_place_medal: Let's get this merged! :rocket:

AdrianFreundQC:
Do you have an ETA on when the next version might get released?

yajo:
I'd like to at least have https://github.com/copier-org/copier/issues/1598 fixed before release. It appeared because of this merge IIRC. It should be quite easy to fix though, I just had very little time lately. Once that one's fixed I'll push the new release. Would you want to contribute it?

yajo:
Ah, nevermind. I just opened https://github.com/copier-org/copier/pull/1681. It was an easy one.

PR #1511, opened 2024-02-09
WIP, add pre/post copy and pre/post update tasks

hahuang65:
Hacking around for a potential solution to https://github.com/copier-org/copier/issues/240

Would love some feedback on whether we think this is a feasible solution or not.

If so, I need a little guidance on how to write some tests for the update tasks (they seem a bit difficult to test as we need to set up some Git repos), as well as how to test the difference between a pre and a post task.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -220,11 +220,27 @@ class Worker:
             features.add("jinja_extensions")
         if self.template.tasks:
             features.add("tasks")
+        if self.template.pre_copy:
+            features.add("pre_copy")
+        if self.template.pre_update:
+            features.add("pre_update")
+        if self.template.post_copy:
+            features.add("post_copy")
+        if self.template.post_update:
+            features.add("post_update")
         if mode == "update" and self.subproject.template:
             if self.subproject.template.jinja_extensions:
                 features.add("jinja_extensions")
             if self.subproject.template.tasks:
                 features.add("tasks")
+            if self.subproject.template.pre_copy:
+                features.add("pre_copy")
+            if self.subproject.template.pre_update:
+                features.add("pre_update")
+            if self.subproject.template.post_copy:
+                features.add("post_copy")
+            if self.subproject.template.post_update:
+                features.add("post_update")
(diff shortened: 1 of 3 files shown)


sisp (commented):
Thanks for taking the initiative, @hahuang65! :pray:

Just a design feedback to start with: Having new top-level keywords `_{pre,post}_{copy,update}` looses the context that those are actually task definitions. How about extending the `_tasks` syntax similar to #326 and #1510?

```yaml
_tasks:
  # implicitly a post-copy task
  - touch hello.txt
  # explicitly a pre-copy task using Jinja
  - "{% if _stage == 'before_copy' %}touch hello.txt{% endif %}"
  # uses environment variable `STAGE={before,after}_{copy,update}`
  - ./task.sh
  # new convenient syntax (#326, #1510)
  - command: touch hello.txt
```

This would be backwards compatible when using the old syntax and add convenience for conditional tasks when using the new syntax.

hahuang65:
thanks for the feedback @sisp!

Does `_tasks` already support templated code in it? To me, adding that may make the yaml file have less sections, but certainly adds complexity to the existing sections. More to maintain and more to test. I don't believe the syntax to be quite readable too.

It also makes it much more difficult to simply port an existing shell script into the `_tasks` section.

Those are just my thoughts. I'm happy to oblige with your design suggestion, especially if it's already been pre-established.

sisp:
@hahuang65 Yes, `_tasks` supports templating already; see [its docs](https://copier.readthedocs.io/en/v9.1.1/configuring/#tasks).

As it turns out, #1510 is already adding the new, more convenient syntax not only for migrations but also for tasks. It might make sense to wait for it to be merged and then build this PR upon it.

sisp:
@hahuang65 If you're interested in continuing work on this PR, you may now build upon the implementation of #1510 that we just merged.

hahuang65:
@sisp exciting! Unfortunately, probably for the next 2 to 3 months I will be busy. If someone would like to take over, feel free. Otherwise, I'll re-visit in a few months.

PR #1513, opened 2024-02-11
remove 'decorator' dependency

danieleades:
besides removing an unneeded dependency (always good), using a 'native' decorator function allows for creating a fully-typed decorator

- decorated methods are now fully type annotated
- unblocks enabling stricter mypy configuration
- helps solve the errors in #1506...

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -47,8 +47,9 @@ CLI help generated from `copier --help-all`:
 copier --help-all
 ```
 """
+from __future__ import annotations
 
-import functools
+import inspect
 import sys
 from os import PathLike
 from pathlib import Path
@@ -68,7 +69,6 @@ P = ParamSpec("P")
 
 
 def _handle_exceptions(method: Callable[P, int]) -> Callable[P, int]:
-    @functools.wraps(method)
     def inner(*args: P.args, **kwargs: P.kwargs) -> int:
         try:
             try:
@@ -82,6 +82,9 @@ def _handle_exceptions(method: Callable[P, int]) -> Callable[P, int]:
             print(colors.red | "\n".join(error.args), file=sys.stderr)
             # DOCS https://github.com/copier-org/copier/issues/1328#issuecomment-1723214165
             return 0b100
+
+    # See https://github.com/copier-org/copier/pull/1513
+    inner.__signature__ = inspect.signature(method, eval_str=True)  # type: ignore[attr-defined]
 
     return inner
 
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -84,18 +84,20 @@ class CopierApp(cli.Application):
     """The Copier CLI application."""
 
     DESCRIPTION = "Create a new project from a template."
-    DESCRIPTION_MORE = dedent(
-        """\
-            Docs in https://copier.readthedocs.io/
-
-            """
-    ) + (
-        colors.yellow
-        | dedent(
+    DESCRIPTION_MORE = (
+        dedent(
             """\
+            Docs in https://copier.readthedocs.io/
+            """
+        )
+        + (
+            colors.yellow
+            | dedent(
+                """\
                 WARNING! Use only trusted project templates, as they might
                 execute code with the same level of access as your user.\n
                 """
+            )
         )
     )
     VERSION = copier_version()
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -47,6 +47,7 @@ CLI help generated from `copier --help-all`:
 copier --help-all
 ```
 """
+
 import sys
 from os import PathLike
 from pathlib import Path
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -47,7 +47,7 @@ CLI help generated from `copier --help-all`:
 copier --help-all
 ```
 """
-import inspect
+
 import sys
 from os import PathLike
 from pathlib import Path
@@ -56,38 +56,28 @@ from typing import Callable
 
 import yaml
 from plumbum import cli, colors
-from typing_extensions import ParamSpec
 
 from .errors import UnsafeTemplateError, UserMessageError
 from .main import Worker
 from .tools import copier_version
 from .types import AnyByStrDict, OptStr, StrSeq
 
-P = ParamSpec("P")
 
-
-def _handle_exceptions(method: Callable[P, int]) -> Callable[P, int]:
-    def inner(*args: P.args, **kwargs: P.kwargs) -> int:
+def handle_exceptions(method: Callable[[], None]) -> int:
+    """Handle keyboard interruption while running a method."""
+    try:
         try:
-            try:
-                return method(*args, **kwargs)
-            except KeyboardInterrupt:
-                raise UserMessageError("Execution stopped by user")
-        except UserMessageError as error:
-            print(colors.red | "\n".join(error.args), file=sys.stderr)
-            return 1
-        except UnsafeTemplateError as error:
-            print(colors.red | "\n".join(error.args), file=sys.stderr)
-            # DOCS https://github.com/copier-org/copier/issues/1328#issuecomment-1723214165
-            return 0b100
-
-    # See https://github.com/copier-org/copier/pull/1513
-    if sys.version_info >= (3, 10):
-        inner.__signature__ = inspect.signature(method, eval_str=True)  # type: ignore[attr-defined]
-    else:
-        inner.__signature__ = inspect.signature(method)  # type: ignore[attr-defined]
-
-    return inner
+            method()
+        except KeyboardInterrupt:
+            raise UserMessageError("Execution stopped by user")
+    except UserMessageError as error:
+        print(colors.red | "\n".join(error.args), file=sys.stderr)
+        return 1
+    except UnsafeTemplateError as error:
+        print(colors.red | "\n".join(error.args), file=sys.stderr)
+        # DOCS https://github.com/copier-org/copier/issues/1328#issuecomment-1723214165
+        return 0b100
+    return 0
 
 
 class CopierApp(cli.Application):
@@ -254,7 +244,6 @@ class CopierCopySubApp(_Subcommand):
         help="Overwrite files that already exist, without asking.",
     )
 
-    @_handle_exceptions
     def main(self, template_src: str, destination_path: str) -> int:
         """Call [run_copy][copier.main.Worker.run_copy].
 
@@ -274,8 +263,7 @@ class CopierCopySubApp(_Subcommand):
             defaults=self.force or self.defaults,
             overwrite=self.force or self.overwrite,
         ) as worker:
-            worker.run_copy()
-        return 0
+            return handle_exceptions(worker.run_copy)
 
 
 @CopierApp.subcommand("recopy")
@@ -322,7 +310,6 @@ class CopierRecopySubApp(_Subcommand):
         help="Skip questions that have already been answered",
     )
 
-    @_handle_exceptions
     def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
         """Call [run_recopy][copier.main.Worker.run_recopy].
 
@@ -340,8 +327,7 @@ class CopierRecopySubApp(_Subcommand):
             overwrite=self.force or self.overwrite,
             skip_answered=self.skip_answered,
         ) as worker:
-            worker.run_recopy()
-        return 0
+            return handle_exceptions(worker.run_recopy)
 
 
 @CopierApp.subcommand("update")
@@ -394,7 +380,6 @@ class CopierUpdateSubApp(_Subcommand):
         help="Skip questions that have already been answered",
     )
 
-    @_handle_exceptions
     def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
         """Call [run_update][copier.main.Worker.run_update].
 
@@ -414,5 +399,4 @@ class CopierUpdateSubApp(_Subcommand):
             skip_answered=self.skip_answered,
             overwrite=True,
         ) as worker:
-            worker.run_update()
-        return 0
+            return handle_exceptions(worker.run_update)
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -48,57 +48,60 @@ copier --help-all
 ```
 """
 
+import functools
 import sys
 from os import PathLike
 from pathlib import Path
 from textwrap import dedent
+from typing import Callable
 
 import yaml
-from decorator import decorator
 from plumbum import cli, colors
+from typing_extensions import ParamSpec
 
 from .errors import UnsafeTemplateError, UserMessageError
 from .main import Worker
 from .tools import copier_version
 from .types import AnyByStrDict, OptStr, StrSeq
 
+P = ParamSpec("P")
 
-@decorator
-def handle_exceptions(method, *args, **kwargs):
-    """Handle keyboard interruption while running a method."""
-    try:
+
+def _handle_exceptions(method: Callable[P, int]) -> Callable[P, int]:
+    @functools.wraps(method)
+    def inner(*args: P.args, **kwargs: P.kwargs) -> int:
         try:
-            return method(*args, **kwargs)
-        except KeyboardInterrupt:
-            raise UserMessageError("Execution stopped by user")
-    except UserMessageError as error:
-        print(colors.red | "\n".join(error.args), file=sys.stderr)
-        return 1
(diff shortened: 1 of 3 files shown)


sisp (changes requested):
It looks like you ran `poetry lock` after you had removed the `decorator`/`types-decorator` dependencies from `pyproject.toml`, which resulted in a full relock of the entire dependency tree. To avoid mixing concerns, could you please restore the old version of `poetry lock` and run `poetry lock --no-update` instead? The changes to `poetry.lock` should be limited to removing the `decorator`-related dependencies then.

danieleades:
@sisp changes to lockfile reverted

poetry.lock
```diff
@@ -1232,24 +1220,15 @@ files = [
...
     {file = "PyYAML-6.0.1-cp311-cp311-win_amd64.whl", hash = "sha256:bf07ee2fef7014951eeb99f56f39c9bb4af143d8aa3c21b1677805985307da34"},
-    {file = "PyYAML-6.0.1-cp312-cp312-macosx_10_9_x86_64.whl", hash = "sha256:855fb52b0dc35af121542a76b9a84f8d1cd886ea97c84703eaa6d88e37a2ad28"},
-    {file = "PyYAML-6.0.1-cp312-cp312-macosx_11_0_arm64.whl", hash = "sha256:40df9b996c2b73138957fe23a16a4f0ba614f4c0efce1e9406a184b6d07fa3a9"},
-    {file = "PyYAML-6.0.1-cp312-cp312-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:a08c6f0fe150303c1c6b71ebcd7213c2858041a7e01975da3a99aed1e7a378ef"},
-    {file = "PyYAML-6.0.1-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:6c22bec3fbe2524cde73d7ada88f6566758a8f7227bfbf93a408a9d86bcc12a0"},
-    {file = "PyYAML-6.0.1-cp312-cp312-musllinux_1_1_x86_64.whl", hash = "sha256:8d4e9c88387b0f5c7d5f281e55304de64cf7f9c0021a3525bd3b1c542da3b0e4"},
-    {file = "PyYAML-6.0.1-cp312-cp312-win32.whl", hash = "sha256:d483d2cdf104e7c9fa60c544d92981f12ad66a457afae824d146093b8c294c54"},
-    {file = "PyYAML-6.0.1-cp312-cp312-win_amd64.whl", hash = "sha256:0d3304d8c0adc42be59c5f8a4d9e3d7379e6955ad754aa9d6ab7a398b59dd1df"},
```
sisp, line 1252: Which version of Poetry are you using? I wonder why, e.g., these locked version have been removed. 🤔 Specifically here, all Python 3.12 wheel files have been removed; this can't be good.

copier/cli.py
```diff
@@ -48,57 +48,60 @@
...
             """
-        )
-        + (
-            colors.yellow
-            | dedent(
-                """\
+    ) + (
+        colors.yellow
+        | dedent(
+            """\
```
sisp, line 101: I don't think there should be any formatting changes here in this PR. Is perhaps your IDE using a more recent version of Ruff than Copier?

sisp (commented):
I have a few inline questions.

Do you happen to know why some tests are failing now? Off the top of my head, I'm not sure why these changes would fail those tests. 🤔

poetry.lock
danieleades: i did think that was a touch odd

```
poetry --version
Poetry (version 1.7.1)
```

copier/cli.py
danieleades: that's likely the case. I wouldn't normally use lint-on-save anyway, but the `.vscode/settings.json` in copier's repo is intrusive.

configuring the repo to use an extension for formatting this way (instead of, say, `poetry run ruff format .`) effectively means you're using an unpinned formatter version and exposing PRs to these kinds of changes.

danieleades:
> I have a few inline questions.
> 
> Do you happen to know why some tests are failing now? Off the top of my head, I'm not sure why these changes would fail those tests. 🤔

i'm very confused about the test failures, because i don't see those locally...

copier/cli.py
sisp: Could you please revert those lines?

We should revise the VS Code settings in a separate PR. Perhaps there's a way to disable using the version of Ruff bundled with its VS Code extension.

poetry.lock
sisp: Odd. Could you try running `poetry update pexpect pyyaml` and check whether those deleted lines get re-added? Might be a Poetry bug ...

danieleades:
so this turned out to be a bit of a rabbit hole.

I can fix the issue with `plumbum` easily enough by adding `inner.__signature__ = inspect.signature(method)`:

```python
def _handle_exceptions(method: Callable[P, int]) -> Callable[P, int]:
    def inner(*args: P.args, **kwargs: P.kwargs) -> int:
        try:
            try:
                return method(*args, **kwargs)
            except KeyboardInterrupt:
                raise UserMessageError("Execution stopped by user")
        except UserMessageError as error:
            print(colors.red | "\n".join(error.args), file=sys.stderr)
            return 1
        except UnsafeTemplateError as error:
            print(colors.red | "\n".join(error.args), file=sys.stderr)
            # DOCS https://github.com/copier-org/copier/issues/1328#issuecomment-1723214165
            return 0b100

    # See https://github.com/copier-org/copier/pull/1513
    inner.__signature__ = inspect.signature(method)  # type: ignore[attr-defined]

    return inner
```

that works perfectly

until i add `from __future__ import annotations`...

Adding that changes the behaviour of the `inspect.signature` method - https://bugs.python.org/issue43355, which is probably what was causing the issue in #1506 in the first place.
I can resolve *that* by adding `eval_str=True` to `inspect.signature`, but that flag is only supported in python 3.10+

danieleades:
@sisp would you kindly trigger the CI for me?

or better yet, merge the fairly trivial #1505 so that CI will auto-trigger for my PRs?

poetry.lock
danieleades: > poetry update pexpect pyyaml

no effect

sisp:
Thanks for tracing this problem, @danieleades! 🙇

So IIUC, with and without the `decorator` library is `from __future__ import annotations` a problem because it requires `inspect.signature`'s `eval_str=True`, which isn't available before Python 3.10. Correct?

So, #1506 is blocked in any case as long as we need to support Python 3.8 & 3.9, which we need to keep supporting until EOL. This means, we can't use a decorator if we want #1506. Right? Perhaps we could replace the decorator by a context manager?

```python
def main(self, ...):
    with _handle_exceptions():
        ...
```

It would be less elegant but should solve the problem unless I'm missing something. WDYT?

danieleades:
> Thanks for tracing this problem, @danieleades! 🙇
> 
> So IIUC, with and without the `decorator` library is `from __future__ import annotations` a problem because it requires `inspect.signature`'s `eval_str=True`, which isn't available before Python 3.10. Correct?
> 
> So, #1506 is blocked in any case as long as we need to support Python 3.8 & 3.9, which we need to keep supporting until EOL. This means, we can't use a decorator if we want #1506. Right? Perhaps we could replace the decorator by a context manager?
> 
> ```python
> def main(self, ...):
>     with _handle_exceptions():
>         ...
> ```
> 
> It would be less elegant but should solve the problem unless I'm missing something. WDYT?

i think a context manager would be less 'magic' to be honest, so might be worth considering

I think this can be made to work as is, just that `from __future__ import annotations` can't be added to this specific file until after dropping support for python 3.9. Or maybe it's possible to upstream a fix into `plumbum.cli`?

it's *less* clear now that this PR adds value, but i personally would argue its still a marginal improvement

sisp:
If an upstream fix in `plumbum.cli` would help, that might be a good way forward. Omitting `from __future__ import annotations` from this file might be okay, too, although I tend to prefer consistency.

I'm a bit undecided about the best way forward (also regarding unblocking #1506). Any opinions, @copier-org/maintainers?

1. Drop `decorators`/`types-decorators` and use a context manager for handling exceptions.
1. Omit `from __future__ import annotations` from this file and wait until Python 3.9 reaches EOL.
1. Attempt to find an upstream fix in `plumbum.cli` to support decorators with `from __future__ import annotations` without requiring `inspect.signature`'s `eval_str` argument.
1. Other ideas?

danieleades:
> 4\. Other ideas?

just using a plain method to handle errors seems pretty clean to me. Have updated the PR to use this approach

copier/cli.py
danieleades: see https://github.com/copier-org/copier/pull/1514

turns out the ruff extension prefers the local ruff version *if present* and falls back to the bundled version. Since Ruff wasn't in the dev dependencies different contributors could get different versions depending on their setup

danieleades:
>If an upstream fix in plumbum.cli would help, that might be a good way forward

i did open an upstream issue, but to be honest i think sidestepping the decorator approach entirely is probably the way forward here. It's easier to reason about, and is actually less lines of code

danieleades:
@sisp will you squash merge this, or would you prefer me to squash the commits? What's the preference in this repo?

sisp:
Both squash-merge and rebase-merge are fine. I'd squash-merge this PR, but I'd prefer waiting just a little longer for some feedback from at least one of the other @copier-org/maintainers.

The minor downside of the function approach is that it only handles exceptions for the wrapped method call but not for the surrounding context manager that instantiates the `Worker` class. Currently, not exception should occur there, but there's a slight difference between the decorator and the function approach nevertheless.

pawamoy:
I agree with @sisp: the worker context manager will let exceptions bubble up on exit, and the function call approach does not catch those. So I'd recommend using a context manager for handling exceptions instead instead.

sisp:
@pawamoy Do you happen to have an idea why some locked `pyyaml` wheel files for Python 3.12 are removed by Poetry when running `poetry lock --no-update`?

pawamoy:
@sisp I don't! I stopped using Poetry more than a year ago :slightly_smiling_face: 
The removed entries are from other systems. Looks like the lock file stops being cross-platform or something? Dependabot might do something differently :shrug:

sisp:
Thanks, @pawamoy! :pray:

@danieleades, could you try restoring `poetry.lock` from `master` as @pawamoy described?

danieleades:
> Thanks, @pawamoy! :pray:
> 
> @danieleades, could you try restoring `poetry.lock` from `master` as @pawamoy described?

I have tried that, doesn't help.
Might be to do with my environment - I'm not using the Nix environment.

If I get the time tonight I'll have a play around with it

danieleades:
> I agree with @sisp: the worker context manager will let exceptions bubble up on exit, and the function call approach does not catch those. So I'd recommend using the context manager approach for handling exceptions instead.

i've used the function approach to also wrap the context manager. It's starting to get a *tiny* bit unwieldy.

it's not at all clear to me how you'd do this with a context manager, as a context manager doesn't allow you to either return a value or re-raise exceptions. What would a context manager based approach look like?

sisp:
Reraising seems to be possible when the `__exit__` method returns `False`: https://stackoverflow.com/a/18399069 But I had only thought about catching exceptions via the context manager but not about returning a corresponding CLI return code. 🤔

I think your current approach works. Just one more idea: Would keeping the original decorator approach by replacing the `decorator` library by [`wrapt`](https://github.com/GrahamDumpleton/wrapt) work? `wrapt` doesn't seem to have type hints though. 🙁

danieleades:
> Reraising seems to be possible when the `__exit__` method returns `False`

sort of. That doesn't give you the opportunity to change the type of the exception. the current code does this-

```python
        try:
            return method(*args, **kwargs)
        except KeyboardInterrupt:
            raise UserMessageError("Execution stopped by user")
```

i can't think how you can *both* re-raise the `UserMessageError` *and* return `False`.

While i think of it, this code snippet should probably be:

```python
        try:
            return method(*args, **kwargs)
        except KeyboardInterrupt as e:
            raise UserMessageError("Execution stopped by user") from e
```

danieleades:
another approach that would work would be to go back to the decorator approach (without the `decorator` dependency) and add a second inner method directly in the class, like so-

```python
def handle_exceptions(method: Callable[P, int]) -> Callable[P, int]:
    def inner(*args: P.args, **kwargs: P.kwargs) -> int:
        try:
            try:
                return method(*args, **kwargs)
            except KeyboardInterrupt:
                raise UserMessageError("Execution stopped by user")
        except UserMessageError as error:
            print(colors.red | "\n".join(error.args), file=sys.stderr)
            return 1
        except UnsafeTemplateError as error:
            print(colors.red | "\n".join(error.args), file=sys.stderr)
            # DOCS https://github.com/copier-org/copier/issues/1328#issuecomment-1723214165
            return 0b100

    return inner

# ...

    @handle_exceptions
    def _main(self, destination_path: cli.ExistingDirectory = ".") -> int:
        with self._worker(
            dst_path=destination_path,
            defaults=self.force or self.defaults,
            overwrite=self.force or self.overwrite,
            skip_answered=self.skip_answered,
        ) as worker:
            worker.run_recopy()
        return 0

    def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
        """Call [run_recopy][copier.main.Worker.run_recopy].

        Parameters:
            destination_path:
                Only the destination path is needed to update, because the
                `src_path` comes from [the answers file][the-copier-answersyml-file].

                The subproject must exist. If not specified, the currently
                working directory is used.
        """
        return self._main(destination_path: cli.ExistingDirectory)
```

sisp:
Thanks for being persistent in trying to solve this problem, @danieleades! 🙏

This last idea has a bit too much duplicate code for my taste. In comparison, I'd prefer the `inner()` closure.

Have you tried `wrapt` instead of `decorator` for keeping the decorator-based approach? Does this work at runtime? And are the missing type hints okay in this case, as we don't call the decorated method ourselves? (I don't have access to a laptop right now. 🙁)

danieleades:
> Have you tried wrapt instead of decorator for keeping the decorator-based approach?

I haven't- though it does obviate the benefit of eliminating a dependency by introducing a new one

sisp:
But perhaps a new one that _works_ in combination with `from __future__ import annotations` unlike `decorator`?

danieleades:
> But perhaps a new one that _works_ in combination with `from __future__ import annotations` unlike `decorator`?

it may well work, but it may not be necessary to introduce a dependency at all if it only saves 2-3 lines of code.
All else being equal, less dependencies is better

danieleades:
> Okay, you convinced me. Let's keep the `inner()` closure approach and drop the `decorator` dependency. 👍 I'll be happy to merge the PR, just some minor cleanup left.
> 
> Could you please try to solve the `poetry.lock` issue and undo the formatting changes around `DESCRIPTION_MORE` which shouldn't be related to this PR? And I believe the `__all__` dunder variable addition isn't strictly related to this PR, right? I think it makes sense to add them to all modules, but in a dedicated PR.

i've updated as best i can, unfortunately i can't get the poetry lock behaviour to match the nix environment, and i'm having trouble setting up the nix environment because `direnv` is not playing nicely with my zsh shell.

Would you kindly update this PR by checking it out, updating the lockfile, and committing the change back to this branch?
That should be the last piece to get this merged!

pawamoy:
If you look at this comment: https://github.com/copier-org/copier/pull/1513#issuecomment-1938231660, you'll see that you can simply revert any changes to the lock file, and remove the two blocks marked with `...` lines in the comment.

danieleades:
> If you look at this comment: https://github.com/copier-org/copier/pull/1513#issuecomment-1938231660, you'll see that you can simply revert any changes to the lock file, and remove the two blocks marked with `...` lines in the comment, as well as copy-paste the last block.

Are you sure? The lock file contains a hash. Do you know that those items are excluded from the hash?

pawamoy:
Yep it was the right hash. I've pushed a commit anyway :smile: 
There's two more lines that have been removed to put back, I'll do that too.

danieleades:
> Yep it was the right hash. I've pushed a commit anyway :smile: 
> There's two more lines that have been removed to put back, I'll do that too.

Cool, I wasn't sure what was included in that hash and have no way to compare!

pawamoy:
OK done! The diff looks nice.

pawamoy (approved):

sisp (approved):
Perfect! :ok_hand: Thanks for the detailed investigation and also final polishing! :pray:

PR #1523, opened 2024-02-21
style: add 'complexity' lint group to ruff config

danieleades:
adds the 'complexity' lint group to the ruff config.

I believe this brings the ruff config inline with the flake8 config, and therefore makes the flake8 linting redundant.

for reference, the flake8 config is:

```ini
[flake8]
ignore = E203, E501, W503, B950
max-line-length = 88
select = C,E,F,W,B
per-file-ignores=
    __init__.py:F401

```

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -823,7 +823,7 @@ class Worker:
         self._apply_update()
         self._print_message(self.template.message_after_update)
 
-    def _apply_update(self):
+    def _apply_update(self):  # noqa: C901
         git = get_git()
         subproject_top = Path(
             git(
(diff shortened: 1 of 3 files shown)


danieleades:
I'm a little surprised to see the behaviour differs. Maybe I'm missing something

yajo:
It says we have a couple of overly complex functions. However both seem ok to me. could you please configure ruff to allow them, or add exclusion comments?

danieleades:
> It says we have a couple of overly complex functions. However both seem ok to me. could you please configure ruff to allow them, or add exclusion comments?

can do.
To be fair, I agree with Ruff's assessment that these functions are overly-complex and in need of a refactor. But i find it surprising that it appears to calculate the complexity differently to flake8

PR #1525, opened 2024-02-22
style: add ruff 'ARG' lint group

danieleades:


devtasks.py
```diff
@@ -42,7 +42,7 @@ def dev_setup():
         local["poetry"]("install")
 
 
-def lint(recycle_container=False):
+def lint(_recycle_container=False):
```
sisp, line 45: Since this argument is unused, we should remove it instead of renaming it:

```suggestion
def lint():
```

sisp (changes requested):

devtasks.py
danieleades: done

sisp (approved):
LGTM, @danieleades! :tada: Thanks! :pray:

PR #1526, opened 2024-02-22
feat: add support for validating multi-select choice answers

sisp:
I've added support for validating answers of multi-select choice questions. While `select` prompts (Copier single-select choice questions) only support per-choice validation, [`checkbox` prompts support answer validation](https://github.com/tmbo/questionary/blob/c894eeffbeb2f1448a108163998ba95b72a8e32a/questionary/prompts/checkbox.py#L32) similar to `text` prompts. Implementation-wise, there's a difference between `text` prompt validation and `checkbox` prompt validation, as `text` prompt validation supports raising `prompt_toolkit.validation.ValidationError` exceptions in the `validate` callable while `checkbox` prompt validation requires the `validate` callable to return an error message string or `True` when the answer is valid. Fortunately, also `text` prompt validation supports this signature, so there's a unified `validate` callable for both prompt types now.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -435,7 +435,9 @@ class Worker:
                 answer = question.parse_answer(result.init[var_name])
                 # Try to validate the answer value if the question has a
                 # validator.
-                question.validate_answer(answer)
+                err_msg = question.validate_answer(answer)
+                if err_msg:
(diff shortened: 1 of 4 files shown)


yajo (approved):
Excellent, thanks!

Issue #1441, opened 2023-11-29
Make `mypy copier --strict` pass


bswck:
### Actual Situation

`mypy copier --strict` fails with 68 errors in 7 files (checked 11 source files).
Checked on compiled mypy 1.5.1, Python 3.8.17. 

### Desired Situation

`mypy copier --strict` passes successfully.

### Proposed solution

Resolve all errors reported by mypy.
A followup to #1431.

bswck:
I want to see [autorefine](https://github.com/bswck/autorefine) in the action of semi-auto-resolving this, as soon as it's ready.

Lujeni:
Hey @bswck ! Thanks for this nice project

Is it possible to propose a batch of small pull requests in order to fix specific mypy errors or would you prefer a single PR to fix all of them?

bswck:
> Hey @bswck ! Thanks for this nice project
> 
> Is it possible to propose a batch of small pull requests in order to fix specific mypy errors or would you prefer a single PR to fix all of them?

Thanks for reaching out. If you want to fix this issue, then I guess you might make a single pull request with bulk fixes in logically separated commits and in line with [the contributing guide](https://github.com/copier-org/copier/blob/master/CONTRIBUTING.md#pull-request-guidelines).

If you're asking about _autorefine_, it will aim to organize fixes of linting and typechecking errors into groups that can be separated into small pull requests for the ease of reviewing and preserving easily bisectable, clean, logically grouped SCM history.

_autorefine_ is a very big idea, but I am currently jobless and have no time nor money to take the effort all by myself. If you want to help, please reach out to me.

sisp:
Resolved via #1527.

PR #1527, opened 2024-02-23
Enable 'strict' Mypy linting

danieleades:
enables strict mypy linting

note that i've suppressed and allow-listed some existing errors.
This allow-list should not be allowed to grow over time, and ideally would be incrementally shrunk until it can be removed entirely.

The error suppressions can be resolved by refactoring, and by creating upstream type stubs for third-party libraries that are untyped in typeshed (out of scope of this PR!)


danieleades:
"just" 49 errors to go!

danieleades:
remaining errors:

```
copier/tools.py:192: error: Missing type parameters for generic type "Match"  [type-arg]
copier/template.py:385: error: Missing type parameters for generic type "dict"  [type-arg]
copier/main.py:866: error: Item "None" of "Template | None" has no attribute "url"  [union-attr]
copier/main.py:867: error: Item "None" of "Template | None" has no attribute "commit"  [union-attr]
copier/main.py:889: error: Argument 2 to "migration_tasks" of "Template" has incompatible type "Template | None"; expected "Template"  [arg-type]
copier/main.py:911: error: Argument "data" to "replace" of "Worker" has incompatible type "Mapping[str, Any]"; expected "dict[str, Any]"  [arg-type]
copier/main.py:914: error: Item "None" of "Template | None" has no attribute "url"  [union-attr]
copier/main.py:993: error: Argument 2 to "migration_tasks" of "Template" has incompatible type "Template | None"; expected "Template"  [arg-type]
copier/cli.py:83: error: Class cannot subclass "Application" (has type "Any")  [misc]
copier/cli.py:105: error: Class cannot subclass "Application" (has type "Any")  [misc]
copier/cli.py:158: error: Untyped decorator makes function "data_switch" untyped  [misc]
copier/cli.py:176: error: Untyped decorator makes function "data_file_switch" untyped  [misc]
tests/test_dirty_local.py:7: error: Module "pytest_gitconfig" does not explicitly export attribute "GitConfig"  [attr-defined]
tests/conftest.py:11: error: Module "pytest_gitconfig" does not explicitly export attribute "GitConfig"  [attr-defined]
Found 14 errors in 6 files (checked 46 source files)
```

danieleades:
@sisp, @yajo this is now ready for review

pyproject.toml
```diff
@@ -59,6 +59,8 @@ pytest-xdist = ">=2.5.0"
 types-backports = ">=0.1.3"
 types-pyyaml = ">=6.0.4"
 types-psutil = "*"
+types-pygments = "^2.17.0.20240310"
+types-colorama = "^0.4.15.20240311"
```
pawamoy, line 63: I'd suggest avoiding upper bounds (and shortening to just major.minor, but that's just my own taste):

```suggestion
types-pygments = ">=2.17"
types-colorama = ">=0.4"
```

copier/cli.py
```diff
@@ -80,35 +80,32 @@ def _handle_exceptions(method: Callable[[], None]) -> int:
...
-class CopierApp(cli.Application):
+class CopierApp(cli.Application):  # type: ignore[misc]
     """The Copier CLI application."""
 
     DESCRIPTION = "Create a new project from a template."
-    DESCRIPTION_MORE = (
-        dedent(
-            """\
+    DESCRIPTION_MORE = dedent(
+        """\
```
pawamoy, line 88: These formatting changes again :smile: 
I suppose we should review and merge #1557 first, to get rid of this once and for all.

danieleades: ha i'm really struggling with this dev environment. keeps picking up my system version of ruff and applying these changes!

danieleades: or i can clean them up manually when everything else is reviewed, i don't really mind

copier/main.py
```diff
@@ -766,7 +792,7 @@ def run_recopy(self) -> None:
                 f"from `{self.subproject.answers_relpath}`."
             )
         with replace(self, src_path=self.subproject.template.url) as new_worker:
-            return new_worker.run_copy()
+            new_worker.run_copy()
```
pawamoy, line 795: If `new_worker.run_copy()` returns something, that would be a breaking change, even if we didn't advertise the return value because of our `None` return annotation. Need to decide whether we're OK with the breaking change or if we should update the return annotation to reflect the actual return value.

copier/cli.py
pawamoy: I think you can also disable auto-formatting in your editor settings while working on this (be careful not to include changes to VSCode's settings.json in the PR though). Unless you mean that the precommit tasks use your system Ruff instead of the Copier environment one? In any case, don't bother too much, this is not fun and I can revert it myself before we merge :slightly_smiling_face:

copier/main.py
danieleades: i'm not sure what you mean?
`new_worker.run_copy()` doesn't return anything, and there's no return statement in that function anyway

tests/test_dirty_local.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 from plumbum import local
 from plumbum.cmd import git
-from pytest_gitconfig import GitConfig
+from pytest_gitconfig.plugin import GitConfig
```
pawamoy, line 7: Did mypy complain? The [docs](https://pypi.org/project/pytest-gitconfig/) show that we should import from `pytest_gitconfig` directly, and the [code](https://github.com/noirbizarre/pytest-gitconfig/blob/main/src/pytest_gitconfig/__init__.py) correctly imports `GitConfig` in the top-level module.

By the way the docs show that this import should be guarded under `if TYPE_CHECKING`, does it apply to us here as well? Should we `from __future__ import annotations` in all tests module (that's what I do in my own projects)?

copier/main.py
pawamoy: Great, so the condition in my comment is never true, and we can discard the whole comment :slightly_smiling_face:

tests/test_dirty_local.py
danieleades: > Did mypy complain? The [docs](https://pypi.org/project/pytest-gitconfig/) show that we should import from `pytest_gitconfig` directly, and the [code](https://github.com/noirbizarre/pytest-gitconfig/blob/main/src/pytest_gitconfig/__init__.py) correctly imports `GitConfig` in the top-level module.

mypy complains because `GitConfig` is not explicitly re-exported from `pytest_gitconfig`. That could be fixed with an upstream change


> By the way the docs show that this import should be guarded under `if TYPE_CHECKING`, does it apply to us here as well?

not required, but something i could add in a future PR

> Should we `from __future__ import annotations` in all tests module (that's what I do in my own projects)?

also not required, since ruff can detect when it is needed and automatically add it

danieleades:
@pawamoy i could probably just keep tinkering with this for ever, but i think this is a place where I can draw a line under it. This is now ready for review

danieleades:
> @sisp, @yajo this is now ready for review

ping!

pawamoy (commented):
IMO it looks good, except for the style changes in cli.py. I'll revert them and push a fixup.

copier/main.py
```diff
@@ -1,4 +1,5 @@
 """Main functions and classes, used to generate or update projects."""
+
```
sisp, line 2: Those empty lines were introduced by Ruff v0.3.0 (IIRC). Would you mind reverting those (and the other occurences), too? They aren't related to this PR.

pyproject.toml
```diff
@@ -135,9 +137,32 @@ known-first-party = ["copier"]
...
 plugins = ["pydantic.mypy"]
-warn_no_return = false
+
+[[tool.mypy.overrides]]
+module = [
+  "coverage.tracer",
+  "funcy",
+  "pexpect.*",
+  "poethepoet.app",
+  "plumbum.*",
```
sisp, line 149: Would you mind moving this line up, so the modules are sorted?

pyproject.toml
```diff
@@ -59,6 +59,8 @@ pytest-xdist = ">=2.5.0"
 types-backports = ">=0.1.3"
 types-pyyaml = ">=6.0.4"
 types-psutil = "*"
+types-pygments = ">=2.17"
+types-colorama = ">=0.4"
```
sisp, line 63: Would you mind sorting the dependencies? I just noticed that `types-pyyaml` was already in the "wrong" place, too. I'd say, just sort them all now.

poetry.lock
sisp, line 1: We've had a similar issue before: Some entries (for `pyyaml` and `pexpect`) get deleted for no obvious reason, and I don't think they should. If @danieleades can't fix it, would you mind fixing it again, @pawamoy? (I don't have a laptop right now.)

sisp (changes requested):
Fantastic work, @danieleades! 👌 Just a few minor remarks from my side.

poetry.lock
pawamoy: Urrh, my bad, didn't think about expanding poetry.lock's diff. One of the inconvenient of GitHub collapsing it by default, even for small diffs... Will do.

pawamoy: Done.

danieleades:
what's causing those test failures?

sisp (changes requested):
The style changes in `cli.py` are still present (cc @pawamoy) and also the introduced empty line between a module doctstring and the first import, which is not needed by the Ruff version locked via Nix (IIRC).

pawamoy:
Reverted style changes in cli.py. Will now revert blank lines at the top of modules. Race to the finish!

pawamoy:
Done.

sisp:
> what's causing those test failures?

I _suspect_ that postponed annotations don't play well with Plumbum in `cli.py`. Could you try reverting back to classical type annotations and see whether tests pass again?

danieleades:
> > what's causing those test failures?
> 
> I _suspect_ that postponed annotations don't play well with Plumbum in `cli.py`. Could you try reverting back to classical type annotations and see whether tests pass again?

strange. i've not made any changes since these tests were last passing

copier/cli.py
```diff
@@ -158,14 +158,14 @@ def __init__(self, executable: PathLike) -> None:
...
-    @cli.switch(
+    @cli.switch(  # type: ignore[misc]
         ["-d", "--data"],
         str,
         "VARIABLE=VALUE",
         list=True,
         help="Make VARIABLE available as VALUE when rendering the template",
     )
-    def data_switch(self, values: StrSeq) -> None:
+    def data_switch(self, values: Iterable[str]) -> None:
```
sisp, line 168: Should this really be an `Iterable` and not a `Sequence`?

copier/cli.py
```diff
@@ -108,7 +108,7 @@ class CopierApp(cli.Application):  # type: ignore[misc]
 class _Subcommand(cli.Application):  # type: ignore[misc]
     """Base class for Copier subcommands."""
 
-    def __init__(self, executable: PathLike[str]) -> None:
+    def __init__(self, executable: "PathLike[str]") -> None:
```
danieleades, line 111: ah i think this was the issue, not sure when this got reverted

copier/cli.py
pawamoy: It could be a sequence, but it doesn't *need* to, we only iterate on it and do not access its element by index.

pawamoy (approved):
LGTM

copier/cli.py
pawamoy: @danieleades I think it's better if you let the person asking the question mark a comment as resolved :slightly_smiling_face:

sisp (approved):
LGTM, too! :tada: Thanks for the fantastic work, @danieleades and @pawamoy! :pray:

sisp:
Resolves #1441.

PR #1556, opened 2024-03-14
perf: add PERF lints

danieleades:


sisp (approved):
LGTM! :ok_hand: Thanks, @danieleades! :pray:

PR #1557, opened 2024-03-17
style: tighten ruff lints

danieleades:
- tighten ruff linting
- normalise doc comments and some formatting

note that this applies ruff 'preview' formatting as a one off. This is compatible with the default ruff formatting

### DIFF
--- a/devtasks.py
+++ b/devtasks.py
@@ -52,9 +52,10 @@ def lint() -> None:
             )
         except ProcessExecutionError:
             _logger.info(
-                "Couldn't create copier-lint-v1 container, probably because a previous one exists. "
-                "Remove it if you want to recycle it. Otherwise, this is OK."
+                "Couldn't create copier-lint-v1 container, probably because a previous "
+                "one exists. Remove it if you want to recycle it. Otherwise, this is "
+                "OK."
             )
         runner["container", "start", "--attach", "copier-lint-v1"] & TEE
     except ProcessExecutionError as error:
-        raise SystemExit(error.errno)
+        raise SystemExit(error.errno) from error
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -239,7 +239,6 @@ def test_good_cli_run_dot_config(
 
     Added based on the discussion: https://github.com/copier-org/copier/discussions/859
     """
-
     src = template_path_with_dot_config(str(config_folder))
 
     with local.cwd(src):
--- a/copier/__init__.py
+++ b/copier/__init__.py
@@ -3,7 +3,7 @@
 Docs: https://copier.readthedocs.io/
 """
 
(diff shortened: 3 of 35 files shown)


devtasks.py
```diff
@@ -78,9 +79,10 @@ def lint():
             )
         except ProcessExecutionError:
             _logger.info(
-                "Couldn't create copier-lint-v1 container, probably because a previous one exists. "
-                "Remove it if you want to recycle it. Otherwise, this is OK."
+                "Couldn't create copier-lint-v1 container, probably because a previous "
+                "one exists. Remove it if you want to recycle it. Otherwise, this is"
+                "OK."
```
pawamoy, line 84: I thought the strings were split automatically by Ruff, but I see here that a space is missing at the end of `is` or at the beginning of `OK`, so I suppose you split them manually?

danieleades: i split them manually. i think black --preview can automatically re-split strings?

tests/test_cli.py
```diff
@@ -133,23 +127,21 @@ def test_data_file_parsed_by_question_type(
     tmp_path_factory: pytest.TempPathFactory,
 ) -> None:
     src, dst, local = map(tmp_path_factory.mktemp, ("src", "dst", "local"))
-    build_file_tree(
-        {
-            (src / "copier.yml"): (
-                """\
+    build_file_tree({
```
pawamoy, line 130: Is this a manual change too? I would find it surprising if Black/Ruff concatenated again `({` together instead of adding newlines and indentation levels.

danieleades: both black and ruff do this in --preview mode. likely to move into non-preview eventually but black doesn't move fast

pawamoy: Interesting, OK, thanks :slightly_smiling_face:

pawamoy: Just to be clear: did you manually run Ruff with preview mode enabled? If yes: is the preview mode enabled in our CI (pre-commit / poe lint / GitHub workflows / etc.)?

If it is not enabled, then I think we should revert all these style changes, and let Ruff format the code again later when such formatting is out of preview mode.

Otherwise, we could consider enabling preview mode, but I'll leave that decision to other @copier-org/maintainers.

pawamoy: Locally, I reverted the changes in tests, ran `poe lint` again, and it didn't style the code as what we see in this PR.

danieleades: What happens if you

1. Clone this PR
2. Run 'poe lint' without reverting any other changes
3. Commit and push back to this PR

?

danieleades: Can do. Will need to find some time to manually revert- hopefully tonight

PR #1558, opened 2024-03-18
style: enable more ruff lints

danieleades:
enables additional ruff lints

this should not be merged yet- the `FURB` lint group is preview only, so pre-commit will need to be updated to use the `--preview` flag in order to use these lints. I've got no idea how to configure this using flake/nix

I've run ruff over this locally with that flag, so the changes have been applied, but we'll want to make sure they continue to be checked in CI


yajo (approved):
Looks good. Could you solve the conflicts please? Anything left before getting out of draft?

pawamoy:
Same as #1557, I suggest doing that again (changes to pyproject.toml + ruff format) in a new PR and closing this one.

PR #1561, opened 2024-03-22
feat(tasks): allow skipping tasks with the  flag

### LINKED ISSUE #1415: Ability to define safe/unsafe modes and force to ignore _tasks
### Actual Situation

If we define `_tasks`, the end user will need to use `--UNSAFE/--trust` to generate the project from the template.
In some situation, the `_tasks` commands are just convenience.

More precisely, in my case:
```yaml
_tasks:
    - "git init -b main && git add -A && git commit -m 'Initial commit'"
    - "pre-commit install"
```

### Desired Situation

I would like the end-user to be able to use the template without running any unsafe feature.

### Proposed solution

The end-user could be able to use a flag (e.g. `--no-script`) to not run any unsafe feature but still create the project.

From the template's creator side, an additional feature would be to mark the `_tasks` as `required/optional` to control if the end-user is allowed to copy the project without the tasks.

noirbizarre:
This PR add a `-T/--skip-tasks` flag allowing to skip template tasks execution.

Some known use cases I have:
- while testing my template, I want to be able to perform multiple rendering with the extensions actives but without running tasks (which may be slow or depending on connection...)
- when applying a template with some bootstrap tasks on an existing project, I want to be able to skip the bootstrap tasks because I know they will fail
- when applying a third party template that I trust, I want to only skip tasks for some reason (known to break, not compatible with the environment...)

Should fix #1415

### DIFF
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -157,6 +157,11 @@ class _Subcommand(cli.Application):  # type: ignore[misc]
             "Allow templates with unsafe features (Jinja extensions, migrations, tasks)"
         ),
     )
+    skip_tasks = cli.Flag(
+        ["-T", "--skip-tasks"],
+        default=False,
+        help="Skip template tasks execution",
+    )
 
     @cli.switch(  # type: ignore[misc]
         ["-d", "--data"],
--- a/copier/main.py
+++ b/copier/main.py
@@ -162,6 +162,9 @@ class Worker:
 
         skip_answered:
             When `True`, skip questions that have already been answered.
+
+        skip_tasks:
+            When `True`, skip template tasks execution.
     """
 
     src_path: str | None = None
@@ -182,6 +185,7 @@ class Worker:
     context_lines: PositiveInt = 3
     unsafe: bool = False
     skip_answered: bool = False
+    skip_tasks: bool = False
 
     answers: AnswersMap = field(default_factory=AnswersMap, init=False)
     _cleanup_hooks: list[Callable[[], None]] = field(default_factory=list, init=False)
@@ -766,7 +770,8 @@ class Worker:
             if not self.quiet:
                 # TODO Unify printing tools
                 print("")  # padding space
-            self._execute_tasks(self.template.tasks)
+            if not self.skip_tasks:
+                self._execute_tasks(self.template.tasks)
         except Exception:
             if not was_existing and self.cleanup_on_error:
                 rmtree(self.subproject.local_abspath)
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1317,7 +1317,7 @@ Skip template [tasks][tasks] execution, if set to `True`.
 
 !!! question "Does it imply `--trust`?"
 
-    This flag does not imply `--trust`, and will do nothing if not used with.
+    This flag does not imply [`--trust`][unsafe], and will do nothing if not used with.
 
 ### `subdirectory`
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1313,7 +1313,7 @@ Skip template [tasks][tasks] execution, if set to `True`.
 
 !!! note
 
-    It only skips tasks, not migration tasks.
+    It only skips [tasks][tasks], not [migration tasks][migrations].
 
 !!! question "Does it imply `--trust`?"
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1303,6 +1303,22 @@ exist.
     {{999999999999999999999999999999999|ans_random|hash('sha512')}}
     ```
 
+### `skip_tasks`
+
+-   Format: `bool`
+-   CLI Flags: `-T`, `--skip-tasks`
+-   Default value: `False`
+
+Skip template tasks execution, if set to `True`.
+
+!!! note
+
+    It only skip tasks, not migration tasks.
+
+!!! question "Does it imply `--trust` ?"
+
+    This flag does not imply `--trust`, and will do nothing if not used with.
+
 ### `subdirectory`
 
 -   Format: `str`
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1313,7 +1313,7 @@ Skip template tasks execution, if set to `True`.
 
 !!! note
 
-    It only skip tasks, not migration tasks.
+    It only skips tasks, not migration tasks.
 
 !!! question "Does it imply `--trust` ?"
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1309,7 +1309,7 @@ exist.
 -   CLI Flags: `-T`, `--skip-tasks`
 -   Default value: `False`
 
-Skip template tasks execution, if set to `True`.
+Skip template [tasks][tasks] execution, if set to `True`.
 
 !!! note
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1315,7 +1315,7 @@ Skip template tasks execution, if set to `True`.
 
     It only skips tasks, not migration tasks.
 
-!!! question "Does it imply `--trust` ?"
+!!! question "Does it imply `--trust`?"
 
     This flag does not imply `--trust`, and will do nothing if not used with.
 
--- a/tests/test_migrations.py
+++ b/tests/test_migrations.py
@@ -26,7 +26,8 @@ SRC = Path(f"{PROJECT_TEMPLATE}_migrations").absolute()
     reason="Windows ignores shebang?",
     strict=True,
 )
-def test_migrations_and_tasks(tmp_path: Path) -> None:
+@pytest.mark.parametrize("skip_tasks", [True, False])
+def test_migrations_and_tasks(tmp_path: Path, skip_tasks: bool) -> None:
     """Check migrations and tasks are run properly."""
     # Convert demo_migrations in a git repository with 2 versions
     src, dst = tmp_path / "src", tmp_path / "dst"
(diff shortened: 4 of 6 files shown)


copier/main.py
```diff
@@ -157,6 +157,12 @@ class Worker:
             When `True`, allow usage of unsafe templates.
 
             See [unsafe][]
+
+        skip_answered:
+            When `True`, skip questions that have already been answered.
```
pawamoy, line 162: Does not seem to belong to this PR?

pawamoy (commented):
Looks solid :slightly_smiling_face: 

Do I understand correctly that passing `--skip-tasks` means I don't have to also pass `--trust`?

noirbizarre:
No, you still have to pass `--trust` (if you want to have jinja extensions loaded), but if you have `--trust --skip-tasks` only extension will be loaded while `--trust` will load extension and execute tasks.
In fact it doesn't touch the `--trust` behavior, just skipping tasks, whether the template is trusted or not (but in the second case, tasks are disabled by security, not from user intent)

copier/main.py
pawamoy: Ah, right :sweat_smile: Yes, please, another PR would be great (and immediately merged).

pawamoy:
OK yes that makes sense.

copier/main.py
noirbizarre: Done there: https://github.com/copier-org/copier/pull/1562 👍🏼 

I'll rebase this one onto it after merge

copier/cli.py
```diff
@@ -157,6 +157,11 @@ def __init__(self, executable: "PathLike[str]") -> None:
             "Allow templates with unsafe features (Jinja extensions, migrations, tasks)"
         ),
     )
+    skip_tasks = cli.Flag(
+        ["-S", "--skip-tasks"],
```
pawamoy, line 161: I still think `-T` would be better, or maybe `-K` (uppercase because that usually means "negative" option). @copier-org/maintainers any opinion on this?

sisp: `-T` seems best IMO, uppercase for negation and `T` for **t**asks (also in case we might have other skip flags in the future).

noirbizarre: PR updated 👍🏼

tests/test_migrations.py
```diff
@@ -41,10 +42,20 @@ def test_migrations_and_tasks(tmp_path: Path) -> None:
...
+    )
     # Check copy was OK
-    assert (dst / "created-with-tasks.txt").read_text() == "task 1\ntask 2\n"
-    assert not (dst / "delete-in-tasks.txt").exists()
+    if not skip_tasks:
+        assert (dst / "created-with-tasks.txt").read_text() == "task 1\ntask 2\n"
+        assert not (dst / "delete-in-tasks.txt").exists()
+    else:
+        assert not (dst / "created-with-tasks.txt").exists()
+        assert (dst / "delete-in-tasks.txt").exists()
```
pawamoy, line 58: It's usually more readable to start with non-negated conditions:

```suggestion
    if skip_tasks:
        assert not (dst / "created-with-tasks.txt").exists()
        assert (dst / "delete-in-tasks.txt").exists()
    else:
        assert (dst / "created-with-tasks.txt").read_text() == "task 1\ntask 2\n"
        assert not (dst / "delete-in-tasks.txt").exists()

```

tests/test_migrations.py
```diff
@@ -54,15 +65,22 @@ def test_migrations_and_tasks(tmp_path: Path) -> None:
...
+    )
     # Check update was OK
-    assert (dst / "created-with-tasks.txt").read_text() == "task 1\ntask 2\n" * 2
-    assert not (dst / "delete-in-tasks.txt").exists()
+    if not skip_tasks:
+        assert (dst / "created-with-tasks.txt").read_text() == "task 1\ntask 2\n" * 2
+        assert not (dst / "delete-in-tasks.txt").exists()
+    else:
+        assert not (dst / "created-with-tasks.txt").exists()
+        assert (dst / "delete-in-tasks.txt").exists()
```
pawamoy, line 83: Same suggestion here:

```suggestion
    if skip_tasks:
        assert not (dst / "created-with-tasks.txt").exists()
        assert (dst / "delete-in-tasks.txt").exists()
    else:
        assert (dst / "created-with-tasks.txt").read_text() == "task 1\ntask 2\n" * 2
        assert not (dst / "delete-in-tasks.txt").exists()
```

pawamoy (commented):
Two suggestions regarding readability and then I'll have nothing left to say I think :smile:

tests/test_migrations.py
noirbizarre: I hesitated because depending on projects I contribute to, some prefer
- if non-negated else
- if nominal case else
I chose the 2nd because it was the existing case but I prefer the first one. Submitting the change ASAP

pawamoy (approved):
LGTM!

tests/test_tasks.py
```diff
@@ -53,6 +53,23 @@ def test_copy_tasks(template_path: str, tmp_path: Path) -> None:
...
 
 
+def test_copy_skip_tasks(template_path: str, tmp_path: Path) -> None:
+    copier.run_copy(
+        template_path,
+        tmp_path,
+        quiet=True,
+        defaults=True,
+        overwrite=True,
+        unsafe=True,
```
sisp, line 63: This test should also pass with

```suggestion
        unsafe=False,
```

right? I think we should test it that way to cover that `skip_tasks=True` makes `unsafe=True` unnecessary (when no other unsafe features are used – like here).

sisp (changes requested):

tests/test_tasks.py
noirbizarre: Yes it should but I don't see the point given that we already test that `unsafe=False` disable extensions, tasks and migrations.
That's exactly the point of this test: check that only tasks are skipped when the template is trusted and `--skip-tasks` is enabled. 
This change would just test that `--skip-tasks` don't do anything when the template is not trusted (eg. when tasks are already skipped).
The need to skip tasks only exists for trusted templates (eg. with extensions, tasks and migrations enabled), more precisely the need is to be able to run a template with extensions active but without the running tasks or to update with the extensions and migrations active but without the tasks. 
(Note that I took into account that it is planned that trusts will be persistent at some point, so this test should still pass when it is the case)

noirbizarre: Edit: I just tried to parametrize `unsafe`, but it doesn't run for `unsafe=False`: trying to run a template with `tasks` and without `unsafe` raise `UnsafeTemplateError` (as expected by https://github.com/copier-org/copier/blob/master/tests/test_unsafe.py in fact)

sisp: Ah, I missed the rationale you explained in https://github.com/copier-org/copier/pull/1561#issuecomment-2015137696. With that in mind, the behavior you observe is expected.

I wonder whether `--skip-tasks` should actually affect the `--trust` behavior. When you pass `--skip-tasks`, you avoid the security risk of tasks. So why should `copier copy --skip-tasks ...` fail because of unsafe features when the only unsafe feature of that template is "tasks"? I'm not sure that's intuitive. WDYT, @copier-org/maintainers?

sisp (changes requested):
We also need to [document this setting here](https://github.com/copier-org/copier/blob/5fe373c83e7cc2810f23117c25b3d62f99396fc7/docs/configuring.md?plain=1#L1305).

tests/test_tasks.py
pawamoy: @noirbizarre recalled previously that not only tasks are unsafe, but Jinja extensions too :slightly_smiling_face: 
Therefore `--skip-tasks` **should not** silently toggle `--trust` on.

sisp: Yes, I didn't mean to toggle `--trust` but only ignore tasks when determining the unsafe features of a template when `--skip-trust` is provided.

I think you've convinced me to leave behavior as it is now. Also, we won't have `--skip-{jinja-extensions,migrations}` because skipping those would render the template useless. So considering `--skip-tasks` as a partial `--trust` seems conceptually wrong – adding to your reasoning towards the same conclusion.

noirbizarre:
PR updated with the setting documentation 👍🏼

docs/configuring.md
```diff
@@ -1303,6 +1303,22 @@ exist.
...
+
+-   Format: `bool`
+-   CLI Flags: `-T`, `--skip-tasks`
+-   Default value: `False`
+
+Skip template tasks execution, if set to `True`.
+
+!!! note
+
+    It only skip tasks, not migration tasks.
```
pawamoy, line 1316:
```suggestion
    It only skips tasks, not migration tasks.
```

docs/configuring.md
```diff
@@ -1303,6 +1303,22 @@ exist.
...
+-   CLI Flags: `-T`, `--skip-tasks`
+-   Default value: `False`
+
+Skip template tasks execution, if set to `True`.
+
+!!! note
+
+    It only skips tasks, not migration tasks.
+
+!!! question "Does it imply `--trust` ?"
```
sisp, line 1318: Sorry for being pedantic 🫣:

```suggestion
!!! question "Does it imply `--trust`?"
```

docs/configuring.md
```diff
@@ -1303,6 +1303,22 @@ exist.
     {{999999999999999999999999999999999|ans_random|hash('sha512')}}
     ```
 
+### `skip_tasks`
+
+-   Format: `bool`
+-   CLI Flags: `-T`, `--skip-tasks`
+-   Default value: `False`
+
+Skip template tasks execution, if set to `True`.
```
sisp, line 1312: I think UX would be a bit better with some links:

```suggestion
Skip template [tasks][tasks] execution, if set to `True`.
```

docs/configuring.md
```diff
@@ -1303,6 +1303,22 @@ exist.
...
+
+!!! note
+
+    It only skips tasks, not migration tasks.
```
sisp, line 1316: I think UX would be a bit better with some links:

```suggestion
    It only skips [tasks][tasks], not [migration tasks][migrations].
```

docs/configuring.md
```diff
@@ -1303,6 +1303,22 @@ exist.
...
+
+!!! question "Does it imply `--trust` ?"
+
+    This flag does not imply `--trust`, and will do nothing if not used with.
```
sisp, line 1320: And I think UX would be a bit better with some links:

```suggestion
    This flag does not imply [`--trust`][unsafe], and will do nothing if not used with.
```

sisp (changes requested):
Just some final :crossed_fingers: docs suggestions, @noirbizarre. I'll be happy to merge afterwards.

sisp (approved):
Fantastic, thanks @noirbizarre! :pray:

yajo:
Sorry for being late to the party.

I think that, if you pass this flag, it should be taken into account around here to avoid marking the template as unsafe if it's only because of tasks but you're skipping them: https://github.com/copier-org/copier/blob/70aa69f7df430a611a20f940b47174da96b88dfa/copier/main.py#L233-L234

sisp:
@yajo I think that's the same [thought that I had](https://github.com/copier-org/copier/pull/1561#discussion_r1537995998), but we decided against it in that discussion. WDYT?

yajo:
Consider migrations. Even if the template uses migrations, we only mark it as unsafe if the current update will trigger them.

In the same fashion, if it has tasks but you're ignoring them, there's no unsafety produced by tasks. Other unsafety criteria should still be honored, but not that one.

PR #1565, opened 2024-03-27
feat: support zip template from local or http endpoint

tianzhipeng-git:
support zip template from local or http endpoint:


yajo:
Why? 🤔

Issue #1574, opened 2024-04-05
Improved logic for `when` for handling initial questions


stereobutter:
**Disclaimer** This feature request/proposal is a spin-off from https://github.com/copier-org/copier/issues/1459 because I felt that that issue didn't get the real, underlying problem across.

# Use Case
When setting up a new project via `copier copy` I'd like to be able and prompt the user for some initial information and then skip that question on any following `copier update`. This is both useful for answers that cannot be changed by updating a template (requiring the use of `copier recopy`) or rendering conditional files when the project is setup for the first time and not on updating.

# The Problem with the current behaviour of`when`
Take as en example the following question:

```yaml
# copier.yaml
id:
  type: str
  help: "Enter an ID for your project"
  when: "{{id is not defined}}" 
```

when the project is first created by running `copier copy` the variable `id` is not defined and the user will be asked to enter an id and the answer is recorded

```yaml
# .copier-answers.yml
_commit: <some commit hash>
id: <some-id-the-user-entered>
...
```

when the project is then later updated via `copier update` the question is skipped - nice, just what we wanted. However when we look at the recorded answers we'll notice that the previous answer is now gone

```yaml
# .copier-answers.yml
_commit: <some other commit hash>
```

That is because (as per documentation) *"If a question is skipped, its answer is not recorded [...]"* which has to be read quite literally and is rather subtle (and I'd argue not what most people would read into that statement.)

If we then run `copier update again` we'll get `ValueError("Question "id" is required")`. The error is not really the point though (to be honest I have been looking at the [code that raises this error](https://github.com/copier-org/copier/blob/70aa69f7df430a611a20f940b47174da96b88dfa/copier/main.py#L466) for more than 15 minutes but I'm still not sure why it does what it does). 

The point I'd argue is that the current behaviour is both a foot gun and impedes legitimate use cases that could be handled with `when` (like the example above). I could not think of any use case where the current behaviour is what users would want to happen.

# Proposal
Change the [implementation of when](https://github.com/copier-org/copier/blob/70aa69f7df430a611a20f940b47174da96b88dfa/copier/main.py#L443C17-L443C38) so it only hides the answer from the answer file if the answer is new (and thus does not remove a previous answer like seen above). 

# Questions
* do other people feel this would be an improvement to `copier` and using `when` in this way should be a supported use case?
* would this change be considered backwards compatible (ignoring https://xkcd.com/1172/)?
[...]

PR #1582, opened 2024-04-14
feat: leave previous answers intact when skipping question

stereobutter:
With this PR skipping questions using a falsy `when` doesn't modify the answers file anymore. This enables the usage of `when: "{{answer is not defined}}"` to only ask a question on `copier copy` and skip the question on any future `copier update` while keeping the original answer recorded. 

```yaml
# copier.yaml
year_of_first_publication:
    type: str
    when: "{{year_of_first_publication is not defined}}"
```

resolves https://github.com/copier-org/copier/issues/1574

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -1,4 +1,5 @@
 """Main functions and classes, used to generate or update projects."""
+
 from __future__ import annotations
 
 import os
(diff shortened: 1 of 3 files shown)


stereobutter:
Oh and I also tried to make the docs for `when` a little bit clearer with regards to what skipping a question actually means (under the new behaviour this PR introduces).

sisp:
Thanks for your initiative, @stereobutter! :pray: Unfortunately, I don't think this idea works – see https://github.com/copier-org/copier/issues/1574#issuecomment-2054153683. Let's keep the discussion there for now until we have a working design proposal.

yajo:
Thanks for your efforts @stereobutter! 😊 I'll close this pr to keep our list clean, because there have been hopefully better solutions proposed in the issue. Let's continue there until we reach consensus.

PR #1587, opened 2024-04-16
fix(updating): yield merge conflict when both template and project add same file

sisp:
I've stumbled over some suboptimal updating UX when _both_ the template and the project add the _same_ new file where the update algorithm would delete file content added by the project but not present in the template. This PR aims to improve the UX by yielding merge conflicts instead which seems semantically more correct than simply deleting previous project content.

The problem originates from the diff between the fresh project based on the old template and the actual project which includes a "new file mode" patch for any file added by the actual project, which gets applied to the same new file added to the project as part of rendering the new template on top of the project:

https://github.com/copier-org/copier/blob/813c90335276320a17c964a2cd4d4e91d89d8662/copier/main.py#L896-L906

My suggested solution is to identify all files that are added by _both_ the project _and_ the new template and create an empty file in the fresh project based on the old template before extracting the diff, so the patch does not use the "new file mode".

I've added two test cases that – of course – test the new behavior but also exemplify two (simplified) situations that I was facing:

1. Simple (see `test_update_with_new_file_in_template_and_project`) – Both the template and the project add the file `.gitlab-ci.yml[.jinja]`. In this case, the content of `.gitlab-ci.yml` after updating would have been this:

    ```yaml
    tests:
      stage: test
      script:
        - ./test.sh --slow
    ```

    With the change of this PR:

    ```yaml
    tests:
      stage: test
      script:
    <<<<<<< before updating
        - ./test.sh
    pages:
      stage: deploy
      script:
        - ./deploy.sh
    =======
        - ./test.sh --slow
    >>>>>>> after updating
    ```

1. Complex (see `test_update_with_new_file_in_template_and_project_via_migration`) – The project adds content to `.gitlab-ci.yml` on top of what the first version of the template provides. In a new version of the template, `.gitlab-ci.yml.jinja` is moved to `.gitlab/ci/main.yml.jinja` and `.gitlab-ci.yml.jinja` now includes the generated `.gitlab/ci/main.yml`. To retain the project's changes/additions to `.gitlab-ci.yml`, I created a pre-update migration task that copies `.gitlab-ci.yml` (containing those changes/additions)  to `.gitlab/ci/main.yml` and stages it, then Copier applies the new template's version of that file (which was also moved there, but Git doesn't recognize it as status `R` but as `A`). (In case somebody has a better idea on how to do this, I'd be curious. This is the best what I could think of.) In this case, the content of `.gitlab/ci/main.yml` after updating would have been this:

    ```yaml
    tests:
      stage: test
      script:
        - ./test.sh --slow
    ```

    With the change of this PR:

    ```yaml
    tests:
      stage: test
      script:
    <<<<<<< before updating
        - ./test.sh
    pages:
      stage: deploy
      script:
        - ./deploy.sh
    =======
        - ./test.sh --slow
    >>>>>>> after updating
    ```

WDYT?

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -11,7 +11,7 @@ from filecmp import dircmp
 from functools import cached_property, partial
 from itertools import chain
 from pathlib import Path
-from shutil import rmtree
+from shutil import copytree, ignore_patterns, rmtree
 from tempfile import TemporaryDirectory
 from types import TracebackType
 from typing import (
@@ -854,8 +854,12 @@ class Worker:
         subproject_subdir = self.subproject.local_abspath.relative_to(subproject_top)
 
         with TemporaryDirectory(
-            prefix=f"{__name__}.old_copy."
-        ) as old_copy, TemporaryDirectory(prefix=f"{__name__}.new_copy.") as new_copy:
+            prefix=f"{__name__}.old_copy.",
+        ) as old_copy, TemporaryDirectory(
+            prefix=f"{__name__}.new_copy.",
+        ) as new_copy, TemporaryDirectory(
+            prefix=f"{__name__}.dst_copy.",
+        ) as dst_copy:
             # Copy old template into a temporary destination
             with replace(
                 self,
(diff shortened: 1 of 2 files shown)


sisp:
I don't understand why Codecov thinks there's such a high coverage drop. :thinking:

pawamoy (commented):
The issue and solution explanations are clear, thanks. The code is well commented. It would be great if you could add a comment similar to what you explained in the PR's body at the beginning of the complex test, to explain what it tries to assert and how.

sisp:
@pawamoy Good idea, thanks! I've force-pushed the amended commit.

pawamoy (approved):
OK that looks good! Not sure about what Codecov is doing, it's a black box to me.

PR #1576, opened 2024-04-07
Do not overwrite identical files

kefirbandi:
If _render_allowed identifies identical source and target, then only returns True if the path refers to a directory, not a file.

PR #1593, opened 2024-04-17
Do not overwrite identical files

kefirbandi:
When copier detects files are identical, they will not be overwritten.

See also: #1092

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -415,7 +415,7 @@ class Worker:
                 quiet=self.quiet,
                 file_=sys.stderr,
             )
-            return True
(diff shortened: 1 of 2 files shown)


sisp:
@kefirbandi What's the need for this PR since there is already #1576? :thinking:

kefirbandi:
> @kefirbandi What's the need for this PR since there is already #1576? 🤔

Sorry, I wanted to replace it, but messed things up. This PR can be simply rejected if I don't find out how to cancel it myself.

kefirbandi:
See if I can close this ...

PR #1596, opened 2024-04-18
Fix repository name collision

### LINKED ISSUE #1595: Name collision which renders repo formally broken and causes several side effects
## What 

Two versions of `tests/demo/doc/mañana.txt` existed in the repository at the same time, each of them with unique filename.

This was possible because of git ill-formed approach to utilize filesystem-dependent filename representations by default: likely, the mentioned file was added twice: both on Linux and macOS and due to filename encoding differences between the systems git permitted it. 

This default can be changed on macOS using:

```shell
$ git config --global core.precomposeunicode true
$ git config --global core.quotePath false
```
and if followed by re-adding affected files, results in fixing the issue.

## Affected software

the mentioned software was installed from `nixpkgs` rev `355cbc482f33f5b07a6bc0d72be862b1ccdb99dd`

- `git version 2.44.0` 
   _minor_:
    * prints warning
    * after fixing the issue in a child branch as per described above, it becomes impossible to switch back to parent branch without hard reset

- gitoxide: `gix 0.35.0`
    * __showstopper__: fails to clone repo properly

- `nix (Nix) 2.21.2` with flakes enabled
    * __showstopper__: fails to install copier using its flake:

```sh
$ nix profile install 'https://flakehub.com/f/copier-org/copier/*.tar.gz'
error:
       … while fetching the input 'https://flakehub.com/f/copier-org/copier/%2A.tar.gz'

       error: creating file '/nix/store/p1k8x1b6arrk91qrva2l3yzga9d1iqhw-source/tests/demo/doc/mañana.txt': File exists
```

### Template

N/A

### To Reproduce

Besides what's been already described, please see aggregated versions info below. 

Also see pinned versions in your `flake.lock`

```sh
$ sw_vers
ProductName:		macOS
ProductVersion:		14.4.1
BuildVersion:		23E224

$ nix --version
nix (Nix) 2.21.2

$ cat ~/.config/nix/registry.json
{
  "flakes": [
    {
      "from": {
        "id": "nix",
        "type": "indirect"
      },
      "to": {
        "lastModified": 1712161137,
        "narHash": "sha256-ObaVDDPtnOeIE0t7m4OVk5G+OS6d9qYh+ktK67Fe/zE=",
        "owner": "NixOS",
        "repo": "nix",
        "rev": "355cbc482f33f5b07a6bc0d72be862b1ccdb99dd",
        "type": "github"
      }
    }
  ],
  "version": 2
}

$ git --version
git version 2.44.0

$ gix --version
gix 0.35.0
```

### Logs

```console
$ nix profile install 'https://flakehub.com/f/copier-org/copier/*.tar.gz'
error:
       … while fetching the input 'https://flakehub.com/f/copier-org/copier/%2A.tar.gz'

       error: creating file '/nix/store/p1k8x1b6arrk91qrva2l3yzga9d1iqhw-source/tests/demo/doc/mañana.txt': File exists
```

### Expected behavior

it should not happen :)

_No response_

### Operating system

macOS

14.4.1

### Copier version

N/A

### Python version

N/A

### Installation method

pipx+pypi

### Additional context

nix flakes (not in the list)

ink-splatters:
Two versions of `tests/demo/doc/mañana.txt` existed in the repository at the same time, each of them with unique filename.

This was possible because of git ill-formed approach utilizing filesystem-dependent filename representations by default.

It's likely that the mentioned file was added twice: both on Linux and macOS, and due to filename encoding differences between the systems, git permitted it.

This PR fixes #1595 by:
- first removing the affected file
- changing faulty git defaults ( basically - enabling UTF-8 support)
- re-adding the file

### DIFF
--- a/tests/demo/doc/mañana.txt
+++ b/tests/demo/doc/mañana.txt
@@ -1,0 +1,2 @@
+UTF-8（ユーティーエフはち、ユーティーエフエイト）はISO/IEC 10646(UCS)とUnicodeで使える8ビット符号単位の文字符号化形式及び文字符号化スキーム。
+正式名称は、ISO/IEC 10646では‘UCS Transformation Format 8’、Unicodeでは‘Unicode Transformation Format-8’という。両者はISO/IEC 10646とUnicodeのコード重複範囲で互換性がある。RFCにも仕様がある[1]。


yajo:
Tbh I never liked having that file around. The file is used in a couple of tests:

https://github.com/copier-org/copier/blob/69662e3a7c9c59920da2316326bd2cf923aae188/tests/test_exclude.py#L57-L75
https://github.com/copier-org/copier/blob/69662e3a7c9c59920da2316326bd2cf923aae188/tests/test_copy.py#L180
https://github.com/copier-org/copier/blob/69662e3a7c9c59920da2316326bd2cf923aae188/tests/test_copy.py#L234-L246

The tests were skipped on non-linux OS.

I think we should refactor those tests and create the src template in a tmp dir during the test using `build_file_tree`.

Do you think you could do that?

yajo:
The most weird thing is that tests still pass, so I guess they were not doing what I thought 🤔

PR #1597, opened 2024-04-18
fix: render `default` list items for multi-select choice questions

### LINKED ISSUE #1594: default option for choices with multiselect
### Describe the problem

When enabling multiselect, the default value is not selected

### Template

```yaml
python_version:
  type: str
  help: What version of python are you targeting?
  default: "3.11"
  choices:
    - "3.8"
    - "3.9"
    - "3.10"
    - "3.11"
    - "3.12"
 
github_runner_python_version:
  type: str
  help: Which Python versions do you want to use on your Github Runner?
  default: "{{ python_version }}"
  multiselect: true
  choices:
    - "3.8"
    - "3.9"
    - "3.10"
    - "3.11"
    - "3.12"
 ```

### To Reproduce

Use the supplied template and copy, see that the version is not selected

### Logs

_No response_

### Expected behavior

`github_runner_python_version` already has the `python_version` selected

_No response_

### Operating system

Linux

Debian bookworm

### Copier version

copier 9.2.0

### Python version

Python 3.11.8

### Installation method

pip+pypi

### Additional context

_No response_

sisp:
I've fixed a bug related to rendering templated list items which was discovered when trying to template an item in the default value list of a multi-select choice question (see #1594). The root cause was not directly related to rendering the default value but rather caused by the `Question.render_value()` method passing any non-string value through. To resolve this problem fundamentally, I've extended the `Question.render_value()` method to recurse into a list and render each list item individually. I think we should add support for recursing into a `dict` in a follow-up PR and, thus, add support for rendering even deeply nested value in YAML data (e.g. actual, i.e. non-string, YAML choice items and default values to `type: yaml` questions).

Fixes #1594.

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -432,7 +432,11 @@ class Question:
             template = self.jinja_env.from_string(value)
         except TypeError:
             # value was not a string
-            return value
+            return (
+                [self.render_value(item) for item in value]
+                if isinstance(value, list)
+                else value
+            )
         try:
             return template.render({**self.answers.combined, **(extra_answers or {})})
         except UndefinedError as error:
--- a/tests/test_templated_prompt.py
+++ b/tests/test_templated_prompt.py
@@ -362,3 +362,65 @@ def test_templated_prompt_with_conditional_choices(
     for iac in iac_choices:
         tui.expect_exact(iac)
     tui.sendline()
+
+
+def test_multiselect_choices_with_templated_default_value(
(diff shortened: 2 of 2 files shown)


pawamoy (approved):

sisp:
@pawamoy Sorry, I needed for force-push again because I had overlooked the added empty line by a recent Ruff version at the top of `user_data.py`.

sisp:
Ah, no renewed approval is needed. :shrug:

Issue #1547, opened 2024-02-25
Organization avatar cut off


sisp:
I noticed that the `copier-org` organization avatar is cut off at the right:

![image](https://github.com/copier-org/.github/assets/2206639/1e038f69-4760-4ea5-bc05-92f970f5658b)

Also, it seems the "C" from "Copier" is present in the avatar at the right, just barely visible. We should remove the text and add padding to give the logo-only image a 1:1 aspect ratio.

I don't have permissions to change it. Would you mind fixing it, @yajo?

sisp:
I can extract the avatar PNG from the SVG logo on Monday and post it here.

sisp:
Here is the avatar PNG with padding, 1:1 aspect ratio, 512x512px resolution and a transparent background:

![copier-org avatar](https://github.com/copier-org/copier/assets/2206639/89539e94-0a73-4494-a93c-082362e13559)

I hope this works fine. :crossed_fingers:

sisp:
@yajo Do you happen to have a moment to update the org avatar?

yajo:
Ah sorry I forgot! Fixed now.

PR #1599, opened 2024-04-20
docs: include logo

yajo:
It should help fixing https://github.com/copier-org/copier/issues/1547.


sisp:
@yajo FYI, it seems auto-merge requires one approval although you can merge without approval.

PR #1604, opened 2024-04-24
ci: use macOS 13 for Python 3.8 and 3.9

sisp:
I've (hopefully :crossed_fingers:) fixed CI by pinning macOS 13 for Python 3.8 and 3.9, as `macos-latest` has been bumped from macOS 12 (x64-based) to macOS 14 (ARM-based), and macOS 14 does not offer pre-built ARM-binaries for Python 3.8 and 3.9. So, as solution, I suggest to only use `macos-latest` for Python 3.10+. Once Python 3.9 has reached EOL, we can revert this slightly ugly condition again.

See https://github.com/copier-org/copier/pull/1602#pullrequestreview-2019446447 for the original analysis of this problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,19 @@ jobs:
           - ubuntu-latest
           - windows-latest
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        # HACK: Use macOS 13 for Python 3.8 and 3.9, as `macos-latest` (macOS 14+) does
+        # not offer binaries for these Python versions. Remove this temporary solution
+        # once binaries become available or when these Python versions reach EOL.
+        exclude:
+          - os: macos-latest
+            python-version: "3.8"
+          - os: macos-latest
+            python-version: "3.9"
+        include:
+          - os: macos-13
+            python-version: "3.8"
+          - os: macos-13
+            python-version: "3.9"
     runs-on: ${{ matrix.os }}
     steps:
       # HACK https://github.com/actions/cache/issues/315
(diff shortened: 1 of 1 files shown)


sisp:
Oh, I just noticed the CI jobs are still called `build (macos-latest, 3.8)` and `build (macos-latest, 3.9)`, which is of course wrong. Let me see if I can fix this, too.

pawamoy (approved):
I don't mind CI showing `macos-latest`. If you manage to correct it though, that's cool :slightly_smiling_face:

sisp:
I've changed the approach to overriding the macOS version for Python 3.8 and 3.9 by [excluding](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations)  and re-[adding](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations) matrix configurations to first filter out the two Python versions under macOS and then adding them with macOS 13 again. This is a bit more verbose but much more explicit IMO. The cache key needed to be extended to include the architecture of the runner, so a previous macOS x64 cache isn't restored on a macOS ARM runner. I took the opportunity to reformat the cache keys to use a `-` separator, which seems more common according to [GitHub's documentation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action). I also took the liberty and added a nasty `# prettier-ignore` comment above the cache key format it manually because Prettier's line wrapping is (IMHO) really ugly and difficult to read.

WDYT (once CI hopefully passes :crossed_fingers:)?

pawamoy (commented):
I like it :+1: A small thing you could add is a TODO comment (or HACK since it seems often used in Copier) to remind us that it's a temporary solution while waiting for pre-built ARM-binaries for Python 3.8 and 3.9 on MacOS 14, or waiting for Python 3.8 and 3.9 to be EOL. Such comments are easy to grep when a Python version gets EOL, or generally when doing some cleanup. That's what I do in my own repositories :slightly_smiling_face:

sisp:
Good idea, done. :slightly_smiling_face: WDYT?

sisp:
These obsolete CI jobs `build (macos-latest, 3.8)` and `build (macos-latest, 3.9)` are still there and hanging, probably because they don't exist in the new CI config anymore. But they prevent merging. 😕

sisp:
Superseded by #1608.

PR #1633, opened 2024-05-08
feat: Allow copier-templates-extensions as an extra

henryiii:
This allows pipx users to use this:

```bash
pipx install copier[templates-extensions]
```

Or this:

```bash
pipx run copier[templates-extensions] <args>
```

Instead of this:

```bash
pipx install copier
pipx inject copier copier-templates-extensions
```

I think this was mentioned a while back, but never added.

I'm not very familiar with non-PEP 621 backends (Poetry. Just Poetry.), so hopefully I got the syntax correct. Also not sure if you want a min version here.


sisp (changes requested):
LGTM! But the `poetry.lock` file needs to be updated:

```shell
poetry lock --no-update
```

henryiii:
```
So, because copier depends on copier-templates-extensions (*)
 and copier is 0.0.0, version solving failed.
```

henryiii:
(The stuff about limits is bogus, because older versions used to have Python caps, that's not the correct solution and would result in the wrong solve)

henryiii:
I could probably just change the dummy version number, not sure if there's a better way to fix this with the poetry-dynamic-versioning hack.

sisp:
I assume you installed the `poetry-dynamic-versioning` plugin, right? If so, it seems that Poetry's dependency resolver does not consider the version computed by the plugin but instead simply uses the version from the metadata. I wonder whether the plugin could be extended to let Poetry use the computed version during dependency resolution. Then, hardcoding a sufficiently high though arbitrary version in our metadata would not be necessary.

henryiii:
> I assume you installed the poetry-dynamic-versioning plugin, right?

Ah, no, I was just using my brew-installed poetry, which doesn't support adding plugins. I just unlinked my poetry with brew, then installed poetry with pipx and injected the plugin. That fixed it. Other tools (like PDM or Hatch) usually respect your `build-system.requires`, but Poetry plugins apparently don't.

sisp:
Nice! The RTD build is failing because `poetry-dynamic-versioning` is not installed there. This should be fixed by adding `poetry self add "poetry-dynamic-versioning[plugin]"` to `.readthedocs.yml` after Poetry is installed. I'm not sure why the two `CI / flake-check ...` checks are failing though.

henryiii:
Okay, fixed readthedocs. On the nix jobs (CI / flake-check), it's still giving:
```
Run copier --version
/Users/runner/work/_temp/ede4b13e-8f0d-4491-9ff7-7771f5b39c64.sh: line 1: copier: command not found
```

sisp:
The `JRMurr/direnv-nix-action@v3` action fails in the `Nix Environment setup` step with the following error:

```
  Because copier-templates-extensions (0.3.0) depends on copier (>=6)
   and no versions of copier-templates-extensions match !=0.3.0, every version of copier-templates-extensions requires copier (>=6).
  So, because copier depends on copier-templates-extensions (*)
   and copier is 0.0.0, version solving failed.
```

I guess the `poetry-dynamic-versioning` plugin is missing. Will try to look into it some more.

yajo:
One could want to add any jinja extension to the Copier env. I think I like it better without this PR. Sorry...😅

pawamoy:
I'm wondering if copier-templates-extensions shouldn't be integrated into Copier directly, since we have the `--trust` flag now. That would solve the "you have to inject an extension into the env" issue, and we wouldn't need the extra here.

sisp:
Makes sense to me.

yajo:
You mean deprecate the entire project and integrate missing features here?

Yeah, maybe it makes more sense... I never had the need to use it, to be honest. Maybe we can discuss it in a separate issue?

For now, I'll close this. Having this dependency adds some unneeded burden here. FWIW I'd have to fix the flake and the lock file conflicts, things I didn't need to do until now 😆 

Just inject the extension in the pyenv like any other and you're good to go with it.

Thanks anyway!

pawamoy:
@henryiii, you can now tell your users to use `uv`:

```bash
uv tool install copier --with copier-templates-extensions
```

:smile:

PR #1641, opened 2024-05-17
build(lint): use local `ruff` executable

sisp:
This PR contains a few related commits regarding Ruff and pre-commit. Before, Ruff was a Nix dependency and configured as a pre-commit hook via [git-hooks.nix](https://github.com/cachix/git-hooks.nix/). This setup has been a bit problematic:

1.Ruff was either not available to non-Nix users or they were using a different version than the one locked by Nix (e.g. via the VS Code extension).

1. Dependabot cannot update Nix dependencies, so we tend to use an outdated Ruff version.
1. The Nix-locked Ruff version is not obvious, as Nix bundles its packages under a distribution release.

To address these problems and improve our (contributors') developer experience, I've made Ruff a regular Python development dependency managed via Poetry and configured a [custom hook](https://github.com/cachix/git-hooks.nix/?tab=readme-ov-file#custom-hooks) in `flake.nix` which uses the local `ruff` executable installed via Poetry. In a separate commit, I've added a similar custom hook for running Ruff's formatter. The custom hook definitions are based on [Ruff's official pre-commit hook definitions](https://github.com/astral-sh/ruff-pre-commit/blob/v0.4.4/.pre-commit-hooks.yaml) and [Ruff's own use thereof](https://github.com/astral-sh/ruff/blob/v0.4.4/.pre-commit-config.yaml#L58-L65). In another separate commit, I've run Ruff's linter and formatter on our code base.

I think this setup is better calibrated to ease contributions also by non-Nix users while retaining the advantages of Nix.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,17 +1,1 @@
-# SEE https://github.com/direnv/direnv/wiki/Python#poetry
-layout_poetry() {
-    VIRTUAL_ENV=$(poetry env info --path 2>/dev/null ; true)
-
-    if [[ -z $VIRTUAL_ENV || ! -d $VIRTUAL_ENV ]]; then
-        log_status "No virtual environment exists. Executing \`poetry install\` to create one."
-        poetry install --with dev,docs
-        VIRTUAL_ENV=$(poetry env info --path)
-    fi
-
-    PATH_add "$VIRTUAL_ENV/bin"
-    export POETRY_ACTIVE=1
-    export VIRTUAL_ENV
-}
-
 use flake . --impure --accept-flake-config --extra-experimental-features 'nix-command flakes'
-layout_poetry
--- a/copier/__main__.py
+++ b/copier/__main__.py
@@ -1,4 +1,5 @@
 """Copier CLI entrypoint."""
+
 from .cli import CopierApp
 
 # HACK https://github.com/nix-community/poetry2nix/issues/504
--- a/copier/cli.py
+++ b/copier/cli.py
@@ -84,21 +84,18 @@ class CopierApp(cli.Application):  # type: ignore[misc]
     """The Copier CLI application."""
 
     DESCRIPTION = "Create a new project from a template."
-    DESCRIPTION_MORE = (
-        dedent(
-            """\
+    DESCRIPTION_MORE = dedent(
+        """\
             Docs in https://copier.readthedocs.io/
 
             """
-        )
-        + (
-            colors.yellow
-            | dedent(
-                """\
+    ) + (
+        colors.yellow
+        | dedent(
+            """\
                 WARNING! Use only trusted project templates, as they might
                 execute code with the same level of access as your user.\n
                 """
-            )
         )
     )
     VERSION = copier_version()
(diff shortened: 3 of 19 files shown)


sisp:
@yajo I'm a bit stuck here, probably because I don't have enough Nix experience. I can't seem to get the custom pre-commit hooks to find the `ruff` executable installed by Poetry into its virtual env. I wonder whether pre-commit hooks are executed in a clean environment where the environment variables exposed by devenv aren't available. Using a native Nix Ruff package via poetry2nix doesn't seem like a good alternative because [poetry2nix needs a manual update with a hash for each Ruff version](https://github.com/nix-community/poetry2nix/pull/1655). See also https://github.com/nix-community/poetry2nix/blob/master/docs/edgecases.md#errors-that-are-related-to-rust-and-cargo and the subsection "Use `preferWheels` and call it a day".

Any ideas how to make this work? And WDYT in general about this idea?

yajo:
I don't think this will really fix the problem. With this patch, users still need nix to develop as it uses it to configure pre-commit, and thus there's not much benefit to package ruff with poetry. It would be much easier to install https://github.com/marketplace/actions/update-nix-flake-lock and have updated dev env automatically. 

FWIW I configured here the stable release of nixpkgs, which is released every 6 months more or less. That's why ruff is so outdated. We could switch to the unstable release where the latest ruff is already packaged. However dealing with unstabilities twice per year seemed easier to me, and having the very latest formatter doesn't make me lose my sleep. It would be easy to get all from stable and only ruff from unstable, if you think it's worth it.

Pre-commit hooks we use include prettier, taplo and alejandra, which poetry cannot handle as they are js and rust. Thus, to remove nix from the picture, we'd have to rely on pre-commit for installation. And that part of pre-commit is doomed. I collected 21 public issues so far here: https://github.com/OCA/oca-addons-repo-template/issues/175#issue-1466067857; compare that with the close to zero maintenance current setup gives us here. I'd be happy to experiment with https://treefmt.com/ which removes that part and makes the rest much faster.

So, back to the basics, what's the problem here? Simply put, nix scares people or they use Windows. But we want frictionless motivated contributions right? We can just mark the pre-commit ci check as optional on PRs and add https://github.com/stefanzweifel/git-auto-commit-action as a post-merge action. We'd have more reformatting commits in our main history, but if someone comes and fixes a typo in our docs, they won't need to set up any dev env (not even with poetry)  for their valid pr to be merged. WDYT?

sisp:
With this change, users wouldn't need Nix to contribute properly formatted Python code, as Ruff would be a Python dev dependency installed via Poetry. Yes, they wouldn't have pre-commit and wouldn't be able to run pre-commit hooks without Nix, but most contributions change only Python code (or docs) and would benefit from having (the correct version of) Ruff available by simply executing `poetry install --with dev`. Without pre-commit, one can run Ruff manually or use an IDE like VS Code with the right extension.

Also, the current git-hooks.nix hook for Ruff doesn't include its formatter, so CI checks don't detect Ruff formatting errors. Sure, at some point the hook will be added there, but asking contributors to reformat their code although CI checks are passing isn't great.

I feel like Nix should be opt-in as much as possible for contributors, and I understand that pre-commit and some hooks can't be available without Nix – that's simply a limitation of a non-Nix dev setup here –, but pulling Ruff into Python space seems like a low-hanging high-impact change (plus we'd have up-to-date versions maintained by Dependabot).

yajo:
I think we can close this in favor of #1676. Closing optimistically but please reopen if you think this still adds value, and we can keep discussing it.

FWIW I plan to update nix flake inputs soon, so ruff and friends will get updated with it. It's been a long time so it will probably have som conflicts and take a bit of time.

Thanks for your love for Copier and efforts on helping our dear contributors! ❤️

PR #1651, opened 2024-05-27
Fix README callout

ofek:


sisp (commented):
Thanks for your PR! The problem with using this syntax is that we use the README on GitHub and also as the landing page of the MkDocs documentation, and Material for MkDocs doesn't recognize the GitHub syntax. I'm not aware of a single syntax that is recognized by both. The only solution I can think of right now is to use the GitHub syntax as you suggested and add a custom MkDocs hook that transforms it to the syntax that Material for MkDocs uses during the MkDocs build.

pawamoy (commented):
Thanks @ofek! Unfortunately I don't think this will render well once on ReadTheDocs. We'd have to write a Markdown extension that supports GitHub call outs (I don't think markdown-callouts does). It wouldn't be hard, just need a bit of time :)

EDIT: I answered before seeing @sisp's answer, oops ^^

README.md
```diff
@@ -109,12 +109,11 @@ Copier targets these main human audiences:
 1.  **Template creators**. Programmers that repeat code too much and prefer a tool to do
     it for them.
 
-    !!! tip
+> [!TIP]
+> Copier doesn't replace the DRY principle... but sometimes you simply can't be
+> DRY and you need a DRYing machine...
```
ofek, line 114:
```suggestion
      ***Tip:*** Copier doesn't replace the DRY principle... but sometimes you simply can't be DRY and you need a DRYing machine...
```

README.md
```diff
@@ -109,12 +109,11 @@ Copier targets these main human audiences:
...
 
-         Copier doesn't replace the DRY principle... but sometimes you simply can't be
-         DRY and you need a DRYing machine...
-
-1.  **Template consumers**. Programmers that want to start a new project quickly, or
+2.  **Template consumers**. Programmers that want to start a new project quickly, or
```
ofek, line 116:
```suggestion
1.  **Template consumers**. Programmers that want to start a new project quickly, or
```

ofek:
I replaced it with standard markdown syntax, let me know what you think!

ofek:
Just let me know!

yajo (approved):
Thanks!

I like more the callout in readthedocs, but at the same time it's true that the README is a landing place for many, so we can make the exception and stick with basic markdown here.

In the rest of the docs, obviously mkdocs styling has priority.

README.md
```diff
@@ -109,10 +109,7 @@ Copier targets these main human audiences:
 1.  **Template creators**. Programmers that repeat code too much and prefer a tool to do
     it for them.
 
-    !!! tip
-
-         Copier doesn't replace the DRY principle... but sometimes you simply can't be
-         DRY and you need a DRYing machine...
+      ***Tip:*** Copier doesn't replace the DRY principle... but sometimes you simply can't be DRY and you need a DRYing machine...
```
sisp, line 112: Prettier would like this change:

```suggestion
    **_Tip:_** Copier doesn't replace the DRY principle... but sometimes you simply can't be DRY and you need a DRYing machine...
```

yajo:
If you rebase, our CI will try to autofix it.

PR #1676, opened 2024-06-25
ci(pre-commit): auto-fix style problems

yajo:
This is another approach to fix the same problem as https://github.com/copier-org/copier/pull/1641 by just letting the CI reformat and push back changes to the origin branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,8 @@ jobs:
           - macos-latest
           - ubuntu-latest
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write # Allow pushing back pre-commit changes
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


yajo:
Hi @sisp! Sorry because lately I've really had little time for Copier. Here's the implementation for https://github.com/copier-org/copier/pull/1641#issuecomment-2121670088. As you can see, I changed a bit the contributing guidelines to reflect how to contribute without Nix.

I'm deliberately pushing a commit that needs reformatting to see the tool in action. I'll close here so when reopening the workflows run again.

From all our pre-commit hooks, only ruff is pythonic, while alejandra, prettier, editorconfig-checker and taplo are rust, nodejs and go. Thus I don't think we're really adding value by removing ruff from that list. I think this change is smaller and adds more value, as now contributions can be done simply from Github UI and contributors won't need to worry about ugly code.

I think it's specially useful for docs contributions. Those tend to be quick and disposable. Asking users to set up a full dev env for that is even worse in this case. And of course, the ruff case is covered also as a side effect.

sisp (commented):
Sorry for the delay. :face_with_peeking_eye:

I think this PR is a good step forward and your making a strong argument about improving the DX for contributors using GitHub's web IDE. :ok_hand: I also agree that all other hooks are not Python-based and thus Ruff would be a special case if we pursued #1641 further. Great work! :medal_sports:

yajo:
This didn't work out well in https://github.com/copier-org/copier/pull/1651 nor in the merged commit https://github.com/copier-org/copier/commit/798d7aa82068079fd8fc5fb68962f3190d763435. I'll have to think something better like https://autofix.ci

PR #1711, opened 2024-07-24
perf(updating): avoid creating subproject copy

### LINKED ISSUE #1709: Template update process on Copier 9.3.0/9.3.1 is slow/indefinite on Windows compared with 9.2.0 for repos with lots of `.gitignore`s (I suspect a new `copytree` operation)
### Describe the problem

My template https://github.com/blakeNaccarato/copier-python is used for my project https://github.com/softboiler/boilercv, among others. Since Copier 9.3.0, I notice behavior where the cached repo contents in `%TEMP%` during e.g. `copier update --vcs-ref=HEAD` is 17GB+, and includes even `.gitignored` content such as `.venv`, and in this case, a `.dvc/cache` directory containing ~2000 files.

I think that this manifests as Copier trying to update from template and somehow starting to make an insanely large diff against the 17GB+ directory copied to `%TEMP%`. It causes a long-lasting hang during the template updating phase, which was at least fifteen minutes long before I cancelled the operation.

In Copier 9.2.0 and prior, I see the cached repo contents over in `%TEMP%` is only ~1GB, as expected, and doesn't venture into `.gitignored` territory. In comparing diffs across your tagged releases, I see the following section where a `copytree` operation seems to take place, only ignoring the `.git` folder. My suspicion is that this may be involved. See below for the permalink to the potentially offending code snippet.

By updating the template with Copier 9.2.0 instead of 9.3.0 or 9.3.1, I get the relatively fast template update operation that I expect.

https://github.com/copier-org/copier/blob/26c811f2f3152549c9f5d908cba79f142c0eccd6/copier/main.py#L936-L948

### Template

https://github.com/blakeNaccarato/copier-python

### To Reproduce

1. Potentially, be on Windows, but this may be OS agnostic
2. Have a repository with lots and lots of `.gitignored` files (perhaps "lots" by number of files, or by size on disk, or both)
3. Run `copier update` on this repository, with Copier 9.3.0 or 9.3.1
4. Wait forever, `copier update` will not finish

### Logs

_No response_

### Expected behavior

Template updating should behave as in Copier 9.2.0, with no major slowdown and no massive `copytree` operation sending ~17GB to `%TEMP%`.

_No response_

### Operating system

Windows

Windows 11

### Copier version

9.3.0

### Python version

CPython 3.11

### Installation method

pipx+pypi

### Additional context

_No response_

sisp:
I've performed an internal change to the update algorithm to avoid creating a subproject copy, which turns out to be a performance regression introduced by #1587.

As of #1587, the update algorithm computes a diff between the _dirty_ HEAD of the subproject's Git repository – dirty because of potentially [staged files during a pre-migration task](https://github.com/copier-org/copier/blob/b7cc803747feae9e8596458c2a2d6c5c75a7b8b2/tests/test_updatediff.py#L1240-L1250) – and a fresh copy based on the old version of the template. Before this PR, a full copy of the subproject directory (excluding the `.git/` directory) was created and the copy was initialized as a new Git repository with one commit comprising all files. For large subprojects, creating a subproject copy incurs non-negligible overhead as reported in #1709.

This PR takes a different approach that does not involve copying the subproject; in fact, it does not even involve fetching shallow copies of local Git repositories. First, a reference of the dirty HEAD of the subproject (after the pre-migration task has run) is obtained by calling `git write-tree`, which creates a tree object without a commit and returns a reference. Also, the HEAD SHA of the temporary Git repository containing a fresh copy from the new template is determined. Then, the diffs involving the dirty HEAD of the subproject or the fresh copy of the new template are computed by referring to these object references. A [trick for comparing local Git repositories](https://archive.kernel.org/oldwiki/git.wiki.kernel.org/index.php/GitTips.html#How_to_compare_two_local_repositories) makes computing the diffs efficient by avoiding fetching shallow copies from local repositories.

Honestly, I'm not super happy with this solution because the update algorithm is getting more and more complex and it feels like we're increasingly stretching the limits of Git. But it keeps the test suite passing while avoiding the subproject copy. :shrug:

WDYT?

Fixes #1709.

### DIFF
--- a/copier/tools.py
+++ b/copier/tools.py
@@ -232,3 +232,38 @@ def escape_git_path(path: str) -> str:
         lambda match: "".join(f"\\{whitespace}" for whitespace in match.group()),
         path,
     )
+
+
+def get_git_objects_dir(path: Path) -> Path:
+    """Get the absolute path of a Git repository's objects directory."""
+    # FIXME: A lazy import is currently necessary to avoid circular imports with
+    # `errors.py`.
+    from .vcs import get_git
+
+    git = get_git()
+    return Path(
+        git(
+            "-C",
+            path,
+            "rev-parse",
+            "--path-format=absolute",
+            "--git-path",
+            "objects",
+        ).strip()
+    )
+
+
+def set_git_alternates(*repos: Path, path: Path = Path(".")) -> None:
+    """Set Git alternates to borrow Git objects from other repositories.
+
+    Alternates are paths of other repositories' object directories written to
+    `$GIT_DIR/objects/info/alternates` and delimited by the newline character.
+
+    Args:
+        *repos: The paths of repositories from which to borrow Git objects.
+        path: The path of the repository where to set Git alternates. Defaults
+            to the current working directory.
+    """
+    alternates_file = get_git_objects_dir(path) / "info" / "alternates"
+    alternates_file.parent.mkdir(parents=True, exist_ok=True)
+    alternates_file.write_bytes(b"\n".join(map(bytes, map(get_git_objects_dir, repos))))
--- a/copier/main.py
+++ b/copier/main.py
@@ -11,7 +11,7 @@ from filecmp import dircmp
 from functools import cached_property, partial
 from itertools import chain
 from pathlib import Path
-from shutil import copytree, ignore_patterns, rmtree
+from shutil import rmtree
 from tempfile import TemporaryDirectory
 from types import TracebackType
 from typing import (
@@ -54,6 +54,7 @@ from .tools import (
     normalize_git_path,
     printf,
     readlink,
+    set_git_alternates,
 )
 from .types import (
     MISSING,
@@ -923,9 +924,7 @@ class Worker:
             prefix=f"{__name__}.old_copy.",
         ) as old_copy, TemporaryDirectory(
             prefix=f"{__name__}.new_copy.",
-        ) as new_copy, TemporaryDirectory(
-            prefix=f"{__name__}.dst_copy.",
-        ) as dst_copy:
+        ) as new_copy:
             # Copy old template into a temporary destination
             with replace(
                 self,
@@ -941,10 +940,14 @@ class Worker:
             self._execute_tasks(
                 self.template.migration_tasks("before", self.subproject.template)  # type: ignore[arg-type]
             )
+            # Create a Git tree object from the current (possibly dirty) index
+            # and keep the object reference.
+            with local.cwd(subproject_top):
+                subproject_head = git("write-tree").strip()
             with local.cwd(old_copy):
                 self._git_initialize_repo()
-                git("remote", "add", "real_dst", "file://" + str(subproject_top))
-                git("fetch", "--depth=1", "real_dst", "HEAD")
+                # Configure borrowing Git objects from the real destination.
+                set_git_alternates(subproject_top)
                 # Save a list of files that were intentionally removed in the generated
                 # project to avoid recreating them during the update.
                 # Files listed in `skip_if_exists` should only be skipped if they exist.
@@ -954,7 +957,8 @@ class Worker:
                     "-r",
                     "--diff-filter=D",
                     "--name-only",
-                    "HEAD...FETCH_HEAD",
+                    "HEAD",
+                    subproject_head,
                 ).splitlines()
                 exclude_plus_removed = list(
                     set(self.exclude).union(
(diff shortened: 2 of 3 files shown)


sisp:
Well, it's working except on Windows. I'll investigate tomorrow.

sisp:
Now, Windows is happy, too. I missed the Windows-specific alternate object directory separator in the Git documentation (`;` for Windows, `:` for Unix): https://git-scm.com/docs/git#Documentation/git.txt-codeGITALTERNATEOBJECTDIRECTORIEScode

This is ready for a critical review now. :balloon:

pawamoy:
> Honestly, I'm not super happy with this solution because the update algorithm is getting more and more complex and it feels like we're increasingly stretching the limits of Git. But it keeps the test suite passing while avoiding the subproject copy. 🤷

Yep, not great for future maintainers, but what other option do we have :confused:? Git is pretty well documented, and you commented the code well too, so I guess it's OK.

sisp:
I've rebased this PR onto the tip of `master` and factored out enhancing the `git` command with additional object directories into a separate function, as we need it more than once now.

hparfr:
Hi, 

There is also a regression in 9.3.0, 9.3.1, `copier update` fails if there is a directory in the destination with no read/write access for the current user (on linux). (because of `copytree`)

This PR is solves it.

copier/tools.py
```diff
@@ -232,3 +233,24 @@ def escape_git_path(path: str) -> str:
...
+        git_cmd: The Git command.
+        *object_dirs: The Git object directories.
+
+    Returns:
+        LocalCommand: The Git command that searches for Git objects in the
+            specified Git object directories.
+    """
+    return git_cmd.with_env(
+        GIT_ALTERNATE_OBJECT_DIRECTORIES=(";" if OS == "windows" else ":").join(
+            str(object_dir / ".git" / "objects") for object_dir in object_dirs
```
hparfr, line 254: `.git` may be a  file with this kind of contents:
`gitdir: pathtogitdir`

if the repo has been initialized with : 
```sh 
$ git init --separate-git-dir ../pathtogitdir
```

So `(object_dir / ".git" / "objects")` don't work.

hparfr (changes requested):

copier/tools.py
sisp: Great catch, thanks a ton! :ok_hand: I'll work on supporting a separate Git directory.

sisp:
I've added a test that involves updating a project with a separate Git directory. To make this test pass, I've replaced the manual construction of the Git objects directory path by a subprocess call of `git` that returns the correct path also when a separate Git directory is used.

While working on this fix, I gained a better understanding of Git object sharing, mainly from these resources:

* https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt-objectsinfoalternates
* https://docs.gitlab.com/ee/development/git_object_deduplication.html

I had been a little unhappy about using the environment variable `GIT_ALTERNATE_OBJECT_DIRECTORIES` and introducing a helper function that constructs this environment variable and assigns it to a Plumbum `git` command. Then, I realized that Git object sharing can be configured also by adding paths to other repositories' objects directories to `$GIT_DIR/objects/info/alternate`. The result is the same, but the semantics of configuring a Git repository with alternate object stores feels conceptually more sound to me than calling `git` with a particular environment variable.

So, the new implementation involves configuring alternate object directories in the temporary repository created by the update algorithm. Then, a regular `git diff-tree` command works with object references borrowed from linked repositories.

sisp:
Oh, Windows isn't working – again. I'll check what to do about it.

sisp:
It's working on Windows now, too. Git is sensitive to newline characters in the `.git/objects/info/alternates` file; it must be the `\n` character on both Linux/macOS and Windows, not `\r\n` on Windows. I hadn't realized that `Path(...).write_text("\n")` writes `\r\n` on Windows, that's what was keeping the Windows CI jobs from passing.

I think this PR is ready for another review.

hparfr (approved):
tested on linux; it works well with a separate gitdir. many thanks @sisp

sisp:
Thanks again for bringing up the `--separate-git-dir` case and for confirming it's working now, @hparfr. :pray:

yajo (approved):
I can see the benefits although honestly I can't say I fully understand the change 😆 

But all tests are happy, so I'm too. I guess it would be valuable to have a test that makes sure there's not a performance regression in this regard. But I don't know how that could be done without testing the actual implementation, which isn't something generally good.

So it's OK for me. I let you hit merge, just in case there's something else missing. As always, thanks! ❤️

sisp:
In essence, instead of fetching copies of other Git repositories, their object stores are linked (via entries in the `.git/objects/info/alternates` file) which avoids duplicating objects. Then, a foreign object can be referenced by its object ID. This only works for local repositories, but Copier's update algorithm operates on local repositories only (after cloning the old and new template repositories), so all good.

Thanks for the positive feedback. :pray: I'll press merge now.

PR #1716, opened 2024-07-30
fix: support Git config without user identity

### LINKED ISSUE #1714: missing config when using git clone
### Describe the problem

When copier sees a template dir that's a dirty git repo, it will run git clone to a temp dir, and then try to commit dirty changes.
However, since no author is passed by copier, git will fail if no global user.name/user.email is set.

### Template

- init a git repo
- commit some files
- add untracked files

### To Reproduce

`copier /path/to/template .`

### Logs

```console
Command line: | /usr/bin/git --git-dir=.git --work-tree=/home/x/templates/pyproject commit -m 'Copier automated commit for draft changes' --no-verify --no-gpg-sign
Stderr:       | Author identity unknown
              | 
              | *** Please tell me who you are.
              | 
              | Run
              | 
              |   git config --global user.email "you@example.com"
              |   git config --global user.name "Your Name"
              | 
              | to set your account's default identity.
              | Omit --global to set the identity only in this repository.
              | 
```

### Expected behavior

Since the commit is temporary, the author information is not important.
I would expect copier to pass something like `--author='copier <copier@example.com>'` to `git-commit`

_No response_

### Operating system

Linux

debian

### Copier version

9.3.1

### Python version

3.12

### Installation method

pipx+pypi

### Additional context

_No response_

sisp:
I've fixed a problem that caused Copier to fail when Git didn't know the user identity (e.g. via system/global `user.{name,email}` settings) while generating a project based on a dirty template.

The problem was caused by a `git commit` call happening during the `copier.vcs.clone()` call on a dirty template:

https://github.com/copier-org/copier/blob/78c484379804bfdd20a652e2ebb5e5c940b9b828/copier/vcs.py#L192-L200

When the user identity wasn't known to Git, it failed with the following error:

```
Author identity unknown

*** Please tell me who you are.
 
Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.
```

Instead of setting `user.{name,email}` before our internal commits and unsetting them afterwards, I extended the `git` command created by `copier.vcs.get_git()` to include the [Git environment variables](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing) `GIT_{AUTHOR,COMMITTER}_{NAME,EMAIL}`, which take precedence over config files, such that all our internal `git commit` calls are made with user identity information. At the same time, I've unset the user identity information in our test suite to make sure we don't implicitly depend on this information being available. To make `git commit` calls in our tests, I've created a new `git` helper which includes user identity information via the same environment variables as before but with a different name and email to distinguish our internal Copier Git user from the test user.

Fixes #1714.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -1104,8 +1104,6 @@ class Worker:
 
     def _git_commit(self, message: str = "dumb commit") -> None:
         git = get_git()
-        git("config", "user.name", "Copier")
-        git("config", "user.email", "copier@copier")
         # 1st commit could fail if any pre-commit hook reformats code
         # 2nd commit uses --no-verify to disable pre-commit-like checks
         git(
@@ -1124,8 +1122,6 @@ class Worker:
             "--no-gpg-sign",
             "--no-verify",
         )
-        git("config", "--unset", "user.name")
-        git("config", "--unset", "user.email")
 
 
 def run_copy(
--- a/copier/vcs.py
+++ b/copier/vcs.py
@@ -17,10 +17,18 @@ from plumbum.machines import LocalCommand
 from .errors import DirtyLocalWarning, ShallowCloneWarning
 from .types import OptBool, OptStrOrPath, StrOrPath
 
+GIT_USER_NAME = "Copier"
+GIT_USER_EMAIL = "copier@copier"
+
 
 def get_git(context_dir: OptStrOrPath = None) -> LocalCommand:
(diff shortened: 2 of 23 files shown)


pawamoy (approved):
Looking good :) also allowed some cleanup do that's nice :)

Issue #1718, opened 2024-07-31
`skip_if_exists` interaction with deleted files on update


lkubb:
### Describe the problem

When a path that is affected by `skip_if_exists` is deleted in a generated project, an update regenerates it anyways.

### Template

https://github.com/salt-extensions/salt-extension-copier (sadly requires `--trust`, but this could be reproduced easily with a simple demo template)

### To Reproduce

1. `copier copy --trust --defaults --data project_name=foo --data author='Foo Bar' --data author_email='foo@b.ar' https://github.com/salt-extensions/salt-extension-copier foo`
2. `cd foo; git init && git add . && git commit -m "initial commit"`
3. `rm tests/functional/modules/test_foo.py && git commit -am rm`
4. `copier update --trust --skip-answered`
5. `git status` [shows new, untracked `tests/functional/modules/test_foo.py`]

### Expected behavior

The path to stay absent if it was deleted after project generation, as it's usually the case.

### Operating system

macOS

### Operating system distribution and version

14.6

### Copier version

copier 9.1.0

### Python version

CPython 3.11

### Installation method

pipx+pypi

### Additional context

The following code recreates the already deleted file:

https://github.com/copier-org/copier/blob/6fd3ad4af688dae998c0fa77e186e4b4666da7b3/copier/main.py#L954-L963

The diff contains the deleted file, it would be removed usually:

https://github.com/copier-org/copier/blob/6fd3ad4af688dae998c0fa77e186e4b4666da7b3/copier/main.py#L1006-L1010

But `git apply` is told to ignore changes to it, hence it remains:

https://github.com/copier-org/copier/blob/6fd3ad4af688dae998c0fa77e186e4b4666da7b3/copier/main.py#L1029-L1033

A solution could be to:
1. Create another diff between the old copy and the real dest (edit: dest copy, the code changed a bit between 9.1 and current HEAD) with `--diff-filter=D`
2. When applying the changes, first run `git apply --include <skip_if_exists-path-1> [...]` with the second diff on stdin (only if there are any `skip_if_exists` paths)
3. Then continue as usual.

sisp:
Thanks for raising this problem, @lkubb! :pray:

[...]

PR #1719, opened 2024-08-01
fix(updating): exclude deleted paths on update

lkubb:
~~Previously, paths that matched a pattern in `skip_if_exists` and that were deleted in the generated project were recreated during each update.~~

~~These deletions are contained in the diff between old vanilla generated project and the (pre-update) current one. They are not executed though since the paths are excluded from the diff application.~~

Previously, if a file that was deleted intentionally in a generated project was changed in the template, an update would have regenerated it (but not if it did not have any changes).

This patch ensures that update behavior is consistent for paths that are not listed in `skip_if_exists` - don't regenerate deleted paths during `update` - by acquiring a list of deleted paths and excluding them in the copy operations that are relevant for acquiring the update diff. `skip_if_exists` paths are not excluded, their presence is always ensured, even by `copier update` (https://github.com/copier-org/copier/issues/1718#issuecomment-2273622414).

Fixes: https://github.com/copier-org/copier/issues/1721

Thanks a lot for maintaining this very neat tool btw!

### DIFF
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -9,7 +9,6 @@ from typing import Literal
 import pexpect
 import pytest
 import yaml
-from pathspec.patterns.gitwildmatch import GitWildMatchPattern
 from plumbum import local
 
 from copier.cli import CopierApp
@@ -558,34 +557,41 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
 @pytest.mark.parametrize(
     "file_name",
     (
-        "normal_file",
+        "skip_normal_file",
         pytest.param(
-            "unicode_âñ",
+            "skip_unicode_âñ",
             marks=pytest.mark.xfail(
                 platform.system() in {"Darwin", "Windows"},
                 reason="OS without proper UTF-8 filesystem.",
             ),
         ),
-        "file with whitespace",
-        " leading_whitespace",
-        "trailing_whitespace ",
-        "   multi_whitespace   ",
+        "skip file with whitespace",
+        " skip_leading_whitespace",
+        "skip_trailing_whitespace ",
+        "   skip_multi_whitespace   ",
         pytest.param(
-            "\tother_whitespace\t\\t",
+            "\tskip_other_whitespace\t\\t",
             marks=pytest.mark.skipif(
                 platform.system() == "Windows",
                 reason="Disallowed characters in file name",
             ),
         ),
         pytest.param(
-            "back\\space",
+            "\a\f\n\t\vskip_control\a\f\n\t\vcharacters\v\t\n\f\a",
             marks=pytest.mark.skipif(
                 platform.system() == "Windows",
                 reason="Disallowed characters in file name",
             ),
         ),
         pytest.param(
-            "!special",
+            "skip_back\\slash",
+            marks=pytest.mark.skipif(
+                platform.system() == "Windows",
+                reason="Disallowed characters in file name",
+            ),
+        ),
+        pytest.param(
+            "!skip_special",
             marks=pytest.mark.skipif(
                 platform.system() == "Windows",
                 reason="Disallowed characters in file name",
@@ -597,15 +603,16 @@ def test_skip_update_deleted(
     file_name: str, tmp_path_factory: pytest.TempPathFactory
 ) -> None:
     """
-    Ensure that ``skip_if_exists`` does not interfere with
-    deleted paths during updates (i.e. they stay deleted).
-    Issue #1718
+    Ensure that paths in ``skip_if_exists`` are always recreated
+    if they are absent before updating.
+    This was discussed in issue #1718.
     """
     src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
     with local.cwd(src):
         build_file_tree(
             {
-                "copier.yaml": f"_skip_if_exists: ['{GitWildMatchPattern.escape(file_name)}']",
+                "copier.yaml": "_skip_if_exists: ['*skip*']",
                 "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_yaml }}",
                 file_name: "1",
                 "another_file": "foobar",
@@ -627,7 +634,8 @@ def test_skip_update_deleted(
         git("add", ".")
         git("commit", "-m1")
     run_update(dst, overwrite=True)
-    assert not skip_me.exists()
+    assert skip_me.exists()
+    assert skip_me.read_text() == "1"
 
 
 @pytest.mark.parametrize(
@@ -669,7 +677,7 @@ def test_skip_update_deleted(
             ),
         ),
         pytest.param(
-            "back\\space",
+            "back\\slash",
             marks=pytest.mark.skipif(
                 platform.system() == "Windows",
                 reason="Disallowed characters in file name",
@@ -683,7 +691,7 @@ def test_skip_update_deleted(
             ),
         ),
         pytest.param(
-            "not_wildmatch*",
+            "dont_wildmatch*",
             marks=pytest.mark.skipif(
                 platform.system() == "Windows",
                 reason="Disallowed characters in file name",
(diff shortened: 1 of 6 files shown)


sisp:
Thanks for raising this problem and offering a PR to fix it, @lkubb! :pray:

I wonder whether we could take a different approach involving programmatic use of [`exclude`](https://copier.readthedocs.io/en/latest/configuring/#exclude) to avoid unnecessarily re-creating files in the first place rather than deleting re-created files as a post-processing step. Similar to your current implementation, we need to determine which files in the `skip_if_exists` list have been deleted by the user. But if we moved building this list above [overwriting the subproject with the new template version](https://github.com/copier-org/copier/blob/e5403c449f8b746533080e769bf137bd32b698a3/copier/main.py#L954-L964), then we could pass it via the [`exclude` argument](https://github.com/copier-org/copier/blob/e5403c449f8b746533080e769bf137bd32b698a3/copier/main.py#L184) to that worker and also to the worker [creating a fresh copy with the new template version](https://github.com/copier-org/copier/blob/e5403c449f8b746533080e769bf137bd32b698a3/copier/main.py#L965-L974). As a results, the [diff](https://github.com/copier-org/copier/blob/e5403c449f8b746533080e769bf137bd32b698a3/copier/main.py#L1009-L1017) should not contain files from `skip_if_exists`, as they don't get created at all.

I haven't tested this idea, so I might have overlooked something.

WDYT, @lkubb?

lkubb:
> Thanks for raising this problem and offering a PR to fix it, @lkubb! 🙏

@sisp Thanks for taking the time to consider my proposal. :) Responding to your issue comment caused me to find another related inconsistency (imho), https://github.com/copier-org/copier/issues/1721.

> I wonder whether we could take a different approach involving programmatic use of [`exclude`](https://copier.readthedocs.io/en/latest/configuring/#exclude) to avoid unnecessarily re-creating files in the first place rather than deleting re-created files as a post-processing step.

Your proposal additionally allows to fix https://github.com/copier-org/copier/issues/1721 elegantly by excluding all deleted paths (if solution (a) from the issue is deemed acceptable).

> I haven't tested this idea, so I might have overlooked something.

An issue I can see with your proposal in comparison to mine is that tasks (not migrations, here it's an issue for both) might assume the presence of a specific file, but that's common to all `exclude` usage.

I will implement your proposal as described above, fixing the new issue as well. Happy to iterate on this if there is some disagreement. Thanks again for your time!

tests/test_updatediff.py
```diff
@@ -554,6 +554,80 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+    """
+    Ensure that deleted paths are not regenerated during updates,
+    even if the template has changes in that path.
+    Issue #1721
+    """
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "copier.yaml": "_skip_if_exists: [skip_me]",
```
sisp, line 602: This seems to be a leftover from the first test.

```suggestion
```

tests/test_updatediff.py
```diff
@@ -554,6 +554,80 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+    updated_file.unlink()
+    with local.cwd(dst):
+        git("init")
+        git("add", ".")
+        git("commit", "-m1")
+    with local.cwd(src):
+        build_file_tree({"updated_file": "bar"})
+        git("commit", "-am2")
+        git("tag", "2.0.0")
+    run_update(dst, skip_answered=True, overwrite=True)
```
sisp, line 627: There are no questions, so `skip_answered` isn't needed.

```suggestion
    run_update(dst, overwrite=True)
```

tests/test_updatediff.py
```diff
@@ -554,6 +554,80 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+    answers_file = dst / ".copier-answers.yml"
+    answers = yaml.safe_load(answers_file.read_text())
+    assert skip_me.read_text() == "1"
+    assert answers["_commit"] == "1.0.0"
+    skip_me.unlink()
+    with local.cwd(dst):
+        git("init")
+        git("add", ".")
+        git("commit", "-m1")
+    run_update(dst, skip_answered=True, overwrite=True)
```
sisp, line 588: There are no questions, so `skip_answered` isn't needed.

```suggestion
    run_update(dst, overwrite=True)
```

tests/test_updatediff.py
```diff
@@ -554,6 +554,80 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
     assert not (dst / "skip_me.rej").exists()
 
 
+def test_skip_update_deleted(tmp_path_factory: pytest.TempPathFactory) -> None:
```
sisp, line 557: Could you parametrize this test to also test files with special characters that need to be escaped internally for `exclude` to make sure they are actually escaped?

tests/test_updatediff.py
```diff
@@ -554,6 +554,80 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+    skip_me.unlink()
+    with local.cwd(dst):
+        git("init")
+        git("add", ".")
+        git("commit", "-m1")
+    run_update(dst, skip_answered=True, overwrite=True)
+    assert not (dst / "skip_me").exists()
+
+
+def test_update_deleted_path(tmp_path_factory: pytest.TempPathFactory) -> None:
```
sisp, line 592: Could you parametrize this test to also test files with special characters that need to be escaped internally for `exclude` to make sure they are actually escaped?

sisp (changes requested):
Looks great! Just a few minor remarks, and we need to escape the file names to be excluded (see inline comment).

tests/test_updatediff.py
```diff
@@ -554,6 +554,80 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+    with local.cwd(dst):
+        git("init")
+        git("add", ".")
+        git("commit", "-m1")
+    with local.cwd(src):
+        build_file_tree({"updated_file": "bar"})
+        git("commit", "-am2")
+        git("tag", "2.0.0")
+    run_update(dst, skip_answered=True, overwrite=True)
+    assert not (dst / "updated_file").exists()
```
yajo, line 628: Reading the test, I deleted the file but the template changed it, and it's not being skipped. Then I'd expect a merge conflict here. Isn't it?

lkubb: [Copier currently just recreates the file from scratch](https://github.com/copier-org/copier/issues/1721). What you're describing is possible fix (b) from the issue.

From a semantic and practical point of view, I think it makes more sense to ignore the file instead of creating a merge conflict each time the template is updated there. An explicit deletion has a different meaning than changing/emptying a file's contents.

tests/test_updatediff.py
```diff
@@ -554,6 +554,196 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+        ),
+    ),
+)
+def test_skip_update_deleted(
+    file_name: str, tmp_path_factory: pytest.TempPathFactory
+) -> None:
+    """
+    Ensure that paths in ``skip_if_exists`` are always recreated
+    if they are absent before updating.
+    This was discussed in issue #1718.
```
yajo, line 608:
```suggestion
```

tests/test_updatediff.py
```diff
@@ -554,6 +554,196 @@ def test_skip_update(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+        ),
+    ),
+)
+def test_update_deleted_path(
+    file_name: str, tmp_path_factory: pytest.TempPathFactory
+) -> None:
+    """
+    Ensure that deleted paths are not regenerated during updates,
+    even if the template has changes in that path.
+    Issue #1721
```
yajo, line 708:
```suggestion
```

yajo (changes requested):
Thanks! This looks excellent!

Would you please remove those references to issues, so I can merge?

yajo (approved):

sisp (approved):
Excellent work, @lkubb, thanks! :ok_hand:

And thanks for taking over, @yajo; I have in on vacation.

Issue #1725, opened 2024-08-07
Exclude on update


msharp9:
### Actual Situation

I have some files I want generated on copy, but then I want them to be ignored thereafter. I would like the files included on copy but excluded on update.  I can do this now, with a regular `copier copy` then passing in exclude when running update, `copier update -x files`, but it's an extra hassle to type out all the files to exclude on the command line everytime I update.

### Desired Situation

A user runs `copier copy`.  Some template files are deleted, renamed, completely changed based on use case. User runs `copier update` and those files are ignored.

### Proposed solution

If the copier.yml template included an `_exclude_on_update` option that would be ideal.

pawamoy:
I suppose it's like `_skip_if_exist`, but more general (no condition of existence, just ignore on updates).

msharp9:
> I suppose it's like `_skip_if_exist`, but more general (no condition on existence, just ignore on updates).

Yes, but also _skip_if_exist has other uses, like ignoring files when migrating a repo that already exists from another template like cookiecutter.

yajo:
Skip if exists should work for you here. Even if the file exists from a previous cookiecutter template, the point is that you want to do nothing with the file if it exists. In practical terms, you declare that the template doesn't care about that file contents. Where those contents come from, or whether you're copying or updating shouldn't make a difference here.

If you really want that file from the template, then remove it before the first copy.

Please feel free to reopen if you still find a valid use case for the requirement. Thank you!

msharp9:
@yajo sorry, migrating from cookiecutter isn't my use case. I was just pointing out that _skip_if_exists does indeed work for that.  My request is still the opposite of that use case. So I'm a bit confused, did you read the original request?

> the point is that you want to do nothing with the file if it exists. 

No, the point is I want to do nothing with the file after it was initially created, regardless if it exists.

git itself actually does this, so it's a good example.  When you run `git init` it creates a hidden subdirectory .git in your repo.  If you look under .git/hooks, you'll find a host of sample hooks, the most commonly used is probably `pre-commit.sample`.  To actually use it, it isn't uncommon for a user to just rename it to `pre-commit` so `git commit` catches the hook. Some users might delete the samples all together since they "know better" and "it's a waste of space". Regardless, once the file has been generated, we don't ever want to touch it, regenerate it, or what have you.  We want to exclude it regardless of if it exists.  `_skip_if_exists` does not work here.

msharp9:
@yajo I can't reopen this.  Should I submit another request?

[...]

PR #1732, opened 2024-08-12
Add Exclude on Update

msharp9:
This addresses:

https://github.com/copier-org/copier/issues/1725

It looks like @yajo suggested solution was a bit more involved, but I had already started this, and just thought to share.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -954,6 +954,7 @@ class Worker:
             # Do a normal update in final destination
             with replace(
                 self,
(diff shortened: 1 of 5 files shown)


yajo:
Thanks for sharing! I like it because it is more scoped and straightforward. However I think the solution from #1733 covers more cases with about the same effort.

I'll keep this one open until we see how that one progresses, and close it later if I decide to merge #1733 finally.

Regards.

Issue #1625, opened 2024-05-03
Allow templating items in `_exclude`


jzazo:
### Actual Situation

I would like to exclude files / folders conditionally in the `copier.yaml` file. Currently, if you add a file/folder to the exclude list, it is always excluded, even if there is a suffixed file that converts. Therefore, the exclusion list applies to the result (this has been discussed in past issues).

My real use case is that I have a few configuration files that apply to my template, and may or not apply to the destination depending on a certain variable. If they don't apply to the destination I would like to exclude them and not copy irrelevant files. If they apply, I would not exclude and normal overwrite rules apply.

### Desired Situation

Being able to exclude files/folders conditionally would allow having config files in my template that are copied over conditionally on variables. This would be a new powerful feature.

### Proposed solution

I would like to have something in the `copier.yaml` that would allow me to exclude files/folders:
```yaml
_exclude:
  - "{% if variable is defined %}myconfig.yml{% endif %}"
```

I think this solution would not impact current behavior of exlusions applying to final names. It just modifies applying the exclusion conditionally.

pawamoy:
Are you aware that Copier allows to do that thanks to file names?

A file named `{% if some_condition %}filename.ext{% ending %}.jinja` will only be created if the `some_condition` evaluates to true.

jzazo:
Not exactly, because the file in question is configuring the template, it cannot have any other name. That file has to exist as is, and excluded on copy conditionally. At the moment I have to overwrite it and make it clean not to copy my template's config over, but it is not used in the destination. Does it make sense?

pawamoy:
Oh, IIUC you have files that are used both for maintaining the template, and as files to copy to destination projects. Interesting. I'd recommend keeping two separate copies indeed, and using the subdirectory feature to separate them. Just my opinion but if you need to change config for your template you might not want this change to apply to projects too? Better explicit than implicit?

pawamoy:
Other than that, no strong opinion :) If other maintainers think this feature is worth it, I won't mind.

jzazo:
Yeah, let me give you an example. I have a `acl/access.yml` file that specifies who has right permissions in the template repo. When I create a copy of the template to apply to that same server, I ask the pertinent question of who should have access, and format a brand new file from `acl/access.yml.jinja` file, which will replace the template's original. If I am not creating a project for that server, I don't want to have this file at all in the new repo, so I would exclude the folder.

[...]

Issue #377, opened 2021-03-26
Tasks run from wrong folder on update with 6.0.0a6


chris-sanders:
I switched from using copier 5.1.0 to the latest 6.0.0a6 to avoid copier deleting a folder that it didn't create. However with the v6 version I'm having an issue with tasks getting executed by invoke during a copier update.

With version 5 the tasks run as expected, but with v6 during an update tasks that call invoke don't run from the destination folder but from the template sub-folder, where it can't find the expected files.

The template can be found at the v6 branch here: https://github.com/charmed-kubernetes/pytest-operator-template/tree/v6

You can run the `tox -e integration` and you'll see that it fails when it invokes one of the scripts because it can't find the file, which is in the destination folder the task is just not being run in the correct location.

Yajo:
Thanks for testing the alpha and for reporting, I'll add this to the v6 roadmap.

chris-sanders:
I've been looking at this more and I've now triggered it on the v5 aslo. But I think I see what's actually happening. As part of the update process the template is cloned into an empty directory and then diffed against the actual final directory. During the inital copy into the tmp directory it's running the tasks and I have tasks which expect files in the destination directory which fail.

I'm not sure how to best handle that. I can catch the errors and just exit clean, but I actually wanted to fail if the desgination folder doesn't match expectation. I'm applying this template as a delta on top of another folder structure and If the structure doesn't match what I expect I would like it to fail.

Yajo:
There are a couple of ideas that come to my mind.

1. Use [invoke's `--search-root`](http://docs.pyinvoke.org/en/stable/invoke.html#cmdoption-r) to specify where to look for it. [I use it in one of our templates](https://github.com/Tecnativa/doodba-copier-template/blob/72b97c1bafa69b9474ff948f08914e12d3a326f3/copier.yml#L50) and it works fine.
2. > I actually wanted to fail if the destination folder doesn't match expectation

   Currently Copier only runs tasks after the copy. What you want is probably a better system that runs them before, to see if the project expectations are matched. That progress is being tracked in #240 and I hope it's ready for v6 final.

Yajo:
> You can run the `tox -e integration` and you'll see that it fails when it invokes one of the scripts because it can't find the file, which is in the destination folder the task is just not being run in the correct location.

Tests fail with something that doesn't seem related to copier:

<details>

```
╰─ pipx run tox -e integration
[...]

PR #1733, opened 2024-08-13
feat: Add `_operation` variable

lkubb:
* Makes `exclude` configuration templatable.
* Adds a `_copier_operation` variable to the rendering contexts for `exclude` and `tasks`, representing the current operation - either `copy`~~, `recopy`~~ or `update`.
* ~~Adds the (semi-)undocumented `_copier_python`, `_copier_conf.os` and `_folder_name` variables to the docs in a second commit (since I restructured them anyways)~~ ~~extracted into https://github.com/copier-org/copier/pull/1856, will rebase once it's merged~~ rebased

Notes:
This was proposed here: https://github.com/copier-org/copier/issues/1718#issuecomment-2282643624

I hope the way it's implemented and tested is as intended by @yajo. ~~The tests are quite synthetic,~~ I would expect the usual application to be `_copier_operation (!/=)= "update"`.

Fixes: https://github.com/copier-org/copier/issues/1725
Fixes: https://github.com/copier-org/copier/issues/1625
([mostly](https://github.com/copier-org/copier/pull/1733#issuecomment-2621951039)) Fixes: https://github.com/copier-org/copier/issues/240

Alternative to https://github.com/copier-org/copier/pull/1732 (regarding update exclusions only).

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -7,9 +7,10 @@ import platform
 import subprocess
 import sys
 from contextlib import suppress
+from contextvars import ContextVar
 from dataclasses import asdict, field, replace
 from filecmp import dircmp
-from functools import cached_property, partial
+from functools import cached_property, partial, wraps
 from itertools import chain
 from pathlib import Path
 from shutil import rmtree
@@ -65,6 +66,8 @@ from .types import (
     AnyByStrMutableMapping,
     JSONSerializable,
     LazyDict,
+    Operation,
+    ParamSpec,
     Phase,
     RelativePath,
     StrOrPath,
@@ -73,6 +76,29 @@ from .user_data import AnswersMap, Question, load_answersfile_data
 from .vcs import get_git
 
 _T = TypeVar("_T")
+_P = ParamSpec("_P")
+
+_operation: ContextVar[Operation] = ContextVar("_operation")
+
+
+def as_operation(value: Operation) -> Callable[[Callable[_P, _T]], Callable[_P, _T]]:
+    """Decorator to set the current operation context, if not defined already.
+
+    This value is used to template specific configuration options.
+    """
+
+    def _decorator(func: Callable[_P, _T]) -> Callable[_P, _T]:
+        @wraps(func)
+        def _wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _T:
+            token = _operation.set(_operation.get(value))
+            try:
+                return func(*args, **kwargs)
+            finally:
+                _operation.reset(token)
+
+        return _wrapper
+
+    return _decorator
 
 
 @dataclass(config=ConfigDict(extra="forbid"))
@@ -248,7 +274,7 @@ class Worker:
         for method in self._cleanup_hooks:
             method()
 
-    def _check_unsafe(self, mode: Literal["copy", "update"]) -> None:
+    def _check_unsafe(self, mode: Operation) -> None:
         """Check whether a template uses unsafe features."""
         if self.unsafe or self.settings.is_trusted(self.template.url):
             return
@@ -327,8 +353,10 @@ class Worker:
         Arguments:
             tasks: The list of tasks to run.
         """
+        operation = _operation.get()
         for i, task in enumerate(tasks):
             extra_context = {f"_{k}": v for k, v in task.extra_vars.items()}
+            extra_context["_operation"] = operation
 
             if not cast_to_bool(self._render_value(task.condition, extra_context)):
                 continue
@@ -358,7 +386,7 @@ class Worker:
                 / Path(self._render_string(str(task.working_directory), extra_context))
             ).absolute()
 
-            extra_env = {k.upper(): str(v) for k, v in task.extra_vars.items()}
+            extra_env = {k[1:].upper(): str(v) for k, v in extra_context.items()}
             with local.cwd(working_directory), local.env(**extra_env):
                 subprocess.run(task_cmd, shell=use_shell, check=True, env=local.env)
 
@@ -625,7 +653,14 @@ class Worker:
     @cached_property
     def match_exclude(self) -> Callable[[Path], bool]:
         """Get a callable to match paths against all exclusions."""
-        return self._path_matcher(self.all_exclusions)
+        # Include the current operation in the rendering context.
+        # Note: This method is a cached property, it needs to be regenerated
+        # when reusing an instance in different contexts.
+        extra_context = {"_operation": _operation.get()}
+        return self._path_matcher(
+            self._render_string(exclusion, extra_context=extra_context)
+            for exclusion in self.all_exclusions
+        )
 
     @cached_property
     def match_skip(self) -> Callable[[Path], bool]:
@@ -928,6 +963,7 @@ class Worker:
         return self.template.local_abspath / subdir
 
     # Main operations
+    @as_operation("copy")
     def run_copy(self) -> None:
         """Generate a subproject from zero, ignoring what was in the folder.
 
@@ -938,6 +974,11 @@ class Worker:
 
         See [generating a project][generating-a-project].
         """
+        with suppress(AttributeError):
+            # We might have switched operation context, ensure the cached property
+            # is regenerated to re-render templates.
+            del self.match_exclude
+
         self._check_unsafe("copy")
         self._print_message(self.template.message_before_copy)
         with Phase.use(Phase.PROMPT):
@@ -967,6 +1008,7 @@ class Worker:
             # TODO Unify printing tools
             print("")  # padding space
 
+    @as_operation("copy")
     def run_recopy(self) -> None:
         """Update a subproject, keeping answers but discarding evolution."""
         if self.subproject.template is None:
@@ -977,6 +1019,7 @@ class Worker:
         with replace(self, src_path=self.subproject.template.url) as new_worker:
             new_worker.run_copy()
 
+    @as_operation("update")
     def run_update(self) -> None:
         """Update a subproject that was already generated.
 
@@ -1024,6 +1067,11 @@ class Worker:
             print(
                 f"Updating to template version {self.template.version}", file=sys.stderr
             )
+        with suppress(AttributeError):
+            # We might have switched operation context, ensure the cached property
+            # is regenerated to re-render templates.
+            del self.match_exclude
+
         self._apply_update()
         self._print_message(self.template.message_after_update)
 
--- a/copier/main.py
+++ b/copier/main.py
@@ -356,7 +356,7 @@ class Worker:
         operation = _operation.get()
         for i, task in enumerate(tasks):
             extra_context = {f"_{k}": v for k, v in task.extra_vars.items()}
-            extra_context["_operation"] = operation
+            extra_context["_copier_operation"] = operation
 
             if not cast_to_bool(self._render_value(task.condition, extra_context)):
                 continue
@@ -656,7 +656,7 @@ class Worker:
         # Include the current operation in the rendering context.
         # Note: This method is a cached property, it needs to be regenerated
         # when reusing an instance in different contexts.
-        extra_context = {"_operation": _operation.get()}
+        extra_context = {"_copier_operation": _operation.get()}
         return self._path_matcher(
             self._render_string(exclusion, extra_context=extra_context)
             for exclusion in self.all_exclusions
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -962,6 +962,18 @@ to know available options.
 
 The CLI option can be passed several times to add several patterns.
 
+Each pattern can be templated using Jinja.
+
+!!! example
+
+    Templating `exclude` patterns using `_operation` allows to have files
+    that are rendered once during `copy`, but are never updated:
+
+    ```yaml
+    _exclude:
+        - "{% if _operation == 'update' -%}src/*_example.py{% endif %}"
+    ```
+
 !!! info
 
     When you define this parameter in `copier.yml`, it will **replace** the default
@@ -1421,6 +1433,8 @@ configuring `secret: true` in the [advanced prompt format][advanced-prompt-forma
 exist, but always be present. If they do not exist in a project during an `update`
 operation, they will be recreated.
 
+Each pattern can be templated using Jinja.
+
 !!! example
 
     For example, it can be used if your project generates a password the 1st time and
@@ -1571,6 +1585,9 @@ other items not present.
         - [invoke, end-process, "--full-conf={{ _copier_conf|to_json }}"]
         # Your script can be run by the same Python environment used to run Copier
         - ["{{ _copier_python }}", task.py]
+        # Run a command during the initial copy operation only, excluding updates
+        - command: ["{{ _copier_python }}", task.py]
+          when: "{{ _operation == 'copy' }}"
         # OS-specific task (supported values are "linux", "macos", "windows" and `None`)
         - command: rm {{ name_of_the_project }}/README.md
           when: "{{ _copier_conf.os in  ['linux', 'macos'] }}"
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -974,6 +974,9 @@ Each pattern can be templated using Jinja.
         - "{% if _copier_operation == 'update' -%}src/*_example.py{% endif %}"
     ```
 
+    The difference with [skip_if_exists][] is that it will never be rendered during
+    an update, no matter if it exitsts or not.
+
 !!! info
 
     When you define this parameter in `copier.yml`, it will **replace** the default
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -966,12 +966,12 @@ Each pattern can be templated using Jinja.
 
 !!! example
 
-    Templating `exclude` patterns using `_operation` allows to have files
+    Templating `exclude` patterns using `_copier_operation` allows to have files
     that are rendered once during `copy`, but are never updated:
 
     ```yaml
     _exclude:
-        - "{% if _operation == 'update' -%}src/*_example.py{% endif %}"
+        - "{% if _copier_operation == 'update' -%}src/*_example.py{% endif %}"
     ```
 
 !!! info
@@ -1587,7 +1587,7 @@ other items not present.
         - ["{{ _copier_python }}", task.py]
         # Run a command during the initial copy operation only, excluding updates
         - command: ["{{ _copier_python }}", task.py]
-          when: "{{ _operation == 'copy' }}"
+          when: "{{ _copier_operation == 'copy' }}"
         # OS-specific task (supported values are "linux", "macos", "windows" and `None`)
         - command: rm {{ name_of_the_project }}/README.md
           when: "{{ _copier_conf.os in  ['linux', 'macos'] }}"
--- a/docs/creating.md
+++ b/docs/creating.md
@@ -150,7 +150,7 @@ The current phase, one of `"prompt"`,`"tasks"`, `"migrate"` or `"render"`.
 
 Some variables are only available in select contexts:
 
-### `_operation`
+### `_copier_operation`
 
 The current operation, either `"copy"` or `"update"`.
 
--- a/docs/creating.md
+++ b/docs/creating.md
@@ -146,6 +146,16 @@ The current phase, one of `"prompt"`,`"tasks"`, `"migrate"` or `"render"`.
     You may encounter this phase when rendering outside of those phases,
     when rendering lazily (and the phase notion can be irrelevant) or when testing.
 
+## Variables (context-dependent)
+
+Some variables are only available in select contexts:
+
+### `_operation`
+
+The current operation, either `"copy"` or `"update"`.
+
+Availability: [`exclude`](configuring.md#exclude), [`tasks`](configuring.md#tasks)
+
 ## Variables (context-specific)
 
 Some rendering contexts provide variables unique to them:
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -1,0 +1,90 @@
+import json
+from pathlib import Path
+
+import pytest
+from plumbum import local
+
+import copier
+
+from .helpers import build_file_tree, git_save
+
+
+def test_exclude_templating_with_operation(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    """
+    Ensure it's possible to create one-off boilerplate files that are not
+    managed during updates via `_exclude` using the `_operation` context variable.
+    """
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    template = "{% if _operation == 'update' %}copy-only{% endif %}"
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "copier.yml": f'_exclude:\n - "{template}"',
+                "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_yaml }}",
+                "copy-only": "foo",
+                "copy-and-update": "foo",
+            }
+        )
+        git_save(tag="1.0.0")
+        build_file_tree(
+            {
+                "copy-only": "bar",
+                "copy-and-update": "bar",
+            }
+        )
+        git_save(tag="2.0.0")
+    copy_only = dst / "copy-only"
+    copy_and_update = dst / "copy-and-update"
+
+    copier.run_copy(str(src), dst, defaults=True, overwrite=True, vcs_ref="1.0.0")
+    for file in (copy_only, copy_and_update):
+        assert file.exists()
+        assert file.read_text() == "foo"
+
+    with local.cwd(dst):
+        git_save()
+
+    copier.run_update(str(dst), overwrite=True)
+    assert copy_only.read_text() == "foo"
+    assert copy_and_update.read_text() == "bar"
+
+
+def test_task_templating_with_operation(
+    tmp_path_factory: pytest.TempPathFactory, tmp_path: Path
+) -> None:
+    """
+    Ensure that it is possible to define tasks that are only executed when copying.
+    """
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    # Use a file outside the Copier working directories to ensure accurate tracking
+    task_counter = tmp_path / "task_calls.txt"
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "copier.yml": (
+                    f"""\
+                    _tasks:
+                        -   command: echo {{{{ _operation }}}} >> {json.dumps(str(task_counter))}
+                            when: "{{{{ _operation == 'copy' }}}}"
+                    """
+                ),
+                "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_yaml }}",
+            }
+        )
+        git_save(tag="1.0.0")
+
+    copier.run_copy(str(src), dst, defaults=True, overwrite=True, unsafe=True)
+    assert task_counter.exists()
+    assert len(task_counter.read_text().splitlines()) == 1
+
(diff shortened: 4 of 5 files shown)


copier/main.py
```diff
@@ -234,7 +238,7 @@ def _cleanup(self) -> None:
         for method in self._cleanup_hooks:
             method()
 
-    def _check_unsafe(self, mode: Literal["copy", "update"]) -> None:
+    def _check_unsafe(self, mode: Operation) -> None:
```
yajo, line 241: No need to pass ˋmodeˋ if it is in ˋself.operationˋ, right?

tests/conftest.py
```diff
@@ -12,6 +12,10 @@
 
 from .helpers import Spawn
 
+pytest_plugins = [
+    "tests.templates",
+]
```
yajo, line 17: I think the way tests are written is a bit more theoretical and complex than necessary.

To test the contents of ˋ_copier_conf.operationˋ, you could just do a single test that prints its output in a file. Then copy, recopy and update in sequence, and assert on each step the changes on the file.

In particular, you're not testing the use of this feature under ˋ_excludeˋ for updates, which is the specific use case we want to support. This could be a separate test with a separate template.

Since those would be just 2 tests, you wouldn't need to share these extra fixtures accross modules, and that'd make the tests easier to read and follow.

So would you please refactor the tests following this method?

copier/main.py
```diff
@@ -60,6 +60,9 @@
 _T = TypeVar("_T")
 
 
+Operation = Literal["copy", "recopy", "update"]
```
yajo, line 63: ˋrecopyˋ is just a shortcut to ˋcopyˋ from the same source as the last time. Maybe ˋrecopyˋ should be removed from this list?

Ping @copier-org/maintainers

yajo (changes requested):
The docs would need to reflect this change too.

copier/main.py
lkubb: Initially I did it like that, but noticed that this would cause a behavior change (at least in theory):

During `_apply_update`, `self.operation` is `update`, but it calls on `run_copy` several times, which would pass `copy` to `_check_unsafe` before this patch.

copier/main.py
lkubb: I don't have a preference here. Left it as-is for now since most applications will just care about `update` vs everything else, but I'm fine with `recopy` == `copy` as well.

sisp: I think `copy` & `update` should be enough for now because `recopy` is just `copy` with some minor usage convenience. If it turns out later that distinguishing `copy` and `recopy` makes sense, we can easily add `recopy`. But unless there's a clear use case for `recopy`, I'd not add it prematurely.

lkubb: Alright, I removed `recopy` now. :)

lkubb:
@yajo Is there anything I still need to do here? Just making sure you noticed the changes. :)

copier/main.py
```diff
@@ -68,6 +68,9 @@
 _T = TypeVar("_T")
 
 
+Operation = Literal["copy", "update"]
```
yajo, line 71: Please move this to `copier.types` module

copier/main.py
yajo: Yes. However, that is correct. You will notice that there are calls to `replace`. In those calls, you can replace some configuration for the sub-worker that is created. Could you please try doing it that way?

copier/main.py
```diff
@@ -203,6 +206,7 @@ class Worker:
     unsafe: bool = False
     skip_answered: bool = False
     skip_tasks: bool = False
+    operation: Operation = "copy"
```
yajo, line 209: thought: I find this a bit counter-intuitive. We have to pass an operation now to the worker. The operation could be something like "update" for example. But then we can `run_copy()` or `run_update()`.

It seems like duplicated features and inconsistent API behavior. If this is the case now, I think we should have a unified `run_operation()` method that runs either one or the other behind the scenes.

When the workers are created, they should be with the correct operation, and from there onwards use that unless `replace()` changes it.

tests/test_context.py
```diff
@@ -0,0 +1,77 @@
...
+    assert ctx_file.read_text() == "copy"
+    # Ensure the file is regenerated on update.
+    # If we left it, an update would detect custom changes
+    # that would be reapplied, i.e. an update would leave us with `copy`.
+    ctx_file.unlink()
+    with local.cwd(dst):
+        git_save()
+    getattr(copier, f"run_{operation}")(str(dst), overwrite=True)
+    expected = "copy" if operation == "recopy" else operation
+    assert ctx_file.read_text() == expected
```
yajo, line 40: thought: So, let's say that we copy one template and render the value. The file will contain `copy`.

Now we run it again. The file contains `update`. The change is OK because the replay renders `copy` and detects the change.

However, what would happen if we run it yet again? The replay will render `copy`, but in reality it is `update`. Now the next one is also `update`.

Triple-thinking this... are we introducing here a way to make replays non-reproducible and thus maybe making the update algorithm unreliable? Because there's no way Copier could know if last play was a copy or an update.

CC @copier-org/maintainers. Should we just reject this feature to avoid shooting our own feet?

yajo (changes requested):
Sorry for taking so long to review! I don't have a lot of time and this one required deep thinking.

copier/main.py
lkubb: > We have to pass an operation now to the worker.

We mustn't pass it to the worker though, it's an internal reference that has to be `copy` by default, only overridden by calls to `replace` in `run_update`.

I wanted to declare this field as `init=False` for that reason, but that does not work together with `replace` by default (see the last paragraph on [dataclasses.replace](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace)). This can be fixed by using a custom `replace` function though.

copier/main.py
lkubb: Sorry, I'm not sure I'm following. First, let me sum up:

* Introducing `_copier_conf.operation` means we have an attribute on the worker representing the current high-level (user-requested) operation.
* You're proposing to use this reference for `_check_unsafe` instead of the parameter.
* I noted that doing this will change how `_check_unsafe` behaves during the individual copy operations that run during an update, where the high-level operation is `update`, but the low-level one is `copy`.

I'm already using `replace` for overriding the operation during `update`. Are you saying the high-level operation during the individual copy operations should be `copy`? Because that would mean `_copier_conf.operation` is *always* `copy`, i.e. defeat the purpose of this feature.

tests/test_context.py
lkubb: Disregarding that `_copier_conf.operation` is not intended to be used for template content changes, this test for checking the value of `operation` only works because the file containing `{{ _copier_conf.operation }}`

1. is deleted
2. is marked as `skip_if_exists`, which means it will be regenerated during `update` if deleted.

What would happen without it being deleted is the following:

1. `copier copy` renders the template, it contains `copy`
2. `copier recopy` overwrites the template, it contains `copy`
3. `copier update` updates the template, it still contains `copy`. This is valid for each subsequent update.

Why (3):
During `_apply_update`, `old_copy` and `new_copy` are both rendered with `_copier_conf.operation == "update"`. The file contains `copy` though, which is detected as a custom change and reapplied.

`_copier_conf.operation` should be documented as a meta variable, intended for templating `copier.yml` configuration. It cannot be used to influence template contents.

sisp (commented):
Sorry for taking so long to give feedback. :face_with_peeking_eye: This PR really needs some deep thinking.

Adding the attribute `operation` to the `Worker` class seems conceptually wrong to me because we'd make the operation mode part of its state while it also has methods to execute either operation (`run_copy()` and `run_update()`). Thus, it would be possible to set the `operation` attribute to `"update"` and call `run_copy()` (or set the `operation` attribute to `"copy"` and call `run_update()`) which would lead to incorrect behavior. This indicates a bad design to me.

How about using a [context variable](https://docs.python.org/3/library/contextvars.html) to manage the operation context? This is a simplified example of what I mean:

```python
from __future__ import annotations

import sys
from contextvars import ContextVar
from dataclasses import dataclass, replace
from functools import wraps
from typing import Callable, Literal, TypeVar

if sys.version_info >= (3, 10):
    from typing import ParamSpec
else:
    from typing_extensions import ParamSpec


Operation = Literal["copy", "update"]
P = ParamSpec("P")
R = TypeVar("R")

_operation: ContextVar[Operation] = ContextVar("_operation")


def as_operation(value: Operation) -> Callable[[Callable[P, R]], Callable[P, R]]:
    def _decorator(func: Callable[P, R]) -> Callable[P, R]:
        @wraps(func)
        def _wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
            token = _operation.set(_operation.get(value))
            try:
                return func(*args, **kwargs)
            finally:
                _operation.reset(token)

        return _wrapper

    return _decorator


@dataclass
class Worker:
    template: str

    @as_operation("copy")
    def run_copy(self) -> None:
        print(f"Worker({self.template}).run_copy(): {_operation.get()}")

    @as_operation("update")
    def run_update(self) -> None:
        print(f"Worker({self.template}).run_update(): {_operation.get()}")
        replace(self, template="old").run_copy()
        replace(self, template="new").run_copy()


worker = Worker("new")

print("$ copier copy ...")
worker.run_copy()
# -> Worker(new).run_copy(): copy

print("$ copier update ...")
worker.run_update()
# -> Worker(new).run_update(): update
# -> Worker(old).run_copy(): update
# -> Worker(new).run_copy(): update
```

The `Worker.run_copy()` and `Worker.run_update()` methods both set their operation context via the decorator `as_operation()`. The decorator wraps the decorated method and sets the operation context to the provided value _only when the operation context has not been set yet_; if the operation context has been set, it remains as is. Thus, a call of `Worker.run_update()` sets the operation context to `"update"`, and when `Worker.run_copy()` gets called within this context, the operation context will remain `"update"`.

Using `ContextVar` is thread-safe. Also, the following requirement stated in the Python docs

> **Important:** Context Variables should be created at the top module level and never in closures. [Context](https://docs.python.org/3/library/contextvars.html#contextvars.Context) objects hold strong references to context variables which prevents context variables from being properly garbage collected.

is met.

I agree that the Jinja variable `_copier_conf.operation` only seems to make sense when rendering template expressions in `copier.yaml`. And although the update seems to produce correct results, the fact that the `_copier_conf.operation` value is unstable during an update (while rendering files etc.), it is asking for trouble IMO. So, how about omitting the `_copier_conf.operation` value in the Jinja context while rendering files, directories, and path names? We'd need to document this behavior of course.

WDYT, @lkubb @yajo?

lkubb:
> Sorry for taking so long to give feedback. 🫣 This PR really needs some deep thinking.

@sisp Thank you for your thoughtful feedback and sorry for taking my time to reply as well.

> Adding the attribute `operation` to the `Worker` class seems conceptually wrong to me because we'd make the operation mode part of its state while it also has methods to execute either operation (`run_copy()` and `run_update()`).

Agreed, I wasn't really happy with it myself, also because it meant the forced inclusion of the operation in inappropriate contexts. Part of this design was caused by the tight coupling of the Copier worker with the renderer imho, I wasn't sure how to solve it properly without a major refactoring.

> Thus, it would be possible to set the `operation` attribute to `"update"` and call `run_copy()` (or set the `operation` attribute to `"copy"` and call `run_update()`) which would lead to incorrect behavior. This indicates a bad design to me.

Note that it's only the former situation that leads to incorrect behavior, the latter is the expected path since the `run_update` method replaces the `operation` value itself.

> How about using a [context variable](https://docs.python.org/3/library/contextvars.html) to manage the operation context? This is a simplified example of what I mean:

Thanks a lot for your well thought-out proposal. I agree this design is superior and am currently implementing it. I'm still contemplating a detail since it aggravates a flaw I just noticed in the `exclude` rendering logic by decoupling the context switch implementation from the worker:

Since `match_excludes` is a cached property whose value now depends on the `_operation`, a changed `_operation` is not reflected in its value if an instance has already been used in a different context. This is not an issue in practice currently since a context switch always coincides with the creation of a new worker instance via `replace`, but it's definitely a gotcha in case this changes at some point.

[Note: An analog of this is currently present in `match_skip`, especially if externally mutated instance attributes need to be considered in worker functions].

I might need to address this by `del`eting the property's cached value in the `finally` block of the context manager to force regeneration, which is still a bit smelly, or turning it into a cached function instead (less smelly, but a breaking change).

> So, how about omitting the `_copier_conf.operation` variable in the Jinja context while rendering files, directories, and path names? We'd need to document this behavior of course.

👍

tests/test_context.py
```diff
@@ -0,0 +1,94 @@
...
+    with local.cwd(dst):
+        git_save()
+
+    copier.run_recopy(dst, defaults=True, overwrite=True, unsafe=True)
+    assert dst_file.exists()
+
+    dst_file.unlink()
+
+    copier.run_update(dst, defaults=True, overwrite=True, unsafe=True)
+    assert not dst_file.exists()
```
sisp, line 94: I _think_ this sequence of commands and assertions is valid, but the update algorithm also runs tasks by calling `Worker.run_copy()` internally, so I'm not sure I feel comfortable with this test setup. How about counting the number of times the task is called by appending, e.g., a character or line to a file and asserting the number of characters or lines?

The task command could be

```yaml
command: [echo task >> task_calls.txt]
```

(the use of `echo <string> >> <filename>` should be platform-agnostic) and we could assert the number of lines.

WDYT?

sisp (changes requested):
Thanks for updating this PR, @lkubb! :+1: This looks very good, just a couple of quite minor requests and ideas from my side.

docs/creating.md
```diff
@@ -75,27 +75,68 @@ their usage.
...
+
+The current operation, either `copy` or `update`.
+
+Availability: [`exclude`](configuring.md#exclude), [`tasks`](configuring.md#tasks)
+
+## Variables (context-specific)
+
+Some rendering contexts provide variables unique to them:
+
+-   [`migrations`](configuring.md#migrations)
```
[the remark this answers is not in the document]
lkubb: Sure :) https://github.com/copier-org/copier/pull/1856

copier/main.py
```diff
@@ -243,7 +269,7 @@ def _cleanup(self) -> None:
         for method in self._cleanup_hooks:
             method()
 
-    def _check_unsafe(self, mode: Literal["copy", "update"]) -> None:
+    def _check_unsafe(self, mode: Operation) -> None:
```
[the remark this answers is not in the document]
lkubb: I'm fine with either, just wanted to point out that in my mind there is a minor semantical difference:

`operation` is always the user-requested, high-level operation.
`mode` in this function can be `copy` during an update since `run_copy` is called inside `run_update`, which in turn calls `_check_unsafe` with `copy`.

tests/test_context.py
```diff
@@ -0,0 +1,94 @@
...
+
+import pytest
+from plumbum import local
+
+import copier
+
+from .helpers import build_file_tree, git_save
+
+
+def test_exclude_templating_with_operation(tmp_path_factory: pytest.TempPathFactory) -> None:
```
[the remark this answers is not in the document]
lkubb: I often wondered the same when creating contributions for Copier, `ruff` doesn't seem too eager to reformat. `black` seems a bit more strict.

tests/test_context.py
```diff
@@ -0,0 +1,94 @@
...
+    assert boilerplate.read_text() == "foo"  # This file is excluded from updates
+    assert other_file.read_text() == "bar"
+
+    # After using the worker for an `update` operation, reuse it for a `copy` again.
+    # This checks that the cached `match_exclude` property is regenerated
+    # after a context switch back from update to copy.
+    copier.run_copy(str(src), dst2, defaults=True, overwrite=True, vcs_ref="1.0.0")
+    for filename in ("boilerplate", "other_file"):
+        assert (dst2 / filename).exists()
+        assert (dst2 / filename).read_text() == "foo"
```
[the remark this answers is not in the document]
lkubb: Yeah, it's a completely theoretical issue at the moment as described [here](https://github.com/copier-org/copier/pull/1733#issuecomment-2463617397). This was intended to warn future developers in case this is changed at some point. Should I remove it?

tests/test_context.py
lkubb: Good catch. I took your suggestion and ensured the target path is outside of the Copier working directories, which should provide an accurate execution count.

mschoettle:
> Adds an _operation variable to the rendering contexts for exclude and tasks, representing the current operation - either copy, recopy or update.

What this PR adds is exactly what I am missing. 

I recently started using copier for an infrastructure project using compose, env files etc. with various different configuration options. 

There are some things, such as questions/answers and tasks, that I don’t want to ask/run on update. And, other tasks should only run when an update is performed. 

I am wondering what the rationale is behind not making the operation available to questions (conditions), if I read the above correctly. Would it be possible to use it in `when`?

lkubb:
@mschoettle 

> There are some things, such as questions/answers and tasks, that I don’t want to ask/run on update. And, other tasks should only run when an update is performed.

For the latter I'd recommend using migrations. They can be configured to run independent of the involved versions.
 
> I am wondering what the rationale is behind not making the operation available to questions (conditions), if I read the above correctly. Would it be possible to use it in `when`?

No, the reason is that `when` == `False` variables are not saved, so you'd lose previously provided answers during an update. If I understood correctly, there's a workaround described in https://github.com/copier-org/copier/issues/1574#issuecomment-2067980359

tests/test_context.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    task = {
+        "command": ["{{ _copier_python }}", "-c", "from pathlib import Path; Path('foo').touch(exist_ok=False)"],
+        "when": "{{ _operation == 'copy'}}",
+    }
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "copier.yml": f"_tasks: {json.dumps([task])}",
```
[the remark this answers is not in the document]
lkubb: After applying this, `editorconfig-check` complained about indentation, which is why the formatting is a bit weird now. The path needs to be JSON-escaped on Windows because the YAML parsing fails otherwise.

copier/main.py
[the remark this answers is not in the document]
sisp: Makes sense, let's keep `mode`.

docs/configuring.md
```diff
@@ -893,6 +893,18 @@ to know available options.
...
+Each pattern can be templated using Jinja.
+
+!!! example
+
+    Templating `exclude` patterns using `_operation` allows to have files
+    that are rendered once during `copy`, but are never updated:
+
+    ```yaml
+    _exclude:
+        - "{%- if _operation == 'update' -%}src/*_example.py{%- endif %}"
```
sisp, line 905: Since there's no need to strip whitespaces, let's keep it simple and avoid confusing readers as to why we're showing this expression with the marker:

```suggestion
        - "{% if _operation == 'update' %}src/*_example.py{% endif %}"
```

tests/test_context.py
```diff
@@ -0,0 +1,90 @@
...
+def test_exclude_templating_with_operation(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    """
+    Ensure it's possible to create one-off boilerplate files that are not
+    managed during updates via `_exclude` using the `_operation` context variable.
+    """
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    template = r"{%- if _operation == 'update' %}copy-only{%- endif %}"
```
sisp, line 21: Just to keep it simple and avoid using a raw string and stripping whitespaces when not actually needed:

```suggestion
    template = "{% if _operation == 'update' %}copy-only{% endif %}"
```

docs/creating.md
```diff
@@ -75,27 +75,68 @@ their usage.
...
+
+The name of the project root directory.
+
+## Variables (context-dependent)
+
+Some variables are only available in select contexts:
+
+### `_operation`
+
+The current operation, either `copy` or `update`.
```
sisp, line 134:
```suggestion
The current operation, either `"copy"` or `"update"`.
```

sisp (approved):
Fantastic work, @lkubb! :ok_hand: LGTM! :tada:

pawamoy:
Do we agree that this PR also fixes the long-standing https://github.com/copier-org/copier/issues/240?

sisp:
@pawamoy I think it addresses part of #240, but pre/post copy/update granularity is not addressed. There's #1511, but it seems stale.

@lkubb @copier-org/maintainers A thought that originates from https://github.com/copier-org/copier/pull/1948#pullrequestreview-2578227878 (and subsequent comments): I think we should rename `_operation` to `_copier_operation`.

pawamoy:
If I read #240 correctly, pre/post were specific to migrations, and are now handled with a `_stage` variable being `"before"` or `"after"`. Tasks get the same `_stage` variable with value `"task"`.

I'm not sure what pre-copy, post-copy, pre-update, or post-update would mean exactly. IMO this PR solves #240.

Let me also link again to [my comment above](https://github.com/copier-org/copier/pull/1733#discussion_r1922313252) about a third update-diff operation, which I think is super important to have.

copier/main.py
sisp: The current implementation, which uses a [context variable and decorator](https://github.com/copier-org/copier/pull/1733#pullrequestreview-2397086710), exposes `_operation = "update"` to the copy tasks, so Copier's internals of re-generating a fresh project aren't leaked to the template developer. So, a task can be limited to running only on `copier copy` with `{% if _operation == "copy" %}./run-task.sh{% endif %}`. Does this address your concern?

pawamoy: What about update tasks that will run in temporary copies then? :thinking: My point is:

- `copier copy` -> `copy` op -> all good
- `copier update` -> `update` op (temp copy) then `update` op again (actual update) -> update tasks run in temp copy too, might cause issues

But I checked #240 again and I only see people with `git init`-like use-cases, i.e. they generally want to run something at `copy` and not at `update`, not the opposite. So for now the current changes might be enough :slightly_smiling_face:

pawamoy: OK, here's an example: https://github.com/copier-org/copier/issues/377#issuecomment-808465792.

> I've been looking at this more and I've now triggered it on the v5 aslo. But I think I see what's actually happening. As part of the update process the template is cloned into an empty directory and then diffed against the actual final directory. During the initial copy into the tmp directory it's running the tasks and I have tasks which expect files in the destination directory which fail.

> I'm not sure how to best handle that. I can catch the errors and just exit clean, but I actually wanted to fail if the destination folder doesn't match expectation. I'm applying this template as a delta on top of another folder structure and If the structure doesn't match what I expect I would like it to fail.

The user here would configure their task as "update" tasks. But they would therefore run in the temporary copy during an update, where they would fail.

sisp:
Ah, you're right, @pawamoy. I got confused. :confused: But looking at https://copier.readthedocs.io/en/stable/updating/#how-the-update-works (and thinking about the detailed implementation), I think are pre-/post-migrations are equivalent to pre-/post-update tasks for `copier update`. So the only thing we seem to be missing is a pre-copy task. Right? :thinking:

copier/main.py
sisp: Your overview of the situation is very helpful, thanks! Then, I see two scenarios and solutions:

1. Run tasks only after `copier copy` (e.g., `git init` – i.e., no task run after internal temporary copy):

    ```yaml
    _tasks:
      - command: git init
         when: "{{ if _operation == 'copy' }}"
    ```

1. Run tasks only after `copier update` (e.g., "resolve and install"):

    ```yaml
    _migrations:
      - ./resolve-and-install.sh
    ```

I must admit that this situation is not entirely obvious. We might need to add documentation to clarify it better.

pawamoy: Yep it's a bit counter-intuitive to have to use a migration to avoid running update tasks in the temp copy. Docs would be a good first step!

pawamoy:
> So the only thing we seem to be missing is a pre-copy task. Right?

Seems like it, yes! I feel like this deserves a proper entry in the docs, something like "Pre/post hooks", which is how Cookiecutter named them, and will be familiar to most users. This entry would document that you can use migration as pre/post-update hooks (with examples), tasks as post-copy or post-update hooks, and... pre-copy hooks are not supported yet. The docs entry should emphasize that for simple message display, users can set `message_after_copy` and related settings. Let me copy this in a new docs issue.

lkubb:
>  I think we should rename `_operation` to `_copier_operation`.

@sisp Done and rebased.

Given the many comments this PR has accumulated over the half year it has been open, I'll try to give a final summary of what it provides and where it fits in the Copier hook system.

<details>
<summary>This PR</summary>

1) Adds templating functionality to `_exclude` items
2) Adds a `_copier_operation` context variable for the `_tasks` and `_exclude` rendering contexts, which indicates whether `copier (re)copy` or `copier update` was invoked by the user.

(1+2) allows to provide dumb boilerplate files which will be excluded during updates
(2) additionally relates to hooks, see `after a copy` below
</details>


If you need your logic to run...

<details>
<summary>before/after any `update`</summary>
Use a `migration` without specifying a version constraint and with appropriate `when` (or even `when: 'true'` to run both before and after). Example:

```yaml
_migrations:
  - this_script_runs_after_all_updates.sh
  - command: this_script_runs_before_all_updates.sh
  - command: this_script_runs_before_and_after_all_updates.sh
    when: "true"
```
</details>

<details>
<summary>before a `copy`</summary>
Come up with an ugly hack possibly involving a custom extension or provide a patch to core, not officially supported atm
</details>

<details>
<summary>after a `copy`</summary>
Use a `task` with `when: '{{ _copier_operation == "copy" }}'` Example:

```yaml
_tasks:
  - command: this_script_runs_after_copy_operations_only.sh
    when: "{{ _copier_operation == 'copy' }}"
```
</details>

> [!NOTE]
> All of the hook suggestions apply to the **generated project** only, not the two intermediate copies involved in `update` operations. If you need a task to run on these intermediate copies as well, specify a plain `task`. Off the top of my head, I can't come up with any use case for this, it seems fragile.


Is there anything I can do to help carry this PR over the line? It has been stalled for a while now and the release of Copier 9.5.0 broke the ugly hack I was using for not having `_copier_operation`:

```yaml
_tasks:
  # Don't run this when updating (updating copies the template twice).
  # This breaks when recopying.
  - >-
    {%- if
      not _copier_conf.answers.last
      and '.new_copy.' not in _copier_conf.dst_path.name
      and '.old_copy.' not in _copier_conf.dst_path.name -%}
        "{{ _copier_python }}" "{{ "{src}{sep}tasks{sep}initialize.py".format(src=_copier_conf.src_path, sep=_copier_conf.sep) }}" init
    {%- endif -%}
```

Specifically https://github.com/copier-org/copier/pull/1880, which added this line and thereby removed `_copier_conf.answers.last` from the context:

https://github.com/copier-org/copier/blob/e444514e5d714a318cf434bcb7fec13d21dfc93c/copier/main.py#L382

docs/configuring.md
```diff
@@ -962,6 +962,18 @@ to know available options.
...
+
+!!! example
+
+    Templating `exclude` patterns using `_copier_operation` allows to have files
+    that are rendered once during `copy`, but are never updated:
+
+    ```yaml
+    _exclude:
+        - "{% if _copier_operation == 'update' -%}src/*_example.py{% endif %}"
+    ```
```
yajo, line 975: I think this deserves an extra explanation:

```suggestion
    ```

    The difference with [skip_if_exists][] is that it will never be rendered during
    an update, no matter if it exitsts or not.
```

yajo (changes requested):
I think there's a needed small extra explanation in the docs. Other than that, if it looks good to you, it's good to me :D

yajo (approved):

PR #1739, opened 2024-08-20
fix: parse new answer when `--skip-answered` is used

### LINKED ISSUE #1728: Bool argument is handled as string in 'recopy' operation
### Describe the problem

An argument which is correctly handled as bool in `copy` and `update` becomes a string in `recopy` operation.

### Template

See #1727 for template

### To Reproduce

See #1727 for initial steps

```copier recopy -A --data with_internal1=false --data show_internal_test_options=true```

The result is:
```
modified   .copier-answers.yml
@@ -2,5 +2,5 @@
 _commit: 0fa9e2f
 _src_path: ../copier_ignored_arg_bug
 project_name: X
-with_internal1: true
+with_internal1: 'false'
 ```

```
modified   f1.txt
@@ -1 +1 @@
-with_internal1: True
+with_internal1: false
```

### Logs

_No response_

### Expected behavior

The argument declared bool stays a bool

_No response_

### Operating system

Linux

Fedora

### Copier version

9.3.1

### Python version

3.11.8

### Installation method

pip+pypi

### Additional context

_No response_

sisp:
I've fixed a bug where a recopy/update with `-A, --skip-answered` did not parse a new answer passed via `-d, --data`. The solution was simple: Skipping a question because of `-A, --skip-answered` needs to happen _after_ parsing a new answer via `-d, --data` takes place, so that using the new answer takes precedence over skipping it.

Fixes #1728.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -480,9 +480,6 @@ class Worker:
                 else:
                     if question.validate_answer(answer):
                         del result.last[var_name]
-            # Skip a question when the user already answered it.
-            if self.skip_answered and var_name in result.last:
-                continue
             # Skip a question when the skip condition is met.
             if not question.get_when():
                 # Omit its answer from the answers file.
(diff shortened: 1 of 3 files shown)


yajo (approved):
Thanks !

PR #1750, opened 2024-08-25
Add the jinja2-slugify extension

offbyone:
While #370 addressed a subset of the requested extensions in #208, it did not in fact pull in cookiecutter, which means that the slugify extension that cookiecutter offers didn't make it.

Thus, this PR, which gets the tip version of [jinja2-slug](https://github.com/xsteadfastx/jinja2-slug) and adds it to the default extension set.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -547,6 +547,7 @@ class Worker:
         paths = [str(self.template.local_abspath)]
         loader = FileSystemLoader(paths)
         default_extensions = [
+            "jinja2_slug.SlugExtension",
             "jinja2_ansible_filters.AnsibleCoreFiltersExtension",
         ]
         extensions = default_extensions + list(self.template.jinja_extensions)
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -985,7 +985,6 @@ on them, so they are always installed when Copier is installed.
     ```yaml title="copier.yml"
     _jinja_extensions:
         - jinja_markdown.MarkdownExtension
-        - jinja2_slug.SlugExtension
         - jinja2_time.TimeExtension
     ```
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -475,6 +475,25 @@ PyYAML = "*"
 test = ["pytest", "pytest-cov"]
 
 [[package]]
+name = "jinja2-slug"
+version = "0.2.0"
+description = "Jinja2 Extension for creating slugs"
+optional = false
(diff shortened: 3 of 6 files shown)


pawamoy:
Thanks for the PR @offbyone. The point of the `_jinja_extensions` setting was that users could then load any extension they want, given they are available in the Copier environment. I don't think adding this particular slugify filter by default honors this initial idea, as template writers can easily specify it themselves in their template configuration.

By the way, having to list the dependency with a Git URL because the package on PyPI is outdated is not a good sign of maintenance, so I'd be reluctant to depend on it.

offbyone:
The reason I did it this way, instead of using the custom jinja extensions, is that documenting pipx injection for something as basic to templating as a slugify function felt like maybe solving the wrong problem. I'm fine if you prefer otherwise, but "converting a human name into programming language friendly names" feels like table stakes for a templating system; it honestly surprised me it wasn't built in in the first place! 

Anyway, no worries, I'll stick with the hacky solution I already have, just figured it might be nice to have something built in.

pawamoy:
Thank you, that makes sense :slightly_smiling_face: I wouldn't be against endorsing a third-party filter or hosting our own, but:

1. different projects have different needs (some projects will need more powerful slugify filters for example), so we won't be able to please everyone anyway and this will be subject to issues and other feature requests (more maintenance)
2. same and even more maintenance if we maintain our own filter

An alternative to documenting `pipx inject` and similar ways of installing the extensions is to publish a wrapper package that depends on both Copier and required extensions, and calls Copier transparently. But I agree it's not ideal either.

offbyone:
On thing that might sort it for me in a more durable way is if I could do something like cookiecutter's local extension code; is there a way to do that in copier?

offbyone:
Heh. I note that `pipx inject` or other custom installations still make an appearance there :D 

For my specific needs, those are dealbreakers, but that's okay; I'm able to work within the limitations I have here, so I'm okay to proceed with the status quo. I appreciate the responses either way.

yajo:
Thanks for your contribution!

I understand this would pave the way for people that come from cookiecutter. However, the more dependencies we include, the more maintenance burden we have.

So, I'll kindly close this PR because it adds no value to the Copier project itself. For your specific templates where you want specific extensions, just add some instructions in the README on how to install those extensions in a fashion that makes your template usable.

Thanks anyway!

PR #1770, opened 2024-09-10
fix: validate answers to secret questions

sisp:
I've fixed the answer validation to apply also to secret questions. This problem came up in an internal discussion with a colleague who stumbled over secret questions not being validated.

In general, I see no reason why answers to secret questions should not be validated. Secret questions are a special case regarding project updates though. We require a secret question to have a default value because we need _some_ value to exist for replaying a fresh project based on the old template using the recorded answers, which don't include secrets. For this to continue working, it's crucial that the default value of a secret question passes the validator – but this rule should actually apply to all questions because otherwise `copier copy --defaults ...` would be useless. Just saying.

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -387,7 +387,7 @@ class Question:
             result["multiline"] = self.get_multiline()
             if placeholder := self.get_placeholder():
                 result["placeholder"] = placeholder
-        if questionary_type in {"input", "checkbox"}:
+        if questionary_type in {"input", "checkbox", "password"}:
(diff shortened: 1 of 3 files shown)


yajo (approved):
Beautiful ! thank you 😊

mschoettle:
Thanks for this, I just came across this.

Question regarding defaults for secrets and project updates: If the default is a random password that satisfies the validator, it would then overwrite the previously supplied password. Is that correct?

sisp:
Correct.

Issue #1474, opened 2024-01-09
Using --data with a multiselect based choice


chrisdaish:
### Describe the problem

Follow on from: https://github.com/orgs/copier-org/discussions/1473

### Template

```
fruit:
    type: str
    help: "Fruit?"
    choices:
        - apple
        - pear
        - banana
    multiselect: true
```

### To Reproduce

```
$ copier copy --data 'fruit=apple' ../copier-issue .
...
ValueError: Invalid choice
```

### Expected behavior

Primary:
--data is able to find the 'apple' from the 'fruit' multiselect based choice, and select it.

Secondary:
To support multiple options perhaps multiple --data entries are required akin to:

```
--data 'fruit=apple' --data 'fruit=banana'
```

or comma separated

```
--data 'fruit=apple, banana'
```

or an array?

```
--data 'fruit=[apple, banana]'
```

### Operating system

macOS

### Operating system distribution and version

Sonoma 14.2.1

### Copier version

copier 9.1.0

### Python version

Python 3.9.6

### Installation method

pip+pypi

noirbizarre:
My 2 cts (as the author of the multiselect feature):

Given https://github.com/copier-org/copier/blob/master/copier/cli.py#L162-L178 is already a list, and current behavior is "last value overrides previous", I would avoid the multiple flags approach as it would mean different behavior and different parsing strategies.

Between `--data 'fruit=apple,banana'` and `--data 'fruit=[apple, banana]'`, I would chose the first one based on this:
- I think comma separated list are more common as parameter and potentially simpler to parse
- this is very personal, but I don't like CLI expecting JSON-like data: most of the time it's a lot of noise for nothing (accepting JSON but expecting only a subset which most of the time is easier to write in basic cli style), and often double escaping required
- if it's not JSON, it means it needs a custom parser and so more complexity and maintenance than splitting on commas

sisp:
I think comma-separated values won't work in general, e.g. `type: yaml` or `type: json` might contain commas which should not be used to split multiple answer values in a multiselect question. 🤔

yajo:
Have you actually tried the array way? Something like:

    copier copy --data 'fruit=[apple]' ../copier-issue .

sisp:
@yajo I've tried several ways, also with an array, none works. The problem is that the answer passed via CLI is a string, which gets parsed by

[...]

PR #1781, opened 2024-09-20
Document how to update when changing a multiselect base answer

kenden:
Document how to update when changing a multiselect base answer
The documentation to update, changing only one answer  when the question is a multi-select base choice, is missing.

This is from https://github.com/copier-org/copier/issues/1474#issuecomment-2328522026, but it also indicates to create the `.yaml` file in another directory (to avoid error `Destination repository is dirty; cannot continue. Please commit or stash your local changes and retry.`).

### DIFF
--- a/docs/updating.md
+++ b/docs/updating.md
@@ -129,29 +129,14 @@ to go through the whole questionnaire again:
 copier update --defaults --data updated_question="my new answer"
 ```
 
-Note: if that question is a multiselect base choice, e.g.:
-
+Note: it is not yet possible to update a multiselect base choice that way, for example a value that
+looks like this in the file `.copier-answers.yml`:
 ```yaml
 python_tested_versions:
     - "3.10"
     - "3.11"
 ```
-
-then you first need to create a file outside of the directory, e.g.
-`~/tmp/copier_answer.yaml`, with the content you want, e.g.:
-
-```yaml
-python_tested_versions:
-    - "3.10"
-    - "3.11"
-    - "3.12"
-```
-
-and then run the command with the `--data-file` argument:
-
-```shell
-copier update --defaults  --data-file=~/tmp/copier_answer.yaml
-```
+As a workaround, you need to use the `--data_file` option instead. See [here](https://copier.readthedocs.io/en/stable/configuring/#data_file) for usage information.
 
 ## How the update works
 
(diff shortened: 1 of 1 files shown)


yajo (changes requested):
Thank you! I'd be happier to merge the fix rather than documenting the workaround, but at least it's something.

would you please mention the issue and tell something that indicates the reader that this is just a workaround until it's fixed?

docs/updating.md
```diff
@@ -129,6 +129,24 @@ to go through the whole questionnaire again:
 copier update --defaults --data updated_question="my new answer"
 ```
 
+Note: if that question is a multiselect base choice, e.g.:
+  ```yaml
+  python_tested_versions:
+  - '3.10'
+  - '3.11'
+  ```
+then you first need to create a file outside of the directory, e.g. `~/tmp/copier_answer.yaml`, with the content you want, e.g.:
```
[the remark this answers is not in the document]
kenden: Thanks! I made a change to use this instead.

docs/updating.md
```diff
@@ -129,6 +129,15 @@ to go through the whole questionnaire again:
...
 ```
 
+Note: it is not yet possible ([ref. issue](https://github.com/copier-org/copier/issues/1474)) to update a multiselect base choice that way, for example a value that
+looks like this in the file `.copier-answers.yml`:
+```yaml
+python_tested_versions:
+    - "3.10"
+    - "3.11"
+```
+As a workaround, you need to use the `--data_file` option instead. See [here](https://copier.readthedocs.io/en/stable/configuring/#data_file) for usage information.
```
yajo, line 139:
```suggestion
As a workaround, for now you need to use [the `--data-file` option][data_file] instead.
```

docs/updating.md
```diff
@@ -129,6 +129,15 @@ to go through the whole questionnaire again:
 copier update --defaults --data updated_question="my new answer"
 ```
 
+Note: it is not yet possible ([ref. issue](https://github.com/copier-org/copier/issues/1474)) to update a multiselect base choice that way, for example a value that
```
yajo, line 132:
```suggestion
Note: it is not yet possible ([ref. issue](https://github.com/copier-org/copier/issues/1474)) to update a multiselect choice that way, for example a value that
```

yajo (changes requested):

yajo (approved):
I pushed by myself some changes to adapt to the global copywriting style, to avoid bothering you with the details. But the essence is the same. Also, after all, the ˋ--data-fileˋ option should be documented here, no matter if it is as a workaround or not.

Thank you!

PR #1796, opened 2024-10-02
fix(exclude): support negative exclude matching child of excluded parent

### LINKED ISSUE #1794: Negative exclude patterns like `--exclude '*' --exclude '!/path/to/include'` not working
### Describe the problem

According to our [documentation](https://copier.readthedocs.io/en/stable/configuring/#exclude), negative exclude patterns such as

```sh
copier copy --exclude '*' --exclude '!file-i-want' ./template ./destination
```

are supported. As it turns out, a wildcard exclude `--exclude '*'` followed by a negative exclude `--exclude '!/path/to/include'` only works when the latter refers to a top-level file/directory and nested files/directories remain excluded. That's because our [`Worker._render_path()`](https://github.com/copier-org/copier/blob/321aa0a82010b735aa0c8c592372344051d9f2db/copier/main.py#L708-L746) method traverses the filesystem tree top-down, and when a [path matches the exclude pattern](https://github.com/copier-org/copier/blob/321aa0a82010b735aa0c8c592372344051d9f2db/copier/main.py#L736-L738) (e.g., when `--exclude '*'`), the entire subtree is skipped even when one of the paths in the subtree is included via a negative exclude (e.g., `--exclude '!/path/to/include'`), as the negative exclude never gets matched.

### Template

[example.zip](https://github.com/user-attachments/files/17187490/example.zip)

### To Reproduce

1. Extract `example.zip` into a template directory, e.g. `/tmp/copier-exclude-issue/src/`, and change the current working directory to `/tmp/copier-exclude-issue/`.

2. Generate a project as follows:

    ```console
    $ copier copy --exclude '*' --exclude '!/include-me.txt' --exclude '!/dir/include-me.txt' src/ dst/

        create  .
        create  include-me.txt
    ```

3. Observe that `dst/include-me.txt` is generated but `dst/dir/include-me.txt` isn't.

### Logs

```console
n/a
```

### Expected behavior

The file `dst/dir/include-me.txt` should be generated because of the negative exclude `--exclude '!/dir/include-me.txt'`.

_No response_

### Operating system

Linux

Ubuntu 22.04

### Copier version

copier 9.3.1

### Python version

CPython 3.12

### Installation method

pipx+pypi

### Additional context

Original question asked in #1790.

sisp:
I've fixed the filtering of files and directories with exclude patterns. Before, a negative exclude pattern was ignored when some parent path matched the exclude patterns. The reason lies in the custom top-down filesystem walk implemented in `Worker._render_folder()` which didn't traverse the contents of a directory if an exclude pattern matched while a _negative_ exclude pattern also matched a child of this directory.

The solution is based on a different approach to walking the filesystem and excluding paths. Now, an iterator of all paths (files and directories) in the tree rooted at the template's copy root is produced via a recursive extension of `os.scandir()` and then filtered with the exclude patterns, so a subtree isn't pruned during the top-down tree walk. Instead, every path in the tree is considered and filtered individually based on the exclude patterns. Thus, `Worker._render_folder()` no longer implements a tree walk but simply creates a directory at the specified path. Also, `Worker._render_{folder,symlink,file}()` no longer render path templates, as paths need to be rendered prior to filtering with exclude patterns.

I decided to use `os.scandir()` over `os.walk()` or `Path.glob("**/*")` for better performance. It's also worth noting that, unlike before, the new implementation will walk subtrees of, e.g., `.git/`, even when they are excluded by the (default) exclude patterns, which may seem suboptimal. But I don't see a way to avoid it when supporting negative includes because pruning a subtree like before leads to incorrect results. That said, when using the [`subdirectory` setting](https://copier.readthedocs.io/en/stable/configuring/#subdirectory), the subtree of `.git/` is ignored because the root of the tree walk is the configured template subdirectory which doesn't contain the `.git/` directory.

Fixes #1794.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -53,6 +53,7 @@ from .tools import (
     escape_git_path,
     normalize_git_path,
     printf,
+    scantree,
     set_git_alternates,
 )
 from .types import (
@@ -599,23 +600,41 @@ class Worker:
             )
         )
 
-    def _render_file(self, src_abspath: Path) -> None:
+    def _render_template(self) -> None:
+        """Render the template in the subproject root."""
+        for src in scantree(str(self.template_copy_root)):
+            src_abspath = Path(src.path)
+            src_relpath = Path(src_abspath).relative_to(self.template.local_abspath)
+            dst_relpath = self._render_path(
+                Path(src_abspath).relative_to(self.template_copy_root)
+            )
+            if dst_relpath is None or self.match_exclude(dst_relpath):
+                continue
+            if src.is_symlink() and self.template.preserve_symlinks:
+                self._render_symlink(src_relpath, dst_relpath)
+            elif src.is_dir(follow_symlinks=False):
+                self._render_folder(dst_relpath)
+            else:
+                self._render_file(src_relpath, dst_relpath)
+
+    def _render_file(self, src_relpath: Path, dst_relpath: Path) -> None:
         """Render one file.
 
         Args:
-            src_abspath:
-                The absolute path to the file that will be rendered.
+            src_relpath:
+                File to be rendered. It must be a path relative to the template
+                root.
+            dst_relpath:
+                File to be created. It must be a path relative to the subproject
+                root.
         """
         # TODO Get from main.render_file()
-        assert src_abspath.is_absolute()
-        src_relpath = src_abspath.relative_to(self.template.local_abspath).as_posix()
-        src_renderpath = src_abspath.relative_to(self.template_copy_root)
-        dst_relpath = self._render_path(src_renderpath)
-        if dst_relpath is None:
-            return
-        if src_abspath.name.endswith(self.template.templates_suffix):
+        assert not src_relpath.is_absolute()
+        assert not dst_relpath.is_absolute()
+        src_abspath = self.template.local_abspath / src_relpath
+        if src_relpath.name.endswith(self.template.templates_suffix):
             try:
-                tpl = self.jinja_env.get_template(src_relpath)
+                tpl = self.jinja_env.get_template(src_relpath.as_posix())
             except UnicodeDecodeError:
                 if self.template.templates_suffix:
                     # suffix is not empty, re-raise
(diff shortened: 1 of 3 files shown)


sisp:
Tests for Python 3.8 are failing because `Path.readlink()` was added in Python 3.9: https://docs.python.org/3/library/pathlib.html#pathlib.Path.readlink I'll fix this later.

sisp:
Done. I think this PR is ready for review now.

copier/main.py
```diff
@@ -669,41 +690,30 @@ def _render_symlink(self, src_abspath: Path) -> None:
             return
 
         if not self.pretend:
+            dst_abspath = self.subproject.local_abspath / dst_relpath
             # symlink_to doesn't overwrite existing files, so delete it first
             if dst_abspath.is_symlink() or dst_abspath.exists():
                 dst_abspath.unlink()
+            dst_abspath.parent.mkdir(parents=True, exist_ok=True)
```
yajo, line 697: If you add this line also to `_render_file` maybe you can remove `_render_folder`?

sisp: No, i don't think so because we need to be able to render empty directories and `_render_{file,symlink}()` only get called for file paths but not directory paths.

yajo: Then this line can be removed from here?

sisp: No – although it seems like it at first glance – because there is an edge case that is covered by the new test `tests/test_exclude.py::test_exclude_wildcard_negate_nested_file`:

There, we have a symlink `foo/bar/keep-me-symlink.txt` that targets the file `foo/keep-me.txt`. At the same time, we exclude everything via `*` except those two files via negative excludes (which wasn't working prior to this PR). This means, the directories `foo/` and `foo/bar/` are excluded according to the exclude patterns, but `Worker._render_file()` is called for `foo/keep-me.txt` and `Worker._render_symlink()` is called for `foo/bar/keep-me-symlink.txt` without any prior call of `Worker._render_folder()` (for `foo/` or `foo/bar/`). Thus, it's important for `Worker._render_{file,symlink}()` to also create their parent directories even though it seems unnecessary.

sisp: > If you add this line also to `_render_file` maybe you can remove `_render_folder`?

By the way, `_render_file` already has this line:

https://github.com/copier-org/copier/blob/7209b3b35ea82672f3cf9d121ef95c551c7355a2/copier/main.py#L635

yajo (approved):

PR #1797, opened 2024-10-02
Run tests on python 3.13

jairhenrique:


sisp (commented):
Thanks for enabling CI for Python 3.13, @jairhenrique! :pray: Just one question: Why did you re-lock all dependencies? It seems the tests were passing without a re-lock as well. Updating the lock file should be a separate PR IMO.

jairhenrique:
@sisp I run `poetry lock` because tests breaks with `coverage==7.0.5`. After lock only one matrix fail with python 3.13 `build (windows-latest, 3.13)`.

You can see more details here:
- Windows fail https://github.com/jairhenrique/copier/actions/runs/11149852649
- Coverage fail https://github.com/jairhenrique/copier/actions/runs/11149733141

sisp:
I see. Would you mind only updating the locked version for `coverage` instead of refreshing all locks?

```sh
poetry update coverage
```

sisp:
It seems some other dependencies need to be updated as well.

jairhenrique:
> It seems some other dependencies need to be updated as well.

I found it, is a `nodeenv` dependency. Updated!
Now to pass all tests we need a new version of `pywin32`. The project already supports python 3.13, but a new version has not been generated. https://github.com/mhammond/pywin32/pull/2260

yajo:
seems like it's available already.
https://github.com/mhammond/pywin32/blob/842fbef58770d8c3a634b2e6033b95f80d19cbd3/CHANGES.txt#L32

could you update that dependency too please?

.github/workflows/ci.yml
```diff
@@ -48,6 +48,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          allow-prereleases: true
```
sisp, line 51: Python 3.13 has been released, so we don't need to enable pre-releases anymore.

pawamoy: This will be useful for 3.14 anyway, no? Not a huge deal, we can add it back later.

sisp:
It seems GitHub hasn't added Python 3.13 to their CI runners. :see_no_evil: We'll have to wait then ...

jairhenrique:
> It seems GitHub hasn't added Python 3.13 to their CI runners. 🙈 We'll have to wait then – or revert to `prereleases: true` .

I revert the commit

sisp:
I think CI on Windows with Python 3.13 is failing now because the locked `regex` version has no prebuilt wheel for Python 3.13 yet. Would you mind trying to update the lock file with `poetry update regex`?

sisp (approved):
LGTM! :tada: Thanks! :pray:

sisp:
@yajo Could you please configure the CI jobs for Python 3.13 to be required? It's something we can't configure via PR.

yajo:
Done, thanks

PR #1804, opened 2024-10-08
feat: add support for dynamic choices

### LINKED ISSUE #1746: Support dynamic choices
### Actual Situation

I'd like to ask for some choices (my case aws-regions) and in the second step I want that the user selects the primary region from the previously specified region. 

Sadly the choices field is not evaluated as expression

### Desired Situation

What happens right now:
the String is destructed as an array and I can chose a letter from my expression

What should happen:
The user should have a way how to dynamicly specify the choices 

### Proposed solution

I'd like that this works:

```yaml
list_input:
  type: str
  help: "Enter a comma-separated list of items."
  default: "item1,item2,item3"

selected_choice:
  type: str
  help: "Select one of the items from your list."
  choices: "{{ list_input.split(',') | map('trim') | list }}"
```

sisp:
I've added support for dynamic choices by allowing a templated string value for the `choices` key that renders as valid YAML. All choice styles are supported. See also https://github.com/orgs/copier-org/discussions/1193#discussioncomment-10040365.

Resolves #1746 #1193.

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -177,7 +177,7 @@ class Question:
     var_name: str
     answers: AnswersMap
     jinja_env: SandboxedEnvironment
-    choices: Sequence[Any] | dict[Any, Any] = field(default_factory=list)
+    choices: Sequence[Any] | dict[Any, Any] | str = field(default_factory=list)
     multiselect: bool = False
     default: Any = MISSING
     help: str = ""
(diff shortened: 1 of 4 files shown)


yajo (approved):
Beautiful.

sisp:
Codecov is hanging for some reason :thinking:.

yajo:
Ah, codecov... using superpowers then.

PR #1811, opened 2024-10-11
refactor: drop support for Python 3.8

sisp:
I've dropped support for Python 3.8, as it reached its [EOL on October 7, 2024](https://devguide.python.org/versions/#unsupported-versions). 

The cleanup to reduce technical debt involves two ~breaking~notable changes because we're now using features only available on Python 3.9+:

- Use `pathlib.Path.readlink` and remove our custom compatibility function
- Use `typing.Annotated` instead of `typing_extensions.Annotated`, and remove the runtime dependency on `typing-extensions`

@copier-org/maintainers ~Should we make another release prior to merging this PR? And should we consider the changes in this PR non-breaking because Python 3.8 has reached EOL and shouldn't be used anymore?~ On second thought, I think none of the changes should be considered breaking because (a) an EOL Python version shouldn't be used anymore and (b) any proper dependency resolver will not include a Copier release with `requires-python = ">=3.9"` for Python 3.8. I'll remove the breaking change marker and footer from the commit message.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     runs-on: ${{ matrix.os }}
     steps:
       # HACK https://github.com/actions/cache/issues/315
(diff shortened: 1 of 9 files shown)


sisp:
@yajo We also need an update of the repo settings to remove the Python 3.8 CI jobs (same as https://github.com/copier-org/copier/pull/1252#issuecomment-1640504776).

sisp:
This PR is now ready for a review, @copier-org/maintainers.

yajo (approved):

PR #1813, opened 2024-10-11
fix: Don't mark files without conflict markers as unmerged

pawamoy:
Closes #1763.

**Please squash when merging, do not rebase.**

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -1108,12 +1108,12 @@ class Worker:
                         # The 3-way merge might have resolved conflicts automatically,
                         # so we need to check if the file contains conflict markers
                         # before storing the file name for marking it as unmerged after the loop.
-                        # with open(fname) as conflicts_candidate:
-                        #     if any(
-                        #         line.rstrip() in ("<<<<<<< before updating", ">>>>>>> after updating")
-                        #         for line in conflicts_candidate
-                        #     ):
-                        #         conflicted.append(fname)
+                        with open(fname) as conflicts_candidate:
+                            if any(
+                                line.rstrip() in ("<<<<<<< before updating", ">>>>>>> after updating")
+                                for line in conflicts_candidate
+                            ):
+                                conflicted.append(fname)
                     # Forcefully mark files with conflict markers as unmerged,
                     # see SO post: https://stackoverflow.com/questions/77391627/
                     # and Git docs: https://git-scm.com/docs/git-update-index#_using_index_info.
--- a/copier/main.py
+++ b/copier/main.py
@@ -1110,7 +1110,7 @@ class Worker:
                         # before storing the file name for marking it as unmerged after the loop.
                         with open(fname) as conflicts_candidate:
                             if any(
-                                line.rstrip() in ("<<<<<<< before updating", ">>>>>>> after updating")
+                                line.rstrip() in {"<<<<<<< before updating", ">>>>>>> after updating"}
                                 for line in conflicts_candidate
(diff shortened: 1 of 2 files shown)


pawamoy:
Once again having trouble with Nix :sweat_smile: 

```
error: creating directory '/nix/store': Permission denied
error: creating directory '/nix/store': Permission denied
```

pawamoy:
OK now I just have to find out how to write a test for this :melting_face:

lkubb:
> OK now I just have to find out how to write a test for this 🫠

Doing the same change in the template's new version and a generated project before updating should be enough to trigger this bug I think.

lkubb:
@pawamoy It doesn't pass when I run it on latest master:

```
====================================================== FAILURES ======================================================
__________________________ test_3way_merged_files_without_conflicts_are_not_marked_unmerged __________________________
[...]
        with local.cwd(dst):
            lines = git("status", "--porcelain=v1").strip().splitlines()
>           assert not any(
                line.startswith("UU") and normalize_git_path(line[3:]) == filename
                for line in lines
            )
E           assert not True
E            +  where True = any(<generator object test_3way_merged_files_without_conflicts_are_not_marked_unmerged.<locals>.<genexpr> at 0x107543a00>)

tests/test_updatediff.py:91: AssertionError
```

pawamoy:
Thanks for testing! Weird, can't make it fail locally. Have commented the fix to see how the test behaves in CI.

lkubb:
You need to add back what the patch replaced when commenting it out, otherwise the test will pass:

```
conflicted.append(fname)
```

pawamoy:
OMG of course. Thanks :melting_face:

pawamoy:
Properly failing locally without the fix. Ready for review then! Thanks @lkubb, really appreciate the speedy help!

copier/main.py
```diff
@@ -1105,8 +1105,15 @@ def _apply_update(self) -> None:  # noqa: C901
...
                         # Remove rejection witness
                         Path(f"{fname}.rej").unlink()
-                        # Store file name for marking it as unmerged after the loop
-                        conflicted.append(fname)
+                        # The 3-way merge might have resolved conflicts automatically,
+                        # so we need to check if the file contains conflict markers
+                        # before storing the file name for marking it as unmerged after the loop.
+                        with open(fname) as conflicts_candidate:
+                            if any(
+                                line.rstrip() in ("<<<<<<< before updating", ">>>>>>> after updating")
```
pawamoy, line 1113: suggestion: We might want to use global variables for the `before updating` and `after updating` strings, in case we change them in the future.

pawamoy:
Codecov doesn't like me. Here it says -0.11% and when I click on their link it says +0.02%...

copier/main.py
yajo, line 1113: issue: a little nitpick. In general, using sets is better for `in` checks:

```suggestion
                                line.rstrip() in {"<<<<<<< before updating", ">>>>>>> after updating"}
```

yajo (approved):
Small nitpick, but the change looks quite nice. Thanks!

copier/main.py
pawamoy: True! Sometimes though I wonder if hashing each line is more or less performant than just comparing a few characters. Anyway, lets just follow the general rule and use a set :+1: :smile:

trallnag:
Just noticed this in action, works great. Thanks!

PR #1815, opened 2024-10-13
refactor: set default value for `keep_trailing_newline` more idiomatically

sisp:
Just a minor stylistic refactoring.

### DIFF
--- a/copier/template.py
+++ b/copier/template.py
@@ -309,11 +309,10 @@ class Template:
         See [envops][].
         """
         result = self.config_data.get("envops", {})
-        if "keep_trailing_newline" not in result:
-            # NOTE: we want to keep trailing newlines in templates as this is what a
-            #       user will most likely expects as a default.
-            #       See https://github.com/copier-org/copier/issues/464
-            result["keep_trailing_newline"] = True
+        # NOTE: we want to keep trailing newlines in templates as this is what a
+        #       user will most likely expects as a default.
+        #       See https://github.com/copier-org/copier/issues/464
+        result.setdefault("keep_trailing_newline", True)
         return result
 
     @cached_property


pawamoy (approved):
:+1:

yajo:
Just a little nitpicking: this commit should have been labeled as `style:`, to avoid displaying it in the changelog. No big deal, just so you know for next times. Thank you!

sisp:
Thanks for bringing this up! I'm still struggling with conventional commits in some cases like this. The way I understand guides like https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13, this should be a (very simple) refactoring, but I agree it's hardly worth an entry in the changelog. Isn't `style:` only for stylistic changes?

pawamoy:
I agree with @sisp's sentiment, sometimes it's not clear, but I exactly do like @yajo and eventually decide whether `refactor` and `chore`/`ci`/`style` depending on whether the change should appear in the changelog :relaxed:

yajo:
The ones that appear in the changelog are: feat, fix, refactor, perf. The rest won't appear. Ah, and those with the `BREAKING CHANGE:` trailer.

More details here: https://github.com/commitizen-tools/commitizen/blob/36cfcab746fd9a5c1bc495332fc3dcd11db967ae/commitizen/cz/conventional_commits/conventional_commits.py#L34-L40

So yes, the question is: should this appear in the changelog? If not, use [other marker](https://github.com/commitizen-tools/commitizen/blob/36cfcab746fd9a5c1bc495332fc3dcd11db967ae/commitizen/cz/conventional_commits/conventional_commits.py#L195).

sisp:
Okay, got it. 👍

PR #1831, opened 2024-10-23
build(deps): add version specifier for `types-psutil` dependency

sisp:
I've added an alternative version specifier for `types-psutil` which uses PEP440 and not Poetry syntax. The constraint is equivalent, but it also works around a Renovate bug that prevents Poetry lock file updates for dependencies constrained with `*`.

Renovate's implementation of Poetry's versioning scheme uses transformation to SemVer and NPM, as Poetry draws much inspiration from there, but PEP440 is a superset of SemVer, so Renovate's Poetry support has some limitations.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1741,4 +1741,4 @@ type = ["pytest-mypy"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.9"
-content-hash = "2faadc22481219780bbf25b968d5391829a1e2833281e7b0f6947e97d7697008"
+content-hash = "ff0c17bbced2ef047a991d8a2905dcbba444ddae6b0a245be25e1b7d05db52d3"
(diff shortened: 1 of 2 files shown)


pawamoy (approved):
Always good to better adhere to PEP 440 :+1:

pawamoy:
They designed their UX before it was standardized. They wanted the standard to reuse their dependency format, as they (well, let's be direct, Sébastien Eustace) believed it was best for both beginners and experienced users. It was decided otherwise, mainly because PEP 508 and PEP 440 are a pretty established thing, and it wasn't a good thing to start diverging from these standards. I think Sébastien was disappointed at first and camped his grounds for quite some time. Anyway, maintainers recently [merged a PR to support PEP 621](https://github.com/python-poetry/poetry/pull/9135) :slightly_smiling_face:

pawamoy:
Round 3 of the discussion, with links to round 1 and 2: https://discuss.python.org/t/pep-621-round-3/5472.

Issue #1837, opened 2024-10-28
Copier 9.4.x not working in debian:bullseye slim images


ap-wtioit:
### Describe the problem

git 2.30.2 (provided in debian bullseye) does not support `--path-format=absolute` for git rev-parse.

`docker run --rm debian:bullseye-slim bash -c 'set -x; apt update && apt install git -y; git --version; git init .; git -C . rev-parse --git-path objects; git -C . rev-parse --path-format=absolute --git-path objects'`
```
...
+ git --version
git version 2.30.2
+ git init .
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /.git/
+ git -C . rev-parse --git-path objects
.git/objects
+ git -C . rev-parse --path-format=absolute --git-path objects
--path-format=absolute
.git/objects
```
`docker run --rm ubuntu:24.04 bash -c 'set -x; apt update && apt install git -y; git --version; git init .; git -C . rev-parse --git-path objects; git -C . rev-parse --path-format=absolute --git-path objects'`
```
...
+ git --version
git version 2.43.0
+ git init .
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /.git/
+ git -C . rev-parse --git-path objects
.git/objects
+ git -C . rev-parse --path-format=absolute --git-path objects
/.git/objects
```

This leads to the following issues when running `copier update -f --UNSAFE`:
```
Traceback (most recent call last):
  File "/home/copier/.local/bin/copier", line 8, in <module>
    sys.exit(copier_app_run())
  File "/home/copier/.local/share/pipx/venvs/copier/lib/python3.9/site-packages/plumbum/cli/application.py", line 640, in run
[...]

PR #1838, opened 2024-10-28
fix: get_git_objects_dir compatible with git 2.30.2

ap-wtioit:
fixes https://github.com/copier-org/copier/issues/1837 by using pathlib logic to get an absolute path

info @wt-io-it


sisp (approved):
Excellent work, @ap-wtioit! :ok_hand: Thanks for catching and fixing this compatibility issue! :pray: LGTM! :tada:

Issue #1271, opened 2023-07-26
Generate file structures on dynamic loop


yajo:
Just: `{% yield single_var from looped_var %}{{ single_var }}{% endyield %}`

The implementation could be similar to https://github.com/orgs/copier-org/discussions/908#discussioncomment-6505536:

Does `single_var` exist in context?
- yes: do nothing special
- no: raise `YieldException("single_var", looped_var)`

OK, we got the exception. Now, are you rendering a filename?
- Yes: start loop. Inject `single_var` in the context and keep rendering the contents (of the dir/file, recursively).
- No: just let the exception raise and print a nice message "Using `yield` outside of file names isn't supported".

_Originally posted by @yajo in https://github.com/copier-org/copier/discussions/908#discussioncomment-6543747_

kj-9:
@yajo 
Hi, I've created this PR to implement the `yield` tag functionality. Would you kindly review and provide advice?

I'm new to Copier and don't have deep knowledge of the codebase yet, so I'm not entirely confident if this implementation is the best approach.

I'm really looking forward to having this feature available, and I hope this contribution can help move it forward. Any feedback would be greatly appreciated!

PR #1855, opened 2024-11-10
feat: add dynamic file structures in loop using yield-tag

kj-9:
What I did:
- Added a new Jinja extension in `copier/jinja_ext.py` to handle `yield` tags
  - This is a lightweight extension that detects `yield` tags and sets up the necessary context through `jinja_env`
- Updated `_render_path` to return multiple destination paths and their corresponding contexts when `yield` tags are present
  - Internally, delegated the recursive logic for generating paths and contexts from `yield` tags to the `_render_parts` function
- Added `extra_context` parameter to `_render_file` to enable injection of contexts generated from `yield` tags

### DIFF
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -116,6 +116,10 @@ class YieldTagInFileError(CopierError):
     """A yield tag is used in the file content, but it is not allowed."""
 
 
+class MultipleYieldTagsError(CopierError):
+    """Multiple yield tags are used in one path name, but it is not allowed."""
+
+
 # Warnings
 class CopierWarning(Warning):
     """Base class for all other Copier warnings."""
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -115,8 +115,6 @@ class UnsafeTemplateError(CopierError):
 class YieldTagInFileError(CopierError):
     """A yield tag is used in the file content, but it is not allowed."""
 
-    pass
-
 
 # Warnings
 class CopierWarning(Warning):
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -112,6 +112,12 @@ class UnsafeTemplateError(CopierError):
         )
 
 
+class YieldTagInFileError(CopierError):
+    """A yield tag is used in the file content, but it is not allowed."""
+
+    pass
+
+
 # Warnings
 class CopierWarning(Warning):
     """Base class for all other Copier warnings."""
--- a/copier/jinja_ext.py
+++ b/copier/jinja_ext.py
@@ -76,7 +76,7 @@ class YieldExtension(Extension):
         return nodes.CallBlock(
             self.call_method(
                 "_yield_support",
-                [yield_iterable, nodes.Const(yield_name.name)],
+                [nodes.Const(yield_name.name), yield_iterable],
             ),
             [],
             [],
@@ -85,7 +85,7 @@ class YieldExtension(Extension):
         )
 
     def _yield_support(
-        self, yield_iterable: Iterable[Any], yield_name: str, caller: Callable[[], str]
+        self, yield_name: str, yield_iterable: Iterable[Any], caller: Callable[[], str]
     ) -> str:
         """Support function for the yield tag.
 
--- a/copier/jinja_ext.py
+++ b/copier/jinja_ext.py
@@ -10,6 +10,8 @@ from jinja2.ext import Extension
 from jinja2.parser import Parser
 from jinja2.sandbox import SandboxedEnvironment
 
+from copier.errors import MultipleYieldTagsError
+
 
 class YieldEnvironment(SandboxedEnvironment):
     """Jinja2 environment with attributes from the YieldExtension.
@@ -65,6 +67,14 @@ class YieldExtension(Extension):
     def __init__(self, environment: YieldEnvironment):
         super().__init__(environment)
 
+    def preprocess(
+        self, source: str, _name: str | None, _filename: str | None = None
+    ) -> str:
+        """Preprocess hook to reset attributes before rendering."""
+        self.environment.yield_name = self.environment.yield_iterable = None
+
+        return source
+
     def parse(self, parser: Parser) -> nodes.Node:
         """Parse the `yield` tag."""
         lineno = next(parser.stream).lineno
@@ -93,6 +103,15 @@ class YieldExtension(Extension):
         Sets the `yield_name` and `yield_iterable` attributes in the environment then calls
         the provided caller function. If an UndefinedError is raised, it returns an empty string.
         """
+        if (
+            self.environment.yield_name is not None
+            or self.environment.yield_iterable is not None
+        ):
+            raise MultipleYieldTagsError(
+                "Attempted to parse the yield tag twice. Only one yield tag is allowed per path name.\n"
+                f'A yield tag with the name: "{self.environment.yield_name}" and iterable: "{self.environment.yield_iterable}" already exists.'
+            )
+
         self.environment.yield_name = yield_name
         self.environment.yield_iterable = yield_iterable
 
--- a/copier/jinja_ext.py
+++ b/copier/jinja_ext.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Callable, Sequence
+from typing import Any, Callable, Iterable
 
 from jinja2 import nodes
 from jinja2.exceptions import UndefinedError
@@ -12,48 +12,49 @@ from jinja2.sandbox import SandboxedEnvironment
 
 
 class YieldEnvironment(SandboxedEnvironment):
-    """Jinja2 environment with a `yield_context` attribute.
+    """Jinja2 environment with attributes from the YieldExtension.
 
     This is simple environment class that extends the SandboxedEnvironment
     for use with the YieldExtension, mainly for avoiding type errors.
 
     We use the SandboxedEnvironment because we want to minimize the risk of hidden malware
-    in the templates so we use the SandboxedEnvironment instead of the regular one.
-    Of course we still have the post-copy tasks to worry about, but at least
+    in the templates. Of course we still have the post-copy tasks to worry about, but at least
     they are more visible to the final user.
     """
 
-    yield_context: dict[str, Any]
+    yield_name: str | None
+    yield_iterable: Iterable[Any] | None
 
     def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
-        self.extend(yield_context=dict())
+        self.extend(yield_name=None, yield_iterable=None)
 
 
 class YieldExtension(Extension):
-    """`Jinja2 extension for the `yield` tag.
+    """Jinja2 extension for the `yield` tag.
 
-    If `yield` tag is used in a template, this extension sets the `yield_context` attribute to the
-    jinja environment. `yield_context` is a dictionary with the following keys:
-    - `single_var`: The name of the variable that will be yielded.
-    - `looped_var`: The variable that will be looped over.
+    If `yield` tag is used in a template, this extension sets following attribute to the
+    jinja environment:
+    - `yield_name`: The name of the variable that will be yielded.
+    - `yield_iterable`: The variable that will be looped over.
 
-    Note that this extension just sets the `yield_context` attribute but renders template
-    as usual. It is caller's responsibility to use the `yield_context` attribute in the
-    template to generate the desired output.
+    Note that this extension just sets the attributes but renders template as usual.
+    It is caller's responsibility to use the `yield_context` attribute in the template to
+    generate the desired output.
 
-    Example:
-        template: "{% yield single_var from looped_var %}"
-        context: {"looped_var": [1, 2, 3], "single_var": "item"}
+    !!! example
 
-        then,
+        ```python
         >>> from copier.jinja_ext import YieldEnvironment, YieldExtension
         >>> env = YieldEnvironment(extensions=[YieldExtension])
         >>> template = env.from_string("{% yield single_var from looped_var %}{{ single_var }}{% endyield %}")
         >>> template.render({"looped_var": [1, 2, 3]})
         ''
-        >>> env.yield_context
-        {'single_var': 'single_var', 'looped_var': [1, 2, 3]}
+        >>> env.yield_name
+        'single_var'
+        >>> env.yield_iterable
+        [1, 2, 3]
+        ```
     """
 
     tags = {"yield"}
@@ -67,15 +68,15 @@ class YieldExtension(Extension):
         """Parse the `yield` tag."""
         lineno = next(parser.stream).lineno
 
-        single_var: nodes.Name = parser.parse_assign_target(name_only=True)
+        yield_name: nodes.Name = parser.parse_assign_target(name_only=True)
         parser.stream.expect("name:from")
-        looped_var = parser.parse_expression()
+        yield_iterable = parser.parse_expression()
         body = parser.parse_statements(("name:endyield",), drop_needle=True)
 
         return nodes.CallBlock(
             self.call_method(
                 "_yield_support",
-                [looped_var, nodes.Const(single_var.name)],
+                [yield_iterable, nodes.Const(yield_name.name)],
             ),
             [],
             [],
@@ -84,19 +85,15 @@ class YieldExtension(Extension):
         )
 
     def _yield_support(
-        self, looped_var: Sequence[Any], single_var_name: str, caller: Callable[[], str]
+        self, yield_iterable: Iterable[Any], yield_name: str, caller: Callable[[], str]
     ) -> str:
         """Support function for the yield tag.
 
-        Sets the yield context in the environment with the given
-        looped variable and single variable name, then calls the provided caller
-        function. If an UndefinedError is raised, it returns an empty string.
-
+        Sets the `yield_name` and `yield_iterable` attributes in the environment then calls
+        the provided caller function. If an UndefinedError is raised, it returns an empty string.
         """
-        self.environment.yield_context = {
-            "single_var": single_var_name,
-            "looped_var": looped_var,
-        }
+        self.environment.yield_name = yield_name
+        self.environment.yield_iterable = yield_iterable
 
         try:
             res = caller()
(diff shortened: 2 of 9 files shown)


kj-9:
hopefully closes #1271

yajo:
I only reviewed tests in depth so far, and it seems the interface is almost exactly what I meant! Awesome! 

Quoting from https://github.com/copier-org/copier/issues/1271#issue-1821616806:

> are you rendering a filename?
> No: just let the exception raise and print a nice message "Using yield outside of file names isn't supported".

That seems to be the only missing part. At least I saw no test for that. Could you add this please? 

When done I'll review the implementation. This PR is big!

Thank you very much 😊

kj-9:
@yajo 

Thank you for your review!

I've added a new commit that implements the behavior you mentioned - throwing an exception when yield is used for rendering file contents. I've also added the corresponding test case.

Looking forward to your implementation review!

copier/errors.py
```diff
@@ -112,6 +112,12 @@ def __init__(self, features: Sequence[str]):
         )
 
 
+class YieldTagInFileError(CopierError):
+    """A yield tag is used in the file content, but it is not allowed."""
+
+    pass
```
pawamoy, line 118: `pass` is unnecessary if there's a docstring :slightly_smiling_face: 

```suggestion
```

kj-9: thanks! I fixed in [6f4acec](https://github.com/copier-org/copier/pull/1855/commits/6f4acecccb67d0f2e896d81ec0580eb499651dbb) .

copier/jinja_ext.py
```diff
@@ -0,0 +1,109 @@
...
+    This is simple environment class that extends the SandboxedEnvironment
+    for use with the YieldExtension, mainly for avoiding type errors.
+
+    We use the SandboxedEnvironment because we want to minimize the risk of hidden malware
+    in the templates so we use the SandboxedEnvironment instead of the regular one.
+    Of course we still have the post-copy tasks to worry about, but at least
+    they are more visible to the final user.
+    """
+
+    yield_context: dict[str, Any]
```
yajo, line 26: I think that dicts are better suited for variable-length collections, but we know this one will always have two items, so:

```suggestion
    yield_name: str
    yield_iterable: Iterable[Any]
```

This change, of course, implies more changes across the rest of the file, including the docstrings.

copier/jinja_ext.py
```diff
@@ -0,0 +1,109 @@
...
+    If `yield` tag is used in a template, this extension sets the `yield_context` attribute to the
+    jinja environment. `yield_context` is a dictionary with the following keys:
+    - `single_var`: The name of the variable that will be yielded.
+    - `looped_var`: The variable that will be looped over.
+
+    Note that this extension just sets the `yield_context` attribute but renders template
+    as usual. It is caller's responsibility to use the `yield_context` attribute in the
+    template to generate the desired output.
+
+    Example:
```
yajo, line 45:
```suggestion
    !!! example
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,109 @@
...
+        template: "{% yield single_var from looped_var %}"
```
yajo, line 46: The example in the code is self-explanatory. We can remove these lines.
```suggestion
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,109 @@
...
+        context: {"looped_var": [1, 2, 3], "single_var": "item"}
```
yajo, line 47:
```suggestion
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,109 @@
...
+
+        then,
```
yajo, line 49:
```suggestion
        ```python
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,109 @@
...
+        >>> from copier.jinja_ext import YieldEnvironment, YieldExtension
+        >>> env = YieldEnvironment(extensions=[YieldExtension])
+        >>> template = env.from_string("{% yield single_var from looped_var %}{{ single_var }}{% endyield %}")
+        >>> template.render({"looped_var": [1, 2, 3]})
+        ''
+        >>> env.yield_context
```
yajo, line 55: Following the suggestion above:

```suggestion
        >>> env.yield_name, env.yield_iterable
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,109 @@
...
+        {'single_var': 'single_var', 'looped_var': [1, 2, 3]}
```
yajo, line 56:
```suggestion
        ('single_var', [1, 2, 3])
        ```
```

copier/main.py
```diff
@@ -606,19 +604,29 @@ def _render_template(self) -> None:
...
+                if self.match_exclude(dst_relpath):
+                    continue
+                if src.is_symlink() and self.template.preserve_symlinks:
+                    self._render_symlink(src_relpath, dst_relpath)
+                elif src.is_dir(follow_symlinks=follow_symlinks):
+                    self._render_folder(dst_relpath)
+                else:
+                    self._render_file(src_relpath, dst_relpath, extra_context=ctx or {})
+                    if self.jinja_env.yield_context:
+                        raise YieldTagInFileError(
```
yajo, line 620: I think this raise makes more sense inside `self._render_file` before the render.

copier/main.py
```diff
@@ -715,8 +727,89 @@ def _render_folder(self, dst_relpath: Path) -> None:
             dst_abspath = self.subproject.local_abspath / dst_relpath
             dst_abspath.mkdir(parents=True, exist_ok=True)
 
-    def _render_path(self, relpath: Path) -> Path | None:
-        """Render one relative path.
+    def _adjust_rendered_part(self, rendered_part: str) -> str:
+        """Adjust the rendered part if necessary.
+
+        If {{ _copier_conf.answers_file }} becomes the full path,
```
yajo, line 733:
```suggestion
        If `{{ _copier_conf.answers_file }}` becomes the full path,
```

yajo (changes requested):
Finally, the first code review is here!

Apart from what's mentioned, I also miss the documentation.

Thank you!

kj-9:
@yajo 
Thank you for the quick review!

I have added a few commits:
- Fixes according to your comments
- Minor code refactoring

And in the last commit, I added documentation. Specifically, three points:

1. Added a row for the new feature in the comparison table (as far as I know, only copier supports this feature)
2. Added `jinja_ext.py` to the `Reference`
3. Added a section explaining the new feature to `Configuring`.

This is my first time adding documentation, so I'm not sure if it meets your expectations.
Could you please review it again?

docs/configuring.md
```diff
@@ -560,6 +560,78 @@ your_project
 You can either use any separator, like `.`, and replace it with `_copier_conf.sep`, like
 in the example above, or just use `/` in the answer (works on Windows too).
 
+## Loop over lists to generate files and directories
```
yajo, line 563: issue: please move all these docs you're adding to `creating.md`.

docs/configuring.md
```diff
@@ -1748,3 +1820,7 @@ copier update -a .copier-answers.main.yml
 copier update -a .copier-answers.pre-commit.yml
 copier update -a .copier-answers.ci.yml
 ```
+
+```
+
+```
```
yajo, line 1826: issue: delete this, it seems garbage left by mistake.

yajo (changes requested):
Implementation looks very good, thanks!

Please review a couple of suggestions for the docs, so we can merge.

@copier-org/maintainers this is an important feature, would you like to add your reviews?

sisp:
I'll try to give feedback later today.

copier/jinja_ext.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self.extend(yield_name=None, yield_iterable=None)
+
+
+class YieldExtension(Extension):
+    """Jinja2 extension for the `yield` tag.
+
+    If `yield` tag is used in a template, this extension sets following attribute to the
+    jinja environment:
```
pawamoy, line 37: API docs being rendered through MkDocs, which uses Python-Markdown, bullet lists require a blank line before them.

```suggestion
    jinja environment:

```

copier/jinja_ext.py
```diff
@@ -0,0 +1,106 @@
...
+    - `yield_name`: The name of the variable that will be yielded.
+    - `yield_iterable`: The variable that will be looped over.
+
+    Note that this extension just sets the attributes but renders template as usual.
```
pawamoy, line 41:
```suggestion
    Note that this extension just sets the attributes but renders templates as usual.
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,106 @@
...
+    It is caller's responsibility to use the `yield_context` attribute in the template to
```
pawamoy, line 42:
```suggestion
    It is the caller's responsibility to use the `yield_context` attribute in the template to
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,106 @@
...
+    generate the desired output.
+
+    !!! example
+
+        ```python
```
pawamoy, line 47: There's a dedicated language for this: `pycon` (for Python Console).

```suggestion
        ```pycon
```

copier/jinja_ext.py
```diff
@@ -0,0 +1,106 @@
...
+from typing import Any, Callable, Iterable
+
+from jinja2 import nodes
+from jinja2.exceptions import UndefinedError
+from jinja2.ext import Extension
+from jinja2.parser import Parser
+from jinja2.sandbox import SandboxedEnvironment
+
+
+class YieldEnvironment(SandboxedEnvironment):
```
pawamoy, line 14: I wonder if it shouldn't be called `CopierEnvironment`, since it's a custom environment for Copier. Would prevent renaming it later if it is customized further. No strong opinion.

copier/main.py
```diff
@@ -715,8 +728,87 @@ def _render_folder(self, dst_relpath: Path) -> None:
...
+
+        if not extra_context:
+            extra_context = {}
+
+        rendered_part = self._render_string(part, extra_context=extra_context)
+
+        yield_name = self.jinja_env.yield_name
+        if yield_name:
+            for value in self.jinja_env.yield_iterable or ():
+                new_context = {**extra_context, **{yield_name: value}}
```
pawamoy, line 783: The dict creation + unpacking seems unecessary, though I might be missing something.

```suggestion
                new_context = {**extra_context, yield_name: value}
```

copier/main.py
```diff
@@ -715,8 +728,87 @@ def _render_folder(self, dst_relpath: Path) -> None:
...
+
+            return
+
+        part = parts[0]
+        parts = parts[1:]
+
+        if not extra_context:
+            extra_context = {}
+
+        rendered_part = self._render_string(part, extra_context=extra_context)
```
pawamoy, line 778: Looks like this could be moved after the `if yield_name` block.

copier/main.py
```diff
@@ -1114,7 +1188,8 @@ def _apply_update(self) -> None:  # noqa: C901
                         # before storing the file name for marking it as unmerged after the loop.
                         with open(fname) as conflicts_candidate:
                             if any(
-                                line.rstrip() in {"<<<<<<< before updating", ">>>>>>> after updating"}
+                                line.rstrip()
+                                in {"<<<<<<< before updating", ">>>>>>> after updating"}
```
pawamoy, line 1192: There are a few other style changes like this and I wonder if this is because of a discrepancy again between Copier's config and the contributor's editor/IDE configuration.

pawamoy (commented):
A few suggestions, nothing critical, looking good :slightly_smiling_face:

sisp (changes requested):
Thanks for working on this long-awaited feature, @kj-9! :pray: I have a few conceptual remarks. WDYT, @kj-9 and @copier-org/maintainers?

copier/jinja_ext.py
```diff
@@ -0,0 +1,106 @@
...
+    in the templates. Of course we still have the post-copy tasks to worry about, but at least
+    they are more visible to the final user.
+    """
+
+    yield_name: str | None
+    yield_iterable: Iterable[Any] | None
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self.extend(yield_name=None, yield_iterable=None)
```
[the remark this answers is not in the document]
yajo: > In addition, in light of [factoring out Copier's Jinja extensions](https://github.com/copier-org/copier/pull/1809#issuecomment-2408823031), this extension might get factored out of Copier

I don't think this particular extension has any use outside of copier, so it can be an exception.

docs/configuring.md
kj-9: deleted in https://github.com/copier-org/copier/pull/1855/commits/ac042c5d961790645ba3b0e627c9a6283e64f140

copier/jinja_ext.py
kj-9: done in https://github.com/copier-org/copier/pull/1855/commits/ac042c5d961790645ba3b0e627c9a6283e64f140

copier/jinja_ext.py
kj-9: done in https://github.com/copier-org/copier/pull/1855/commits/ac042c5d961790645ba3b0e627c9a6283e64f140

copier/jinja_ext.py
kj-9: done in https://github.com/copier-org/copier/pull/1855/commits/ac042c5d961790645ba3b0e627c9a6283e64f140

copier/jinja_ext.py
kj-9: done in https://github.com/copier-org/copier/pull/1855/commits/ac042c5d961790645ba3b0e627c9a6283e64f140

copier/main.py
kj-9: done in https://github.com/copier-org/copier/pull/1855/commits/ac042c5d961790645ba3b0e627c9a6283e64f140

copier/main.py
kj-9: This diff was displayed even though it had already been merged into the master branch (https://github.com/copier-org/copier/commit/fd23923b5cc903131fa828f4a89b4f457d1750e6). Please check again after merging the master branch and adding a new commit, as the diff has already disappeared.

copier/jinja_ext.py
kj-9: this could be affected by https://github.com/copier-org/copier/pull/1855#discussion_r1899536004
so I leave this for now.

copier/jinja_ext.py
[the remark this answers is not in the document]
kj-9: @sisp 
Thanks for your comment.

Your opinion sounds good. I do not fully understand the context and impact of factoring the Jinja extension. I feel it is a matter of style or philosophy which can be decided by the core maintainers, but the implementation seems straightforward. (so I'm happy to work on in if I need to👍)

but IMHO, as @yajo commented, if we are not going to extract this extension, there doesn't seem to be a significant difference between the current implementation and your suggestion. The subclass is quite small and doesn't seem to be a significant cost.

copier/jinja_ext.py
```diff
@@ -0,0 +1,106 @@
...
+    def _yield_support(
+        self, yield_name: str, yield_iterable: Iterable[Any], caller: Callable[[], str]
+    ) -> str:
+        """Support function for the yield tag.
+
+        Sets the `yield_name` and `yield_iterable` attributes in the environment then calls
+        the provided caller function. If an UndefinedError is raised, it returns an empty string.
+        """
+        self.environment.yield_name = yield_name
+        self.environment.yield_iterable = yield_iterable
```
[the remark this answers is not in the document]
kj-9: You are right. I will add treatment for the case and add some test cases as well.

copier/main.py
pawamoy: Oh, side-effect, OK :+1:

copier/main.py
pawamoy: Perfect, thank you! :ok_hand:

copier/jinja_ext.py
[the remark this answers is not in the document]
yajo: Yes, the yield tag is only to be used in file names, and not within templates, and that's part of the design.

For normal Jinja templates, where all data is within the rendering context, the yield tag doesn't make sense. There's the for tag for that.

So I don't think this extension will be outsourced, and if it were in the future we can always do the extraction later when there is some worldwide need for this extension.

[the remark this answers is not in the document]
sisp: Agreed, let's keep the environment subclass. :+1:

copier/jinja_ext.py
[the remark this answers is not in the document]
kj-9: @sisp  @copier-org/maintainers
Added 810f448 to treat multiple yield tags:
- Added an error to be raised when multiple yield tags are detected.
- Added test cases for the error.

Additionally, I found a useful hook called [preprocess](https://jinja.palletsprojects.com/en/stable/extensions/#jinja2.ext.Extension.preprocess), which called every time rendering starts. modified to use this hook to initializes the yield tag variables to `None`.

Please review again🙏
It seems that all other comments have been closed, so this might be the last commit.👍

copier/jinja_ext.py
```diff
@@ -0,0 +1,126 @@
...
+    - `yield_name`: The name of the variable that will be yielded.
+    - `yield_iterable`: The variable that will be looped over.
+
+    Note that this extension just sets the attributes but renders templates as usual.
+    It is the caller's responsibility to use the `yield_context` attribute in the template to
+    generate the desired output.
+
+    !!! example
+
+        ```pycon
```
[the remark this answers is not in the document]
yajo: I didn't know about it, sorry. It's good then.

yajo (approved):

lbenka:
Hi, any idea when this is gonna be released in the main version, seems like it's only on master.

pawamoy:
Copier doesn't have a schedule for releases, new releases are made when yajo finds the time to do so, and believes the time is right, mostly :shrug: A better question would be "Are there any blockers I could help with before a new release is published?" :pray: :blush:

yajo:
Sorry for the delay, I've been so busy lately...

I'll release when I have time, but you can install from master too in the mean time

PR #1878, opened 2024-12-03
build(vscode-extensions): update recommendations

yajo:
Using builtin prettier and ruff extensions is of no help because we use devenv.

Instead, pre-commit and nix-ide extensions integrate perfectly out of the box.

Thus, updating recommendations.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,9 +1,9 @@
 {
   "recommendations": [
-    "charliermarsh.ruff",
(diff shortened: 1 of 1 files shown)


sisp (approved):
I still don't have a working Nix environment 🫣, but those extensions make sense to me and I trust you've verified they are working. The pre-commit extension looks interesting even without Nix, I'll need to try it out. 🤩

PR #1925, opened 2025-01-08
refactor: Separate part of run_copy logic into a new generate method

blagoySimandov:
The `run_copy` method contained logic for both asking the questions and running the project generation. This made it impossible to do something between the two actions (after asking the questions and before running generation). Dividing the generation into it's own separate method allows for this.


sisp:
I'm not convinced by the benefit of this change. Could you elaborate on the motivation?

PR #1940, opened 2025-01-20
feat(settings): add user settings support with `defaults` values and trusts

noirbizarre:
Hello 👋🏼 

This PR adds support for user settings with initial support for:
- user defaults (#235)
- trusted repositories

Settings are stored in `<CONFIG_ROOT>/settings.yml` where `<CONFIG_ROOT>` is the
standard configuration directory for your platform:

- `$XDG_CONFIG_HOME/copier` (`~/.config/copier ` in most cases) on Linux as defined by
    [XDG Base Directory Specifications](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
- `~/Library/Application Support/copier` on macOS as defined by
    [Apple File System Basics](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html)
- `%USERPROFILE%\AppData\Local\copier` on Windows as defined in
    [Known folders](https://docs.microsoft.com/en-us/windows/win32/shell/known-folders)

This location can be overridden by setting the `COPIER_SETTINGS` environment variable.

The settings look like:
```yaml
defaults:
    user_name: "John Doe"
    user_email: john.doe@acme.com
trust:
    - https://github.com/account/template.git
    - https://github.com/org/
    - ~/templates/
```

### DIFF
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -139,3 +139,7 @@ class DirtyLocalWarning(UserWarning, CopierWarning):
 
 class ShallowCloneWarning(UserWarning, CopierWarning):
     """The template repository is a shallow clone."""
+
+
+class MissingSettingsWarning(UserWarning, CopierWarning):
+    """Settings path has been defined but file is missing."""
--- a/copier/main.py
+++ b/copier/main.py
@@ -46,6 +46,7 @@ from .errors import (
     YieldTagInFileError,
 )
 from .jinja_ext import YieldEnvironment, YieldExtension
+from .settings import Settings
 from .subproject import Subproject
 from .template import Task, Template
 from .tools import (
@@ -58,13 +59,7 @@ from .tools import (
     scantree,
     set_git_alternates,
 )
-from .types import (
-    MISSING,
-    AnyByStrDict,
-    JSONSerializable,
-    RelativePath,
-    StrOrPath,
-)
+from .types import MISSING, AnyByStrDict, JSONSerializable, RelativePath, StrOrPath
 from .user_data import DEFAULT_DATA, AnswersMap, Question
 from .vcs import get_git
 
@@ -192,6 +187,7 @@ class Worker:
     answers_file: RelativePath | None = None
     vcs_ref: str | None = None
     data: AnyByStrDict = field(default_factory=dict)
+    settings: Settings = field(default_factory=Settings.from_file)
     exclude: Sequence[str] = ()
     use_prereleases: bool = False
     skip_if_exists: Sequence[str] = ()
@@ -467,6 +463,7 @@ class Worker:
             question = Question(
                 answers=result,
                 jinja_env=self.jinja_env,
+                settings=self.settings,
                 var_name=var_name,
                 **details,
             )
@@ -998,11 +995,14 @@ class Worker:
         )
         subproject_subdir = self.subproject.local_abspath.relative_to(subproject_top)
 
-        with TemporaryDirectory(
-            prefix=f"{__name__}.old_copy.",
-        ) as old_copy, TemporaryDirectory(
-            prefix=f"{__name__}.new_copy.",
-        ) as new_copy:
+        with (
+            TemporaryDirectory(
+                prefix=f"{__name__}.old_copy.",
+            ) as old_copy,
+            TemporaryDirectory(
+                prefix=f"{__name__}.new_copy.",
+            ) as new_copy,
+        ):
             # Copy old template into a temporary destination
             with replace(
                 self,
--- a/copier/main.py
+++ b/copier/main.py
@@ -241,7 +241,7 @@ class Worker:
 
     def _check_unsafe(self, mode: Literal["copy", "update"]) -> None:
         """Check whether a template uses unsafe features."""
-        if self.unsafe:
+        if self.unsafe or self.settings.is_trusted(self.template.url):
             return
         features: set[str] = set()
         if self.template.jinja_extensions:
--- a/copier/settings.py
+++ b/copier/settings.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import os
 import warnings
+from os.path import expanduser
 from pathlib import Path
 from typing import Any
 
@@ -19,7 +20,12 @@ ENV_VAR = "COPIER_SETTINGS_PATH"
 class Settings(BaseModel):
     """User settings model."""
 
-    defaults: dict[str, Any] = Field(default_factory=dict)
+    defaults: dict[str, Any] = Field(
+        default_factory=dict, description="Default values for questions"
+    )
+    trust: set[str] = Field(
+        default_factory=set, description="List of trusted repositories or prefixes"
+    )
 
     @classmethod
     def from_file(cls, settings_path: Path | None = None) -> Settings:
@@ -38,3 +44,18 @@ class Settings(BaseModel):
                 f"Settings file not found at {env_path}", MissingSettingsWarning
             )
         return cls()
+
+    def is_trusted(self, repository: str) -> bool:
+        """Check if a repository is trusted."""
+        return any(
+            repository.startswith(self.normalize(trusted))
+            if trusted.endswith("/")
+            else repository == self.normalize(trusted)
+            for trusted in self.trust
+        )
+
+    def normalize(self, url: str) -> str:
+        """Normalize an URL using user settings."""
+        if url.startswith("~"):  # Only expand on str to avoid messing with URLs
+            url = expanduser(url)
+        return url
--- a/copier/settings.py
+++ b/copier/settings.py
@@ -1,0 +1,40 @@
+"""User settings models and helper functions."""
+
+from __future__ import annotations
+
+import os
+import warnings
+from pathlib import Path
+from typing import Any
+
+import yaml
+from platformdirs import user_config_path
+from pydantic import BaseModel, Field
+
+from .errors import MissingSettingsWarning
+
+ENV_VAR = "COPIER_SETTINGS_PATH"
+
+
+class Settings(BaseModel):
+    """User settings model."""
+
+    defaults: dict[str, Any] = Field(default_factory=dict)
+
+    @classmethod
+    def from_file(cls, settings_path: Path | None = None) -> Settings:
+        """Load settings from a file."""
+        env_path = os.getenv(ENV_VAR)
+        if settings_path is None:
+            if env_path:
+                settings_path = Path(env_path)
+            else:
+                settings_path = user_config_path("copier") / "settings.yml"
+        if settings_path.is_file():
+            data = yaml.safe_load(settings_path.read_text())
+            return cls.model_validate(data)
+        elif env_path:
+            warnings.warn(
+                f"Settings file not found at {env_path}", MissingSettingsWarning
+            )
+        return cls()
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -1,0 +1,39 @@
+# Settings
+
+Copier settings are stored in `<CONFIG_ROOT>/settings.yml` where `<CONFIG_ROOT>` is the
+standard configuration directory for your platform:
+
+-   `$XDG_CONFIG_HOME/copier` (`~/.config/copier ` in most cases) on Linux as defined by
+    [XDG Base Directory Specifications](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+-   `~/Library/Application Support/copier` on macOS as defined by
+    [Apple File System Basics](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html)
+-   `%USERPROFILE%\AppData\Local\copier` on Windows as defined in
+    [Known folders](https://docs.microsoft.com/en-us/windows/win32/shell/known-folders)
+
+This location can be overridden by setting the `COPIER_SETTINGS_PATH` environment
+variable.
+
+## User defaults
+
+Users may define some reusable default variables in the `defaults` section of the
+configuration file.
+
+```yaml title="<CONFIG_ROOT>/settings.yml"
+defaults:
+    user_name: "John Doe"
+    user_email: john.doe@acme.com
+```
+
+This user data will replace the default value of fields of the same name.
+
+### Well-known variables
+
+To ensure templates efficiently reuse user-defined variables, we invite template authors
+to use the following well-known variables:
+
+| Variable name | Type  | Description            |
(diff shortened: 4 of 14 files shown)


docs/settings.md
```diff
@@ -0,0 +1,55 @@
...
+stdandrd configuration directory for your platform:
+
+-   `$XDG_CONFIG_HOME/copier` (`~/.config/copier ` in most cases) on Linux as defined by
+    [XDG Base Directory Specifications](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+-   `~/Library/Application Support/copier` on macOS as defined by
+    [Apple File System Basics](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html)
+-   `%USERPROFILE%\AppData\Local\copier` on Windows as defined in
+    [Known folders](https://docs.microsoft.com/en-us/windows/win32/shell/known-folders)
+
+This location can be overridden by setting the `COPIER_SETTINGS` environment variable.
```
sisp, line 13: How about renaming the environment variable to `COPIER_SETTINGS_FILE` to be more specific that its value is the path of the settings file?

docs/settings.md
```diff
@@ -0,0 +1,55 @@
...
+    user_name: "John Doe"
+    user_email: john.doe@acme.com
+```
+
+Those user data will replace the default value of fieds of the same name.
+
+### Well known variables
+
+To ensure templates efficiently reuse user defined variables, we invite template authors
+to use the following well known variables:
```
sisp, line 31: I think it should be:

```suggestion
### Well-known variables

To ensure templates efficiently reuse user-defined variables, we invite template authors
to use the following well-known variables:
```

:nerd_face:

docs/settings.md
```diff
@@ -0,0 +1,55 @@
...
+
+| variable name | type  | description            |
```
sisp, line 33: I think it's nicer to capitalize the table header:

```suggestion
| Variable name | Type  | Description            |
```

pawamoy (changes requested):
Very nice! I like the prefix matching approach for trusted repositories, it's simple and it makes sense.

tests/test_config.py
```diff
@@ -91,6 +92,61 @@ def test_read_data(
     )
 
 
+def test_settings_defaults_precedense(
```
sisp, line 95:
```suggestion
def test_settings_defaults_precedence(
```

:nerd_face:

tests/test_settings.py
```diff
@@ -0,0 +1,95 @@
...
+    file_path = tmp_path / "file.yml"
+    file_path.write_text("defaults:\n  from: file")
+
+    settings = Settings.from_file(file_path)
+
+    assert settings.defaults == {"from": "file"}
+
+
+@pytest.mark.parametrize(
+    "repository,trust,is_trusted",
```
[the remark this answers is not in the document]
sisp: @pawamoy Perhaps we should enable [Ruff's `PT`](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt) rules. :thinking:

copier/settings.py
```diff
@@ -0,0 +1,53 @@
...
+
+ENV_VAR = "COPIER_SETTINGS"
+
+
+class Settings(BaseModel):
+    """User settings model."""
+
+    defaults: dict[str, Any] = Field(
+        default_factory=dict, description="Default values for questions"
+    )
```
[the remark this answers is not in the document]
noirbizarre: Yes, well-known variable ia meant for that, and you can override the `COPIER_SETTINGS` env var (allowing things like having different settings for pro/perso, or by clients...)

copier/settings.py
```diff
@@ -0,0 +1,53 @@
...
+        """Load settings from a file."""
+        if settings_path is None:
+            if env_path := os.getenv(ENV_VAR):
+                settings_path = Path(env_path)
+            else:
+                settings_path = user_config_path("copier") / "settings.yml"
+        if settings_path.is_file():
+            data = yaml.safe_load(settings_path.read_text())
+            return cls.model_validate(data)
+        return cls()
```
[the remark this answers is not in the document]
noirbizarre: Agreed, it needs a warning, and I would say it also needs error handling in case the file is not valid (Pydantic raw except might not be very user friendly)

copier/main.py
```diff
@@ -38,6 +38,8 @@
 from pydantic_core import to_jsonable_python
 from questionary import unsafe_prompt
 
+from copier.settings import Settings
```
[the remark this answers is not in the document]
noirbizarre: Great, I prefer, I'll change that

sisp (changes requested):
Fantastic work, @noirbizarre! :tada: Thanks for working on and contributing these valuable features! :pray:

I have a few minor remarks. And would you mind fixing the formatting errors in the corresponding commit and drop the auto-fix commit? We'll likely want to rebase-merge this PR to keep the two feature commits, so the auto-fix commit would only add some noise to the Git history.

copier/settings.py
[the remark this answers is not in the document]
pawamoy: Right, I didn't think about using the env var to deal with per-template defaults. Not very ergonomic though :sweat_smile:

docs/settings.md
pawamoy: I'll also note that the wrapping pipes (left-most and right-most ones) are not mandatory :slightly_smiling_face:

noirbizarre:
You're welcome ! 🙏🏼 
I take the suggestion commit, and I'll rebase properly everything, that was plan from the start 😉

copier/settings.py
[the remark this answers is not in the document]
noirbizarre: Actually, to me, it's the most efficient: for each job I have a workspace with some specific env vars for every setting file, git config, ...
This way it's set once and for all, and I never have to worry about adding specific parameters to each app call. Plus, it allows proper files separation, which is required sometimes (maybe not for copier) for security or privacy reasons.
However, I can understand it doesn't fit everyone and that it is my use case.
I am totally open to improvements, but I think we can iterate on that.
Are you OK to merge this, try it and iterate on feedback? I am totally ok to work on an alternative layout in another PR while keeping this one

docs/settings.md
noirbizarre: I hesitated with `COPIER_SETTINGS_PATH`, would this be OK?

tests/test_config.py
```diff
@@ -460,6 +517,53 @@ def test_user_defaults(
             },
             {},
             {},
+            {},
+            dedent(
+                """\
+                A string: foo
+                """
+            ),
```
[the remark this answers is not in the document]
noirbizarre: Es, I think the entire file could benefit from a reformat

copier/settings.py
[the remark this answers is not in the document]
noirbizarre: I thought about this, and here's my proposal (that I can submit as another PR as I think it's another extra feature):
- keep the default like this, having a repository as a root requirement would prevent having common reusable values
- have a repositories section to allow overriding settings by repositories
It would look like:
```yaml
# Common default values
defaults:
  key1: value
  key2: value2
# Overrides settings by repositories patterns
repositories:
  https://github.com/org/*:
    default:
      key1: overriden
```
Resulting defaults for `https://github.com/org/repo.git` would be:
```yaml
key1: overriden
key2: value
```
WDYT ?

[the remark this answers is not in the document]
pawamoy: Yes that'd be great! Best of both worlds. No need to make this change for the current PR indeed. Thanks a lot for thinking about it 😄

copier/errors.py
```diff
@@ -139,3 +139,7 @@ class DirtyLocalWarning(UserWarning, CopierWarning):
 
 class ShallowCloneWarning(UserWarning, CopierWarning):
     """The template repository is a shallow clone."""
+
+
+class MissingSettingsWarning(UserWarning, CopierWarning):
+    """Settings path have been defined but file is missing."""
```
sisp, line 145:
```suggestion
    """Settings path has been defined but file is missing."""
```

:nerd_face:

copier/settings.py
```diff
@@ -0,0 +1,61 @@
...
+            if trusted.endswith("/")
+            else repository == self.normalize(trusted)
+            for trusted in self.trust
+        )
+
+    def normalize(self, url: str) -> str:
+        """Normalize an URL using user settings."""
+        if url.startswith("~"):  # Only expand on str to avoid messing with URLs
+            url = expanduser(url)
+        return url
```
sisp, line 61: I wonder whether we should use [`copier.vcs.get_repo()`](https://github.com/copier-org/copier/blob/557c0d614497e9b1b3382d4fd4be0d773d612bff/copier/vcs.py#L94-L127) to normalize trusted repository URLs to be more robust against non-canonical URLs. From the tests below, a (normalized) repository `https://github.com/user/repo.git` wouldn't be considered trusted despite the trust list entry `https://github.com/user/repo`. Won't this lead to confusion?

I realize that `copier.vcs.get_repo()` currently doesn't handle URLs with a trailing slash properly:

```pycon
>>> get_repo("https://github.com/user/repo/")
'https://github.com/user/repo/.git'
```

But this could be fixed. WDYT?

docs/settings.md
```diff
@@ -0,0 +1,56 @@
+# Settings
+
+Copier settings are stored in `<CONFIG_ROOT>/settings.yml` where `<CONFIG_ROOT>` is the
```
sisp, line 3: Should be rename `<CONFIG_ROOT>` to `<COPIER_HOME>` for better clarity and consistency with the code block title `$COPIER_HOME/settings.yml` below?

sisp (changes requested):
Just a few more comments after another review iteration. :slightly_smiling_face:

docs/settings.md
pawamoy: I vote for `<CONFIG_ROOT>` everywhere, or `<CONFIG_FOLDER>`. `$COPIER_HOME` makes it sound like that's a valid environment variable.

docs/settings.md
sisp: LGTM :slightly_smiling_face:

docs/settings.md
noirbizarre: Fixed, `<CONFIG_ROOT>` everywhere 👍🏼

copier/settings.py
noirbizarre: Yes, I noticed that when I started #1941. That's one of the reasons I separated the shortcuts into their own PR.
If this is OK for you, I would prefer handling `get_repo` and URL normalization in #1941 as it is its main point. It would allow having the full picture and craft a solution that handles more cases.

I think we have the opportunity to properly review/complete/fix normalization rules in #1941

noirbizarre:
PR rebased to solve lock conflict

pawamoy (approved):
:ok_hand:

tests/test_settings.py
```diff
@@ -62,10 +63,39 @@ def test_settings_from_param(
     assert settings.defaults == {"from": "file"}
 
 
-def test_settings_definedBut_missing(
+def test_settings_defined_but_missing(
```
sisp, line 66: Could you please move this change to the first commit?

copier/settings.py
sisp: Okay, let's address URL normalization in #1941.

copier/main.py
```diff
@@ -211,14 +211,12 @@ def __enter__(self) -> Worker:
...
-    def __exit__(self, type: None, value: None, traceback: None) -> None:
-        ...
+    def __exit__(self, type: None, value: None, traceback: None) -> None: ...
 
     @overload
     def __exit__(
         self, type: type[BaseException], value: BaseException, traceback: TracebackType
-    ) -> None:
-        ...
+    ) -> None: ...
```
sisp, line 219: Could you please reformat these lines in the first commit, so none of the two commits contain any formatting changes here?

copier/settings.py
```diff
@@ -0,0 +1,61 @@
+"""User setttings models and helper functions."""
```
sisp, line 1:
```suggestion
"""User settings models and helper functions."""
```

:nerd_face:

sisp (changes requested):
Just some final polishing requests from my side. With two independent commits in a PR, we need to be extra careful to associate a change with the right commit. For instance, we have a formatting change in the first commit and the inverse change in the second commit, so there's no diff for this section under `Files changed` but there is a diff in each commit.

noirbizarre:
Updated 👍🏼 
Sorry, the direnv+nix seems broken on my setup, and it was still having an older flake8+black stack in pre-commit (which applied the faulty formatting).
However, the `poe lint` task seems OK, I'll do with it 👍🏼

poetry.lock
sisp, line 1: I almost don't dare requesting one more change, @noirbizarre ... :face_with_peeking_eye:

Would you mind re-generating the `poetry.lock` file with Poetry v1.8 (in the first commit)? The Nix environment does not use Poetry v2.0 yet and running `poetry install` with Poetry v1.8 while `poetry.lock` was generated with Poetry v2.0 raises this warning:

```
The lock file might not be compatible with the current version of Poetry.
Upgrade Poetry to ensure the lock file is read properly or, alternatively, regenerate the lock file with the `poetry lock` command.
```

This my last request. :crossed_fingers:

sisp (changes requested):

poetry.lock
noirbizarre: It's OK, I told myself I need to be careful with Poetry 2.0 and then I forgot 😅
Anyway, poetry version fixed and lock updated 👍🏼

sisp, line 1: @noirbizarre It looks like you've updated the entire lock file. Could you please reset it and run `poetry lock --no-update` to only include the changes related to adding the `platformdirs` package?

sisp (changes requested):

poetry.lock
noirbizarre: Done 👍🏼

sisp (approved):
Thank you very much again for this great PR and your patience with the review cycles, @noirbizarre! 🙏

I think we can merge this PR, only one tiny cosmetic detail I just stumbled over 🫣: Could you fix the second commit message?

```diff
-feat(settings): allows to defined some trusted repositories or prefixes
+feat(settings): allow to define some trusted repositories or prefixes
```

noirbizarre:
No problem, there it is, fixed. (Sorry, I do many typos currently 😅)

sisp:
No worries, thanks again for your efforts and patience!

PR #1948, opened 2025-01-26
feat(context): expose a `_phase` context variable (fix #1883)

### LINKED ISSUE #1883: Expose the current phase in context
### Actual Situation

Extensions have no way of knowing the current phase (aka. questions, render, tasks or migration).

As a consequence, they have to guess which context variable is available and safe for them to use. 

Ex: [`copier-templates-extensions` was guessing the rendering phase with the presence of `_copier_conf` context var](https://github.com/copier-org/copier-templates-extensions/blob/main/src/copier_templates_extensions/extensions/context.py#L46) but is [broken since 9.4](https://github.com/copier-org/copier-templates-extensions/issues/7) due to its heuristic not working anymore

### Desired Situation

Jinja extension can selectively run during the phase they intend to, and they have a safe way of knowing which phase is currently being executed

### Proposed solution

In addition to an `_operation` context variable which has already been proposed and pending merge (#1733), a `_phase` context variable is also added.
Its value can be one of `questions` (or `prompt`), `render`, `migrate` or `tasks`.

noirbizarre:
This PR exposes a `_phase` variable which can be one of `prompt`, `tasks`, `mmigrate` and `render`.
The known phases are grouped into the `copier.types.Phase` enum.

This implements #1883 and allows fixing copier-org/copier-templates-extensions#7

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -65,6 +65,7 @@ from .types import (
     AnyByStrMutableMapping,
     JSONSerializable,
     LazyDict,
+    Phase,
     RelativePath,
     StrOrPath,
 )
@@ -277,10 +278,18 @@ class Worker:
         Files will only be parsed lazily on 1st access. This helps avoiding
         circular dependencies when the file name also comes from a variable.
         """
+
+        def _render(path: str) -> str:
+            with Phase.use(Phase.UNDEFINED):
+                return self._render_string(path)
+
+        # Given those values are lazily rendered on 1st access then cached
+        # the phase value is irrelevant and could be misleading.
+        # As a consequence it is explicitely set to "undefined".
         return LazyDict(
             **{
                 name: lambda path=path: load_answersfile_data(
-                    self.dst_path, self._render_string(path)
+                    self.dst_path, _render(path)
                 )
                 for name, path in self.template.external_data.items()
             }
@@ -375,6 +384,7 @@ class Worker:
             _copier_conf=conf,
             _folder_name=self.subproject.local_abspath.name,
             _copier_python=sys.executable,
+            _copier_phase=Phase.current(),
         )
 
     def _path_matcher(self, patterns: Iterable[str]) -> Callable[[Path], bool]:
@@ -560,7 +570,9 @@ class Worker:
         """
         path = self.answers_file or self.template.answers_relpath
         template = self.jinja_env.from_string(str(path))
-        return Path(template.render(**self.answers.combined))
+        return Path(
+            template.render(_copier_phase=Phase.current(), **self.answers.combined)
+        )
 
     @cached_property
     def all_exclusions(self) -> Sequence[str]:
@@ -928,7 +940,8 @@ class Worker:
         """
         self._check_unsafe("copy")
         self._print_message(self.template.message_before_copy)
-        self._ask()
+        with Phase.use(Phase.PROMPT):
+            self._ask()
         was_existing = self.subproject.local_abspath.exists()
         try:
             if not self.quiet:
@@ -937,12 +950,14 @@ class Worker:
                     f"\nCopying from template version {self.template.version}",
                     file=sys.stderr,
                 )
-            self._render_template()
+            with Phase.use(Phase.RENDER):
+                self._render_template()
             if not self.quiet:
                 # TODO Unify printing tools
                 print("")  # padding space
             if not self.skip_tasks:
-                self._execute_tasks(self.template.tasks)
+                with Phase.use(Phase.TASKS):
+                    self._execute_tasks(self.template.tasks)
         except Exception:
             if not was_existing and self.cleanup_on_error:
                 rmtree(self.subproject.local_abspath)
@@ -1044,9 +1059,10 @@ class Worker:
             ) as old_worker:
                 old_worker.run_copy()
             # Run pre-migration tasks
-            self._execute_tasks(
-                self.template.migration_tasks("before", self.subproject.template)  # type: ignore[arg-type]
-            )
+            with Phase.use(Phase.MIGRATE):
+                self._execute_tasks(
+                    self.template.migration_tasks("before", self.subproject.template)  # type: ignore[arg-type]
+                )
             # Create a Git tree object from the current (possibly dirty) index
             # and keep the object reference.
             with local.cwd(subproject_top):
@@ -1120,7 +1136,7 @@ class Worker:
                 self._git_initialize_repo()
                 new_copy_head = git("rev-parse", "HEAD").strip()
             # Extract diff between temporary destination and real destination
-            # with some special handling of newly added files in both the poject
+            # with some special handling of newly added files in both the project
             # and the template.
             with local.cwd(old_copy):
                 # Configure borrowing Git objects from the real destination and
@@ -1265,9 +1281,10 @@ class Worker:
             _remove_old_files(subproject_top, compared)
 
         # Run post-migration tasks
-        self._execute_tasks(
-            self.template.migration_tasks("after", self.subproject.template)  # type: ignore[arg-type]
-        )
+        with Phase.use(Phase.MIGRATE):
+            self._execute_tasks(
+                self.template.migration_tasks("after", self.subproject.template)  # type: ignore[arg-type]
+            )
 
     def _git_initialize_repo(self) -> None:
         """Initialize a git repository in the current directory."""
--- a/copier/types.py
+++ b/copier/types.py
@@ -1,11 +1,17 @@
 """Complex types, annotations, validators."""
 
+from __future__ import annotations
+
+from contextlib import contextmanager
+from contextvars import ContextVar
+from enum import Enum
 from pathlib import Path
 from typing import (
     Annotated,
     Any,
     Callable,
     Dict,
+    Iterator,
     Literal,
     Mapping,
     MutableMapping,
@@ -75,3 +81,34 @@ class LazyDict:
         if key not in self.done:
             self.done[key] = self.pending[key]()
         return self.done[key]
+
+
+class Phase(str, Enum):
+    """The known execution phases."""
+
+    PROMPT = "prompt"
+    TASKS = "tasks"
+    MIGRATE = "migrate"
+    RENDER = "render"
+    UNDEFINED = "undefined"
+
+    def __str__(self) -> str:
+        return str(self.value)
+
+    @classmethod
+    @contextmanager
+    def use(cls, phase: Phase) -> Iterator[None]:
+        """Set the current phase for the duration of a context."""
+        token = _phase.set(phase)
+        try:
+            yield
+        finally:
+            _phase.reset(token)
+
+    @classmethod
+    def current(cls) -> Phase:
+        """Get the current phase."""
+        return _phase.get()
+
+
+_phase: ContextVar[Phase] = ContextVar("phase", default=Phase.UNDEFINED)
--- a/docs/creating.md
+++ b/docs/creating.md
@@ -136,6 +136,16 @@ variable:
 
 The name of the project root directory.
 
+### `_copier_phase`
+
+The current phase, one of `"prompt"`,`"tasks"`, `"migrate"` or `"render"`.
+
+!!! note
+
+    There is also an additional `"undefined"` phase used when not in any phase.
+    You may encounter this phase when rendering outside of those phases,
+    when rendering lazily (and the phase notion can be irrelevant) or when testing.
+
 ## Variables (context-specific)
 
 Some rendering contexts provide variables unique to them:
--- a/tests/test_migrations.py
+++ b/tests/test_migrations.py
@@ -521,3 +521,32 @@ def test_migration_jinja_variables(
             assert f"{variable}={value}" in vars
         else:
             assert f"{variable}=" in vars
+
+
+def test_copier_phase_variable(tmp_path_factory: pytest.TempPathFactory) -> None:
+    """Test that the Phase variable is properly set."""
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    with local.cwd(src):
+        build_file_tree(
+            {
+                **COPIER_ANSWERS_FILE,
+                "copier.yml": (
+                    """\
+                    _migrations:
+                        - touch {{ _copier_phase }}
+                    """
+                ),
+            }
+        )
+        git_save(tag="v1")
+    with local.cwd(dst):
+        run_copy(src_path=str(src))
+        git_save()
+
+    with local.cwd(src):
+        git("tag", "v2")
+    with local.cwd(dst):
+        run_update(defaults=True, overwrite=True, unsafe=True)
+
+    assert (dst / "migrate").is_file()
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -5,13 +5,14 @@ from __future__ import annotations
 import json
 import warnings
 from collections import ChainMap
+from collections.abc import Mapping, Sequence
 from dataclasses import field
 from datetime import datetime
 from functools import cached_property
 from hashlib import sha512
 from os import urandom
 from pathlib import Path
-from typing import Any, Callable, Literal, Mapping, Sequence
+from typing import Any, Callable, Literal
 
 import yaml
 from jinja2 import UndefinedError
@@ -33,6 +34,7 @@ from .types import (
     AnyByStrMutableMapping,
     LazyDict,
     MissingType,
+    Phase,
     StrOrPath,
 )
 
@@ -464,7 +466,13 @@ class Question:
                 else value
             )
         try:
-            return template.render({**self.answers.combined, **(extra_answers or {})})
+            return template.render(
+                {
+                    **self.answers.combined,
+                    **(extra_answers or {}),
+                    "_copier_phase": Phase.current(),
+                }
+            )
         except UndefinedError as error:
             raise UserMessageError(str(error)) from error
 
--- a/tests/test_answersfile.py
+++ b/tests/test_answersfile.py
@@ -250,3 +250,31 @@ def test_external_data_with_umlaut(
     copier.run_copy(str(src), dst, defaults=True, overwrite=True)
     answers = load_answersfile_data(dst, ".copier-answers.yml")
     assert answers["ext_umlaut"] == "äöü"
+
+
+def test_undefined_phase_in_external_data(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    build_file_tree(
+        {
+            (src / "copier.yml"): (
+                """\
+                _external_data:
+                    data: '{{ _copier_phase }}.yml'
+                key: "{{ _external_data.data.key }}"
+                """
+            ),
+            (src / "{{ _copier_conf.answers_file }}.jinja"): (
+                "{{ _copier_answers|to_nice_yaml }}"
+            ),
+        }
+    )
+    git_save(src, tag="v1")
+
+    (dst / "undefined.yml").write_text("key: value")
+
+    copier.run_copy(str(src), dst, defaults=True, overwrite=True)
+    answers = load_answersfile_data(dst, ".copier-answers.yml")
+    assert answers["key"] == "value"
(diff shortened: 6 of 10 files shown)


tests/test_migrations.py
```diff
@@ -521,3 +521,32 @@ def test_migration_jinja_variables(
             assert f"{variable}={value}" in vars
         else:
             assert f"{variable}=" in vars
+
+
+def test_phase_variable(tmp_path_factory: pytest.TempPathFactory) -> None:
+    """Test that the j-phase variable is properly set"""
```
pawamoy, line 527:
```suggestion
    """Test that the phase variable is properly set"""
```

pawamoy:
Looking super good to me :smile: Thank you for the awesome PRs @noirbizarre!

noirbizarre:
PR updated with the typo fixes.

You're welcome 🙏🏼 It benefits everybody 👌🏼

pawamoy (approved):

sisp (changes requested):
Fantastic work again, @noirbizarre! :pray:

Just a minor stylistic request: It seems Python enums should be upper-case[^1][^2]. I suggest following this guideline.

[^1]: https://mail.python.org/pipermail/python-ideas/2016-September/042340.html
[^2]: https://docs.python.org/3/library/enum.html

noirbizarre:
Updated to upper case 👍🏼

But actually, there is no official recommendation (so no "should"). I, personally, tend to do the opposite and I explicitly chose lower case (and use this convention on all my projects) for multiple reasons:
- `StrEnum` in 3.11 use lower case values with `auto()` which is deviating from the initial `StrEnum` package implementation: https://github.com/python/cpython/issues/115509#issuecomment-2507207990 and is a source of confusion. They would have chosen upper case, there wouldn't be a debate
- this debate has occurred in very language and there are arguments for all casing (some even recommend CamelCase), and as a consequence, almost none of the language provide a formal recommendation
- it makes hard to understand some behavior in Python's StrEnum without relying on the doc or a REPL (there is some general confusion between `Phase['tasks']`, `Phase['TASKS']`, `Phase('tasks')`, `Phase('TASKS')`, `Phase.TASKS == 'tasks'`..., it needs more thinking and practice). I abuse of enums, so not an issue to me, but I see many beginners struggling with this.
- there is no linter enforcing this
- the documentation doesn't even provide a single `StrEnum` example, only int based enums
- this mail is actually the only existing reference to that, it's Guido's preference and the start of a debate on this topic, which he ruled by authority, not consensus, and only by updating the example, no reference to any recommendation
- the enum PEP is actually using lower case: https://peps.python.org/pep-0435/

All lower case as the benefit of consistency and lower cognitive overhead and if I didn't use `StrEnum` which is only available in 3.11, I did it in prevision of.

Nitpicky, I know, but given it was a choice, I wanted to justify it 😉

sisp (approved):
It seems you're right, there isn't a definitive style rule about enums. :face_with_peeking_eye: Thanks for elaborating on this topic (especially the detailed message before you edited it :wink:)! Well, now it's upper-case. Shall we leave it like this?

Just a final thought: Shall we expose the phase via the `_phase` variable (as it is now) or would it be better to rename/move it? E.g.:

- `_copier_phase` – only `_folder_name`, `now`, and `make_secret` seems to not have the `_copier` prefix where `now` and `make_secret` are both deprecated.
- `_copier_conf.phase` – or should we only have static values under `_copier_conf`?

@noirbizarre @copier-org/maintainers Any opinions?

pawamoy:
I tend to use lowercase too, except when the enum contains names that are reserved keywords like `class`! No other choice than going uppercase then. As long as we're consistent within Copier's code base, I'm fine with both.

I'm accessing the phase with `parent["_copier_conf"]["_phase"]` in https://github.com/copier-org/copier-templates-extensions/pull/8, it would feel a bit redundant as `_copier_phase`, but again, consistency is more important IMO.

sisp:
@pawamoy So, would you tend to be in favor of `_copier_phase` rather than `_phase`?

pawamoy:
Ah, wait. This PR adds `_phase` to the dict returned by `_render_context`. But this dict already contains `_copier_conf`, which contains the phase. So do we really need to add it to the upper dict again?

In the same way, the PR adds `_phase` to the context in `_render_value`, which IIRC is used to render values in copier.yml. I don't remember however if `_copier_conf` is accessible when rendering values in copier.yml. If that's the case, then I'll ask the same question as above: do we need the shortcut `{{ _phase }}` if we already have `{{ _copier_conf.phase }}`?

pawamoy:
OK looking at the code I *think* we *need* the phase in `render_value`, because `_copier_conf` is not available there, and we need the phase in context extensions to decide whether to run or not. Whether we need it in `_render_context` is unclear to me, since it's already contained in the `_copier_conf` value, and since no use-cases were suggested for this variable in file names / file contents.

pawamoy:
OK, sorry I was very confused. Here's my final comprehension of the issue (which may be worth adding as documentation or code comments at least).

- During `copier copy`, and prompt rendering, context extensions do not get `_copier_conf`. We can use its absence to decide not to run the context hook.
- During `copier update` however, and prompt rendering, context extensions *do* get `_copier_conf`. We can't use its absence anymore. Instead, we need the phase. It's passed through `parent["_copier_conf"]["phase"]`: we don't strictly need the additional top-level `"_phase"` key-value pair. We could keep it to provide more consistency: a `_phase` / `_copier_phase` top-level key always present. No strong opinions.
- Extensions can now also skip running the context hook during tasks and migrations, thanks to the phase.

noirbizarre:
Oh, I didn't though about all that 😅 
Prefixing by `_copier` makes sense and is consistent with existing ones.
I would go for a root `_copier_phase` as it is not a configuration, just a runtime state.
I added it in `_render_context` just to say that it is always available. I don't like to presume future evolution as well as community usage, so I've been on the dumb "it's the current state, and it's always available approach".

Note that I actually use context extensions during prompt on my projects and actually this PR can maybe help to have a more general approach on this and assume that extensions can run at any phase. (This might later lead to having copier-template-extensions `TemplateExtensionLoader` builtin 🤷🏼‍♂️)

Anyway, PR update with `_copier_phase` naming.

sisp (commented):
I've thought about this PR some more and also about – what @pawamoy noticed – the phase value being automatically exposed as `_copier_conf.phase` because [most `Worker` dataclass attributes](https://github.com/copier-org/copier/blob/57439e5c18616169ae44635fe56d2f76bb06e416/copier/main.py#L338-L339) are automatically [included in `_copier_conf`](https://github.com/copier-org/copier/blob/57439e5c18616169ae44635fe56d2f76bb06e416/copier/main.py#L354). As

> I would go for a root `_copier_phase` as it is not a configuration, just a runtime state.

makes sense to me, I think the phase value should _not_ be accessible under `_copier_conf` though. This PR also reminds me of a similar situation we've had in #1733 where initially a dataclass attribute was used as well and [I suggested to use a context variable](https://github.com/copier-org/copier/pull/1733#pullrequestreview-2397086710) instead. Now, it feels odd to me to use a dataclass attribute here to manage the phase value because it has no meaning while not rendering, prompting or executing tasks/migrations. How about applying an approach similar to https://github.com/copier-org/copier/pull/1733#pullrequestreview-2397086710 in this PR as well?

Slightly off-topic: I just realized that #1733 exposes a top-level `_operation` Jinja variable; we might want to rename it to `_copier_operation` there for consistency.

pawamoy:
I agree with all @sisp's points :slightly_smiling_face: 

Regarding `_copier_conf`, the code says "backwards compatibility, remove it?":

https://github.com/copier-org/copier/blob/57439e5c18616169ae44635fe56d2f76bb06e416/copier/main.py#L334-L339

Not sure what it means exactly, but it seems like we wanted to stray away from this anyway.

sisp:
@pawamoy This comment pre-dates my involvement with Copier, but I don't think the (raw) `Worker` dataclass attributes are very useful in the Jinja context; [useful config values are explicitly added](https://github.com/copier-org/copier/blob/57439e5c18616169ae44635fe56d2f76bb06e416/copier/main.py#L342-L344). I'd be in favor of dropping those lines as part of the next major release

pawamoy:
Fine by me. I'd suggest making sure `_copier_phase` is always passed. It's currently not passed from the `answers_relpath` property, which calls `template.render(**self.answers.combined)` (`template` being the answers file name). Looks like it's the only missing location :slightly_smiling_face:

noirbizarre:
PR update with `answers_relpath` handled 👍🏼 

I saw the deprecation noticed, but it is used by all rendering except those for the prompts.

> I'd be in favor of dropping those lines as part of the next major release.

Agreed, but I would go even further, I would isolate rendering in a single class/module/function/whatever, so future modification of this kind occurs in a single place. It would be easier to maintain and evolve, and easier to test.

sisp:
While resuming a discussion in #1733, @pawamoy [reminded be about the existence of the `_stage` variable in migrations](https://github.com/copier-org/copier/pull/1733#issuecomment-2621666810) that was introduced by #1510. This variable is limited to render contexts related to migrations, but nevertheless it isn't prefixed with `_copier`. Just to make sure we're consistent, would you mind sharing your thoughts once more about naming the variable `_phase` vs. `_copier_phase`? Honestly, I'm torn between prefixing and not prefixing now. :see_no_evil:

noirbizarre:
I think prefixing global variables is consistent with what is already documented: https://copier.readthedocs.io/en/latest/creating/#variables-global
The `_stage` one is one of the scoped variables that currently on exists for migrations: https://copier.readthedocs.io/en/latest/creating/#variables-context-specific 

Having globals prefixed make sense to me (not only in copier but more generally as a sane development anti-collision safety)

It seems part of the current release is being focused on variables and putting some rules on it, so if I follow my reasoning, I would rename the var `_copier_operation` in #1733 as it is a global.

For migration, I think there are 2 choices:
- say they are not global but scoped to migrations and keep them as they are
- say it needs to follow the same rule, refactor migration scoped var into a migration scoped object `_copier_migration` (e.g. `_copier_migration.stage`, `_copier_migration.version_from`...)

Given the 2nd option would be a breaking change, it would force very copier template maintainer to rewrite all migrations and would produce more verbose migrations, I think it would be a lot of work for very few to none benefit except naming rule compliance.
So I would go with the first choice, keep them as they are.

noirbizarre:
Updated: rebased, solved conflicts and adapted to the latest changes.
That was not easy, as 25bc06d01 introduced a loop:
- `Worker.answers_relpath` uses `Worker.answers.combined` as context
- `Worker.answers.combined` uses `Worker.answers.system`
- `Worker.answers.system` is generated by `Worker._system_render_context()`
- `Worker._system_render_context()` builds `_copier_conf` using `Worker.answers_relpath`
As a consequence, the values used to generate `answers_relpath` are those from the previous `Worker._system_render_context()` call.

I found a fix for this PR and added a comment to keep track of it.

sisp:
There are some merge conflicts because of #1993 now, which mostly reverted the system render context change because of #1977 that also affected this PR.

> This PR also reminds me of a similar situation we've had in https://github.com/copier-org/copier/pull/1733 where initially a dataclass attribute was used as well and https://github.com/copier-org/copier/pull/1733#pullrequestreview-2397086710 instead. Now, it feels odd to me to use a dataclass attribute here to manage the phase value because it has no meaning while not rendering, prompting or executing tasks/migrations. How about applying an approach similar to https://github.com/copier-org/copier/pull/1733#pullrequestreview-2397086710 in this PR as well?
>
> &mdash; https://github.com/copier-org/copier/pull/1948#pullrequestreview-2580789541

I still think that the `Worker.phase` attribute isn't the right design. Would you mind trying to use a context variable (https://github.com/copier-org/copier/pull/1733#pullrequestreview-2397086710) instead?

noirbizarre:
Updated:
- conflicts solved
- uses a `ContextVar` instead of an attribute

sisp (changes requested):
Thanks, @noirbizarre!

Sorry for being a little pedantic about the context variable approach. 🫣  But could you remove the default phase value and use a context manager to manage the phase value, such that a particular phase value is only set within an explicit created context? This way, we reduce the risk of forgetting to set a new phase value and accidentally exposing a previously set one, as [`ContextVar.get()`](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar.get) raises a `LookupError` when neither a value is set nor a default value is provided.

E.g.:

```python
with use_phase(phase.TASK):
    self._execute_task(self.template.tasks)
```

(I'm open to better context manager names than `use_phase` by the way. 😄)

noirbizarre:
Updated with a context manager.
However, it's not possible to work without a default value or ~2/3 of the tests are failing, which is expected:
- methods are unit tested outside the lifecycle worker lifecycle so they are out of any known phase and fail with a `LookupError`.
- many properties and methods (especially `answers_relpath` and `_print_message`) are called outside any phase for computation or user feedback and triggers a rendering, so a lookup
- some operations are done in preparation of a phase outside the phase method itself, example, print statement are outside the methods, as the same method can be used differently depending on the phase (ex: `_execute_tasks`)

Not having a default value would mean:
- modifying ~2/3 of the tests
- indenting many statements just to use the context manager to ensure the touched property or the method triggering a render/lookup as side effect is not failing

That would mean having most of the codebase modified just to add a variable, and potential conflict with any pending PR, or any PR merged before this one.

So instead, I added a `Phase.UNDEFINED` as default so tests not relying on it and ran outside the lifecycle don't fail on as an unrelated side effect and the day someone actually hit a case where he needs the `_copier_phase` in a new phase and setting the phase has been forgotten (that's a lot of conditions), it will have an explicit `undefined` as value, it will be easy to identify, and it won't make `copier` failing for him until a fix is released.

WDYT ?

sisp (changes requested):
Thanks for explaining why no default value for the context variable won't work. You're right, and I think introducing `"undefined"` and using it as a default value is a good idea. I also like the implementation of the context manager as a class method of `Phase`. :+1: Well done! :clap:

IIUC, the phase value when [rendering the answers file path of external data](https://github.com/copier-org/copier/blob/6eaa0d7e0959c81479b214eaf73aa55d14d0364d/copier/main.py#L283) depends on the phase in which that particular external data is _first_ accessed due to lazy access. Would it be better to override the phase value with `"undefined"` instead? The meaning of the phase doesn't seem well defined in this case. WDYT?

copier/types.py
```diff
@@ -75,3 +81,35 @@ def __getitem__(self, key: str) -> Any:
...
+
+
+class Phase(str, Enum):
+    """The known execution phases."""
+
+    PROMPT = "prompt"
+    TASKS = "tasks"
+    MIGRATE = "migrate"
+    RENDER = "render"
+    UPDATE = "update"
```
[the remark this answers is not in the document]
noirbizarre: To me it doesn't represent the same thing, the `update` operation is replaying the different existing phases but is also doing some git operations in the `_apply_update` method. So, I hesitated between letting `undefined` as phase for those steps or defining a dedicated phase for that.
I don't have a strong opinion on this, and I am perfectly fine with saying it should stay `undefined` while not in one of the other phases.
I updated the PR to remove it

noirbizarre:
Make sense 👍🏼 
I set it explicitly to `undefined` and added a comment to explain why.

copier/main.py
```diff
@@ -277,14 +278,18 @@ def _external_data(self) -> LazyDict:
...
+        # As a consequence it is explicitely set to "undefined".
+        with Phase.use(Phase.UNDEFINED):
+            return LazyDict(
+                **{
+                    name: lambda path=path: load_answersfile_data(
+                        self.dst_path, self._render_string(path)
+                    )
+                    for name, path in self.template.external_data.items()
+                }
+            )
```
sisp, line 292: This won't work because the answers file path is rendered lazily. The phase context needs to be overridden inside the lamba. But this isn't possible I'm Python AFAIK, so we could introduce a closure instead of the lamda which renders the string inside the phase context.

Could you also add a test case please?

sisp (changes requested):
Almost done, I think. 🤞

copier/main.py
noirbizarre: Done 👍🏼

copier/main.py
```diff
@@ -277,15 +278,21 @@ def _external_data(self) -> LazyDict:
...
+                name: lambda path=path: self._lazy_load_answers(path)
                 for name, path in self.template.external_data.items()
             }
         )
 
+    def _lazy_load_answers(self, path: str) -> AnyByStrDict:
+        """Lazy-Load external answers file."""
+        with Phase.use(Phase.UNDEFINED):
+            return load_answersfile_data(self.dst_path, self._render_string(path))
+
```
sisp, line 295: I don't like the semantics of this method because it suggests that lazily loaded answers should always render the path with `_copier_phase = "undefined"` which isn't a general rule. I was thinking about a closure – see my other inline comment.

copier/main.py
```diff
@@ -277,15 +278,21 @@ def _external_data(self) -> LazyDict:
...
+        # As a consequence it is explicitely set to "undefined".
         return LazyDict(
             **{
-                name: lambda path=path: load_answersfile_data(
-                    self.dst_path, self._render_string(path)
-                )
+                name: lambda path=path: self._lazy_load_answers(path)
                 for name, path in self.template.external_data.items()
             }
         )
```
sisp, line 289: I'd use a closure to deal with this special case locally:

```python
def _render(path: str) -> str:
    with Phase.use(Phase.UNDEFINED):
        return self._render_string(path)

# Given those values are layzily rendered on 1st access then cached
# the phase value is irrelevant and could be misleading.
# As a consequence it is explicitely set to "undefined".
return LazyDict(
    **{
        name: lambda path=path: load_answersfile_data(
            self.dst_path, _render(path)
        )
        for name, path in self.template.external_data.items()
    }
)
```

copier/main.py
```diff
@@ -277,15 +278,21 @@ def _external_data(self) -> LazyDict:
         Files will only be parsed lazily on 1st access. This helps avoiding
         circular dependencies when the file name also comes from a variable.
         """
+        # Given those values are layzily rendered on 1st access then cached
```
sisp, line 281:
```suggestion
        # Given those values are lazily rendered on 1st access then cached
```

:nerd_face:

sisp (changes requested):

copier/types.py
```diff
@@ -75,3 +81,34 @@ def __getitem__(self, key: str) -> Any:
...
             self.done[key] = self.pending[key]()
         return self.done[key]
+
+
+class Phase(str, Enum):
+    """The known execution phases."""
+
+    PROMPT = "prompt"
+    TASKS = "tasks"
+    MIGRATE = "migrate"
```
sisp, line 91: Admittedly totally subjective, but would

```suggestion
    MIGRATIONS = "migrations"
```

be more consistent with

```python
TASKS = "tasks"
```

without introducing inconsistency with the other members?

noirbizarre: I don't really have an opinion on the enum members semantic.
They were named after the action being executed in that phase, `prompt`, `migrate` and `render` are actions. `tasks` is short for "execute tasks".
I can rename them whatever suit you.
But for consistency, if you go for `migrations`, aka. the type of object dealt in the phase, I would go for: 
- ~`prompt`~ > `questions`
- `tasks`
- ~`migrate`~ > `migrations`
- `render` > ??? (hard to describe the object there instead of the action)

sisp: Thanks for sharing your thoughts. Let's keep the action semantics.

docs/creating.md
```diff
@@ -136,6 +136,15 @@ variable:
...
 The name of the project root directory.
 
+### `_copier_phase`
+
+The current phase, one of `"prompt"`,`"tasks"`, `"migrate"` or `"render"`.
+
+!!! note
+
+    There is also an additional `"undefined"` phase used when not in any phase, mostly
+    for testing purpose.
```
sisp, line 146: Just a final thought here: I think the `"undefined"` phase isn't mostly used for testing purposes. It's used when rendering, e.g., settings in `copier.yaml` including `exclude` items, `skip_if_exists` items, `subdirectory`. Perhaps we could rephrase this a little?

noirbizarre: Updated with a more accurate description.

sisp (approved):
Awesome! 🤩 Huge thanks for adding this feature and being incredibly patient with my pedantic review feedback and the many iterations involving code updates and discussions! 🙇 Fantastic work, @noirbizarre! 🏆

PR #1970, opened 2025-02-17
fix: decode answers file content explicitly as UTF-8

### LINKED ISSUE #1963: Copier is messing up Umlaut on template update
### Describe the problem

When entering an Umlaut (ö,ü,ä) to an answer everything is written correctly. But on updating the template, copier is messing it up.

### Template

New `template` with git initialized (and tagged)

`{{_copier_conf.answers_file}}.jinja`:
```jinja
{{ _copier_answers|to_nice_yaml -}}
```

`copier.yml`:
```yml
umlaut:
  type: str
  help: Type an umlaut (ö,ü,ä).
```

`umlaut.md.jinja`:
```jinja
{{ umlaut }}
```

### To Reproduce

1.  `copier copy -d umlaut=ä .\template\ .\test\` (PowerShell)
2. `cd test`
3. `git init`, `git add .`, `git commit -m 'first'`
4. `copier update --trust`

### Logs
On update you will see:

```console
Updating to template version 0.2.0
🎤 Type an umlaut (ö,ü,ä).
   Ã¤
```

The `.copier-answers.yml`-file and the `umlaut.md` file will be overwritten with `Ã¤`

### Expected behavior

No overwriting. Keep `ä` in `.copier-answers.yml`-file and the `umlaut.md` file

_No response_

### Operating system

Windows

Windows 11

### Copier version

9.4.1

### Python version

3.13.2

### Installation method

local build

### Additional context

Independent of build method, here `uvx` was used.

sisp:
I've fixed a string encoding/decoding bug on Windows that broke updating a project with recorded non-ASCII character answers.

According to my investigation, `pathlib.Path.read_text()` calls [`io.text_encoding()`](https://github.com/python/cpython/blob/aa845af9bb39b3e2ed08bbb00a8e932a97be8fc0/Lib/pathlib/_local.py#L776), and when no encoding value is provided (i.e., `encoding=None`) then a [default encoding is determined](https://github.com/python/cpython/blob/fc8c99a8ce483db23fa624592457e350e99193f6/Lib/_pyio.py#L56-L59). When I reproduced the error on a quite clean Windows 11, `sys.flags.utf8_mode == 0`, so `io.text_encoding(encoding=none)` returns `"locale"` and `locale.getencoding()` returns `"cp1252"`. At the same time, the [`to_nice_yaml` filter passes the serialized YAML string through `to_text()`](https://gitlab.com/dreamer-labs/libraries/jinja2-ansible-filters/-/blob/58e352012f1b0bd6eb68cd02a4618f9af21d61bb/jinja2_ansible_filters/core_filters.py#L98) without passing an explicit value to the [`encoding` argument which defaults to `"utf-8"`](https://gitlab.com/dreamer-labs/libraries/jinja2-ansible-filters/-/blob/58e352012f1b0bd6eb68cd02a4618f9af21d61bb/jinja2_ansible_filters/module_utils.py#L196). Hence, when the answers file content is rendered as `{{ _copier_answers|to_nice_yaml }}`, it is _always_ encoded as UTF-8, but on Windows 11 its content was decoded as CP-1252.

The solution is to _always_ decode the answers file content as UTF-8.

I believe we'll need to fix this problem also for [loading from external data files](https://github.com/copier-org/copier/pull/1880) and for [loading user settings](https://github.com/copier-org/copier/pull/1940). I'll send follow-up PRs if/where needed once this one has been merged.

Fixes #1963.

### DIFF
--- a/copier/subproject.py
+++ b/copier/subproject.py
@@ -56,7 +56,7 @@ class Subproject:
         """Get last answers, loaded raw as yaml."""
         try:
             return yaml.safe_load(
-                (self.local_abspath / self.answers_relpath).read_text()
+                (self.local_abspath / self.answers_relpath).read_text("utf-8")
             )
         except OSError:
             return {}
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -1768,3 +1768,36 @@ def test_gitignore_file_unignored(
     # Otherwise, it should succeed.
     run_update(dst_path=dst, overwrite=True)
     assert "_commit: v3" in (dst / ".copier-answers.yml").read_text()
+
+
+def test_update_with_answers_with_umlaut(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "copier.yml": (
+                    """\
+                    umlaut:
+                        type: str
+                    """
+                ),
+                "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_nice_yaml }}",
+            }
+        )
+        git_init("v1")
+        git("tag", "v1")
+
+    run_copy(str(src), dst, data={"umlaut": "äöü"}, overwrite=True)
+    answers_file = dst / ".copier-answers.yml"
+    answers = yaml.safe_load(answers_file.read_text("utf-8"))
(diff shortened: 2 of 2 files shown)


sisp:
I realized that the user settings file doesn't seem to be affected by this encoding/decoding issue because it isn't shared across developer machines with potentially different operating systems unlike the answers file. Hence, when the user settings file is encoded with the system's default encoding scheme, there should be no problem as Copier is currently using that same encoding scheme to decode the file content.

PR #1982, opened 2025-02-22
enable stricter ruff configuration

danieleades:


pawamoy (approved):
Thanks so much @danieleades for opening a new PR this quick :rocket: 

LGTM.

I'm really not a fan of E501, especially with 80-characters width, which is quite short and triggers a lot of those, but let's *not* address this here :+1:

danieleades:
> Thanks so much @danieleades for opening a new PR this quick 🚀
> 
> LGTM.
> 
> I'm really not a fan of E501, especially with 80-characters width, which is quite short and triggers a lot of those, but let's _not_ address this here 👍

If you feel that way, it makes sense to increase that character limit _before_ merging this PR since the lines will wrap differently. Do you have a specific width in mind?

pawamoy:
I generally use 120. Let see what other maintainers think.

sisp:
I have don't have a strong opinion on line length, but I tend to prefer sticking with (sane) defaults. Ruff defaults to 88. Do you feel it's a too low value?

pawamoy:
With 4-space indents, I think it's too low, yes. There are many, many strings and snippets that get split or exploded on multiple lines because we're in a condition in a loop in a method in a class. But I understand why people sometimes prefer sticking to 80/88, it makes it easier to read code side by side in multiple panes. So, no strong opinion. If none of us has strong opinions, lets keep the current line length :slightly_smiling_face:

copier/errors.py
```diff
@@ -72,7 +72,7 @@ class ExtensionNotFoundError(UserMessageError):
     """Extensions listed in the configuration could not be loaded."""
 
 
-class CopierAnswersInterrupt(CopierError, KeyboardInterrupt):
+class CopierAnswersInterruptError(CopierError, KeyboardInterrupt):
```
sisp, line 75: I think we should add an alias

```python
# Backwards compatibility
CopierAnswersInterrupt = CopierAnswersInterruptError
```

to retain backwards compatibility, just in case somebody relies on catching this exception.

tests/test_conditional_file_name.py
```diff
@@ -69,7 +69,7 @@ def test_answer_changes(
     with local.cwd(src):
         build_file_tree(
             {
-                "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_nice_yaml }}",
+                "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_nice_yaml }}",  # noqa: E501
```
sisp, line 72: I think we could avoid such rule exceptions:

```suggestion
                "{{ _copier_conf.answers_file }}.jinja": (
                    "{{ _copier_answers|to_nice_yaml }}"
                ),
```

Several more cases like this below.

sisp (changes requested):
@danieleades Looks great! Just some minor remarks from my side.

PR #1984, opened 2025-02-23
style: use pathlib

danieleades:


copier/vcs.py
```diff
@@ -190,12 +189,12 @@ def clone(url: str, ref: str | None = None) -> str:
             _clone = _clone["--filter=blob:none"]
     _clone()
     # Include dirty changes if checking out a local HEAD
-    if ref in {None, "HEAD"} and os.path.exists(url) and Path(url).is_dir():
+    url_abspath = Path(url).absolute()
+    if ref in {None, "HEAD"} and url_abspath.exists() and url_abspath.is_dir():
```
pawamoy, line 193: I believe `is_dir` implies `exists`.

```suggestion
    if ref in {None, "HEAD"} and url_abspath.is_dir():
```

tests/test_symlinks.py
```diff
@@ -312,10 +311,10 @@ def test_update_file_to_symlink(tmp_path_factory: pytest.TempPathFactory) -> Non
...
     with local.cwd(src):
         # test updating a symlink
-        os.remove("bbbb.txt")
-        os.symlink("aaaa.txt", "bbbb.txt")
-        os.remove("cccc.txt")
-        with open("cccc.txt", "w+") as f:
+        os.remove("bbbb.txt")  # noqa: PTH107
+        os.symlink("aaaa.txt", "bbbb.txt")  # noqa: PTH107
+        os.remove("cccc.txt")  # noqa: PTH107
+        with open("cccc.txt", "w+") as f:  # noqa: PTH123
```
pawamoy, line 317: Why are these not updated and ignored?

pawamoy (commented):
Thanks. I guess it's good to be more consistent about using pathlib.

tests/test_symlinks.py
danieleades: i think i thought pathlib wasn't adding a lot of value here. it's not, but i think it's worth it for consistency. will update

copier/vcs.py
pawamoy: Thanks for making sure of it!

sisp (approved):
LGTM, @danieleades! :tada: Thanks! :bow: I also prefer `pathlib` over `os`/`os.path`. :+1:

Issue #1990, opened 2025-02-27
raise a specific exception when a task fails


tsvikas:
### Actual Situation

when a task fails, copier raises `subprocess.CalledProcessError`. 

### Desired Situation

raise a specific exception, such as TaskFailed
this will allow user or external tools that run copier to separate between exceptions caused by tasks, and those caused by code issues.

### Proposed solution

in `copier/main.py:369`, replace the subprocess internal check with code that checks the return code and raises `TaskFailed` with the task command and task returncode.

I can write the code if needed.

tsvikas:
i can go with any name :).
notice that if the task is a linter / checker, it might return 1 when it means "i found issues in your code" and not "i failed to run"

PR #1991, opened 2025-02-27
✨ Add TaskError exception for subprocess management

tsvikas:
Introduce the TaskError exception to handle subprocess failures gracefully. 

solves #1990

### DIFF
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -3,7 +3,8 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import TYPE_CHECKING, Sequence
+from subprocess import CompletedProcess
+from typing import TYPE_CHECKING, Self, Sequence
 
 from .tools import printf_exception
 from .types import PathSeq
@@ -120,6 +121,36 @@ class YieldTagInFileError(CopierError):
 
 class MultipleYieldTagsError(CopierError):
     """Multiple yield tags are used in one path name, but it is not allowed."""
+
+
+class TaskError(CopierError):
+    """Exception raised when a task fails."""
+
+    def __init__(
+        self,
+        task_command: str | Sequence[str],
+        returncode: int,
+        stdout: str | bytes | None,
+        stderr: str | bytes | None,
+    ):
+        self.task_command = task_command
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+        message = f"Task {task_command!r} returned non-zero exit status {returncode}."
+        super().__init__(message)
+
+    @classmethod
+    def from_process(
+        cls, process: CompletedProcess[str] | CompletedProcess[bytes]
+    ) -> Self:
+        """Create a TaskError from a CompletedProcess."""
+        return cls(
+            task_command=process.args,
+            returncode=process.returncode,
+            stdout=process.stdout,
+            stderr=process.stderr,
+        )
 
 
 # Warnings
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -22,6 +22,12 @@ class CopierError(Exception):
 class UserMessageError(CopierError):
     """Exit the program giving a message to the user."""
 
+    def __init__(self, message: str):
+        self.message = message
+
+    def __str__(self) -> str:
+        return self.message
+
 
 class UnsupportedVersionError(UserMessageError):
     """Copier version does not support template version."""
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -151,7 +151,9 @@ class TaskError(subprocess.CalledProcessError, UserMessageError):
         )
         message = f"Task {command!r} returned non-zero exit status {returncode}."
         if self.stderr:
-            stderr_str = self.stderr.decode() if isinstance(self.stderr, bytes) else self.stderr
+            stderr_str = (
+                self.stderr.decode() if isinstance(self.stderr, bytes) else self.stderr
+            )
             message += f"\nStderr:\n{stderr_str.strip()}"
         UserMessageError.__init__(self, message)
 
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -136,7 +136,7 @@ class MultipleYieldTagsError(CopierError):
     """Multiple yield tags are used in one path name, but it is not allowed."""
 
 
-class TaskError(UserMessageError, subprocess.CalledProcessError):
+class TaskError(subprocess.CalledProcessError, UserMessageError):
     """Exception raised when a task fails."""
 
     def __init__(
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import subprocess
 from pathlib import Path
 from subprocess import CompletedProcess
 from typing import TYPE_CHECKING, Self, Sequence
@@ -129,7 +130,7 @@ class MultipleYieldTagsError(CopierError):
     """Multiple yield tags are used in one path name, but it is not allowed."""
 
 
-class TaskError(UserMessageError):
+class TaskError(UserMessageError, subprocess.CalledProcessError):
     """Exception raised when a task fails."""
 
     def __init__(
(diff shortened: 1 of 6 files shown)


copier/errors.py
```diff
@@ -120,6 +123,36 @@ class MultipleYieldTagsError(CopierError):
...
+        task_command: str | Sequence[str],
+        returncode: int,
+        stdout: str | bytes | None,
+        stderr: str | bytes | None,
+    ):
+        self.task_command = task_command
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+        message = f"Task {task_command!r} returned non-zero exit status {returncode}."
```
sisp, line 140: - Rename `task_command` to `command`, as it's implied by the exception name that the command is associated with a task.
- I'd surround the command in the message with backticks to improve readability when the command contains spaces.

```suggestion
        command: str | Sequence[str],
        returncode: int,
        stdout: str | bytes | None,
        stderr: str | bytes | None,
    ):
        self.command = command
        self.returncode = returncode
        self.stdout = stdout
        self.stderr = stderr
        message = f"Task `{command!r}` returned non-zero exit status {returncode}."
```

sisp (changes requested):
Thanks for your contribution, @tsvikas! :pray:

I have two minor suggestions (see inline comment). 

IIUC, a failing task will raise the `TaskError` exception and the message will be shown, but the error from the task command will be hidden when running Copier via CLI, as it's not printed anymore but only accessible via the `stdout`/`stderr` attributes of the exception object. I think we should catch this exception in

https://github.com/copier-org/copier/blob/a731d2b679e49fe1f83326b5639822faebd46571/copier/cli.py#L66-L80

and print the message plus `stdout`/`stderr`.

copier/errors.py
tsvikas: the `!r` already add quotemarks to the output

tsvikas: i will change the `.command` as suggested

sisp: > the `!r` already add quotemarks to the output

Oh, I didn't know. 🤦 Good work then! 👏

copier/errors.py
```diff
@@ -120,6 +123,36 @@ class MultipleYieldTagsError(CopierError):
     """Multiple yield tags are used in one path name, but it is not allowed."""
 
 
+class TaskError(CopierError):
```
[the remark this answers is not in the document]
sisp: Good idea! 👌

tsvikas:
> Thanks for your contribution, @tsvikas! 🙏
> 
> I have two minor suggestions (see inline comment).
> 
> IIUC, a failing task will raise the `TaskError` exception and the message will be shown, but the error from the task command will be hidden when running Copier via CLI, as it's not printed anymore but only accessible via the `stdout`/`stderr` attributes of the exception object. I think we should catch this exception in
> 
Actually, the `stdout`/`stderr` will be None, since we did not capture them in `subprocess.run()`, and the output of the process still goes to the screen.

So we only need to output the message itself (`"Task {command!r} returned non-zero exit status {returncode}."`).
I believe that inheriting from `UserMessageError` will take care of that nicely?

tsvikas:
I did all of the requested changes, except inheriting from `CalledProcessError` (see my comment above) - waiting for more input for that.
Hopefully i didn't do too much mess in the comments/review here 🙏

tsvikas:
Done :)

copier/errors.py
```diff
@@ -1,7 +1,11 @@
 """Custom exceptions used by Copier."""
 
+from __future__ import annotations
```
sisp, line 3: I'd prefer removing this line and reverting the typing changes accordingly, and address this change in a separate PR to avoid mixing concerns here.

copier/errors.py
```diff
@@ -120,6 +130,37 @@ class MultipleYieldTagsError(CopierError):
     """Multiple yield tags are used in one path name, but it is not allowed."""
 
 
+class TaskError(UserMessageError, subprocess.CalledProcessError):
```
sisp, line 133: Just semantics, but I'd prefer changing the order of the base classes because semantically `TaskError` is rather a subclass of `CalledProcessError`.

sisp (changes requested):
@tsvikas I have a few minor remarks.

@copier-org/maintainers What is the correct use of `UserMessageError`? When I check its occurrences, it seems this error is directly raised when it originates from an improper use of Copier that can be fixed by the _user_. One counter-example I've seen is a missing `--unsafe`/`--trust` flag which raises `UnsafeTemplateError` (which isn't a subclass of `UserMessageError`), although this problem can be resolved by a Copier user by adding the flag. What about a task error? Is it an error caused mainly by a template author which can't be fixed by a user? Is subclassing `UserMessageError` appropriate in this case? WDYT?

tsvikas:
> @copier-org/maintainers What is the correct use of `UserMessageError`? When I check its occurrences, it seems this error is directly raised when it originates from an improper use of Copier that can be fixed by the _user_. One counter-example I've seen is a missing `--unsafe`/`--trust` flag which raises `UnsafeTemplateError` (which isn't a subclass of `UserMessageError`), although this problem can be resolved by a Copier user by adding the flag. What about a task error? Is it an error caused mainly by a template author which can't be fixed by a user? Is subclassing `UserMessageError` appropriate in this case? WDYT?

personally i assumed that `UserMessageError` is for a ny exception that should be reported to the user in a message. 
if it is not the case, i will be happy for input about the way the @copier-org/maintainers prefer to use to report something to the user. is it the `printf_exception` function? something else? i don't want to add here a 2nd/3rd way to do something simple.

tsvikas:
Hey @sisp , I saw you pinged the maintainers earlier regarding the `UserMessageError`—do you think it would be helpful to re-ping them, or should we proceed with an alternative approach?
Just want to make sure we align with the project’s conventions before finalizing this. Thanks!

pawamoy:
> What is the correct use of UserMessageError?

I would just be guessing so I'll leave that to @yajo :bow:

sisp:
I'd like to wait for @yajo's opinion, as he is the lead maintainer with a long history developing Copier. All of us are working on Copier as time permits. Let's just wait for him to find time to share his thoughts.

yajo:
IIRC, `UserMessageError` is an error that should send a message to the user, just like the name says. It should not contain a traceback, because that would be confusing.

sisp:
Just to clarify your comment, @yajo: Is it a valid pattern to let _any_ exception (like `TaskError`) inherit from `UserMessageError` that shall print an error message to the _user_ (i.e., errors that can be addressed by a template _user_, not author)? If so, should `UnsafeTemplateError` also inherit from `UserMessageError`? If yes, then we'd need to catch specific subclasses of `UserMessageError` such as `UnsafeTemplateError` in the CLI's exception handler first before catching `UserMessageError` to support exit codes other than `1` for some errors (like `UnsafeTemplateError` for which the CLI exits with code `2`).

I guess the underlying question is: What is the meaning of `UserMessageError` and when to subclass it vs. just subclassing `CopierError`  or other child classes? My confusion originates from an attempt to decouple API and CLI, such that (IMHO) exceptions shouldn't be designed with CLI-related error management in mind; CLI error handling should rather be an afterthought, as a CLI is merely one possible alternative interface to the Python API with its own peculiarities of managing errors.

sisp:
@tsvikas Sorry for the delay. I've thought about the use of `UserMessageError` some more and believe the way you're using it here makes sense. :+1: Just one more thought: The new error handling for CLI users swallows all error information from the task process, which may make it difficult to understand the reason for the error. How about printing also `stderr` in addition to the generic message?

tsvikas:
Thank you

I've implemented printing of `stderr` in `TaskError`. As a reminder, this won’t affect the current case, since the task execution runs without capturing, so both `stdout `and `stderr `are `None `(and were already printed to the screen). But it could be useful for future usages of `TaskError`.

tsvikas:
it seems that the codecov is complaining about those lines being untested, which is true, because any task used currently will not capture the stderr -- this is only a future facing change.
I think that in order to test it we will need to write code that uses `TaskError` in a different way than in the codebase itself, which seems too artificial for a test.
Alternatively we can add ignore for codecov, or revert this change, or revert this change and `assert stderr is None` to prevent wrong usage of `TaskError`.

what do you think we should do?

sisp:
Ah, I missed that `stderr` is always `None` and both `stdout` and `stderr` are printed unconditionally at the moment. Eventually, I think it we should handle task error printing like pre-commit:

* When no error occurs, nothing is printed by default. Printing captured `stdout`/`stderr` is opt-in via the `--verbose` switch. In Copier, `--quiet` and a new `--verbose` switch would be mutually exclusive.
* When an error occurs, captured `stdout`/`stderr` is printed by default. In Copier, we would suppress it when our `--quiet` switch is used.

For our Python API users, `TaskError`'s `stdout`/`stderr` attributes would contain the captured output of a failed task for further processing.

That said, I suggest to revert db1a6dc1106c2f05c4b784dce1ceb54f12123101 to keep this PR focused on introducing the `TaskError` exception and address further enhancements in subsequent PRs. Sorry for the confusion.

tsvikas:
done :)

sisp (approved):
Fantastic, thanks for the great work, @tsvikas! :pray:

PR #2000, opened 2025-03-02
Reduce the number of test warnings

noirbizarre:
This PR reduces the number of raised warning during tests:
- from 82 to 2 warnings on Python <= 3.11
- from 84 to 4 warnings on Python >= 3.12

To reduce those warnings, this PR:
- adds missing `pytest.warns()` context manager for expected warnings
- adds missing `pytest.deprecated_call()` on expected deprecation warnings
- migrate all tests containing migrations to the new migration format

Remaining warnings:
- 2 in `test_unsafe.py` on a remaining legacy migration test case (case only exists with legacy format)
- 2 `datetime.utcnow()` on Python 3.12+, tied to the deprecated built-in `now()`, adding a compatibility layer was not worth it given it will be removed

Benefits:
- tests expected warnings
- tests are up to date with the current API
- `pytest` output is uncluttered

This PR does not contain any functional change

### DIFF
--- a/tests/test_legacy_migration.py
+++ b/tests/test_legacy_migration.py
@@ -66,9 +66,14 @@ def test_migrations_and_tasks(tmp_path: Path, skip_tasks: bool) -> None:
         git("add", ".")
         git("commit", "-m1")
     # Update it to v2
-    run_update(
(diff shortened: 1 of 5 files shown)


sisp (approved):
Fantastic work, @noirbizarre! :trophy: Thanks! :pray:

PR #2002, opened 2025-03-03
fix(settings): auto-detect encoding when reading settings file

sisp:
I've fixed loading the settings file to auto-detect the encoding instead of using the system's default encoding.

Without the fix, the added test fails with this error:

```
    def read_text(self, encoding=None, errors=None):
        """
        Open the file in text mode, read it, and close the file.
        """
        encoding = io.text_encoding(encoding)
        with self.open(mode='r', encoding=encoding, errors=errors) as f:
>           return f.read()
E           UnicodeDecodeError: 'cp932' codec can't decode byte 0x98 in position 19: illegal multibyte sequence
```

Inspired by #1998.

/cc @noirbizarre – Just FYI, as you implemented the settings feature.


sisp:
CI is failing with this error:

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v3.0.11`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down
```

See https://github.com/JRMurr/direnv-nix-action/issues/12.

pawamoy (approved):

noirbizarre:
Seems good to me.
To be honest, I never had to handle anything else than `utf-8` for years, it seems mostly a Windows only issue.
But TIL that `yaml.safe_load()` is encoding type detection, good to know 👍🏼

PR #2003, opened 2025-03-03
fix: auto-detect encoding when reading external data file

sisp:
I've fixed loading external data files to auto-detect the encoding instead of using the system's default encoding.

Without the fix, the added test fails with this error:

```
self = <yaml.loader.SafeLoader object at 0x7f20921b6e70>, size = 4096

    def update_raw(self, size=4096):
>       data = self.stream.read(size)
E       UnicodeDecodeError: 'cp932' codec can't decode byte 0x98 in position 7: illegal multibyte sequence
```

Follow-up of #1973. Inspired by #1998.

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -522,7 +522,7 @@ def load_answersfile_data(
 ) -> AnyByStrDict:
     """Load answers data from a `$dst_path/$answers_file` file if it exists."""
     try:
-        with Path(dst_path, answers_file).open(encoding="utf-8") as fd:
+        with Path(dst_path, answers_file).open("rb") as fd:
             return yaml.safe_load(fd)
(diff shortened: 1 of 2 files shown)


sisp:
CI is failing with this error:

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v3.0.11`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down
```

See https://github.com/JRMurr/direnv-nix-action/issues/12.

pawamoy (approved):

pawamoy:
Nice one 👌

PR #2016, opened 2025-03-07
test: use helper function to load answers file data

sisp:
I've improved the test suite to consistently use our helper function `copier.user_data.load_answersfile_data()` for loading answers. There are only a few exceptions where the answers file was written and read as JSON; changing those to YAML would slightly alter the semantics of the test.


pawamoy (approved):
Nice :ok_hand:

PR #2027, opened 2025-03-11
docs: Correct typo (broken _external_data example)

cbrown1234:
I was getting rather confused as to why the examples in the docs for `_external_data` were not working, after checking the tests it seems there was a typo


sisp (approved):
Good catch! 👌 Thanks for fixing the docs! :pray:

PR #2032, opened 2025-03-13
feat: raise `InteractiveSessionError` when prompting in non-interactive environment

### LINKED ISSUE #1988: Better error message for EOFError exception
### Actual Situation

Thanks for copier!  Keep up the good work.

Related to #666.  I tried to run copier in docker and got the same `EOFError` exception. 

### Desired Situation

I'd like to see a better error message that tells the user what is wrong and how to fix it.

### Proposed solution

Yes, it is a user error now that I know what is wrong.  But, a better error message would have saved me a bit of searching and be more user friendly:

"Input required but no terminal input available.  Use "--defaults" and optionally "--data" as needed to run copier in this environment."

sisp:
I've improved the error handling when prompting in a non-interactive environment, which was failing with `EOFError`. Now, the new `InteractiveSessionError` exception is raised.

Resolves #1988.

### DIFF
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import platform
+import subprocess
 import sys
 from pathlib import Path
 from typing import Any, Dict, List, Mapping, Protocol, Union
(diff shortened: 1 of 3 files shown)


sisp:
Looks like this needs some more thought ... :thinking:

pawamoy:
Ouch! Are you trying to draw something in ASCII?

asciicopier

tests/test_prompt.py
```diff
@@ -997,3 +998,37 @@ def test_secret_validator(
...
+    """Answering a question interactively requires a TTY."""
+    src, dst = question_tree(type="str")
+    process = subprocess.run(
+        (*COPIER_PATH, "copy", str(src), str(dst)),
+        stdin=subprocess.PIPE,  # Prevents interactive input
+        capture_output=True,
+        timeout=10,
+    )
+    assert process.returncode == 1
+    assert b"Input required but no TTY available" in process.stderr
```
rsyring, line 1013: Test the full error message?

sisp: Done :+1:

sisp:
@pawamoy No ASCII art anymore. :smile:

To have consistent error handling of prompts, I changed implementation of the confirmation prompt for overriding a file to use `questionary` instead of `plumbum.cli.terminal`, which is more consistent anyway, as we use `questionary` for question prompts. It seems that not only `pexpect` is having issues with `prompt-toolkit` on Windows but also `subprocess.run()`, so I marked the tests as expected to fail on Windows.

pawamoy (approved):
LGTM!

PR #2034, opened 2025-03-13
Improve instruction for multiselection variable

### LINKED ISSUE #2033: Multiselection hints are too obscure
### Actual Situation

Image

- `<space> to select` - select what?
- `<a> to toggle` - toggle what?
- `<i> to invert` - invert what?

Initially, I thought that all these actions refer to **a (one) selection**. But in this case, all these actions would do the same thing.

Only after trying and experimenting with a toy template I understood everything.

### Desired Situation

Let's address this bit by bit.

What actions do:

- `<space> to select` - toggles a single option under the arrow.
- `<a> to toggle` - either:
    - Selects every option.
    - Deselects every option.
- `<i> to invert` - inverts current assignments. What was selected become deselected, what was deselected became selected.

Scopes:

- `<space> to select` - single choice.
- `<a> to toggle` - all choices.
- `<i> to invert` - all choices.

### Proposed solution

Maybe rewrite it like this:

- `<space> to select` - `<space> toggle`
- `<a> to toggle` - `<a> select all`
- `<i> to invert` - `<i> invert selection`

If we came to consensus, I can make a PR

InAnYan:
Closes #2033.

The instruction was co-authored with @pawamoy.

Currently, I have two issues with this code:

1. Instruction is too big.
2. Is it okay to leave the instruction like this? Like a literal.

*And yes, the instruction should be wrapped in parenthesis and shouldn't include a preceding space.*

### DIFF
--- a/copier/user_data.py
+++ b/copier/user_data.py
@@ -406,7 +406,9 @@ class Question:
         if self.choices:
             if self.multiselect:
                 questionary_type = "checkbox"
-                result["instruction"] = "(Use arrow keys to move, <space> (un)select, <a> (un)select all, <i> invert selection, <enter> confirm)"
+                result["instruction"] = (
+                    "(Use arrow keys to move, <space> (un)select, <a> (un)select all, <i> invert selection, <enter> confirm)"
+                )
             else:
                 questionary_type = "select"
             result["choices"] = self._formatted_choices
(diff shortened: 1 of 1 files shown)


yajo:
Please could you post a screenshot of how it looks now?

InAnYan:
image

sisp:
Sorry for the delay. :face_with_peeking_eye: I don't mind a new line for the instructions. Just an alternative idea for the arrow keys: How about using Unicode arrow characters, i.e. "↓↑ keys" instead of "arrow keys"?

Issue #918, opened 2023-01-14
Formalize copier.yml schema (or use JSON schema)


bollwyvl:
**Is your feature request related to a problem? Please describe.**

I'd like to be able to validate a `copier.yml` without loading it in copier.

**Describe the solution you'd like**

Create a normalized, versioned JSON schema, checked into the repo, and available for tools. It could be authored in yaml, but having it checked in as actual JSON, somewhere, is important for downstreams.

Use this as an up-front validator when parsing a file.

**Describe alternatives you've considered**

Optionally, allow opting a `copier.yml` into being a JSON Schema itself, as there is already robust tooling for this in many languages. A number of features (`oneOf`, `default`) are already handled by JSON schema. Templating (though conditionals schema, `if`, and other things are possible).

**Additional context**

sisp:
I've actually been thinking about something similar a while ago. I was looking for IDE autocompletion for `copier.yml` in VS Code which can be done via JSON schema.

Copier uses [Pydantic](https://github.com/pydantic/pydantic) in some places although (I think) a bit inconsistently. I've quietly been working on refactoring the [`user_data.Question`](https://github.com/copier-org/copier/blob/f8ae8d6e8ed8ec5333abe4b20ad53bc87123bff0/copier/user_data.py#L142-L414) class into several Pydantic models to get a proper/detailed domain model of the `copier.yml` content (which may still contain Jinja templates) as well as the rendered content. The Pydantic model of the `copier.yml` file could then be used to validated the file upon loading it into Copier, and it can also be used to generate a [JSON schema](https://docs.pydantic.dev/usage/schema/), making it the single source of truth.

What I've been working on doesn't work yet although I've come quite far already. When I've made enough progress, I'll send a PR, but I can't provide a timeline.

bollwyvl:
Yeah, pydantic is... fine. But indeed, the language servers definitely like proper schema, so actually shipping and using a versioned schema at runtime is useful in its own right. 

But, having schema exist external to Pydantic has certain benefits, such as being able to use [hypothesis-jsonschema](https://github.com/python-jsonschema/hypothesis-jsonschema) to _test_ the pydantic implementation. And pydantic has some... quirks in its generation of schema, for sure. 

Having been down the "iteratively apply schema for templated yaml" (and/or, the much worse YAML tagged/magic comment): there's not a lot you can do but intercept right after `(ruamel_)yaml.(safe_)load`.

[...]

PR #2039, opened 2025-03-14
refactor!: replace `jinja2-ansible-filters` by `jinja2-copier-extension`

### LINKED ISSUE #1398: pyyaml-include is GPL3, doesn't that poison your MIT project?

sisp:
I've replace the `jinja2-ansible-filters` dependency with our own reimplementation ([`jinja2-copier-extension`](https://github.com/copier-org/jinja2-copier-extension)) of most of the Jinja2 filters by relying on the Ansible filters documentation as much as possible to avoid copying/rewriting GPL-licensed code. The tests are based on a combination of Ansible filters documentation and some thinking of my own.

According to https://github.com/copier-org/jinja2-copier-extension/pull/1, which I've released as [v0.1.0](https://github.com/copier-org/jinja2-copier-extension/releases/tag/v0.1.0):

> This is the set of re-implemented filters from `jinja2-ansible-filters`:
> 
> - [x] `ans_groupby`
> - [x] `ans_random`
> - [x] `b64decode`
> - [x] `b64encode`
> - [x] `basename`
> - [x] `bool`
> - [x] `checksum`
> - [ ] `comment` – This filter seems a bit involved, and I'm not sure how relevant it is for Copier.
> - [x] `dirname`
> - [x] `expanduser`
> - [x] `expandvars`
> - [x] `extract`
> - [x] `fileglob`
> - [x] `flatten`
> - [x] `from_json`
> - [x] `from_yaml`
> - [x] `from_yaml_all`
> - [x] `hash`
> - [x] `mandatory`
> - [x] `md5`
> - [x] `quote`
> - [x] `random_mac`
> - [x] `realpath`
> - [x] `regex_escape`
> - [x] `regex_findall`
> - [x] `regex_replace`
> - [x] `regex_search`
> - [x] `relpath`
> - [x] `sha1`
> - [x] `shuffle`
> - [x] `splitext`
> - [x] `strftime`
> - [ ] `subelements` – This filter seems a bit involved, and I'm not sure how relevant it is for Copier.
> - [x] `ternary`
> - [x] `to_datetime`
> - [x] `to_json`
> - [x] `to_nice_json`
> - [x] `to_nice_yaml`
> - [x] `to_uuid`
> - [x] `to_yaml`
> - [x] `type_debug`
> - [x] `win_basename`
> - [x] `win_dirname`
> - [x] `win_splitdrive`
> 
> At this point, 2 filters are missing.

I'm marking this PR as a draft for now because it's a breaking change at least because two filters from `jinja2-ansible-filters` are missing, potentially also because the reimplementation might not behave 100% identically (although I tried to be thorough). Thus, after merging this PR, we'd need to make a major release. But to reduce the number of major releases (and breaking changes for our users), I suggest collecting a few more PRs with breaking changes (e.g., [revising Copier's public API](https://github.com/orgs/copier-org/discussions/1250), possibly [modeling `copier.yaml` content with Pydantic](https://github.com/copier-org/copier/issues/918#issuecomment-1400554449), perhaps more). I'll be happy to keep them up-to-date until we're ready for batch merging and making the major release.

The way I see it, `jinja2-copier-extension` (actually, no different to `jinja2-ansible-filters`) is essentially factored out code from Copier, as the registered Jinja2 filters (and their exact behavior) are directly exposed as template features of Copier. Hence, I believe we should pin the version of `jinja2-copier-extension` in our dependency manifest to have control over the exposed Jinja2 filters and their exact behavior. When a new release of `jinja2-copier-extension` becomes available, its level of change may affect also Copier's level of change, e.g. a major release may require a major Copier release if a breaking change in `jinja2-copier-extension` is also breaking for Copier's users (e.g., when it cannot or should not be compensated by Copier). In addition, I suggest to include updates of `jinja2-copier-extension` version pins (and perhaps even links to its release logs) in our changelog. For this reason, I've added a package rule to our Renovate config that overrides the semantic commit type for `jinja2-copier-extension` updates to `fix`. I've drawn some inspiration for this approach from [Pydantic's dependency on `pydantic-core`](https://github.com/pydantic/pydantic/blob/b9fb3f110d6653653fb48a44430e7d2fca886ae9/pyproject.toml#L49).

(That said, I'm wondering whether we should also pin the `jinja2` package (in a different PR, of course), as its behavior is directly exposed to Copier template authors. :thinking:)

Sorry for the long delay since submitting #1809.

Supersedes #1809.
Fixes #1398.


pyproject.toml
```diff
@@ -32,7 +32,7 @@ colorama = ">=0.4.6"
 dunamai = ">=1.7.0"
 funcy = ">=1.17"
 jinja2 = ">=3.1.5"
-jinja2-ansible-filters = ">=1.3.1"
+jinja2-copier-extension = "0.1.0"
```
yajo, line 35: IMHO we should put lower bounds only.

This way we leave downstream packagers to do their work in a more easy way. On our side, we still supply the nix packages for whoever wants officially supported pinnings. Or they can reuse our poetry.lock in any other way.

yajo (approved):
Apart from that comment, I think the improvement is cool. I don't mind doing 1 or 10 major releases. They're free 🤷🏼‍♂️ . We can become the next google chrome, no problem 😆

pyproject.toml
sisp: Hm, but we don't test the behavior of Jinja filters from `jinja2-copier-extension` in Copier, so if we change a filter in an incompatible way without pinning in our manifest, the identical Copier version will behave differently when installed before and after that release of `jinja2-copier-extension`.

Concretely, I'm planning to revise some API inconsistencies, that I retained from `jinja2-ansible-filters` for backwards compatibility, in `jinja2-copier-extension` in a future release. For example:

```diff
 def do_regex_findall(
     string: str,
-    # TODO(sisp): Rename argument to `pattern` for consistency with other filters.
-    regex: str,
+    pattern: str,
-    # TODO(sisp): Require flags to be keyword-only arguments.
+    *,
     multiline: bool = False,  # noqa: FBT001 FBT002
     ignorecase: bool = False,  # noqa: FBT001 FBT002
 ) -> list[str] | list[tuple[str, ...]]:
```

Renaming `regex` to `pattern` _may_ be a breaking change in case somebody passes the pattern as a keyword argument. And making `multiline` and `ignorecase` keyword-only arguments _may_ be breaking in case somebody passes those arguments as positional arguments. If we release these changes as v0.2.0 and use a lower bound in our `pyproject.toml`

```toml
jinja2-copier-extension = ">=0.1.0"
```

then next time somebody re-installs Copier (e.g., even just `pipx run copier` with an expired package cache suffices) a previously working Copier template may suddenly fail to work.

Isn't the difference in this case that the public API (in the sense of registered filters in a Jinja environment that loads the `jinja2-copier-extension` extension) is also Copier's public API (in the same sense), whereas the public API of other dependencies is used merely internally in Copier?

mibelbahri:
UP

sisp:
Thanks for keeping this on my radar, @mibelbahri. :+1: I'm working towards a major release with several (potentially) breaking changes, this PR will be included. I'm trying to get there ASAP. :crossed_fingers:

PR #2067, opened 2025-04-02
fix(updating): ignore last answer of `when: false` questions

### LINKED ISSUE #2057: When running `copier update`, responses to questions with "when==false" retain the last answers instead of reverting to the defaults.
### Describe the problem

Hi,

In the [documentation](https://copier.readthedocs.io/en/stable/configuring/#advanced-prompt-formatting), it states:
> - when: Condition that, if false, skips the question. ...  
> If a question is skipped, its answer is not recorded, but its default value is available in the render context.

However, when `copier update` is executed, the answers to skipped questions persist as their last values instead of updating to the default values.

### Template

https://github.com/kubotty/copier-default-debug

### To Reproduce

1. Apply the template:
```bash
mkdir copier-default-check
cd copier-default-check
git init
copier copy https://github.com/kubotty/copier-default-debug.git -r 0.4.0 .
```
```
🎤 First question
Yes
🎤 Second question
Yes

Copying from template version 0.4.0
    create  file.md
    create  README.md
    create  .copier-answers.yml
```

Check the content of `file.md`:
```bash
cat file.md
```
```
- True
- True
- Yes
- Yes
```

2. Update and change the answer, but it retains the last answer:
```bash
git add .
git commit -m v0.4.0
copier update -r 0.5.0
```
```
Updating to template version 0.5.0
🎤 First question
No
```
```bash
cat file.md
```
```
- False
- True
- No
- Yes
```

### Logs

```console

```

### Expected behavior

The default values should be reflected in the updates:
```bash
copier update -r 0.5.0
    Updating to template version 0.5.0
    🎤 First question
    No
```
After checking the content of `file.md`, it should show:
```bash
cat file.md
```
Expected Output:
```
- False
- False
- No
- No
```

_No response_

### Operating system

Windows

Windows 11 Enterprize

### Copier version

9.6.0

### Python version

3.12

### Installation method

pipx+pypi

### Additional context

_No response_

sisp:
I've fixed a bug that caused an update or recopy involving the switch of a question's `when` value from `when: true` to `when: false` to reuse the recorded answer, specifically for computed values the default value wasn't used to re-compute the answer. 

To fix it, we now delete the last answer of a `when: false` question.

Fixes #2057.

### DIFF
--- a/copier/main.py
+++ b/copier/main.py
@@ -563,6 +563,10 @@ class Worker:
             if not question.get_when():
                 # Omit its answer from the answers file.
                 self.answers.hide(var_name)
+                # Delete last answers to re-compute the answer from the default
+                # value (if it exists).
+                if var_name in self.answers.last:
+                    del self.answers.last[var_name]
                 # Skip immediately to the next question when it has no default
                 # value.
                 if question.default is MISSING:
--- a/tests/test_recopy.py
+++ b/tests/test_recopy.py
@@ -130,3 +130,50 @@ def test_recopy_dont_validate_computed_value(
     run_recopy(dst, overwrite=True)
     answers = load_answersfile_data(dst)
     assert "computed" not in answers
+
+
+def test_conditional_computed_value(tmp_path_factory: pytest.TempPathFactory) -> None:
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    build_file_tree(
+        {
+            src / "copier.yml": (
+                """\
+                first:
+                    type: bool
+
+                second:
+                    type: bool
+                    default: "{{ first }}"
+                    when: "{{ first }}"
+                """
+            ),
+            src / "{{ _copier_conf.answers_file }}.jinja": (
+                "{{ _copier_answers|to_nice_yaml }}"
+            ),
+            src / "log.txt.jinja": "{{ first }} {{ second }}",
+        }
+    )
+    git_save(src)
+
+    run_copy(str(src), dst, data={"first": True}, defaults=True)
+    answers = load_answersfile_data(dst)
+    assert answers["first"] is True
+    assert answers["second"] is True
+    assert (dst / "log.txt").read_text() == "True True"
+
+    git_save(dst, "v1")
+
+    run_recopy(dst, data={"first": False}, overwrite=True)
+    answers = load_answersfile_data(dst)
+    assert answers["first"] is False
+    assert "second" not in answers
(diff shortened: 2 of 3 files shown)


pawamoy (commented):
Looking good. At this point though I wonder if this isn't a breaking change for some users who expect the current behavior :thinking:

sisp:
Perhaps another argument towards considering it a fix: When I run `copier update` a second time, then the `when: false` answer gets re-computed because the answers file does not contain the answer anymore. This is bad because `copier update` without changing answers or using a different template version should be a no-op.

pawamoy:
Re-reading #1574, which sounds like yet another request to **control** what users can do (a thing I generally disagree with, even in the name of UX), I'd lean towards considering this PR a necessity, yes.

Your last comment confuses me though. Can we describe current vs. new behavior again?

### Current behavior

False then false again:

- A question gets `when: false`
- Its value is computed thanks to the default
- Answer is not recorded (good IIUC)
- Answer is therefore recomputed on update (good IIUC)

False then true:

- A question gets `when: false`
- Its value is computed thanks to the default
- Answer is not recorded (good IIUC)
- Question gets `when: true` on update
- Answer is provided and recorded (good IIUC)

True then true again:

- Question asked first time
- Answer recorded (good IIUC)
- Question asked again on update
- Answer updated (or not) and recorded (good IIUC)

True then false:

- Question asked first time
- Answer recorded (good IIUC)
- Question not asked on update
- Last answer re-used (**not good** IIUC)

### New behavior

Only the "true then false" case changes, to:

- Question asked first time
- Answer recorded (good IIUC)
- Question not asked on update
- Last answer deleted, value recomputed from default (**good** IIUC)

Your last comment seems to imply that the "false then false again" case is wrong, because unanswered questions will get a re-computed value (from the default) each time the project is updated. But this PR doesn't seem to change that :thinking:

pawamoy:
> Sorry for the confusion

No no, no worries here! Just wanted to make sure I understand :smile: 

> the update algorithm implementation is not easy enough

You're damn right! It's far from easy. That is exactly why I had to lay down the  different cases :blush: 

> So, should we treat this PR as a bug fix rather than a breaking change (of previously unintended behavior)? WDYT?

Yes, I  believe we should!

sisp:
Great, then, if you hit approve, we can get this merged. :tada:

pawamoy:
Oh, right, sorry😅

pawamoy (approved):

PR #2069, opened 2025-04-03
refactor: deprecate access to non-public API

sisp:
I've made a step towards clarifying Copier's public API (#1250) by renaming internal modules with a leading `_` and re-exposing the API with deprecation warnings on non-public API import. This way, the current API surface remains unchanged, but imports of any internal-only modules and members will raise a deprecation warning. For example:

```pycon
>>> from copier.main import Worker
<stdin>:1: DeprecationWarning: Importing from `copier.main` is deprecated. This module is intended for internal use only and will become inaccessible in the future. If you have any questions or concerns, please raise an issue at <https://github.com/copier-org/copier/issues>.
<stdin>:1: DeprecationWarning: Importing `Worker` from `copier.main` is deprecated. This module member is intended for internal use only and will become inaccessible in the future. If you have any questions or concerns, please raise an issue at <https://github.com/copier-org/copier/issues>.
```

I've split this refactoring into two commits where the first commit renames modules and the second commit re-exposes the API with deprecation warnings. The reason for this split is that Git doesn't seem to detect file renames in this case, so a combined commit is impossible to review. Thus, please review the commits individually.

The suggested public API is this:

```python
from copier import run_copy, run_recopy, run_update
from copier.errors import ... # all exceptions
from copier.settings import Settings # needed for passing settings to `run_{copy,recopy,update}(settings=...)`
```

WDYT?


pawamoy:
Yes! That's exactly what I'm doing for all my existing projects. Well, instead of renaming modules to `_module`, I actually move them under an `_internal` folder, but the result is the same. Completely in favor of this. This helps tremendously in the long run when we have to move objects around.

copier/_deprecation.py
```diff
@@ -0,0 +1,57 @@
...
+        f"Importing `{member}` from `{module}` is deprecated. This module member is "
+        "intended for internal use only and will become inaccessible in the future. "
+        f"{_issue_note}",
+        DeprecationWarning,
+        stacklevel=3,
+    )
+
+
+def deprecate_member(member: str, module: str, new_import: str) -> None:
+    """Deprecate a module member with a new import statement as with a warning.
```
pawamoy, line 44:
```suggestion
    """Deprecate a module member with a new import statement with a warning.
```

pawamoy:
In my own deprecation messages I generally add "import from <there> instead", so that it's clear for users how they should rewrite their imports.

Also in my own projects I go full ["hidden module layout"](https://mkdocstrings.github.io/griffe/guide/users/recommendations/public-apis/#module-layout) but I understand it might not be desirable for Copier. I'm fine with continuing to expose a few submodules as public API :slightly_smiling_face:

copier/_deprecation.py
sisp: Kudos for your sharp eye! :clap: :trophy: Thanks! :pray:

sisp:
> Yes! That's exactly what I'm doing for all my existing projects.

I had seen the module-level `__getattr__()` trick for these kinds of deprecations and layout changes, but to be honest, I took your recent public API cleanup in `mkdocstrings` as inspiration. :wink:

> Also in my own projects I go full ["hidden module layout"](https://mkdocstrings.github.io/griffe/guide/users/recommendations/public-apis/#module-layout) [...]

Thanks for sharing this article, this is a really amazing summary of what "public API" means and how to manage it in Python! :star::star::star:

Two remarks about the current approach:

1. I noticed that something like

    ```python
    if TYPE_CHECKING:
        from copier._main import *  # noqa: F403
    ```

    seems necessary to retain static types when importing from deprecated modules. I didn't see this snippet neither in Griffe's public API guide nor in your recent public API cleanup in `mkdocstrings`.

1. I was experimenting with applying the [`@warnings.deprecated`](https://docs.python.org/3/library/warnings.html#warnings.deprecated) decorator to internal module members, as VS Code (via Pylance) highlights such deprecations with strikethrough rendering, e.g., <code>from copier.main import <del>Worker</del></code>. But I haven't been able to get this to work for imported/re-exported members, e.g. in `copier/main.py`:

    ```python
    from typing import TYPE_CHECKING

    if TYPE_CHECKING:
        from typing_extensions import deprecated

        # Idea #1
        from copier._main import Worker

        Worker = deprecated("Internal use only")(Worker)


        # Idea #2
        from copier._main import Worker as _Worker

        Worker = deprecated("Internal use only")(_Worker)
    ```

    None of the ideas are working for different reasons. I think idea 2 should actually work, but Pylance treats the `Worker` alias as a variable and not a function, and I suspect that strikethrough rendering is only shown for deprecated functions/classes but not for variables (perhaps because deprecated variables aren't well defined?).

    If you happen to have another idea, I'm curious whether this could be achieved somehow. But perhaps I'm overdoing it, runtime warnings are likely sufficient anyway.

pawamoy:
Thanks for the kind words :heart: 

1. Ah, good point. I didn't notice that when type-checking, but it makes sense :thinking: Seems like a good idea to do this type-checking wildcard import, yes.
2. Yeah `warnings.deprecated` is quite limited, like many of the other third-party libraries than offer to handle deprecations. I actually have a blog post in the works, doing an exhaustive survey of all these libs and their shortcomings. But anyway: you won't be able to decorate attributes with `warnings.deprecated` indeed, it's only meant for classes and functions/methods. Which means you can't statically mark an **object location** as deprecated (something I didn't think about before! thank you so much for bringing this use-case to my attention, I'll include that in my blog post :smile:)

pawamoy:
I guess if you really wanted IDEs to strike through deprecated locations, you'd have to wrap the object into a proxy of some kind:

```python
# in copier.main
import warnings
from copier._main import Worker as BaseWorker


@warnings.deprecated("...")
class Worker(BaseWorker):
    """Deprecated. Use `copier.Worker` instead."""
```

pawamoy:
Yes, you'd have to either duplicate the signature, or declare the function dynamically, playing the type system, but then you'd have to assign to an attribute again, defeating the decorator.

```python
def replicate_signature(func: Callable[[param spec stuff], R]) -> Callable[[param spec stuff], R]:
    @warnings.deprecated("...")
    def replicated(...): ...
    return replicated

# decorator defeated
function = replicate_function(original_function)
```

sisp:
Thanks for sharing your thoughts! :pray: Well, I guess runtime warnings are enough then. :smile:

sisp:
It's my understanding that @pawamoy agrees with this PR although there's no formal approval. I'll merge this PR, so the next release will raise deprecation warnings.

sisp:
Yes, via private module imports, but we won't guarantee stability. Which symbols are you relying on?

Issue #2070, opened 2025-04-04
Windows11: Location of settings.yml incorrect


roteiro:
### Describe the problem

OS: Win11
Reading https://copier.readthedocs.io/en/stable/settings/ I've put my User defaults under 
`%USERPROFILE%\AppData\Local\copier\settings.yml`
However those defaults were ignored until i moved that file to 
`%USERPROFILE%\AppData\Local\copier\copier\settings.yml`

### Template

-

### Logs

```console

```

### Expected behavior

-

### Operating system

Win11

### Operating system distribution and version

Win11

### Copier version

copier 9.6.0

### Python version

CPython 3.13.2

### Installation method

uv tool install

pawamoy:
Thanks for the report and PR @roteiro! Makes sense, and looking good.

PR #2071, opened 2025-04-04
fix: settings dir with deduplicated copier subdir

roteiro:
platformdirs implicitely sets appauthor to appname if not explicitely set to false, resulting in an dir with <CONFIG_ROOT>/copier/copier

Closes https://github.com/copier-org/copier/issues/2070

### DIFF
--- a/copier/settings.py
+++ b/copier/settings.py
@@ -35,7 +35,20 @@ class Settings(BaseModel):
             if env_path:
                 settings_path = Path(env_path)
             else:
-                settings_path = user_config_path(appname="copier", appauthor=False) / "settings.yml"
+                settings_path = (
+                    user_config_path("copier", appauthor=False) / "settings.yml"
+                )
+                # NOTE: Remove after a sufficiently long deprecation period.
+                if OS == "windows":
+                    old_settings_path = user_config_path("copier") / "settings.yml"
+                    if old_settings_path.is_file():
+                        warnings.warn(
+                            f"Settings path {old_settings_path} is deprecated. "
+                            f"Please migrate to {settings_path}.",
+                            DeprecationWarning,
+                            stacklevel=2,
+                        )
+                        settings_path = old_settings_path
         if settings_path.is_file():
             data = yaml.safe_load(settings_path.read_bytes())
             return cls.model_validate(data)
--- a/copier/settings.py
+++ b/copier/settings.py
@@ -28,6 +28,10 @@ class Settings(BaseModel):
         default_factory=set, description="List of trusted repositories or prefixes"
     )
 
+    def _default_settings_path() -> Path:
+        return user_config_path("copier", appauthor=False) / "settings.yml"
+
+
     @classmethod
     def from_file(cls, settings_path: Path | None = None) -> Settings:
         """Load settings from a file."""
@@ -36,8 +40,7 @@ class Settings(BaseModel):
             if env_path:
                 settings_path = Path(env_path)
             else:
-                settings_path = (
-                    user_config_path("copier", appauthor=False) / "settings.yml"
+                settings_path = _default_settings_path()
                 )
                 # NOTE: Remove after a sufficiently long deprecation period.
                 if OS == "windows":
--- a/copier/settings.py
+++ b/copier/settings.py
@@ -31,7 +31,6 @@ class Settings(BaseModel):
     def _default_settings_path() -> Path:
         return user_config_path("copier", appauthor=False) / "settings.yml"
 
-
     @classmethod
     def from_file(cls, settings_path: Path | None = None) -> Settings:
         """Load settings from a file."""
@@ -41,7 +40,7 @@ class Settings(BaseModel):
                 settings_path = Path(env_path)
             else:
                 settings_path = cls._default_settings_path()
-                
+
                 # NOTE: Remove after a sufficiently long deprecation period.
                 if OS == "windows":
                     old_settings_path = user_config_path("copier") / "settings.yml"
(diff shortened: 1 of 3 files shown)


copier/settings.py
```diff
@@ -35,7 +35,7 @@ def from_file(cls, settings_path: Path | None = None) -> Settings:
             if env_path:
                 settings_path = Path(env_path)
             else:
-                settings_path = user_config_path("copier") / "settings.yml"
+                settings_path = user_config_path(appname="copier", appauthor=False) / "settings.yml"
```
sisp, line 38: Copier users on Windows may already use the `<CONFIG_ROOT>/copier/copier` path. I agree that it should be `<CONFIG_ROOT>/copier` on all platforms, but we should retain backwards compatibility and raise a deprecation warning:

```suggestion
                settings_path = (
                    user_config_path("copier", appauthor=False) / "settings.yml"
                )
                # NOTE: Remove after a sufficiently long deprecation period.
                if OS == "windows":
                    old_settings_path = user_config_path("copier") / "settings.yml"
                    if old_settings_path.is_file():
                        warnings.warn(
                            f"Settings path {old_settings_path} is deprecated. "
                            f"Please migrate to {settings_path}.",
                            DeprecationWarning,
                            stacklevel=2,
                        )
                        settings_path = old_settings_path
```

For this, you need to add the import `from .tools import OS`.

roteiro: @sisp Thanks for the improvement! Since the `copier/copier` path was undocumented and recently introduced with `v9.5.0` I think a rather short deprecation method would suffice.

Forgive my ignorance, this is my first pull request ever :) 
How is future code removal handled, simply through the `# NOTE`? Would it make sence to specify a target version for removal in the comment?

pawamoy: I *cannot* miss this occasion to advertise [Yore](https://github.com/pawamoy/yore) :smile:

pawamoy (approved):
LGTM, thanks for the update :slightly_smiling_face:

tests/test_settings.py
```diff
@@ -24,6 +25,11 @@ def test_settings_from_default_location(settings_path: Path) -> None:
 
     assert settings.defaults == {"foo": "bar"}
 
+@pytest.mark.skipif(OS != 'windows')
```
sisp, line 28: Could you please use `platform.system()` do identify the OS instead of `copier.tools.OS`? We use the latter for internal OS identification, which may be [patched for testing](https://github.com/copier-org/copier/blob/f842729d998961bf10a88d36940b75cc4641d150/tests/test_tasks.py#L164).

```suggestion
@pytest.mark.skipif(platform.system() != "Windows")
```

sisp (changes requested):

tests/test_settings.py
```diff
@@ -8,6 +8,7 @@
 
 from copier.errors import MissingSettingsWarning
 from copier.settings import Settings
+from copier.tools import OS
```
sisp, line 11: @roteiro You still need to remove this line and import `platform`.

roteiro: You are too fast for me :)
I know, there are still several errors, i'll ping you when its ready for review. Sorry for taking your time.

sisp: Ah, sorry, I didn't want to be pushy. :see_no_evil: No worries, feel free to ping me when you're ready. :ok_hand:

roteiro:
@sisp ok, should be ready for review now. The added test and mypy passed locally. There were some failures on the test suite but they seem unrelated to my changes?
Please also have a look if I got the staticmethod right, I'm not that experienced with OOP. 
I added a short note in the documentation about the deprecation of the old settings path, please check if that is necessary or can be removed

docs/settings.md
```diff
@@ -10,6 +10,10 @@ standard configuration directory for your platform:
 -   `%USERPROFILE%\AppData\Local\copier` on Windows as defined in
     [Known folders](https://docs.microsoft.com/en-us/windows/win32/shell/known-folders)
 
+!!! note
+
+    Windows only: `%USERPROFILE%\AppData\Local\copier\copier` was the standard configuration directory until v9.6.0. This standard configuration directory is deprecated and will removed in a future version.
```
sisp, line 15:
```suggestion
    Windows only: `%USERPROFILE%\AppData\Local\copier\copier` was the standard configuration directory until v9.6.0. This standard configuration directory is deprecated and will be removed in a future version.
```

copier/settings.py
```diff
@@ -13,6 +13,7 @@
 from pydantic import BaseModel, Field
 
 from .errors import MissingSettingsWarning
+from .tools import OS
```
sisp, line 16:
```suggestion
```

sisp (changes requested):
LGTM! :tada: Just some minor remarks, and Ruff and Prettier will want to reformat a little, but we can leave this to the autofix action.

copier/settings.py
sisp: Sorry, I confused the files. 🤦 This import is needed, but the one in the test file needs to be removed.

roteiro:
> LGTM! 🎉 Just some minor remarks, and Ruff and Prettier will want to reformat a little, but we can leave this to the autofix action.
Thanks for the guidance, I quite learned something with this small contribution. I assume you will merge with a proper commit message for the changelog generation?

copier/settings.py
sisp: This import is missing now. Could you add iy again, please?

sisp (approved):
LGTM! :tada: Thanks for identifying and fixing this bug, @roteiro! :pray:

PR #2079, opened 2025-04-09
build: migrate from Poetry to uv

sisp:
I've migrated our package and dependency management from Poetry to [uv]. This includes the migration from [poetry2nix] to [uv2nix].

## Motivation

- We have an increasing number of failing Nix-based CI jobs after updating dependencies (e.g., [#1949](https://github.com/copier-org/copier/pull/1949), [#2055](https://github.com/copier-org/copier/pull/2055), [#2056](https://github.com/copier-org/copier/pull/2056), ...). The root cause appears to be poetry2nix for a few reasons including [lack of support for Poetry v2 and lack of overrides maintenance](https://github.com/nix-community/poetry2nix?tab=readme-ov-file#maintenance-status-%EF%B8%8F-unmaintained-%EF%B8%8F) plus uncertain maintenance status. The maintainer of poetry2nix is also the author of [pyproject.nix] and uv2nix, and recommends to migrate.
- uv has gained rapid adoption in the community with some benefits (IMHO) compared to Poetry. For example:
    - uv eases Python interpreter installation across all major operating systems without the need for separate tools. This may not be relevant to developers using our Nix development environment, but not all contributors to Copier use Nix.
    - uv implements PEP standards where possible including [PEP 621 metadata][pep621] and [PEP 735 dependency groups][pep735], whereas Poetry v1 uses a proprietary metadata and dependency specification and v2 has adopted only PEP 621 metadata at the time of writing. Implementing standards increases ecosystem compatibility, e.g. [Renovate's Poetry integration has known issues][renovate-poetry-issues], as Renovate internally processes Poetry's dependency specifiers as NPM specifiers, which only works properly for SemVer versions but PEP440 is a superset of SemVer. Renovate supports uv via its [`pep621` manager][renovate-pep621-manager] without such limitations, merely [uv workspaces and PEP 723 scripts lockfile maintenance are not yet supported][renovate-uv-issues] at the time of writing, which we're not using.
    - uv's lock file is more merge-friendly than [Poetry's][poetry-issue-lockfile-merge-friendly]. IMO, development dependencies should not only be pinned via a lock file but also in the dependency manifest, i.e. in `pyproject.toml`, because:
        - We control the exact version and can ensure compatible configuration. With lower-bounded versions, the dependency resolver may downgrade a dependency without us noticing unless we study the lock file changes diligently.
        - We don't need to maintain an accurate version lower bound. Strictly speaking, if we declare a lower version bound – even for a development dependency –, we'd need to ensure that all versions that satisfy this constraint work. This is difficult and extremely cumbersome to achieve with, e.g., linters and formatters because different versions may result in different linting and formatting results. If we pin such development dependencies, we're not suggesting that a range of versions yields identical results, and the maintenance effort is minimized.
    When pinning development dependencies in the mainfest with Poetry, multiple simultaneous PRs with updates cannot be merged easily because merging one such PR results in a merge conflict of the other PRs, as Poetry's `content-hash` value in `poetry.lock` is changed in every such PR. As uv's lock file doesn't contain this kind of information, PRs can be merged without recalculating the lock file.

## Remarks

I've split the migration into two commits to ease PR review:

1. Migration from Poetry to uv
2. Style fixes by formatters and linters

I'm still struggling quite a bit with Nix, so, @yajo, please review my Nix-related changes particularly carefully. I've tried retaining the different Nix interfaces to the best of my ability including support for `nix-shell` via `flake-compat`. I haven't managed to test the `direnv` integration, perhaps you could double-check this is still working correctly.

~~I'm anticipating CI to fail, so I'm marking this PR as a draft until I've resolved all issues.~~ Surprisingly, CI is working instantly. :exploding_head: :smile:

[pep621]: https://peps.python.org/pep-0621/
[pep735]: https://peps.python.org/pep-0735/
[poetry2nix]: https://github.com/nix-community/poetry2nix
[poetry-issue-lockfile-merge-friendly]: https://github.com/python-poetry/poetry/issues/496
[pyproject.nix]: https://github.com/pyproject-nix/pyproject.nix
[renovate-pep621-manager]: https://docs.renovatebot.com/modules/manager/pep621/
[renovate-poetry-issues]: https://docs.renovatebot.com/modules/manager/poetry/#additional-information
[renovate-uv-issues]: https://github.com/renovatebot/renovate/issues?q=is%3Aissue%20state%3Aopen%20uv%20label%3Amanager%3Apep621%2Cmanager%3Apep723
[uv]: https://docs.astral.sh/uv/
[uv2nix]: https://pyproject-nix.github.io/uv2nix/


sisp:
@yajo Could you please review this PR, as you're the Nix expert? :pray: Please let me know if you need some clarification about changes and decisions.

The RTD job is failing because there seems to be an upstream problem: https://github.com/readthedocs/readthedocs.org/issues/12093

pawamoy:
I did not take time to review yet, but I'm all for this migration to uv. I've been using it for more than a year now, and it's great.

> we'd need to ensure that all versions that satisfy this constraint work. This is difficult and extremely cumbersome to achieve

Less difficult thanks to `uv lock --resolution lowest-direct` for example, but yeah, still a bit cumbersome to set up, both in CI and locally.

Also, lock files were just standardized (see https://discuss.python.org/t/pep-751-one-last-time/77293/150), and uv will likely support the standard very soon (while Poetry will probably lag behind?).

sisp:
> [...] and uv will likely support the standard very soon (while Poetry will probably lag behind?).

I'm not so sure about that: https://github.com/astral-sh/uv/issues/12584#issue-2960561808

pawamoy:
Oh well. Thanks for the link!

sisp:
RTD is working again and I fixed an error in `.readthedocs.yml`.

yajo:
You've read my mind. I had idea of doing exactly this change.

We can follow another approach if you want. These days I'm having very very few time to dedicate to Copier, and you guys are actually keeping it well maintained. And I'm the most Nix user here. So we can drop Nix packaging from here and just leave the devshell. What would that mean? That we still provide a shared set of dev tools (python, uv, prettier, etc.), but the final copier package wouldn't be published here anymore.

These days [Copier is packaged in nixpkgs](https://search.nixos.org/packages?channel=24.11&show=copier&from=0&size=50&sort=relevance&type=packages&query=Copier), so it wouldn't be a big problem anymore. Keeping the package well maintained is easier there, where everybody knows nix deeply.

sisp:
I wouldn't have suggested it myself, as you seem to use Nix a lot. But if providing a Nix package in this project isn't a priority for you, I'm happy to simplify the Nix setup as you suggested. I'll update the PR tomorrow.

I'm sorry to read that you're time for Copier is so limited these days. I've always appreciated your opinion and thoughts about changes. :pray:

.github/workflows/flakehub-publish-tagged.yml
sisp, line 1: This file can be deleted, as we don't build a Flake package anymore.

sisp (commented):
I've updated this PR to use only a Nix devshell using devenv and removed the Flake package, as @yajo suggested. As a result, there's no need to use uv2nix anymore.

Would you mind reviewing this PR to make sure my Nix changes are correct, @yajo? :pray:

sisp:
I'll merge this PR once CI has passed to move forward and unblock pending Renovate PRs. We can always improve the Nix setup in the future, as it's only relevant for development now.

PR #2089, opened 2025-04-23
ci(release): migrate to PyPI trusted publishing

sisp:
I've migrated our release workflow from using PyPI access tokens to [trusted publishing](https://docs.pypi.org/trusted-publishers/).

A few remarks about advice on https://docs.pypi.org/trusted-publishers/security-model/ that I've followed:

* I've created a dedicated workflow `release.yml` that is trusted by pypi.org. But it seems that dependence on jobs from another workflow (here, `build` and `flake-check` from `ci.yml`) via `needs` isn't supported. The only alternative I've found is dependence on the completion and success of a complete workflow via [`workflow_run`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run) and `job.<job_id>.if: ${{ github.event.workflow_run.conclusion == 'success' }}`, but I haven't found a Python project that implements this approach. I think it's sufficiently safe to assume that the tests are passing when making a release, but I'm open to hearing different opinions.
* I've limited the scope of the publishing jobs by using separate `build` and `publish`/`publish-test` jobs where only the latter have write permissions.
* I've created dedicated environments `pypi`/`pypi-test` scoped to tags with the name pattern `v*`.
* I've created tag protection rules.

As before, the distribution files are uploaded to test.pypi.org when a tag is pushed. In a follow-up PR, I plan to change this setup to upload to test.pypi.org also on the default branch (`master`) to test the publishing process more regularly (with some inspiration from [`github.com/iterative-ai/dvc:.github/workflows/build.yaml`](https://github.com/iterative/dvc/blob/99d9ca714de61518bd672ff647ed1ad7fb56f1d1/.github/workflows/build.yaml#L50-L73)).

**Todos after merging this PR:**

- [ ] Configure trusted publishing on pypi.org
- [ ] Configure trusted publishing on test.pypi.org

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   pull_request:
   push:
     branches: [master]
-    tags: ["*"]
   workflow_dispatch:
     inputs:
       pytest_addopts:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,6 @@ on:
   pull_request:
   push:
     branches: [master]
-    tags: ["*"]
   workflow_dispatch:
```
sisp, line 7: Disabling this workflow for pushed tags should resolve the following problem with Codecov:

> Coverage data is based on head [8e9a87f (15 uploads)](https://app.codecov.io/gh/copier-org/copier/commit/8e9a87f79d9a5d6d4ece14fa0c96912ee2373cb2) compared to base [f215a23 (30 uploads)](https://app.codecov.io/gh/copier-org/copier/commit/f215a237441171a84aadc829d695ba3928fc71d9)

30 instead of 15 uploads are cause by the two CI runs triggered by the pushed release commit to `master` and the tag that references that same commit.

pawamoy (approved):
Thanks @sisp , looking good to me 👍

.github/workflows/release.yml
```diff
@@ -0,0 +1,82 @@
+name: release
+
+on:
+  push:
+    branches: [master]
```
sisp, line 5: I just noticed in a final check that this line isn't needed at the moment, I probably introduced it while thinking about publishing to test.pypi.org from the `master` branch.

```suggestion
```

PR #2127, opened 2025-05-05
docs: include full _copier_conf attributes list

wgordon17:


wgordon17:
Here's the funny thing, I hadn't even _noticed_ the introduction sentence with

```
_copier_conf includes a representation of the current Copier [Worker](https://copier.readthedocs.io/en/stable/reference/main/#copier._main.Worker) object
```

until I went back and looked _after_ I had found the `Attributes` table. Now that I _know_ it's there, this PR is really just intended to give some additional clarity (feel free to close as well if you feel that the extra `Note` shouldn't be necessary ¯\\\_(ツ)_/¯ )

sisp:
Thanks for aspiring to improve the docs! :bow: I agree that this the render context isn't documented as well as it could be. There's only one concern I have with this addition: `copier.main` will become an internal-only module and `copier.main.Worker` will become an internal-only class. So the API reference will become unavailable. IMO, we should model public interfaces (like the `copier.yml` content specification and the render context) with Pydantic to have declarative and explicit representations as single source of truth of information, static typing, runtime validation, and (exportable) JSON schema. I've resumed work on this long-standing idea of mine, but it's a quite massive change. In the meantime, how about documenting the render context explicitly and manually? We'll need to keep it synchronous with the actual implementation,, but changes are quite infrequent.

wgordon17:
Hey @sisp, thanks for the review. If y'all are good with ensuring that the docs are just always in-sync, I'm _absolutely_ happy to help copy over the data. Let me do a quick fix, and I'll let you give it another review 😊

wgordon17:
@sisp I think this should be ready for a review now. Direct link to docs preview: https://copier--2127.org.readthedocs.build/en/2127/creating/#_copier_conf

A few things that I adjusted (if you're OK with that)
- Alphabetized the attributes to ideally make it easier to skim, versus the previous order matching `_main.py`
- Included `Defaults` on the rows where I could easily see what the defaults were in `_main.py`
- Added a `NOTE` in `_main.py` to hopefully provide a reminder to folks that any changes there need to _also_ be reflected on the, now static, Attributes table in docs

wgordon17:
@sisp Just checking if you're good with the changes?

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
 
-`_copier_conf` includes a representation of the current Copier
-[Worker][copier.main.Worker] object, also slightly modified:
+`_copier_conf` includes a slightly modified representation of the current Copier Worker
+object:
 
 -   It only contains JSON-serializable data.
 -   You can serialize it with `{{ _copier_conf|to_json }}`.
 -   ⚠️ It contains secret answers inside its `.data` key.
 -   Modifying it doesn't alter the current rendering configuration.
```
sisp, line 105: How about deleting this entire snippet, as referring to the `Worker` class and the relationship between the render context and the `Worker` class is an implementation detail?

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `quiet`            | `bool`                     | When `True`, disable all output.<br>See the [`quiet`][] setting for related information.                                                                                                                                                                   |
+| `sep`              | `str`                      | The operating system-specific directory separator.                                                                                                                                                                                                         |
+| `settings`         | `pydantic.BaseModel`       | Specified settings that control copier's behavior.<br>See the [`settings`][] page for related information.                                                                                                                                                 |
```
wgordon17, line 122: @sisp Thoughts on the content for this one?

docs/creating.md
```diff
@@ -127,8 +127,8 @@ Attributes:
...
-| `vcs_ref`          | `str | None`               | The VCS tag/commit of the template, `None` if the template is not VCS-tracked.<br>See the [`vcs_ref`][] setting for related information.                                                                                                                   |
-| `vcs_ref_hash`     | `str | None`               | The VCS commit hash of the template, `None` if the template is not VCS-tracked.                                                                                                                                                                            |
+| `vcs_ref`          | `str                       | None`                                                                                                                                                                                                                                                      | The VCS tag/commit of the template, `None` if the template is not VCS-tracked.<br>See the [`vcs_ref`][] setting for related information. |
```
wgordon17, line 130: @sisp we might have to disable `autoformat` for this PR (not sure if you can). Leaving the `\` in `str \| None` caused the `\` to actually _print_ in the assembled docs, however removing it causes `autoformat` to think that it's a column separator and add extra whitespace which then breaks the table layout 🤦‍♂️

docs/creating.md
sisp: How about referring to the [`copier.settings.Settings`](https://github.com/copier-org/copier/blob/e2ce0d4f13f439c6d9467fc4bf0e0af5f6a60989/copier/settings.py#L21) model class via `[copier.settings.Settings][]` (thanks to mkdocstrings)?

```suggestion
| `settings`         | [copier.settings.Settings][] | Specified settings that control copier's behavior.<br>See the [`settings`][] page for related information.                                                                                                                                                                                                                                                                                         |
```

(The other rows will need re-formatting because of the wider column ...)

docs/creating.md
sisp: Mhhh, not the first time that Prettier is causing some problems with Markdown formatting. How about using HTML and entity escaping in this particular case?

```suggestion
| `vcs_ref`          | <code>str &vert; None</code>                                                                                                                                                                                                                                                            | The VCS tag/commit of the template, `None` if the template is not VCS-tracked.<br>See the [`vcs_ref`][] setting for related information. |
```

Not great, but perhaps better than disabling Prettier altogether for the table.

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `quiet`            | `bool`                     | When `True`, disable all output.<br>See the [`quiet`][] setting for related information.                                                                                                                                                                   |
+| `sep`              | `str`                      | The operating system-specific directory separator.                                                                                                                                                                                                         |
+| `settings`         | `pydantic.BaseModel`       | Specified settings that control copier's behavior.<br>See the [`settings`][] page for related information.                                                                                                                                                 |
```
wgordon17:
```suggestion
| `settings`         | `[Settings][copier.settings.Settings]`       | Specified settings that control copier's behavior.<br>See the [`settings`][] page for related information.                                                                                                                                                 |
```

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `quiet`            | `bool`                       | When `True`, disable all output.<br>See the [`quiet`][] setting for related information.                                                                                                                                                                   |
+| `sep`              | `str`                        | The operating system-specific directory separator.                                                                                                                                                                                                         |
+| `settings`         | `pydantic.BaseModel`         | Specified settings that control copier's behavior.<br>See the [`settings`][] page for related information.                                                                                                                                                 |
```
wgordon17:
```suggestion
| `settings`         | `[Settings][copier.settings.Settings]`         | Specified settings that control copier's behavior.<br>See the [`settings`][] page for related information.                                                                                                                                                 |
```

docs/creating.md
wgordon17: Trying with 

```
`[Settings][copier.settings.Settings]`
```

wgordon17: That didn't work, trying again with `<code>[Settings][copier.settings.Settings]</code>`

wgordon17: There we go, that worked! @sisp up to you if we stick with the shorter `[Settings][copier.settings.Settings]` or go with the longer `[copier.settings.Settings][]` ¯\\\_(ツ)_/¯

sisp: Is the problem with `[copier.settings.Settings][]` that it doesn't get rendered as inline code? If so, I imagine ``[`copier.settings.Settings`][]`` or ``[`copier.settings.Settings`][copier.settings.Settings]`` should work – not tested though, as I'm commenting from mobile right now.

wgordon17: Yup, _perfect_! 👌 Which do you prefer `Settings` (A)  vs `copier.settings.Settings` (B) as the display value?

A SCR-20250607-qqea

B SCR-20250607-qqrn

sisp: I think (B) is better because it's more explicit.

wgordon17: Great, since (B) is what's currently in the PR. I'll mark this one resolved then 👍

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `dst_path`         | `Path`                         | Destination path where to render the subproject.<br>⚠️ When [updating a project][updating-a-project], it may be a temporary directory, as Copier's update algorithm generates fresh copies using the old and new template versions in temporary locations. |
+| `exclude`          | `Sequence[str]`                | Specified additional [file exclusion patterns][patterns-syntax].<br>See the [`exclude`][] setting for related information.                                                                                                                                 |
+| `os`               | `str`                          | The detected operating system, either `"linux"`, `"macos"`, `"windows"` or `None`.                                                                                                                                                                         |
```
sisp: I just noticed that the type is actually a literal; this also allows to shorten the description:

```suggestion
| `os`               | <code>Literal["Linux", "macos", "windows"] &vert; None</code> | The detected operating system, `None` if it could not be detected.                                                                                                                                                                         |
```

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `quiet`            | `bool`                         | When `True`, disable all output.<br>See the [`quiet`][] setting for related information.                                                                                                                                                                   |
+| `sep`              | `str`                          | The operating system-specific directory separator.                                                                                                                                                                                                         |
+| `settings`         | [`copier.settings.Settings`][] | Specified settings that control copier's behavior.<br>See the [`settings`][] page for related information.                                                                                                                                                 |
```
sisp: Perhaps clarify the description a little to indicate that these are general user settings unspecific to a particular template, e.g.:

```suggestion
| `settings`         | [`copier.settings.Settings`][] | General user settings.<br>See the [`settings`][] page for related information.                                                                                                                                                 |
```

wgordon17:
@sisp 2 questions/thoughts, looking at 

image

1. Why is `Linux` capitalized while the others are lowercase? Could/should that be normalized? 🤔 Likely a breaking change 😞 
2. Do you want to use `&nbsp;` to keep everything on one-line? Would definitely make the column wider 😞 Not sure if you have a preference ¯\\\_(ツ)_/¯

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `unsafe`           | `bool`                                                        | When `True`, allow usage of unsafe templates.<br>See the [`unsafe`][] setting for related information.                                                                                                                                                     |
+| `use_prereleases`  | `bool`                                                        | When `True`, `vcs_ref`/`vcs_ref_hash` may refer to a prerelease version of the template.<br>See the [`use_prereleases`][] setting for related information.                                                                                                 |
+| `user_defaults`    | `dict[str, Any]`                                              | Specified user defaults that may override a template's defaults during question prompts.                                                                                                                                                                   |
```
wgordon17, line 129: Should we point to https://copier--2127.org.readthedocs.build/en/2127/settings/#user-defaults? What would that be, `See the [User defaults][settings.user-defaults] setting for related information.`

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `os`               | <code>Literal["Linux", "macos", "windows"] &vert; None</code> | The detected operating system, `None` if it could not be detected.                                                                                                                                                                                         |
+| `overwrite`        | `bool`                                                        | When `True`, overwrite files that already exist, without asking.<br>See the [`overwrite`][] setting for related information.                                                                                                                               |
+| `pretend`          | `bool`                                                        | When `True`, produce no real rendering.<br>See the [`pretend`][] setting for related information.                                                                                                                                                          |
```
wgordon17, line 119:
```suggestion
| `pretend`          | `bool`                                                        | When `True`, perform no real rendering.<br>See the [`pretend`][] setting for related information.                                                                                                                                                          |
```
?

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `exclude`          | `Sequence[str]`                                               | Specified additional [file exclusion patterns][patterns-syntax].<br>See the [`exclude`][] setting for related information.                                                                                                                                 |
+| `os`               | <code>Literal["Linux", "macos", "windows"] &vert; None</code> | The detected operating system, `None` if it could not be detected.                                                                                                                                                                                         |
+| `overwrite`        | `bool`                                                        | When `True`, overwrite files that already exist, without asking.<br>See the [`overwrite`][] setting for related information.                                                                                                                               |
```
wgordon17, line 118:
```suggestion
| `overwrite`        | `bool`                                                        | When `True`, overwrite files that already exist without asking.<br>See the [`overwrite`][] setting for related information.                                                                                                                               |
```
?

docs/creating.md
```diff
@@ -95,27 +95,40 @@ suitable to [autoupdate your project safely][the-copier-answersyml-file]:
...
+| `data`             | `dict[str, Any]`                                              | Answers to the questionnaire, defined in the template, provided via CLI (`-d,--data`) or API (`data`).<br>See the [`data`][] setting for related information.<br>⚠️ May contain secret answers.                                                            |
+| `defaults`         | `bool`                                                        | When `True`, use default answers to questions.<br>See the [`defaults`][] setting for related information.                                                                                                                                                  |
+| `dst_path`         | `Path`                                                        | Destination path where to render the subproject.<br>⚠️ When [updating a project][updating-a-project], it may be a temporary directory, as Copier's update algorithm generates fresh copies using the old and new template versions in temporary locations. |
```
wgordon17, line 115:
```suggestion
| `dst_path`         | `Path`                                                        | Destination path where to render the subproject.<br>⚠️ When [updating a project][updating-a-project], this may be a temporary directory, as Copier's update algorithm generates fresh copies using the old and new template versions in temporary locations. |
```

wgordon17:
Ok, threw a few more "consistency" suggestions in here as well, WDYT?

docs/creating.md
sisp: [`user_defaults` in the render context](https://github.com/copier-org/copier/blob/e2ce0d4f13f439c6d9467fc4bf0e0af5f6a60989/copier/_main.py#L411) is a [`Worker` class attribute](https://github.com/copier-org/copier/blob/e2ce0d4f13f439c6d9467fc4bf0e0af5f6a60989/copier/_main.py#L231) that can be passed via API only, but it's not the same as [`Settings.defaults`](https://github.com/copier-org/copier/blob/e2ce0d4f13f439c6d9467fc4bf0e0af5f6a60989/copier/settings.py#L24-L26). It may make sense to document `user_defaults` in `docs/configuring.md` in a separate PR.

docs/creating.md
sisp: Sure, but we should synchronize the descriptions and the `Worker` class attributes in a separate PR – wherever it makes sense.

docs/creating.md
sisp: Sure, but we should synchronize the descriptions and the `Worker` class attributes in a separate PR – wherever it makes sense.

docs/creating.md
sisp: Sure, but we should synchronize the descriptions and the `Worker` class attributes in a separate PR – wherever it makes sense.

sisp (commented):
> 1. Why is `Linux` capitalized while the others are lowercase? Could/should that be normalized? 🤔 Likely a breaking change 😞 

OMG, that's totally my error. I was typing on mobile and it seems typing correction capitalized it without me noticing. It should be lowercase of course. Thanks for catching this! :1st_place_medal:

> 2. Do you want to use `&nbsp;` to keep everything on one-line? Would definitely make the column wider 😞 Not sure if you have a preference ¯\\\_(ツ)_/¯ 

It does look better without the line wrap, but then the description column might get squeezed too much. I'd say, your call. We can always change it later. Thanks for paying attention to such details though! :bow:

wgordon17:
@sisp Odd, it doesn't seem to be respecting my `&nbsp;`...oh well, for something we're not entirely sold on, it's probably not worth debugging ¯\\\_(ツ)_/¯ I've lowercased `linux`...perhaps it's finally time to ship it? 🚀

sisp (approved):
Looks great! :star: Thank you very much for working so diligently on this PR! :bow: Let's merge it. :wink:

Issue #2075, opened 2025-04-07
Fix visual conflict resolution for Jetbrains IDEs


thurse93:
### Actual Situation

Hi,

~~I have a problem with the way JetBrains IDEs interprets the conflict markers created by copier compared to cruft. Here's an example conflict file with copier:~~

The Conflict View of JetBrains IntelliJ IDEA seems to be broken when resolving conflicts caused by copier, at least compared to cruft.

Following conflict file:

```
<<<<<<< before updating
- Preamble added in project
- from Template
- Addendum added in project
=======
- from Template
- some other info from template
>>>>>>> after updating
```

Results in the following conflict view window in IntellJ IDEA:

![Image](https://github.com/user-attachments/assets/2694e974-f1ee-446f-a9cc-d8dbe8968988)

The problem is that I cannot select and apply changes from "Your version" since the are somehow automerged in the Result.

cruft on the other hand produces this file in its conflict resolution process:

```
<<<<<<< ours
- Preamble added in project
- from Template
- Addendum added in project
=======
- from Template
- some other info from template
>>>>>>> theirs
```

Which results in a more correct conflict view in IDEA:

![Image](https://github.com/user-attachments/assets/2e1c6835-1d17-489d-b1cd-4826e0b8bd81)

### Desired Situation

~~The conflict markers created by copier and cruft should be identical and as such be treated equally in Jetbrains IDEs.~~

~~Of course it would be desirable that IntelliJ has a more text agnostic conflict view but I did not find any indicator on whether thats feasable or if that's because the `ours` / `theirs` scheme is more of a git standard?~~

~~Also I'm not sure whether there may be other tools besides IntelliJ relying on `ours` / `theirs`.~~

Okey I was a bit too quick with my conclusions. Seems that the naming of the git markers is not the main point but the way copier does its conflict resolution. Seems that the problem has been resolved for IDEs like VSCode but not for JetBrains?: https://github.com/copier-org/copier/issues/1350

### Proposed solution

~~I would suggest making the conflict marker text configurable, so that the conflict marker output can be tweaked.~~

~~Could be free text or a fixed switch between `before updating` & `after updating` or `ours` & `theirs`.~~

At the moment I have no Idea how to fix this because I couldn't really grasp the problem yet.

pawamoy:
Hi @thurse93, thanks for the report!

[...]

PR #2129, opened 2025-05-06
fix: correct git stage order on merge conflicts

thurse93:
Fixes the stage order in which the git index gets updated on merge conflicts. This is a preparation PR for @pawamoy.

This change doesn't seem to break any tests, however im not familiar enough with the code base to write one that catches this condition.

fixes https://github.com/copier-org/copier/issues/2075


pawamoy (approved):
Tried locally in VSCode, looking good :slightly_smiling_face: Thanks!

PR #2136, opened 2025-05-08
feat: add `--vcs-ref=:current:` feature to reference the current ref when updating

ytausch:
Sometimes, a user might only want to update the answers to their questions, but not the template version.
As far as I am aware, copier does not support that currently.

In this PR, I add a `--answers-only` flag to `copier update` which fills exactly this gap.

Note about tests: I am a bit overwhelmed by your testing setup, could you maybe please point me to the right places to add corresponding tests? Thanks in advance!

Cc @pavelzw

### DIFF
--- a/copier/_types.py
+++ b/copier/_types.py
@@ -6,7 +6,7 @@ import sys
 from collections.abc import Iterator, Mapping, MutableMapping, Sequence
 from contextlib import contextmanager
 from contextvars import ContextVar
-from enum import Enum, auto
+from enum import Enum
 from pathlib import Path
 from typing import (
     Annotated,
@@ -128,7 +128,20 @@ _phase: ContextVar[Phase] = ContextVar("phase", default=Phase.UNDEFINED)
 
 
 class VcsRef(Enum):
-    CURRENT = auto()
+    CURRENT = ":CURRENT"
     """A special value to indicate that the current ref of the existing
     template should be used.
     """
+
+    @classmethod
+    def try_parse(cls, value: str | None) -> VcsRef | str | None:
+        """Try to parse a string into a special VcsRef value or return the string.
+
+        None is returned as is.
+        """
+        if value is None:
+            return None
+        try:
+            return cls(value.upper())
+        except ValueError:
+            return value
--- a/copier/_types.py
+++ b/copier/_types.py
@@ -128,20 +128,7 @@ _phase: ContextVar[Phase] = ContextVar("phase", default=Phase.UNDEFINED)
 
 
 class VcsRef(Enum):
-    CURRENT = ":CURRENT:"
+    CURRENT = ":current:"
     """A special value to indicate that the current ref of the existing
     template should be used.
     """
-
-    @classmethod
-    def try_parse(cls, value: str | None) -> VcsRef | str | None:
-        """Try to parse a string into a special VcsRef value or return the string.
-
-        None is returned as is.
-        """
-        if value is None:
-            return None
-        try:
-            return cls(value.upper())
-        except ValueError:
-            return value
--- a/copier/_types.py
+++ b/copier/_types.py
@@ -128,7 +128,7 @@ _phase: ContextVar[Phase] = ContextVar("phase", default=Phase.UNDEFINED)
 
 
 class VcsRef(Enum):
-    CURRENT = ":CURRENT"
+    CURRENT = ":CURRENT:"
     """A special value to indicate that the current ref of the existing
     template should be used.
     """
--- a/copier/_types.py
+++ b/copier/_types.py
@@ -6,7 +6,7 @@ import sys
 from collections.abc import Iterator, Mapping, MutableMapping, Sequence
 from contextlib import contextmanager
 from contextvars import ContextVar
-from enum import Enum
+from enum import Enum, auto
 from pathlib import Path
 from typing import (
     Annotated,
@@ -125,3 +125,10 @@ class Phase(str, Enum):
 
 
 _phase: ContextVar[Phase] = ContextVar("phase", default=Phase.UNDEFINED)
+
+
+class VcsRef(Enum):
+    CURRENT = auto()
+    """A special value to indicate that the current ref of the existing
+    template should be used.
+    """
--- a/copier/__init__.py
+++ b/copier/__init__.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any
 
 from . import _main
 from ._deprecation import deprecate_member_as_internal
+from ._types import VcsRef as VcsRef
 
 if TYPE_CHECKING:
     from ._main import *  # noqa: F403
--- a/copier/_cli.py
+++ b/copier/_cli.py
@@ -403,6 +403,7 @@ class CopierUpdateSubApp(_Subcommand):
         ["-O", "--only-answers"],
         default=False,
         help="Only update copier answers, do not update the template version",
+        excludes=["--vcs-ref"],
     )
 
     def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
@@ -424,7 +425,6 @@ class CopierUpdateSubApp(_Subcommand):
                 context_lines=self.context_lines,
                 defaults=self.defaults,
                 skip_answered=self.skip_answered,
-                update_answers_only=self.only_answers,
                 overwrite=True,
             ) as worker:
                 worker.run_update()
--- a/copier/_cli.py
+++ b/copier/_cli.py
@@ -399,8 +399,8 @@ class CopierUpdateSubApp(_Subcommand):
         default=False,
         help="Skip questions that have already been answered",
     )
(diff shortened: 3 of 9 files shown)


sisp (commented):
Thanks for suggesting this feature and working on a PR, @ytausch! :pray:

I've checked a few popular CLI tools and it seems that it's more common to place an adverb like "only" in front, i.e. `--only-answers` in this case.

About the short flag `-u`, what's the rationale behind "u"? (Not saying I have another/better suggestion off the top of my head ...)

Updating only answers without a new template version (and without explicitly specifying the current version via the `--vcs-ref` flag) is conceptually like running `copier update --vcs-ref CURRENT ...` where `CURRENT` is a hypothetical special value that refers to the current template version of the project. This doesn't work at the moment, and `CURRENT` is a valid arbitrary branch name, so we can't treat it as a special value in case somebody decides to use it as a branch name. But what I'm trying to say is that this feature almost exists already without introducing a new flag. For the sake of pursuing this avenue further, special values to refer to the current template version might be `""` (although perhaps too ambiguous and unspecific) or `<CURRENT>` (although it would need to be quoted at least on Unix because `<` and `>` are treated as redirect operators). WDYT about this idea in general, @ytausch @pawamoy?

In case we decide against reusing the `--vcs-ref` flag, I lean towards reusing the `vcs_ref` attribute of the `Worker` class instead of adding a new `answers_only` attribute to reduce special-casing. For example, we could introduce an enum with (for now) one member:

```python
from enum import auto, Enum

class VCSRef(Enum):
    CURRENT = auto()
```

Then, when the `--only-answers` flag is used, we can pass `vcs_ref=VCSRef.CURRENT` to the `Worker` class constructor and condition the shortcut in the `Worker.template` property on the `self.vcs_ref` value instead of `self.update_answers_only`. WDYT, @pawamoy?

If we add the `--only-answers` with the enum-based implementation, please update the [`vcs_ref` setting](https://copier.readthedocs.io/en/stable/configuring/#vcs_ref) documentation (`Format: str | VCSRef`), and I suggest to mention the `--only-answers` flag in a new paragraph there.

Also, the `--only-answers` flag would need to be declared mutually exclusive with `--vcs-ref`.

Regarding tests, please add a test case to [`tests/test_updatediff.py`](https://github.com/copier-org/copier/blob/master/tests/test_updatediff.py). I think the [`test_update_with_skip_answered_and_new_answer`](https://github.com/copier-org/copier/blob/477777cd19dd15adf627d9087594c0670ce6d05b/tests/test_updatediff.py#L1579-L1603) case might serve as a good starting point for you. The test case sketch may be roughly this:

1. Create a template v1 with one question with a Git tag `v1`.
2. Evolve the template to v2 by modifying it a little, e.g. by adding a file, and tag it as `v2`.
3. Generate a project with template `v1` (answer passed via `--data`/`data={...}`) and commit the files. Check the recorded answer and template version (must be `v1`)
4. Update the project with `--only-answers`/`vcs_ref=VCSRef.CURRENT` (in case that's the implementation decision) and a different answer (passed via `--data`/`data={...}`). Check the recorded answer and template version (must still be `v1`).

ytausch:
Thanks a ton for your elaborate response, @sisp!

> I've checked a few popular CLI tools and it seems that it's more common to place an adverb like "only" in front, i.e. --only-answers in this case.

Fixed!

> About the short flag -u, what's the rationale behind "u"? 

There is none really. How about `-O` for only?

> copier update --vcs-ref CURRENT ..

I had a similar thought when implementing this, but decided against it because I felt `--only-answers` could be a pretty common use case where `--vcs-ref <CURRENT>` (or similar) might be too complicated for users who are not too familiar with the tool. But I don't have any strong opinion on this.

Resuing `vcs_ref` in the Worker seems fine for me. I'll have a look into testing in the next days!

ytausch:
Everything should now be implemented!

pawamoy:
@sisp

> Then, when the --only-answers flag is used, we can pass vcs_ref=VCSRef.CURRENT to the Worker class constructor and condition the shortcut in the Worker.template property on the self.vcs_ref value instead of self.update_answers_only. WDYT, @pawamoy?

Yep, sounds good. I do like better the shortcut approach with a dedicated flag, rather than reusing `--vcs-ref` :+1:

sisp (changes requested):
This is excellent work, @ytausch! :1st_place_medal: Just a few minor remarks from my side.

sisp (changes requested):
I just stumbled over a little missing piece: We need to export the `VcsRef` enum as part of the public Python API to support using this new feature via API. We're currently deprecating non-public API, so there's some noise. To expose `VcsRef` in a future-proof way, could you please add the following line to `copier/__init__.py`?

```diff
 ...
 from ._deprecation import deprecate_member_as_internal
+from ._types  import VcsRef as VcsRef
```

sisp (approved):
Excellent work! :1st_place_medal: Thanks again for contributing this feature! :pray:

sisp:
I almost hit "merge", but then I was rethinking the commit message and stumbled over the semantics of the `--only-answers` flag. Is the semantics of this flag actually about updating only answers, or is it rather about keeping/pinning/_not updating_ the template version? For example, wouldn't `copier update --only-answers --skip-answered` be an allowed but odd combination of flags, as this would be a no-op? I'm just wondering whether a flag name related to not updating the template version would be more precise and less prone to becoming inaccurate in the future than `--only-answers`. Perhaps `copier update --keep-version` or `copier update --no-version` or something like that? :thinking:

pavelzw:
i would prefer `--keep-version` over `--no-version`

ytausch:
I agree and renamed everything to `--keep-version`. Thank you so much for making this one of the smoothest collaborations I've ever seen :D

pawamoy:
Whatever is easier to maintain in the long term :) No strong opinion for me.

ytausch:
Same for me!

sisp:
Thanks for your feedback. TBH, I lean towards reusing `--vcs-ref` now because it reuses existing functionality and implies no more than it's actually doing which seems the most future-proof option to me. Regarding the sentinel, I found two variants that should satisfy all requirements: `.current` and `:current`. Esthetically (but highly subjectively) I might even prefer `:current` (`copier update --vcs-ref :current` or equivalently `copier update --vcs-ref=:current`), perhaps because it resembles container image tags a little, I don't know. Any preferences?

ytausch:
`:current` seems fine.

sisp:
Okay, let's use `:current:` then. :+1:

ytausch:
Done. Do we have to change something in `_render_context` if `vcs_ref` can now be a special `VcsRef` value?

ytausch:
With this approach, we would still lose some type safety. How about passing `self.resolved_vcs_ref` as `vcs_ref` to the jinja context instead? I could imagine it is more interesting what vcs_ref we are actually on instead of what was passed via the command line.

sisp:
Fair point. Or how about passing the _resolved_ ref as `_copier_conf.vcs_ref` and waiting for somebody to actually have a use case that requires knowing the original value? We're already doing this for [`_copier_conf.src_path`](https://github.com/copier-org/copier/blob/7ccb4247af407693319fbfa4a32d968ab61c8a95/copier/_main.py#L398).

ytausch:
Jup, this was my idea :) Changed it!

copier/_types.py
```diff
@@ -125,3 +125,23 @@ def current(cls) -> Phase:
...
+        """Try to parse a string into a special VcsRef value or return the string.
+
+        None is returned as is.
+        """
+        if value is None:
+            return None
+        try:
+            return cls(value.upper())
+        except ValueError:
+            return value
```
sisp, line 147: Do you think this classmethod is sufficiently relevant to become part of the public API? With the current API deprecations, I'm trying quite hard to minimize the public API surface to ease internal refactoring safely.

How about creating a slightly more general-purpose internal helper in `_tools.py` instead?

```python
from __future__ import annotations

from enum import Enum
from typing import TypeVar

_T = TypeVar("_T")
_E = TypeVar("_E", bound=Enum)
    

def try_enum(enum_type: type[_E], value: _T) -> _E | _T:
    try:
        return enum_type(value)
    except:
        return value
```

Then, in `_cli.py` we could call it as `vcs_ref=try_enum(VcsRef, self.vcs_ref)` and only the bare `VcsRef` enum would become part of the public API. Just one detail: For this to work, the member value would need to be `":current:"` instead of `":CURRENT:"`.

sisp:
> Jup, this was my idea :)

Ah, sorry, I must have misread your suggestion to pass the resolved ref as `_copier_conf.vcs_ref_resolved`. :facepalm:

copier/_types.py
ytausch: Done!

sisp (approved):
LGTM! :tada: Thanks again for the excellent work :1st_place_medal: and hanging in there with the final adjustments and polishing. :bow:

PR #2151, opened 2025-05-16
fix(vcs): git checkout after clone must contains `-- .`

### LINKED ISSUE #1887: copier copy didn't copy my uncommit file even though I added -r HEAD option
### Describe the problem

copy on repository contains uncommit files doesn't works even when added `-r HEAD` option.

### Template

hello {{name}}

### To Reproduce

1. Create empty directory
2. Create **copier.yaml** contains content below

```yaml
_min_copier_version: 9.4.1
_subdirectory: templates
name:
  type: str
  help: What is your name?
```

3. Create **templates** directory and file named **message.txt**
4. The file should contains `hello {{name}}`

```yaml
$PWD/example:
  copier.yaml: |
    _min_copier_version: 9.4.1
    _subdirectory: templates
    name:
      type: str
      help: What is your name?
  templates/message.txt: |
    hello {{name}}
```

5. Run `copier copy . .temp` and it should works normally
6. Run `git init`
7. Rerun `copier copy . .temp` again and it doesn't works anymore
8. Even with `copier copy -r HEAD . .temp`, it doesn't works as well

CleanShot 2024-12-12 at 17 50 31@2x

### Logs

```console
.../copier/9.4.1/venv/lib/python3.12/site-packages/copier/vcs.py:209: DirtyLocalWarning: Dirty template changes included automatically.
  warn(

Copying from template version 0.0.0.post1.dev0+9e08b53
```

### Expected behavior

it copy templates to target directory.

_No response_

### Operating system

macOS

MacOS 15.1.1

### Copier version

copier 9.4.1

### Python version

Python 3.12.2

### Installation method

pip+pypi

### Additional context

_No response_

kamontat:
Without `-- .` on the end cause git to not checkout latest changes from **.git** directory.

CleanShot 2025-05-16 at 16 27 55@2x

[fixes #1887]

### DIFF
--- a/tests/test_vcs.py
+++ b/tests/test_vcs.py
@@ -93,7 +93,7 @@ def test_local_dirty_clone(
     tmp_path_factory: pytest.TempPathFactory, gitconfig: GitConfig
 ) -> None:
     """
-    When core.fsmonitor is enabled, dirty clone required specific git clone command to works properly.
+    When core.fsmonitor is enabled, normal `git checkout` command won't works.
     """
 
     gitconfig.set({"core.fsmonitor": "true"})
--- a/tests/test_vcs.py
+++ b/tests/test_vcs.py
@@ -12,7 +12,7 @@ from copier import run_copy, run_update
 from copier._main import Worker
 from copier._user_data import load_answersfile_data
 from copier._vcs import checkout_latest_tag, clone, get_git_version, get_repo
-from copier.errors import ShallowCloneWarning
+from copier.errors import DirtyLocalWarning, ShallowCloneWarning
 
 from .helpers import git
 
@@ -85,6 +85,26 @@ def test_local_clone() -> None:
     local_tmp = clone(tmp)
     assert local_tmp
     assert Path(local_tmp, "README.md").exists()
+    shutil.rmtree(local_tmp, ignore_errors=True)
+
+
+@pytest.mark.impure
+def test_local_dirty_clone() -> None:
+    tmp = clone("https://github.com/copier-org/autopretty.git")
+    assert tmp
+    assert Path(tmp, "copier.yml").exists()
+
+    with local.cwd(tmp):
+        Path("unknown.txt").write_text("hello world")
+        Path("README.md").write_text("override string!!")
+
+    with pytest.warns(DirtyLocalWarning):
+        local_tmp = clone(tmp)
+
+    assert local_tmp
+    assert Path(local_tmp, "unknown.txt").exists()
+    assert Path(local_tmp, "README.md").exists()
+    assert Path(local_tmp, "copier.yaml").exists()
     shutil.rmtree(local_tmp, ignore_errors=True)
 
 
--- a/tests/test_vcs.py
+++ b/tests/test_vcs.py
@@ -5,9 +5,9 @@ from pathlib import Path
 from typing import Callable
 
 import pytest
-from pytest_gitconfig.plugin import GitConfig
 from packaging.version import Version
 from plumbum import local
+from pytest_gitconfig.plugin import GitConfig
 
 from copier import run_copy, run_update
 from copier._main import Worker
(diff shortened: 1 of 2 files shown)


sisp (changes requested):
Thanks for fixing this bug, @kamontat ! :pray: Would you mind adding a test case?

tests/test_vcs.py
```diff
@@ -88,6 +88,26 @@ def test_local_clone() -> None:
...
+    with pytest.warns(DirtyLocalWarning):
+        local_tmp = clone(tmp)
+
+    assert local_tmp
+    assert Path(local_tmp, "unknown.txt").exists()
+    assert Path(local_tmp, "README.md").exists()
+    assert Path(local_tmp, "copier.yaml").exists()
+    shutil.rmtree(local_tmp, ignore_errors=True)
+
+
```
sisp: Instead of cloning a remote repo, which makes the test impure, could you build a local repo instead? For example something like this (untested):

```python
def test_local_dirty_clone(tmp_path_factory: pytest.TempPathFactory) -> None:
    src = map(tmp_path_factory.mktemp, ("src",))
    build_file_tree({src / "foo.txt": "bar"})
    git_save(src)
    build_file_tree({src / "foo.txt": "baz", src / "hello.txt": "world"})

    with pytest.warns(DirtyLocalWarning):
        tmp = clone(src)

    assert tmp
    assert Path(tmp, "foo.txt").exists()
    assert Path(tmp, "foo.txt").read_text() == "baz"
    assert Path(tmp, "hello.txt").exists()
    assert Path(tmp, "hello.txt").read_text() == "world"
```

sisp (changes requested):
No problem if the error only occurs on macOS, as our CI workflows run on macOS, Linux, and Windows. But it would be good to know whether the new test fails in our macOS CI workflow without the fix. Could you temporarily push only the test case without the fix to verify this? Afterwards, if the test fails, re-add the fix to check that it makes the test pass.

tests/test_vcs.py
kamontat: @sisp I revert the fixes. Please trigger CI.

sisp:
Odd, the test case is passing on CI macOS although the fix is commented out. So, the test case doesn't seem to reproduce the problem correctly. Does the test fail on your local macOS?

kamontat:
@sisp Do you need any help from me? I would love get this merge because I have to manually update copier code every time I install it.

sisp:
@kamontat Sorry for the delay. I was hoping we'd find a way to test this on the macOS runner. I find it a little suspicious why running the Copier CLI is producing this error for you while the test isn't. :thinking: In the test suite, we're configuring an [isolated Git config](https://github.com/copier-org/copier/blob/457a357b6a84cf962c37fb29020209108e626791/tests/conftest.py#L56-L69) to avoid side effects from the global/system Git config. I wonder whether this might explain the difference on your machine. Not sure whether the test suite will run properly, but could you try commenting out the `default_gitconfig` fixture and running only your added test case?

sisp:
Tests on macOS and Windows are failing now – great! :tada: Thanks for figuring this out! :bow:

You can enable the fix now, and when the tests pass, we can merge it.

kamontat:
@sisp Apply hopefully the ci will pass. Thank you for your help.

kamontat:
> Now, several other tests are failing. 🤔

Any suggestion?

kamontat:
@sisp I need `core.fsmonitor` on some of my project as it significantly improve on performance.
Is it ok, if I implement try-then-fallback if repo is not clone correctly?

So it will be...
- First then without --, then check
- If check failed, try again with --; then check
- If check failed again, then return error

sisp:
Looks like this is working. :+1: Could you add an inline comment above that line with a link to this issue, so we'll remember why the `core.fsmonitor=false` setting is used there? Then, I think we can merge this PR. :tada:

PR #2158, opened 2025-05-19
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

TODOs
- [x] remove `TEMP` commit after confirming that CI picks up typos

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,6 +180,13 @@ tag_format = "v$version"
 update_changelog_on_bump = true
 version = "9.7.1"
 
+[tool.codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = '.git*,*.svg,*.lock,*.css,pyproject.toml,.mypy_cache,.venv'
+check-hidden = true
+ignore-regex = '\bla vie\b'
+ignore-words-list = 'ans'
+
 [tool.hatch.version]
 source = "vcs"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ copier = "copier.__main__:CopierApp.run"
 
 [dependency-groups]
 dev = [
+  "codespell[toml]==2.4.1",
   "commitizen==4.8.0",
   "mypy==1.15.0",
   "pexpect==4.9.0",
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -311,7 +311,7 @@ class Worker:
 
         # Given those values are lazily rendered on 1st access then cached
         # the phase value is irrelevant and could be misleading.
-        # As a consequence it is explicitely set to "undefined".
+        # As a consequence it is explicitly set to "undefined".
         return LazyDict(
             {
                 name: lambda path=path: load_answersfile_data(  # type: ignore[misc]
--- a/copier/_user_data.py
+++ b/copier/_user_data.py
@@ -290,7 +290,7 @@ class Question:
 
         The questionary lib expects some specific data types, and returns
         it when the user answers. Sometimes you need to compare the response
-        to the rendered one, or viceversa.
+        to the rendered one, or vice-versa.
 
         This helper allows such usages.
         """
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -975,7 +975,7 @@ Each pattern can be templated using Jinja.
     ```
 
     The difference with [skip_if_exists][] is that it will never be rendered during
-    an update, no matter if it exitsts or not.
+    an update, no matter if it exists or not.
 
 !!! info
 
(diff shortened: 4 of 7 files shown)


sisp (changes requested):
Thanks for your initiative to avoid spelling errors by introducing codespell, @yarikoptic! :+1:

We're already using [pre-commit](https://pre-commit.com/) (via [devenv](https://github.com/copier-org/copier/blob/7ccb4247af407693319fbfa4a32d968ab61c8a95/flake.nix#L96-L136), which uses [git-hooks.nix](https://github.com/cachix/git-hooks.nix) internally) to run Git hooks, and we have an [autofix GHA workflow](https://github.com/copier-org/copier/blob/7ccb4247af407693319fbfa4a32d968ab61c8a95/.github/workflows/autofix.yml) that commits and pushes changes applied by pre-commit. I think it's preferable to add codespell as a pre-commit hook to leverage the existing autofix workflow and provide spellchecks not only via CI but also during local development.

devenv doesn't offer codespell as a builtin hook, but you can add a [custom hook](https://devenv.sh/reference/options/#git-hookshooks). Derived from the [official codespell pre-commit hook](https://github.com/codespell-project/codespell/blob/ed6185a67247989111d989498d44c81390dab043/.pre-commit-hooks.yaml), I think this is it (untested):

```nix
pre-commit.hooks = {
  # ...
  codespell = {
    enable = true;
    package = null;
    name = "codespell";
    description = "Checks for common misspellings in text files.";
    entry = "uv run codespell";
    types = ["text"];
  };
  # ...
};
```

For this, you need to add codespell as a [dev dependency in `pyproject.toml`](https://github.com/copier-org/copier/blob/7ccb4247af407693319fbfa4a32d968ab61c8a95/pyproject.toml#L47-L64) via uv.

yarikoptic:
uv newb here asking -- `uv add --dev codespell==2.4.1` results in massive change to uv.lock, primarily due to removal of currently present in many entries ` upload-time = ` . I think that is undesired. Is there some option to keep those or smth like that? (uv 0.6.14+15 (6b7f60c1e 2025-04-12))

sisp:
That's because you're using a slightly outdated version of uv. A more recent version uses revision 2 of the lock file schema. Could you try updating uv?

pyproject.toml
```diff
@@ -45,6 +45,7 @@ copier = "copier.__main__:CopierApp.run"
 
 [dependency-groups]
 dev = [
+  "codespell==2.4.1",
```
sisp, line 48: We need to install the [`toml` extra](https://github.com/codespell-project/codespell/blob/ed6185a67247989111d989498d44c81390dab043/pyproject.toml#L50-L52), as codespell relies on `tomli` for Python <3.11 when sourcing the config from `pyproject.toml` (which I prefer, so all good).

```suggestion
  "codespell[toml]==2.4.1",
```

Don't forget to update the the lock file.

pyproject.toml
```diff
@@ -186,3 +187,10 @@ source = "vcs"
 [build-system]
 requires = ["hatchling", "hatch-vcs"]
 build-backend = "hatchling.build"
+
+[tool.codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = '.git*,*.svg,*.lock,*.css,pyproject.toml,.mypy_cache,.venv'
+check-hidden = true
+ignore-regex = '\bla vie\b'
+ignore-words-list = 'ans'
```
sisp, line 196: Just a cosmetic request: Could you please move the `[tool.codespell]` section further up, e.g. after the `[tool.commitizen]` section, so these `[tool.*]` sections are grouped? `[tool.hatch.*]` is a bit special because it's related to `[build-system]`, so I'd keep those two close.

sisp (changes requested):
Looks pretty good. Just a few minor requests from my side.

yarikoptic:
ok, rebased, redid addition, force-pushed. I still left that intentional typo in to see if CI picks it up, so not to be merged, will move into draft

sisp:
It seems to be working as expected:

```
--- devenv:git-hooks:run failed with error: Task exited with status: exit status: 1
--- devenv:git-hooks:run stdout:
0002.25: alejandra................................................................Passed
0001.95: codespell................................................................Failed
0001.95: - hook id: codespell
0001.95: - exit code: 65
0001.95: 
0001.95: CONTRIBUTING.md:227: CONTRIBUING ==> CONTRIBUTING
0001.95: 
0001.72: editorconfig-checker.....................................................Passed
0000.16: prettier.................................................................Passed
0000.10: ruff.....................................................................Passed
0000.04: ruff-format..............................................................Passed
0000.02: taplo....................................................................Passed
--- devenv:git-hooks:run stderr:
```

yarikoptic:
then we must be ready!

Next step for me is finally to learn some `copier` ;-)

PR #2165, opened 2025-05-23
doc: add inheritance and composition section in user docs

francesco086:
Contributes to https://github.com/copier-org/copier/issues/934


docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
+# Template Inheritance and Composition
+
+## What is Inheritance, what is Composition?
```
sisp, line 3: Perhaps this is more concise?

```suggestion
## Inheritance vs. Composition
```

docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
...
+
+Template Inheritance is a way to create a new template based on an existing one. It
```
sisp, line 5: * I'd not capitalize "Inheritance" in a regular sentence.
* How about introducing "parent" and "child" in the first sentence to clarify the relationship for later use?

```suggestion
Template inheritance is a way to create a new (child) template based on an existing (parent) template. It
```

docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
...
+allows you to reuse code and structure from the parent template while adding or
+overriding specific parts in the child template. This is useful for creating variations
```
sisp, line 7: I think it's mostly about _extensions_, not so much _variations_ – perhaps a matter of definition. Variations – in my mind – can often be addressed by conditional questions and conditional rendering in the same template. WDYT?

```suggestion
overriding specific parts in the child template. This is useful for creating extensions
```

docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
...
+of a template without duplicating code.
+
+<!-- TODO: add an example of inheritance -->
+
+Template Composition, on the other hand, is a way to combine multiple templates,
```
sisp, line 12: Like above:

```suggestion
Template composition, on the other hand, is a way to combine multiple templates,
```

docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
...
+independent of each other, into a single new template. In this way you can join the
+functionalities of multiple templates, by adding some "glue code", and avoid repetition.
```
sisp, line 14: Do you have an approach like yours using tasks that run `copier copy` in mind? I've mostly thought of "composition" along the lines of https://copier.readthedocs.io/en/stable/configuring/#applying-multiple-templates-to-the-same-subproject. There, template composition means generating output from multiple templates in the same target directory, but there is no new template that is a composition of other templates.

docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
...
+
+<!-- TODO? (not sure if it relevant...) add an operational distinction between the two -->
+
+## Why copier does not natively support inheritance and composition?
+
+<!-- TODO: add content, re-use content from https://github.com/copier-org/copier/issues/934#issuecomment-1518964035 -->
+
+## Solutions
+
+### Git submodules + jinja template inheritance
```
sisp, line 26:
```suggestion
### Git submodules + Jinja template inheritance
```

:nerd_face:

docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
...
+
+<!-- TODO: Add content-->
+
+### Fork-like approach
+
+<!-- TODO: Add content-->
+
+### Leverage on copier task
```
sisp, line 34: Perhaps something like this?

```suggestion
### Delegated `copier copy` calls via tasks
```

Not sure ... :thinking:

mkdocs.yml
```diff
@@ -12,6 +12,7 @@ nav:
   - Generating a project: "generating.md"
   - Updating a project: "updating.md"
   - Settings: "settings.md"
+  - Template Inheritance and Compostion: "inheritance_and_composition.md"
```
sisp, line 15: See above:

```suggestion
  - Reusing templates: "reusing.md"
```

docs/inheritance_and_composition.md
```diff
@@ -0,0 +1,36 @@
...
+independent of each other, into a single new template. In this way you can join the
+functionalities of multiple templates, by adding some "glue code", and avoid repetition.
+
+<!-- TODO: add an example of composition -->
+
+<!-- TODO? (not sure if it relevant...) add an operational distinction between the two -->
+
+## Why copier does not natively support inheritance and composition?
+
+<!-- TODO: add content, re-use content from https://github.com/copier-org/copier/issues/934#issuecomment-1518964035 -->
```
sisp, line 22: I anticipate that a proper solution might require Copier to support it natively. At least for the fork-like workflow, I think that specified template versions in parent template migrations need to be mapped to child template versions – at the very least. But it's too early to make a definitive statement.

sisp (commented):
Thanks for volunteering to write documentation on this complex topic, @francesco086! :pray: I have a few suggestions, curious what you think.

Regarding solutions: [Meta-templates](https://github.com/copier-org/copier/issues/934#issuecomment-1405492635) is another one, although I don't know how relevant it actually is in practice.

francesco086:
> Regarding solutions: [Meta-templates](https://github.com/copier-org/copier/issues/934#issuecomment-1405492635) is another one, although I don't know how relevant it actually is in practice.

Right! I think it is not really a solution for inheritance and composition tbh, but I added it to the list, and we will discuss it at due time

docs/inheritance_and_composition.md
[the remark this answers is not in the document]
francesco086: My feeling is that `reusing` is too generic, and it is especially confusing in the context of copier. The first thing that would come to my mind is "Of course! I have some code that I want to re-use by creating a template! Let's see how".

What about `composing`, `composing_templates`, or `inheriting_and_composing_templates` (too long perhaps)? Alternatively, `reusing_templates` may be explicit enough.

I leave the final decision to you / the maintainers, just let me know what do you think is best and I will apply 😄

docs/inheritance_and_composition.md
francesco086: Reading carefully what he wrote once again, it seems to me that everything reduces to the UNIX philosophy of "do one thing well", or am I wrong?

sisp:
Thanks for the reminder, @francesco086! :+1:

I definitely see some advantages in the task/migrations-based approach, especially the encapsulation of the parent template. But I struggle with the implementation of a simple scenario like this: Imagine you have a generic Python project template (parent) and a Python FastAPI template (child). Consider, e.g., the `pyproject.toml.jinja` file with PEP 621 project metadata where the parent template declares `dependencies = []` (because it makes no assumptions about runtime dependencies) and the child template declares `dependencies = ["fastapi[standard]"]` (because it generates starter code for a FastAPI app). IIUC, the task/migrations-based approach can only exclude the `pyproject.toml.jinja` file from the parent template and redefine it in the child template although (for the sake of this example) the only difference is the one line about runtime dependencies. Am I right?

francesco086:
Hey @sisp !

Yes, you are absolutely right. 

One would make a copy-paste (or git sparse clone) from the parent, and modify it. The real problem would come with the parents upgrade (notice that since the parent version is pinned, this is under manual control). You would have to do the same procedure mentioned before. You could use diff tools to help yourself, but eventually it would be a manual process. Frankly I don't think one can do much better. In these scenarios usually git gives you conflicts to resolve manually. But I may be mistaken.

In general, I think one should try to avoid the overwriting of a file, so it should be limited to as few files as possible. For example, in the use case you mentioned, one could use dynamic dependencies and only overwrite a requirements file (of course this requires being able to influence the parent template, which is not always granted)
```
[project]
dynamic = ["dependencies"]

[tool.setuptools.dynamic]
dependencies = { file = ["requirements.txt"] }
```
This to say that I think that this drawback has not such strong impact.

francesco086:
Do you have a reference of this fork-like approach for me to document it? It could be my next method to document.

I will also add this limitation you pointed out to the cons.

Issue #1170, opened 2023-05-31
Updating a project fails when Jinja extension with breaking change is used


sisp:
### Describe the problem

Consider the following situation:

1. A template uses v1 of a Jinja extension.
2. A project is generated using this template.
3. A breaking change happens in the Jinja extension which gets released as v2.
4. The template adapts to v2 of the Jinja extension and gets released.
5. The project is updated using the new version of the template. **This may not be possible.**

So what's the problem? Copier's update algorithm re-generates a fresh project using the _old_ version of the template (in this case, the one that uses v1 of the Jinja extension) and also generates a fresh project using the _new_ version of the template (in this case, the one that uses v2 of the Jinja extension). But it is only possible to install one version of a package at a time in a virtual env, either v1 or v2 of the Jinja extension. So when v1 is installed, re-generating a fresh project using the _old_ version of the template works but generating a fresh project using the _new_ version of the template fails. And when v2 is installed, generating a fresh project using the _new_ version of the template works but re-generating a fresh project using the _old_ version of the template fails.

### Template

See the next section for a complete test case.

### To Reproduce

Add this test case to a new file, e.g. `tests/test_jinja_extension_breaking_change.py`, in Copier's test suite:

```python
import sys
from pathlib import Path
from textwrap import dedent

import pytest
from plumbum import local
from plumbum.cmd import git

from copier.main import run_copy, run_update

from .helpers import build_file_tree


def test_update_with_jinja_extension_breaking_change(
    tmp_path_factory: pytest.TempPathFactory, monkeypatch: pytest.MonkeyPatch
) -> None:
    src, dst, ext = map(tmp_path_factory.mktemp, ("src", "dst", "ext"))

    monkeypatch.setattr("sys.path", [str(ext)] + sys.path)

    with local.cwd(ext):
        Path("test_ext").mkdir()
        Path("test_ext", "__init__.py").write_text(
            dedent(
                """\
                from jinja2.ext import Extension

                def to_slug(value):
                    return value.replace(" ", "-")

                class TestExt(Extension):
                    def __init__(self, environment):
                        super().__init__(environment)
                        environment.filters["to_slug"] = to_slug
                """
            )
        )

    with local.cwd(src):
        build_file_tree(
            {
                "copier.yml": (
                    """\
                    _jinja_extensions:
                        - test_ext.TestExt

                    version:
                        type: str
                    """
                ),
[...]

PR #2187, opened 2025-06-12
fix(external_data): replay old external data before updating

yajo:
Before this fix, when a template used local `_external_data`, it could never again be cleanly updated. It always gave conflicts.

This was happening because the old worker that did the replay always happened on an empty directory. Thus it never had external data available. Since external data can alter the render result (just like any other answer), the replay was never reliable.

Now, new data is saved in the answers file to be able to reliably replay including that context. Then, updates can work again as expected and result in meaningful diffs only when there should be.

BREAKING CHANGE: A new `_dst_commit` is added to the answers file, pointing to the commit just before the last template update. This is needed for reliable replays. This means that any copy or update operation will always produce diff when the subproject is git-tracked.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -3,7 +3,7 @@
   "configurations": [
     {
       "name": "Python: debug tests",
-      "type": "python",
+      "type": "debugpy",
       "request": "launch",
       "program": "${file}",
       "console": "internalConsole",
--- a/copier/_cli.py
+++ b/copier/_cli.py
@@ -368,7 +368,7 @@ class CopierUpdateSubApp(_Subcommand):
         from the last Copier execution, including the source template
         (it must be a key called `_src_path`).
 
-        If that file contains also `_commit`, and `destination_path` is a git
+        If that file contains also `_commit`, `_dst_commit` and `destination_path` is a git
         repository, this command will do its best to respect the diff that you have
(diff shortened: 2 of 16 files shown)


yajo:
Indeed, that would solve this too.

yajo:
I found out one way to workaround the issue.

Set up `copier.yaml` like this:

```yml
# Get external data from other template, or any other file
_external_data:
  parent_template: .copier-answers.yml

# Use a computed variable for the data you actually need from there, declared like this
version:
  default: "{{ _external_data.parent_template.version | d(last_version) }}"
  when: false
  type: float
```

Then set up `{{ _copier_conf.answers_file }}.jinja` like this:

```yaml
# Changes here will be overwritten by Copier; do NOT edit manually

# Smoother replays. Notice that `last_version` is NOT in the questionary
last_version: {{ version }}
```

Why this works?
1. When copying for the 1st time, `version` gets its value from the external data, which is available in that moment.
1. When updating, Copier will fail to get the external data because it does not have a notion about when the last copy was done. Thus it will fall back to using the value recorded in `last_version`.
2. After rendering, `last_version` will be recorded as the current version of that render. Thus, next renders will know the last value.

This works for me without introducing any backwards incompatibility, and the day #1170 gets fixed it will also fix this issue. Thus, closing.

PR #2189, opened 2025-06-13
feat: add support for providing serialized answers to multiselect choice questions

### LINKED ISSUE #1474: Using --data with a multiselect based choice

sisp:
I've added support for providing serialized answers to multiselect choice questions.

This feature enables two use cases:

1. The default value can be templated by rendering default choice items as a YAML-style list.
2. An answer can be passed via the CLI flag `--data` as a YAML-style list. Resolves #1474.

The answer to a multiselect choice question is a list of choice items. While it seems that the most obvious solution is to parse the answer using a YAML parser, I think it's not the right approach because the list items should be parsed according to `type: <type>` and not necessarily using a YAML parser, and it's inconsistent with other question types to first parse them as YAML and afterwards parse according to `type: <type>`. But I've found a way to shallowly parse the serialized answer as a YAML-style list without parsing the items. This way, the items remain raw strings, delegating their parsing to the parser according to `type: <type>`. There is one subtlety: YAML strings may be quoted or (sometimes) unquoted. To avoid double-quoting after the shallow list parsing, the quotes of a YAML string are stripped.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -434,3 +434,41 @@ def test_skip_fileexists(template_path: str, tmp_path: Path) -> None:
     assert a_txt.read_text() == "PREVIOUS_CONTENT"
     answers = load_answersfile_data(tmp_path, "altered-answers.yml")
     assert answers["_src_path"] == str(template_path)
+
+
+def test_data_with_multiselect_choice_question(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+
+    build_file_tree(
+        {
+            (src / "copier.yml"): (
+                """\
+                q:
+                    type: int
+                    choices: [1, 2, 3]
+                    multiselect: true
+                """
+            ),
+            (src / "{{ _copier_conf.answers_file }}.jinja"): (
+                """\
+                # Changes here will be overwritten by Copier
+                {{ _copier_answers|to_nice_yaml }}
+                """
+            ),
+        }
+    )
+
+    run_result = CopierApp.run(
+        [
+            "copier",
+            "copy",
+            "--data=q=[1, 3]",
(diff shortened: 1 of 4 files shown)


tests/test_cli.py
```diff
@@ -434,3 +434,41 @@ def test_skip_fileexists(template_path: str, tmp_path: Path) -> None:
...
+                """
+            ),
+        }
+    )
+
+    run_result = CopierApp.run(
+        [
+            "copier",
+            "copy",
+            "--data=q=[1, 3]",
```
yajo, line 467: Suggestion: document this

sisp: What do you mean exactly? I've added documentation for this feature in the docs. Do you mean an inline comment here?

sisp: Ah, sorry, I wasn't thinking straight. You're suggesting to document passing multiselect answers via CLI. :bulb:

Issue #1077, opened 2023-04-04
Copier should support .yaml extension  for .copier-answers.yml


sandeeps311:
Hi Team,
While using a copier update im getting this error 
**Cannot update because cannot obtain old template references from `.copier-answers.yml`.**
I think copier only supports .yml extension for copier-answers.yml,
Pleas check below code reference 
https://github.com/copier-org/copier/blob/17135c5f346bb2daad32c40ae43f8790ac50c270/copier/main.py#L594

ahgraber:
For additional info, https://copier.readthedocs.io/en/latest/configuring/#answers_file implies that the answers file is user-configurable

yajo:
If you're not using the default file name, then you must provide it with the `-a` flag.

sandeeps311:
It would be nice if you added support for both file extensions ".yml" and ".yaml" because there is no difference in functionality or syntax between the two extensions.

yajo:
Sadly, the YAML project only decided that `.yaml` is the standard file extension after Copier was invented. Thus, we keep this old default. But you can choose a different default as long as you use it in your CLI, if that missing "a" is important for you. 😃

PR #2196, opened 2025-06-17
Allow for .copier-answers.yaml to be default in backwards-compatible way

timkpaine:
This is a very small backwards-compatible change.

Copier uses `.copier-answers.yml` as the default, and supports overriding. As discussed in https://github.com/copier-org/copier/issues/1077 `.yml` is the default suffix, not `.yaml`. It makes sense to not change the default, but perhaps it would be ok to start to check `.copier-answers.yaml`? This PR does that very small change, if the user has decided to use the default answers file but with `.yaml` suffix instead of `.yml` suffix, `copier update` will work without `-a`.


copier/_main.py
```diff
@@ -973,9 +973,10 @@ def resolved_vcs_ref(self) -> str | None:
     @cached_property
     def subproject(self) -> Subproject:
         """Get related subproject."""
+        default_answers_file = Path(".copier-answers.yaml") if Path(".copier-answers.yaml").exists() else Path(".copier-answers.yml")
```
sisp: This won't work because the `Path` object passed to the `answers_relpath` argument of `Subproject(local_abspath=..., answers_relpath=...`) is the path of the answers file relative to `local_abspath`. `Path(".copier-answers.yaml").exists()` checks whether `.copier-answers.yaml` exists in the current working directory, which is not guaranteed to be `local_abspath`.

sisp (changes requested):
I'm open to supporting the `.yaml` suffix in a backwards-compatible way, but I'm afraid the solution is a little more involved.

1. See my inline comment about your suggested change.
2. The default answers file name/path is hardcoded in a few places: https://github.com/search?q=repo%3Acopier-org%2Fcopier+%22.copier-answers.yml%22+path%3A%2F%5Ecopier%5C%2F%2F&type=code This should be refactored anyway, but with a primary and secondary file name/path, we can't simply use a default value, as there isn't a single one anymore but rather some logic that first tries reading the primary file name/path and falls back to the secondary when the primary doesn't exist. Perhaps the simplest solution is to encode the default answers file name/path as `None` and only resolve `None` as `.copier-answers.yml` or `.copier-answers.yaml` in [`copier/_user_data.py::load_answersfile_data`](https://github.com/copier-org/copier/blob/680365c5b735dc7ee3258048143c35a46c979797/copier/_user_data.py#L526-L542).

Whatever the solution looks like in the end, I have two requests:
1. I think we should check for the existence of both files (`.yml` suffix and `.yaml` suffix) and raise a warning if both exist, just in case, e.g., a project was migrated from `.yaml` to `.yml` but forgot to remove the `.yaml` file.
1. Please add tests.

WDYT?

timkpaine:
> 1. I think the `.yml` suffix should take precedence over the `.yaml` suffix for maximum backwards compatibility.

Sounds good, unfortunately it will be slightly annoying as we'll have to make sure to look in the "right direction", e.g. in some places you want to check `.yml` first then default to `.yaml`, like when reading, but in others you want to check `.yaml` first then default to `.yml`, like when writing.

> 2. I think we should check for the existence of both files (`.yml` suffix and `.yaml` suffix) and raise a warning if both exist, just in case, e.g., a project was migrated from `.yaml` to `.yml` but forgot to remove the `.yaml` file.

Makes sense to me.

> 3. Please add tests.

Will do

pawamoy:
Awesome presentation @timkpaine! Always nice to see our work presented in conferences :heart_eyes:

tests/test_yaml_answersfile.py
```diff
@@ -0,0 +1,251 @@
...
+                    default: test
+                """
+            ),
+            (root / "{{ _copier_conf.answers_file }}.jinja"): (
+                "{{ _copier_answers|tojson }}"
+            ),
+        }
+    )
+    git_save(root)
+    return root
```
sisp: How about moving this fixture to the `TestCopyWithYamlAnswersFile` class below? This is the only place where it is used.

tests/test_yaml_answersfile.py
sisp, line 1: How about moving the contents of this file to the existing `tests/test_answersfile.py` file to avoid scattering somewhat related tests across multiple files?

sisp, line 1: I acknowledge your efforts to test the answers file resolution thoroughly. But I notice quite a few tests involving rather internal classes and functions which are likely subject to change because of internal refactoring once the public API has been cleaned up. What really matters is the intended answers file resolution behavior for a few user-level scenarios which can be tested with `run_{recopy,update}` calls. How about limiting tests to those scenarios and removing the tests involving non-public API?

copier/_user_data.py
```diff
@@ -585,11 +589,50 @@ def parse_yaml_list(string: str) -> list[str]:
...
+            )
+
+        # .yml takes precedence, fall back to .yaml
+        if yml_exists:
+            answers_file = DEFAULT_ANSWERS_FILE_YML
+        elif yaml_exists:
+            answers_file = DEFAULT_ANSWERS_FILE_YAML
+        else:
+            # Default to .yml when neither exists
+            answers_file = DEFAULT_ANSWERS_FILE_YML
```
sisp: I think we can call `resolve_answersfile_path` here to deduplicate this code block.

sisp (changes requested):
Thanks for reminding me of this PR, @timkpaine! :+1: And sorry for the delay ... 🫣 

I've left a couple of inline comments.

PR #2210, opened 2025-07-03
feat: add support for prompting filesystem paths

sseemayer:
Users might want to prompt for paths in their filesystem with tab completion.

Implement this by creating a new question type "path" that translates to a questionary type "path" in a trivial way since tab completion is already available in that package.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -3,16 +3,15 @@ from __future__ import annotations
 import filecmp
 import json
 import os
-import platform
 import sys
 import textwrap
 from collections.abc import Mapping
 from enum import Enum
 from hashlib import sha1
 from pathlib import Path
-from pexpect.popen_spawn import PopenSpawn
-from typing import Any, Protocol, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Protocol
 
+from pexpect.popen_spawn import PopenSpawn
 from plumbum import local
 from plumbum.cmd import git as _git
 from plumbum.machines import LocalCommand
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -11,7 +11,6 @@ import pexpect
 import pytest
 import yaml
 from coverage.tracer import CTracer
-
 from pexpect.popen_spawn import PopenSpawn
 from plumbum import local
 


sseemayer:
>Do you have a specific use case (in mind) for path questions? I'd be curious to learn about it to get a better feeling for the scenarios in which it may be useful.

At BASF, we use copier to scaffold various bioinformatic workflows written in [Snakemake](https://snakemake.readthedocs.io/en/stable/), i.e. having a generic template that gets instantiated for an individual analysis. These workflows often take files as an input, e.g. the paths at which genetic sequencing data is located, and we copy the files into the right place using tasks. Since we are on a [high performance compute environment ](https://www.basf.com/global/en/who-we-are/innovation/how-we-innovate/our-RnD/Digitalization_in_R-D/supercomputer) with a highly nested folder structure, being able to tab-complete while searching for these paths greatly improves usability.

I will try to make progress on documentation and tests today.

copier/_user_data.py
```diff
@@ -427,6 +427,8 @@ def _validate(answer: str) -> str | Literal[True]:
             result["multiline"] = self.get_multiline()
             if placeholder := self.get_placeholder():
                 result["placeholder"] = placeholder
+        if type_name == "path":
+            questionary_type = "path"
         if questionary_type in {"input", "checkbox", "password"}:
```
hunter-cameron: I think the new "path" type would also need to be added here to attach the validators.

sseemayer: good point, added in 90512a2

sisp (commented):
Thanks for elaborating on your use case and your diligent implementation of tests including even tab completion! :1st_place_medal: :bow:

Regarding the tab completion test: It appears that `pexpect.popen_spawn.PopenSpawn` does not support such advanced interaction. TBH, I'm not familiar with these details of testing terminal interaction, so I asked a popular chatbot about it that replied as follows:

> `PopenSpawn` uses `subprocess.Popen` under the hood. It does not allocate a pseudo-terminal (PTY) by default. But  `prompt_toolkit` expects to run in a TTY (interactive terminal) to enable autocompletion, display, and cursor management. This means: using `PopenSpawn` with `prompt_toolkit` will often disable or break autocompletion and interactive behaviors, because the app thinks it's not running in a terminal.

We use `questionary` for prompting, which builds on top of `prompt-toolkit`, so assuming this chatbot isn't hallucinating, it sounds like our use of `PopenSpawn` is the problem.

To solve this problem, I suggest to use `pexpect.spawn` instead which does work for the tab completion test:

```diff
diff --git a/tests/test_prompt.py b/tests/test_prompt.py
index afc8eab..48bec80 100644
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -10,6 +10,8 @@ from typing import Any, Protocol, Union
 import pexpect
 import pytest
 import yaml
+from coverage.tracer import CTracer
+from pexpect import spawn as pexpect_spawn
 from pexpect.popen_spawn import PopenSpawn
 from plumbum import local
 
@@ -174,10 +176,11 @@ def test_copy_default_advertised(
         assert load_answersfile_data(".").get("_commit") == "v2"
 
 
-def test_path_completion(
-    tmp_path_factory: pytest.TempPathFactory,
-    spawn: Spawn,
-) -> None:
+@pytest.mark.skipif(
+    condition=platform.system() == "Windows",
+    reason="pexpect.spawn does not work on Windows",
+)
+def test_path_completion(tmp_path_factory: pytest.TempPathFactory) -> None:
     """Test that file paths can handle tab completion"""
     src, dst, completedir = map(tmp_path_factory.mktemp, ("src", "dst", "my-directory"))
     with local.cwd(src):
@@ -186,10 +189,13 @@ def test_path_completion(
         git("commit", "--allow-empty", "-m", "v2")
         git("tag", "v2")
     with local.cwd(dst):
+        # Disable subprocess timeout if debugging (except coverage), for commodity
+        # See https://stackoverflow.com/a/67065084/1468388
+        tracer = getattr(sys, "gettrace", lambda: None)()
+        timeout = 10 if not isinstance(tracer, (CTracer, type(None))) else None
         # Copy the v1 template
-        tui = spawn(
-            COPIER_PATH + ("copy", str(src), ".", "--vcs-ref=v1"), timeout=10
-        )
+        cmd = COPIER_PATH + ("copy", str(src), ".", "--vcs-ref=v1")
+        tui = pexpect_spawn(cmd[0], list(cmd[1:]), timeout=timeout)
 
         # Check that default values are maintained
         expect_prompt(tui, "current_location", "path")
```

We can't replace `pexpect.popen_spawn.PopenSpawn` by `pexpect.spawn` altogether, even though [interactive prompt tests are skipped on Windows](https://github.com/copier-org/copier/blob/ecdcae4cf7791af4b9f192788f4b27ec4b349e9e/tests/conftest.py#L22-L27) at the moment, because a few tests are failing with `pexpect.spawn` with an error "Your terminal doesn't support cursor position requests (CPR).".

One important note about path questions: Be careful using these answers in rendered files and with recording them in `.copier-answers.yml` because paths are often not portable across machines. It sounds like you're using answers to path questions only for tasks. In this case, it might be a good idea to use [secret questions](https://copier.readthedocs.io/en/stable/configuring/#secret_questions) to avoid recording answers in the answers file.

sseemayer:
@sisp Thanks a lot for your help! I've implemented your suggested changes and adjusted the type annotation for the `expect_prompt` function. With this, all tests pass now for me locally 🚀 

The pyright-based type checking in my text editor still complains that `timeout` might be `None` when passing into `pexpect_spawn`, because it infers from the [`pexpect.spawn` constructor that that variable defaults to 30 and thus must be int-typed](https://github.com/pexpect/pexpect/blob/fc8f062518b40bd0862aae870cdedf5d9c0c7fc3/pexpect/pty_spawn.py#L36), even though the implementation shows that None is handled correctly. I wanted to ignore that warning using a `# type: ignore` comment but [`test_tools` prevents me from doing so](https://github.com/copier-org/copier/blob/457a357b6a84cf962c37fb29020209108e626791/tests/test_tools.py#L16), so I have kept it as-is.

sseemayer:
rebased onto the current master branch.

tests/test_prompt.py
```diff
@@ -156,6 +177,47 @@ def test_copy_default_advertised(
         assert load_answersfile_data(".").get("_commit") == "v2"
 
 
+@pytest.mark.skipif(
+    condition=platform.system() == "Windows",
+    reason="pexpect.spawn does not work on Windows",
+)
+def test_path_completion(tmp_path_factory: pytest.TempPathFactory) -> None:
+    """Test that file paths can handle tab completion"""
```
sisp:
```suggestion
    """Test that file paths can handle tab completion."""
```

:nerd_face:

tests/test_prompt.py
```diff
@@ -156,6 +177,47 @@ def test_copy_default_advertised(
...
+    src, dst, completedir = map(tmp_path_factory.mktemp, ("src", "dst", "my-directory"))
+    with local.cwd(src):
+        build_file_tree(PATH_TREE)
+        git_save(tag="v1")
+        git("commit", "--allow-empty", "-m", "v2")
+        git("tag", "v2")
+    with local.cwd(dst):
+
```
sisp:
```suggestion
```

:nerd_face:

tests/test_prompt.py
```diff
@@ -156,6 +177,47 @@ def test_copy_default_advertised(
...
+        # Copy the v1 template
+        cmd = COPIER_PATH + ("copy", str(src), ".", "--vcs-ref=v1")
+        tui = pexpect_spawn(cmd[0], list(cmd[1:]), timeout=timeout)
+
+        # Check that default values are maintained
+        expect_prompt(tui, "current_location", "path")
+        tui.sendline()
+
+        # check tab completion of a filesystem path (/path/to/my-direct<TAB> should complete to /path/to/my-directory0)
```
sisp:
```suggestion
        # Check tab completion of a filesystem path (/path/to/my-direct<TAB> should
        # complete to /path/to/my-directory0)
```

:nerd_face:

tests/test_prompt.py
```diff
@@ -156,6 +177,47 @@ def test_copy_default_advertised(
...
+        expect_prompt(tui, "star_location", "path", help="Location of a bonus star")
+        tui.send(str(completedir)[:-4])
+        tui.send(Keyboard.Tab)
+        tui.sendline()
+        tui.sendline()
```
sisp, line 211: Is this second `tui.sendline()` needed? (I haven't run the test case to check ...)

tests/test_prompt.py
```diff
@@ -156,6 +177,47 @@ def test_copy_default_advertised(
...
+
+        tui.expect_exact(pexpect.EOF)
+
+        assert load_answersfile_data(".").get("_commit") == "v1"
+        assert load_answersfile_data(".").get("current_location") == "/dev/warppipe0"
+        assert load_answersfile_data(".").get("star_location") == str(completedir)
```
sisp: Just a minor remark – I'd load the answers file only once:

```suggestion
        answers = load_answersfile_data(".")
        assert answers.get("_commit") == "v1"
        assert answers.get("current_location") == "/dev/warppipe0"
        assert answers.get("star_location") == str(completedir)
```

sisp (changes requested):
Looks great, @sseemayer! :1st_place_medal: I only have some minor polishing remarks. :nerd_face:

tests/test_prompt.py
sseemayer: yes - without it, the test waits indefinitely.

tests/helpers.py
```diff
@@ -11,6 +11,7 @@
 from pathlib import Path
 from typing import Any, Protocol
 
+from pexpect import spawn as pexpect_spawn
```
sisp: It seems [`pexpext.spawn` is exported only on Unix platforms](https://github.com/pexpect/pexpect/blob/fc8f062518b40bd0862aae870cdedf5d9c0c7fc3/pexpect/__init__.py#L79-L82). As the test case is skipped on Windows, I think it's sufficient to fix the import error by importing from `pexpect.pty_spawn` instead:

```suggestion
from pexpect.pty_spawn import spawn as pexpect_spawn
```

sseemayer: Ah, I came up with a different solution but yours is better.

tests/helpers.py
```diff
@@ -11,6 +12,12 @@
 from pathlib import Path
 from typing import Any, Protocol
 
+try:
+    from pexpect.pty_spawn import spawn as pexpect_spawn
+except ModuleNotFoundError as e:
+    # module does not exist on windows. ignore windows failures only
+    if platform.system() != "Windows":
+        raise e
```
sisp: Is mypy happy with using `pexpect.spawnbase.SpawnBase` instead of `PopenSpawn | pexpect_spawn` below? In that case, we could simplify this import to:

```python
if TYPE_CHECKING:
    from pexpect.spawnbase import SpawnBase
```

sseemayer: again, nice idea! All tests pass for me locally with this.

sisp (approved):
Fantastic work, @sseemayer! 🏆 Thanks for contributing this feature and tweaking the test setup to also test autocompletion! 🙇

PR #2213, opened 2025-07-04
chore: set up dev environment using Devbox instead of plain Nix

sisp:
I've migrated our dev environment setup from plain Nix to [Devbox](https://github.com/jetify-com/devbox) (which also leverages Nix internally).

Our GHA workflows have started failing again recently, although we've not changed anything, because of an error when installing devenv:

```
error: Cannot build '/nix/store/q17vd9afap76fs5yximbn7fii45246h1-nix-2.24-devenv.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/6wwykwdm99zl84x82x0jh9lrr9byf3a3-nix-2.24-devenv-man
         /nix/store/c9pigfi3330bj4q4cga76inicbh9sfc2-nix-2.24-devenv-dev
         /nix/store/kgbdfs86q9yhak48m5q23faqdndzr7wc-nix-2.24-devenv-debug
         /nix/store/r8x8wairqqqz295v9rxghsi0cgcj9v7d-nix-2.24-devenv-doc
         /nix/store/w6i7b0zji0840pk418hzjkmnski8c8nq-nix-2.24-devenv
       Last 25 log lines:
       > source root is source
       > Running phase: patchPhase
       > applying patch /nix/store/3bal0hf39zp5p75xz45gihanq9493cvb-ghsa-g948-229j-48j3-2.24.patch
       > patching file src/libstore/local-store.cc
       > Hunk #1 succeeded at 176 (offset -11 lines).
       > Hunk #2 FAILED at 345.
       > Hunk #3 succeeded at 375 (offset -19 lines).
       > 1 out of 3 hunks FAILED -- saving rejects to file src/libstore/local-store.cc.rej
       > patching file src/libstore/unix/build/local-derivation-goal.cc
       > Hunk #1 succeeded at 521 (offset -5 lines).
       > Hunk #2 succeeded at 1112 (offset -5 lines).
       > Hunk #3 succeeded at 1217 (offset -5 lines).
       > Hunk #4 succeeded at 1724 (offset -9 lines).
       > Hunk #5 succeeded at 3012 (offset -15 lines).
       > patching file src/libstore/unix/build/local-derivation-goal.hh
       > patching file src/libutil/file-system.cc
       > Hunk #1 succeeded at 250 (offset 3 lines).
       > Hunk #2 succeeded at 260 (offset 3 lines).
       > Hunk #3 succeeded at 306 (offset 3 lines).
       > Hunk #4 succeeded at 328 (offset 3 lines).
       > Hunk #5 succeeded at 370 (offset 3 lines).
       > Hunk #6 succeeded at 382 (offset 3 lines).
       > Hunk #7 succeeded at 400 (offset 3 lines).
       > patching file src/libutil/file-system.hh
       > Hunk #1 succeeded at 126 (offset -22 lines).
       For full logs, run:
         nix log /nix/store/q17vd9afap76fs5yximbn7fii45246h1-nix-2.24-devenv.drv
error: Cannot build '/nix/store/wfkv59s5l6vpv871qgvdw7pi48lgxlcs-devenv-1.3.1.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/zf664bs59r7g20dp80nd06kqf63c7fq8-devenv-1.3.1
```

I suspect the error originates from the installation of the devenv derivation from the Nix channel `nixos-24.11`  which to my knowledge isn't a unique identifier, so the CI runs aren't reproducible. It might be possible to fix this somehow, but there are some other shortcomings (IMHO) when using plain Nix, including:

- **Fragile Renovate support:** This requires a dedicated scheduled GHA workflow for updating the `flake.lock` file. It would be nicer if Renovate could manage all dependency updates though.
- **Lack of support for explicit package versions:** This means, we can't specify the version of each dev dependency but it's determined by the Nix channel. As a result – for example –, we can't keep the version of uv synchronized between non-Nix and Nix installations using Renovate.

To resolve the above error and reduce the complexity of using plain Nix while retaining the benefits of the vast and language-agnostic Nix package ecosystem, I've given Devbox a try. Devbox seems to be well-supported by Renovate, it supports explicit package versions, and minimizes the exposure of Nix, thereby reducing complexity especially for developers not familiar with Nix.

In contrast to devenv, Devbox doesn't offer native support for pre-commit hooks. Therefore, I've created the `.pre-commit-config.yaml` file manually. Some pre-commit hooks require tools that aren't installed by uv, so running pre-commit won't work without the full dev environment. But as pre-commit is opt-in anyway, I don't think this is a problem.


sisp:
I'll go ahead and merge this to unblock our GHA workflows in other PRs. We can always revert and decide on a different path if needed, but it seems that Devbox works well, the UX is quite nice, and complexity is much lower than before.

Issue #751, opened 2022-08-25
Add support for nested prompts


sisp:
**Is your feature request related to a problem? Please describe.**
I'd like to include zero or more custom PyPI source URLs and optional authentication credentials in a generated Python project (the credentials would be processed by a Copier task after successful project generation). If I were only trying to add _zero or **one**_ of those PyPIs, I could handle it with the currently available prompt types, but with _zero or **more**_ it's more involved. Handling multiple repeated prompts is a separate topic I've raised in #750 already. If that proposal gets accepted, I believe it would be necessary to support nested prompts for this use case whose answers are gathered in a Python `dict`. But nested prompts might also be useful on their own in order to group related answers.

**Describe the solution you'd like**
I imagine a solution could look like this (where the `multiple` field is taken from the proposal in #750):
```yaml
pypi_sources:
  type: struct  # or `dict`/`map`/`nested`/...
  help: Add custom PyPI sources
  fields:
    url:
      type: str
      help: Custom PyPI source URL
    auth:
      type: bool
      help: Requires authentication?
      default: yes
    auth_username:
      type: str
      help: Username
      when: "{% if pypi_sources|last|attr('auth') %}true{% endif %}"
    auth_password:
      type: str
      help: Password
      when: "{% if pypi_sources|last|attr('auth') %}true{% endif %}"
      secret: true
  multiple: true
```

The answers of the nested prompts would be gathered in a Python `dict`. The validation expression, when present, would be executed after every input with the current dict of answers. For validating each field independently, the field must be selected explicitly to avoid possible name collisions in the accessible Jinja context, e.g. `pypi_sources|last|attr('url')` for validating the just entered URL. This way, complex validation logic that depends on previously entered answers would be possible, too.

Deep nesting would be possible but should probably be avoided to keep the questionnaire simple. Regarding UX, perhaps the nesting could be visualized using Unicode art similar to how the `tree` tool prints filesystem trees:

```
🎤 Add custom PyPI sources
├─┬─ 🎤 Custom PyPI source URL
│ │     https://gitlab.com/api/v4/groups/<group_id_1>/-/packages/pypi/simple
│ ├─ 🎤 Requires authentication?
│ │     Yes
│ ├─ 🎤 Username
│ │     sisp
│ └─ 🎤 Password
│       ********
└─┬─ 🎤 Custom PyPI source URL
  │     https://gitlab.com/api/v4/groups/<group_id_2>/-/packages/pypi/simple
[...]

PR #2223, opened 2025-07-08
Nested prompts

dellekappa:
This PR relates to issue #751 

Some reasoning follows:

- implemented the question type `dict` that will be rendered, as expected, as a `dict` in jinja context. This is why I used `dict` instead of `nested`
- on the answers file side, the answers are still a flat list that includes only the leafs of the question tree, but, to be able to traverse the question tree hierarchy, a classical `property.property` synthax has been used. Under the hood this is handled by [dpath](https://github.com/dpath-maintainers/dpath-python). This permits the user to easily specify an answer value from the command line at any hierarchy level.
- the deepness in prompt have been rendered just with indentation. Even if what have been proposed in #751 is much more appealing, is also much harder to implement because requires to know in advance the questions that will be rendered


dellekappa:
@sisp I made a new PR. It is internally a completely different implementation, much easier to mantain and more in line with this sentence in the #751 issue: 

> a nested question class could act as a container type that delegates asking/prompting to its children

implementing #750 on top of this would be much easier.

PR #2276, opened 2025-08-14
fix: disallow render paths outside destination directory

sisp:
WDYT, @pawamoy?


copier/_main.py
```diff
@@ -669,16 +670,16 @@ def jinja_env(self) -> YieldEnvironment:
...
+            "posix": PurePosixPath,
+            "windows": PureWindowsPath,
+            "native": PurePath,
         }
 
         def _pathjoin(
             *path: str, mode: Literal["posix", "windows", "native"] = "posix"
         ) -> str:
-            return separators[mode].join(path)
+            return str(path_type[mode](*path))
```
sisp, line 687: Without this change, an absolute path on POSIX systems would have a double-`/` prefix. While this is not functionally relevant, the path in the `ForbiddenPathError` message would look ugly.

pawamoy: Looks robust :+1:

yajo (approved):

tests/test_copy.py
```diff
@@ -1274,3 +1276,58 @@ def test_copier_phase_variable(tmp_path_factory: pytest.TempPathFactory) -> None
...
+def test_relative_render_path_outside_destination_raises_error(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    root = tmp_path_factory.mktemp("root")
+    (src := root / "src").mkdir()
+    (dst := root / "dst").mkdir()
+    build_file_tree(
+        {
+            root / "forbidden.txt": "foo",
+            src / "{{ pathjoin('..', 'forbidden.txt') }}": "bar",
```
pawamoy, line 1291: This is the problematic line for Windows right? Because of the single-quotes in the filename?

sisp: No, those are fine (I tried on Windows). The current error seems to be caused by the colon in the Windows drive part (`C:`) which isn't allowed in a directory/file name. I think moving this to, e.g., a computed value and using the variable in the Jinja expression might work. I'll try to figure it out over the weekend.

PR #2278, opened 2025-08-14
fix: validate default answer only after user input

omus:
Follow up to #2145. Copier versions before 9.8.0 used to allow invalid defaults as long as the user would modify them to make them valid before continuing. This PR restores that old behavior while still ensuring that `validator` checks are still run on default when `--defaults` is used. 

The rational for using a `default` instead of a `placeholder` in this case is to provide partial input to the answer to reduce the amount of typing required during interactive use.

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -607,6 +607,12 @@ class Worker:
                 raise CopierAnswersInterrupt(
                     self.answers, question, self.template
                 ) from err
+            # Computed values (i.e., `when: false`) are intentionally not
+            # validated at the moment.
+            # https://github.com/copier-org/copier/issues/1779#issuecomment-2365006990
+            # https://github.com/copier-org/copier/pull/1785
+            if question.get_when():
+                question.validate_answer(new_answer)
             self.answers.user[var_name] = new_answer
 
         # Reload external data, which may depend on answers
--- a/copier/_user_data.py
+++ b/copier/_user_data.py
@@ -276,14 +276,7 @@ class Question:
                     result = self.render_value(
                         self.settings.defaults.get(self.var_name, self.default)
                     )
-        result = self.parse_answer(result)
-        # Computed values (i.e., `when: false`) are intentionally not validated
-        # at the moment.
-        # https://github.com/copier-org/copier/issues/1779#issuecomment-2365006990
-        # https://github.com/copier-org/copier/pull/1785
-        if self.get_when():
-            self.validate_answer(result)
-        return result
+        return self.parse_answer(result)
 
     def get_default_rendered(self) -> bool | str | Choice | None | MissingType:
         """Get default answer rendered for the questionary lib.
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -451,6 +451,38 @@ def test_when(
     assert context == {"question_2": "something"}
 
 
+def test_default(tmp_path_factory: pytest.TempPathFactory, spawn: Spawn) -> None:
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    build_file_tree(
+        {
+            (src / "copier.yml"): yaml.dump(
+                {
+                    "_envops": BRACKET_ENVOPS,
+                    "_templates_suffix": SUFFIX_TMPL,
+                    "team": {
+                        "type": "str",
+                        "default": "@copier-org/",
+                        "validator": "[% if not (team | regex_search('^@copier-org/.+$')) %]Must be a copier-org GitHub team[% endif %]",
+                    },
+                }
+            ),
+            (src / "[[ _copier_conf.answers_file ]].tmpl"): (
+                "[[ _copier_answers | to_nice_yaml ]]"
+            ),
+        }
+    )
+    tui = spawn(COPIER_PATH + ("copy", str(src), str(dst)))
+    expect_prompt(tui, "team", "str")
+    tui.expect_exact("@copier-org/")
+    tui.sendline("everyone")
+    tui.expect_exact(pexpect.EOF)
+    answers = load_answersfile_data(dst)
+    assert answers == {
+        "_src_path": str(src),
+        "team": "@copier-org/everyone",
+    }
+
+
 def test_placeholder(tmp_path_factory: pytest.TempPathFactory, spawn: Spawn) -> None:
     src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
     build_file_tree(


sisp:
I think that

```shell
copier copy ...                                         # ✅ (because of valid interactive input)
copier copy --defaults ...                              # ❌ (because of invalid default value)
copier copy --defaults -d team=@copier-org/everyone ... # ✅ (because of valid non-interactive input)
copier copy -d team=@copier-org/everyone ...            # ✅ (because of valid non-interactive input)
```

would be inconsistent. It is not up to a template author to decide which alternative is used by a template user, so they should all work.

> So if we want to ensure that defaults always valid we may want to go as far as always validating the defaults even when the question is skipped.

I tend to agree. Specifying a validator in a question spec is conceptually similar to type refinement in runtime validation libraries (e.g., [Zod's refinement](https://zod.dev/api?id=refinements#refinements)), adding constraints to a base type. Off the top of my head, I can't think of a reason why it should not apply to default values.

That said, I think it's important to be clear about the (current) interplay of `default` and `when`:

```yaml
# copier.yml
q1:
  type: str
  when: false

q2:
  type: str
  default: foo
  when: false

_message_after_copy: |
  q1: {{ q1 is undefined }}
  q2: {{ q2 == 'foo' }}
```

```console
$ copier copy ...
...
q1: True
q2: True
```

When a default value is specified for a `when: false` question (aka "skipped question", or "computed value" if `false` is hardcoded), then the corresponding Jinja variable is defined and its value is the default value. Thus, a `when: false` question with a default value is only skipped during prompting, but its answer does exist in the Jinja context. This behavior probably makes the most sense for questions that are still meaningful in the context of the questionnaire but don't require a user answer (e.g., because the answer can be derived from previous answers). For example:

```yaml
editors:
  type: str
  help: IDE integrations
  choices:
    - vscode
    - pycharm
    - sublime

devcontainer:
  type: bool
  help: Enable dev container support
  default: false
  when: "{{ 'vscode' in editors }}"
```

Here, VS Code and PyCharm have dev container support, so the user may choose to enable it. But the question is skipped when Sublime Text has been selected because it doesn't (seem to) support dev containers. In all cases, the `devcontainer` variable is exposed in the Jinja context because it has a default value. This makes sense because the `.devcontainer/devcontainer.json` file shall be generated whenever dev container support is enabled, only `customizations` may contain IDE-specific settings.

> The other issue I see with enforcing that defaults have to be valid is currently if an invalid default is used accidentally in a template it is easy to miss such mistakes if you questions are skipped with `when` conditions

Yes, there might be use cases where a question should be skipped and its Jinja variable should be undefined (i.e., not present in the render context) because it has no meaningful default value. For example:

```yaml
database_engine:
  type: str
  help: Database engine
  choices:
    - postgres
    - mysql
    - none
  default: postgres

database_url:
  type: str
  help: Database URL
  default: >-
    {%- if database_engine == 'postgres' -%}
    postgresql://user:pass@localhost:5432/dbname
    {%- elif database_engine == 'mysql' -%}
    mysql://user:pass@localhost:3306/dbname
    {%- endif -%}
  when: "{{ database_engine != 'none' }}"
  # Simplified for illustration purposes
  validator: "{% if '://' not in database_url %}Invalid{% endif %}"
```

Here, the default value of the `database_url` question is only defined for the database engines "postgres" and "mysql" but not for "none", and we can't render, e.g., an empty string as the default value for "none" because the validator would fail. We could surround the actual validator condition with `{% if database_engine != 'none' %}...{% endif %}`, but that feels hacky and redundant with the `when` expression. What we'd actually want is to fully skip this question – I'd say by declaring the default value as undefined. An undefined default answer of a skipped question would mean that there is no answer, so the validator wouldn't be applied. With this, IMO it seems perfectly fine to require default values to always pass validation. Support for unsetting default values is a feature I've had in mind for a while, so I took the liberty and drafted a PR: #2286 WDYT?

About your idea to have a partial default value: We'd need to check whether `questionary`/`prompt-toolkit` support this level of UX. But I'm not yet convinced by the need for this feature because this example could be refactored such that the partial default value isn't needed. Do you have a real use case – and if so, are you able to share it?

PJ-Schulz:
Hi @omus @sisp, we maintain use a Copier template for generating Talos projects, and we are **stuck on Copier <9.8** because the new behavior breaks two patterns we rely on heavily. As a result we cannot adopt any of the new features that have landed since 9.8. I wanted to share our use cases (simplified) in case they help the discussion.

## Use case 1 — placeholder tokens inside a default value

```yaml
custom_registry_url:
  type: str
  help: What is the url of the custom registry?
  default: "https://<IP_ADDRESS>:5000"
  validator: ... # skip here for readability, but validate that its a valid url.
```

The idea: show the user the *shape* of the expected value (`https://…:5000`) and force them to replace `<IP_ADDRESS>` with a real IP. The validator then guarantees the IP is inside the pod subnet.

We want to **force** an edit until the validator is happy and do not to use a random value.

## Use case 2 — default derived from context that may violate a constraint

```yaml
preset_type:
  type: str
  help: Which preset type would you like to use?
  choices:
    - iiot
    - customer

name:
  type: str
  help: What is the name?
  default: "{{ preset_type }}-{{ _copier_conf.dst_path.name }}"
  validator: >-
    {% if name | length > 20 %}too long (max 20 characters){% endif %}
```

The default is derived from the preset and the destination folder name. For some combinations the derived default is too long (external constraint: max 20 chars). In that case the user must shorten it.

**Behavior on 9.7.1** (what we want):

```text
$ copier copy template/ long-project-name
  🎤 Which preset type would you like to use?
   customer
  🎤 What is the name?
   customer-long-project-name

too long (max 20 characters)
```
The user sees the derived default, the validator tells them why it's not acceptable (`too long (max 20 characters)`), and they edit until it passes. Perfect UX.

**Behavior on 9.8.0** (broken for us):

```text
$ copier copy template/ long-project-name
  🎤 Which preset type would you like to use?
   customer
  ...
  ValueError: Validation error for question 'name': too long (max 20 characters)
```

Copier crashes with a stack trace **before the `name` question is ever shown to the user**. From the user's perspective they just picked `customer` and got an unrelated traceback. This is a confusing failure mode. As the error is only indirectly related to the question the user last answered. It becomes even more confusing when other questions appear between `preset_type` and `name` that have nothing to do with them at all.

## On the proposed alternatives

- **`placeholder`**: doesn't fit — we need a partially-prefilled value that the user edits, not an empty field with a hint.
- **`UNSET` (#2286)**: doesn't fit either — we want a default rendered into the prompt where the user is forces to change it for making the validator happy.


**A new boolean property on the question that controls whether the default is validated or not**: This would solve both our use cases cleanly. Something like: `validate_default` or `initial_validate`

sisp:
I _imagine_ we could defer validation _in interactive mode_ until the user presses `Enter` to avoid eager validation of the default value. This might allow for an invalid default value that the user _must_ edit. In non-interactive mode (when passing `--defaults`), the default value must be valid or it must be overridden via `-d, --data` or `--data-file`. This would require a default value conditioned on the mode though:

```yaml
custom_registry_url:
  type: str
  help: What is the url of the custom registry?
  default: "{% if _copier_conf.defaults %}https://127.0.0.1:5000{% else %}https://<IP_ADDRESS>:5000{% endif %}"
  # A simple validator just for demo purposes here
  validator: |
    {%- if '<' in custom_registry_url -%}
    Invalid URL
    {%- endif -%}
```

In non-interactive mode, a valid default value remains mandatory because passing just `--defaults` without `-d, --data` or `--data-file` should always work IMHO, while in interactive mode it may be invalid.

Would this be a compromise, @PJ-Schulz @OJ-Schulz @omus?

PR #2352, opened 2025-10-13
fix: resolve path between dst_path and answers_file to handle dst_path not already existing

Aohzan:
Hello,

`--answers-file` must be relative to the destination path, but when this one doesn't exist, the answers file cannot be found, example:

```
copier copy --answers-file ../../../.copier-answers.yml mytemplate my/dest/path
```
will not found the `.copier-answers.yml` file

an other solution (or complementary) will be to search the file in the current dir.

### DIFF
--- a/tests/test_answersfile.py
+++ b/tests/test_answersfile.py
@@ -333,3 +333,49 @@ def test_undefined_phase_in_external_data(
     copier.run_copy(str(src), dst, defaults=True, overwrite=True)
     answers = load_answersfile_data(dst, ".copier-answers.yml")
     assert answers["key"] == "value"
+
+
+
+def test_answersfile_relative_path_nonexistent_dst(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    """Test that relative answers_file paths work when dst_path doesn't exist yet."""
+    src = tmp_path_factory.mktemp("src")
+    # Create a simple template
+    build_file_tree(
+        {
+            (src / "copier.yml"): (
+                """\
+                project_name:
+                  type: str
+                """
+            ),
+            (src / ".copier-answers.yml"): (
+                """\
+                project_name: "test project"
+                """
+            ),
+            (src / "README.txt.jinja"): (
+                "Project: {{ project_name }}"
+            ),
+        }
+    )
+    git_save(src, tag="v1")
+
+    # Destination path that doesn't exist yet
+    dst_dir =  src / "my" / "dest" / "path"
+
+    # Answers file in parent directory (relative path of dst_dir)
+    answers_file = "../../../.copier-answers.yml"
+
+    # This should work even though dst_dir doesn't exist yet
+    copier.run_copy(
+        str(src),
+        dst_dir,
+        answers_file=answers_file,
+        defaults=True,
+        overwrite=True,
+    )
+
+    # Verify the project was created
+    assert (dst_dir / "README.txt").read_text() == "Project: test project"
--- a/copier/_user_data.py
+++ b/copier/_user_data.py
@@ -577,7 +577,8 @@ def load_answersfile_data(
 ) -> AnyByStrDict:
     """Load answers data from a `$dst_path/$answers_file` file if it exists."""
     try:
-        with Path(dst_path, answers_file).open("rb") as fd:
+        answers_path = Path(dst_path, answers_file).resolve()
+        with answers_path.open("rb") as fd:
             return yaml.safe_load(fd)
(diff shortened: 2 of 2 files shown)


tests/test_answersfile.py
```diff
@@ -333,3 +333,49 @@ def test_undefined_phase_in_external_data(
...
+                """\
+                project_name:
+                  type: str
+                """
+            ),
+            (src / ".copier-answers.yml"): (
+                """\
+                project_name: "test project"
+                """
+            ),
```
sisp, line 357: This setup is not representative of a real one.

```suggestion
            (src / "{{ _copier_conf.answers_file }}.jinja"): (
                "{{ _copier_answers|to_nice_yaml }}"
            ),
```

tests/test_answersfile.py
```diff
@@ -333,3 +333,49 @@ def test_undefined_phase_in_external_data(
...
+            (src / "README.txt.jinja"): (
+                "Project: {{ project_name }}"
+            ),
+        }
+    )
+    git_save(src, tag="v1")
+
+    # Destination path that doesn't exist yet
+    dst_dir =  src / "my" / "dest" / "path"
```
sisp, line 366: The result might be same, but this setup is a bit unintuitive because you're generating the project in a subdirectory of the template's sources. I'd follow the usual setup with

```python
```

and generate in `dst =  src / "my" / "dest" / "path"` instead.

tests/test_answersfile.py
```diff
@@ -333,3 +333,49 @@ def test_undefined_phase_in_external_data(
...
+    copier.run_copy(
+        str(src),
+        dst_dir,
+        answers_file=answers_file,
+        defaults=True,
+        overwrite=True,
+    )
+
+    # Verify the project was created
+    assert (dst_dir / "README.txt").read_text() == "Project: test project"
```
sisp, line 381: I think you should also assert that the `.copier-answers.yml` file exists in the expected location.

sisp (changes requested):
Thanks for adding the test! :+1: I've left a few remarks.

tests/test_answersfile.py
Aohzan: where is defined the `_copier_answers`?

tests/test_answersfile.py
Aohzan: you mean `dst_dir = dst / "my" / "dest" / "path"` ? or I miss something

sisp: Sorry, yes, I mistyped.

tests/test_answersfile.py
sisp: This is a Jinja variable available in our render context.

https://copier.readthedocs.io/en/stable/creating/#_copier_answers

Aohzan: yes but where do I define that `project_name: "test project"`

sisp: You can either set it as the default answer in `copier.yml`

```diff
             (src / "copier.yml"): (
                 """\
                 project_name:
                   type: str
+                  default: test project
                 """
             ),
```

and run

```sh
copier.run_copy(..., defaults=True, ...)
```

(as you're doing already) or instead pass the answer via the `data` argument:

```sh
copier.run_copy(..., data={"project_name": "test project"}, ...)
```

Aohzan: Ok but I don't test my case where I want to be dure that the existing answers file is loaded, no?

sisp: Does your use case assume an existing answers file? If so, why? This seems like an unusual pattern. The answers file is generated by Copier to record previous answers plus metadata to enable updating to new template versions. It looks like you're assuming the answers file to exist before the first `copier copy` call. Are you trying to pass answers to `copier copy` from a file instead of via interactive prompting or inline answers via the [`-d,--data`](https://copier.readthedocs.io/en/stable/configuring/#data) flag? If so, perhaps [`--data-file`](https://copier.readthedocs.io/en/stable/configuring/#data_file) is what you're looking for.

Aohzan: one of our use case is to have a template for the repository, then we can add terraform stack in subdirs, and we call the initial answer file to not repeat common questions

sisp: I think `--data-file` might work for this use case as well although you'd be passing an answers file, which is a superset of a typical data file.

sisp: Or [`_external_data`](https://copier.readthedocs.io/en/stable/configuring/#external_data) might be worth checking out.

PR #2354, opened 2025-10-15
fix: updates with floating tags

maganaluis:
## Problem
Copier stores git refs (tags/branches) in `_commit` field of `.copier-answers.yml`. When using moving/floating tags (e.g., `workflows/v1`, `stable/v2`, `latest`), future `copier update` commands fail because the tag now points to a different commit.

### Solution
Added `--resolve-commit-to-sha` flag that stores immutable SHA hashes instead of git refs.

**Implementation:**
- New CLI flag: `--resolve-commit-to-sha`
- Template config option: `_resolve_commit_to_sha: true` in `copier.yml`
- Priority: CLI flag > template config > default (false)

**Usage:**
```bash
# CLI
copier copy --resolve-commit-to-sha template-repo destination

# Template config (copier.yml)
_resolve_commit_to_sha: true
```

**Changes:**
- `copier/_main.py`: Added `resolve_commit_to_sha` field to Worker class and logic in `_answers_to_remember()`
- `copier/_cli.py`: Added CLI flag
- `copier/_template.py`: Added template config support
- `tests/test_resolve_commit_to_sha.py`: Coverage for new functionality

Fixes https://github.com/copier-org/copier/issues/987

## Revision 2

**Solution:** Copier now stores BOTH semantic version and SHA hash, then automatically
chooses which to use during updates.

**Key Changes:**

1. **Dual-Versioning** - Always stores both values:

    - `_commit`: Semantic version (human-readable)
    - `_commit_sha`: SHA hash (machine-reliable)

2. **Automatic Tag Resolution** - Detects and handles floating tags:

    - Floating patterns: `latest`, `stable/*`, `main`, `master`, `develop`, `feat/*`,
      etc.
    - Uses SHA for floating tags, preserves semantic versions for stable tags

3. **Renamed Flag**:
    - `--resolve-commit-to-sha` → `--ignore-git-tags` 
    - Flag controls USAGE during updates, not storage

**Usage:**

```bash
# Automatic 
copier update  # Automatic tag resolution handles everything

# Force SHA usage
copier update --ignore-git-tags

# Template config (copier.yml)
_ignore_git_tags: true
```

**Implementation:**

**Modified Files:**

-   `copier/_main.py`: Automatic tag resolution logic, dual-versioning storage (~40 lines
    reduced)
-   `copier/_subproject.py`: Use SHA for FROM version in diffs
-   `copier/_cli.py`: Renamed flag
-   `copier/_template.py`: Config properties for `ignore_git_tags` and
    `stable_tag_patterns`
-   `tests/test_ignore_git_tags.py`: Complete rewrite with end-to-end floating tag tests

**Resolution Logic:**

```
1. CLI flag --ignore-git-tags → Use SHA
2. Stable semantic version → Use semantic version
3. Floating tag pattern → Use SHA (automatic)
```

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import os
 import platform
-import re
 import stat
 import subprocess
 import sys
@@ -32,7 +31,6 @@ from typing import (
 from unicodedata import normalize
 
 from jinja2.loaders import FileSystemLoader
-from packaging.version import InvalidVersion, parse as parse_version
 from packaging.version import Version
 from pathspec import PathSpec, __version__ as pathspec_version
 from plumbum import ProcessExecutionError, colors
@@ -234,21 +232,6 @@ class Worker:
     # NOTE: attributes are fully documented in [creating.md](../docs/creating.md)
     # make sure to update documentation upon any changes.
 
-    # Floating tag patterns for automatic tag resolution
-    # These patterns identify tags/refs that are likely to move over time
-    _FLOATING_TAG_PATTERNS = [
-        r"^latest$",
-        r"^stable(/.*)?$",
-        r"^main$",
-        r"^master$",
-        r"^develop$",
-        r"^development$",
-        r"^HEAD$",
-        # Branch-like patterns
-        r"^[a-z]+[-_][a-z]+$",  # e.g., feature-branch, dev-test
-        r"^(feat|fix|chore|docs|test|refactor)/.*$",  # conventional branch names
-    ]
-
     src_path: str | None = None
     dst_path: Path = Path()
     answers_file: RelativePath | None = None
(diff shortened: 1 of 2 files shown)


pawamoy:
Thanks! Then yeah I don't see any downsides to recording the commit SHA and using it when we can't parse the version or when there's no new version (except maybe a longer execution time when a project is updated and there's indeed no new version? but that's not very impactful) :+1:

sisp:
Sounds good. Just some additional ideas for the flag name:

- `--ignore-tags` / `--ignore-git-tags`
- `--no-tags` / `--no-git-tags` – a bit similar to `git fetch --no-tags`

maganaluis:
@sisp are there any additional changes we need to make to get this merged?

copier/_main.py
```diff
@@ -218,11 +220,31 @@ class Worker:
...
+        r"^latest$",
+        r"^stable(/.*)?$",
+        r"^main$",
+        r"^master$",
+        r"^develop$",
+        r"^development$",
+        r"^HEAD$",
+        # Branch-like patterns
+        r"^[a-z]+[-_][a-z]+$",  # e.g., feature-branch, dev-test
+        r"^(feat|fix|chore|docs|test|refactor)/.*$",  # conventional branch names
```
sisp: I think that all tags/refs that match these patterns wouldn't be parsed as versions. Do we need to detect floating tags/refs explicitly then? I'd say version parsing should be enough in this case.

PR #2369, opened 2025-10-23
Focus on subproject subdir when assessing dirtiness

### LINKED ISSUE #2362: Allow subprojects to be updated if there are other uncommitted changed outside the subproject
### Actual Situation

I opened this as [a discussion](https://github.com/orgs/copier-org/discussions/2360) but have since realised a minor code change would enable this to be a feature that would mitigate the need for a workaround. I'm opening an issue so I can link a PR to it. 

All background info can be found in the linked discussion. TL;DR: you can't currently update two subdirectories with mutually exclusive template links in the same git commit since _any_ uncommitted workspace changes will raise a "Destination repository is dirty" error, even when they are changes made outside a subdirectory containing the templated project.

### Desired Situation

Being able to update several subdirectories that have separately been generated from a template / templates without having to commit changes on each subdirectory change. I have no issue with not being allowed to run `copier update` a second time on a given subdir without committing.

### Proposed solution

My proposal is to update [this](https://github.com/copier-org/copier/blob/74f2f59df193b61e4b3a1363934595d6c4dbe731/copier/_subproject.py#L44-L46) `is_dirty()` method from:

```py
        if self.vcs == "git":
            with local.cwd(self.local_abspath):
                return bool(get_git()("status", "--porcelain").strip())
```
to:

```py
        if self.vcs == "git":
            with local.cwd(self.local_abspath):
                return bool(get_git()("status", self.local_abspath, "--porcelain").strip())
```

brynpickering:
Fixes #2362 

I've added a test to cover this specific use-case and all seems to be OK with the test suite when run locally!

### DIFF
--- a/tests/test_dirty_local.py
+++ b/tests/test_dirty_local.py
@@ -218,7 +218,7 @@ def test_update_with_gpg_sign(
 
 def test_parallel_projects_in_subdirs(
     tmp_path_factory: pytest.TempPathFactory,
-) -> tuple[Path, Path, Path, Path]:
+) -> None:
     src = tmp_path_factory.mktemp("src")
     parent = tmp_path_factory.mktemp("parent")
     dst1 = parent / "project1"


tests/test_dirty_local.py
```diff
@@ -214,3 +214,66 @@ def test_update_with_gpg_sign(
...
+        git("add", "-A")
+        git("commit", "-m", "initial template commit")
+
+    with local.cwd(parent):
+        git("init")
+
+    # Create projects
+    for dst in [dst1, dst2]:
+        run_copy(str(src), dst, overwrite=True, data={"question": dst.name})
+
```
sisp, line 258: Perhaps add assertions here as well to make sure the contents of `file.txt` after the first copy are different than after the update:

```suggestion

    for dst in [dst1, dst2]:
        assert (dst / "file.txt").read_text() == dst.name

```

WDYT?

brynpickering: Updated and added a couple of other tests in the process

tests/test_dirty_local.py
```diff
@@ -214,3 +214,106 @@ def test_update_with_gpg_sign(
     gitconfig.set({"user.signinkey": "123456", "commit.gpgsign": "true"})
     with pytest.warns(DirtyLocalWarning):
         run_update(dst, defaults=True, overwrite=True)
+
+
+@pytest.fixture
+def setup_parallel_projects(
```
sisp: I think it's more intuitive to name this fixture in noun form and not in imperative mood, as you're accessing the resources created by this fixture later.


```suggestion
def parallel_projects(
```

sisp:
I see your point. Well, I don't have a strong opinion, our test suite is certainly not consistent, as it has grown over years with several maintainers and many contributors. I lean a bit to a single "user journey" style test in this case, but it's just a slight preference. In any case, I really appreciate your work here! :bow:

brynpickering:
@sisp I've squashed the test as requested; hopefully all good to go now.

tests/test_dirty_local.py
```diff
@@ -214,3 +214,67 @@ def test_update_with_gpg_sign(
...
+        assert (dst / "file.txt").read_text() == f"Updated {dst.name}"
+
+    # Verify subdirectories are both now dirty
+    expected = "M project1/.copier-answers.yml\n M project1/file.txt\n M project2/.copier-answers.yml\n M project2/file.txt"
+    with local.cwd(parent):
+        assert git("status", "--porcelain").strip() == expected
+
+    # Test failing on attempting to update our now dirty project subdirectories.
+    for dst in [dst1, dst2]:
+        with pytest.raises(UserMessageError):
```
sisp: Could you also assert part of the exception message to make sure the right `UserMessageError` exception is raised?


```suggestion
        with pytest.raises(UserMessageError, match="Destination repository is dirty"):
```

tests/test_dirty_local.py
```diff
@@ -214,3 +214,67 @@ def test_update_with_gpg_sign(
     gitconfig.set({"user.signinkey": "123456", "commit.gpgsign": "true"})
     with pytest.warns(DirtyLocalWarning):
         run_update(dst, defaults=True, overwrite=True)
+
+
+def test_parallel_projects_in_subdirs(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> tuple[Path, Path, Path, Path]:
```
sisp: The return type is incorrect here:


```suggestion
) -> None:
```

This is the mypy error:

```
tests/test_dirty_local.py:219: error: Missing return statement  [return]
```

sisp (changes requested):

PR #2374, opened 2025-10-25
fix: fix user defaults not reflected if default is `MISSING` in `copier.yml`

### LINKED ISSUE #2373: User defaults in Copier settings (`<CONFIG_ROOT>/settings.yml`) will not be reflected unless `default: ...` is explicitly set in `copier.yml`
### Describe the problem

User defaults in Copier settings (`<CONFIG_ROOT>/settings.yml`) will not be reflected unless `default: ...` is explicitly set in `copier.yml`. I'm not sure if this is the intended behavior, but since it is highly confusing, I would appreciate it if this could be documented.

### Template

`copier.yml`: 
```yaml
github_user:
    type: str
```
`settings.yml`
```yaml
defaults:
    github_user: 34j
```

### To Reproduce

1. Run `copier copy --defaults` would fail `ValueError: Question "github_user" is required`
2. If I add `    default: ''` to the last line in `copier.yml` and run `copier copy --defaults` it would work

### Logs

```console

```

### Expected behavior

The behavior is documented or the problem is fixed

_No response_

### Operating system

Linux

NixOS 25.05

### Copier version

9.10.3

### Python version

CPython 3.13.5

### Installation method

uvx+pypi

### Additional context

This seems to be caused by `get_default()` (#958) returning `MISSING` before checking `self.settings` and I think this can be removed maybe:

https://github.com/copier-org/copier/blob/b9f2d11465842bff09d6917a49b9b6cfc5f31d94/copier/_user_data.py#L275-L276

34j:
Closes #2373
Related: #1940

### DIFF
--- a/copier/_user_data.py
+++ b/copier/_user_data.py
@@ -272,8 +272,6 @@ class Question:
                 try:
                     result = self.answers.user_defaults[self.var_name]
                 except KeyError:
-                    if self.default is MISSING:
-                        return MISSING
                     try:
                         result = self.render_value(
                             self.settings.defaults.get(self.var_name, self.default),
@@ -282,6 +280,8 @@ class Question:
                             },
                         )
                     except UnsetError:
+                        return MISSING
+                    if result is MISSING:
                         return MISSING
         result = self.parse_answer(result)
         # Computed values (i.e., `when: false`) are intentionally not validated


sisp (approved):
Thanks for contributing a fix for this bug, @34j! :bow:

LGTM! :tada:

Issue #1350, opened 2023-10-02
Inline conflicts should use standard git markers


marcelmindemann:
### Actual Situation

I am using copier 8.3.0. When updating a rendered project that diverged from the template, inline markers are added to files with conflicts. These markers are custom to copier and therefore not recognized by git or my IDE.

### Desired Situation

The marker should use the standard git notation and set the status of the destination git repository to "conflicts". Similar to how https://github.com/cruft/cruft leaves the conflicting files after a failed `git apply <patch>`.

### Proposed solution

Would this not solve many problems elegantly? It would not only get rid of the need to use `pre-commit` to prevent pushing conflicting files, as `git` would not allow commiting the working tree in "conflicting files" state. Moreover, my IDE's git integration would help me to solve conflicts visually (using, for example, the brilliant 3-way-merge tool of Jetbrains or VSCode).

pawamoy:
I noticed that VSCode didn't offer the same visual help for Copier markers. Would improve the UX, I agree.

yajo:
I see visual help on VSCode with conflicts generated by Copier. I can use bundled tools to accept or reject changes. Could you please be more specific on what you're missing? Maybe some video or screenshots would help.

pawamoy:
The differences I noticed:

- no "Accept incoming" or "Accept current" buttons in the editor, directly above the conflict
- no previous/next buttons in the top bar (go to previous/next conflict)
- no check for marker presence when staging files within the VCS tab

Colors (blue/green) are there in the editor though.

12rambau:
![image](https://github.com/copier-org/copier/assets/12596392/7a73e684-eb26-4eb0-b5a5-50eafad873f3)

It's working just fine on my end.

pawamoy:
Huh, OK :thinking:  I'll run some experiments :slightly_smiling_face:

yajo:
I see it like @12rambau. Can you guys share a screenshot of the problem?

pawamoy:
![conflict](https://github.com/copier-org/copier/assets/3999221/831cad8d-a8c1-416e-8ed5-40c1e9f03d38)

yajo:
The accept buttons only appear on normal file view. If you're viewing the diff file view, it won't work. That's just normal VSCode behavior. Could you check if that's the problem?

pawamoy:
You're right! I didn't know that. Here are more screenshots.

---

After a Git merge with conflicts:

[...]

PR #2376, opened 2025-10-28
refactor!: introduce new update algorithm based on `git merge`

sisp:
I have introduced a new update algorithm that is centered around the high-level `git merge` command, performing a 3-way merge of a fresh project based on the new template version into the current project with the common ancestor being a fresh project based on the old template version. This is – in a nutshell – what the previous update algorithm was doing using a sequence of low-level Git commands and other hacks. With this refactoring, I'm hoping to simplify the update algorithm implementation, reduce the risk of missed edge cases by relying on a battle-tested high-level command, and paving the way for new opportunities like improving merge conflict handling through custom merge drivers like <https://mergiraf.org> and supporting an update algorithm variant that does not rely on replaying a fresh project based on the old template version (see #1170).

I have marked this PR as a draft because this is a significant (and breaking, see below for more details) change that needs to be thoroughly reviewed and shouldn't be rushed.

According to my research, Copier's update algorithm has a long history, starting with #106 and evolving from a 2-way merge to a 3-way merge with merge conflicts stored first in `.rej` files and later also available as inline markers. Many edge cases have shown up along the way and their fixes contributed to the current complexity. At some point, I realized that the current algorithm is essentially a 3-way merge, implemented with a sequence of low-level commands and hacks. But `git merge` also performs a 3-way merge with just a single command, given that the Git graph contains three commits with the appropriate relationship. This Git graph does not exist naturally in our case, as a Copier template needs to be instantiated before updating a project, and these template instances are not tracked via Git. But it is possible to construct this relationship in a synthetic Git graph which consists of the three commits necessary to perform the 3-way merge. `git merge` can be applied on this synthetic graph, and with some tricks it looks like a familiar merge. Thanks to using plain `git merge`, [delicate Git index manipulation](https://github.com/copier-org/copier/blob/de88305564e99a698289ba2e108edf410f85bb78/copier/_main.py#L1377-L1406) to induce a mid-merge state for merge conflicts, especially to enable [visual merge conflicts in IDEs like VS Code](https://github.com/copier-org/copier/issues/1350), becomes obsolete, aborting a Copier update becomes a simple `git merge --abort` command, and edge cases like [add/add merge conflicts](https://github.com/copier-org/copier/pull/1587) are handled without special treatment. In a way, `copier copy` is like `git clone`, and `copier update` is like `git merge` – but extended to handling parametrized starters with some extra sugar.

This PR introduces a few breaking changes:

- `git merge` can only produce inline conflict markers but no `.rej` files, so I've removed the `conflict` setting. I suppose that nobody uses `.rej` files anymore, as inline conflict markers have been the default setting for a while and are familiar to Git users.
- `git merge` does not support a configurable context size, so I've removed the `context_lines` setting.
- `git merge` does not support full customization of the conflict maker labels. To compensate for this limitation at least a little bit, a temporary ref `copier/after-updating` points to the commit of the fresh project based on the new template version, which is used as the conflict marker label for "incoming" changes. The conflict marker label for the current state is `HEAD`.
- `.gitattributes` settings affect Copier's internal `git merge` call. This may have (unexpected) side effects which are difficult to anticipate. But it may be exactly what we want, glancing at a custom merge driver like <https://mergiraf.org> again.

If we decide to (eventually) merge this PR, I suggest to slip in deprecation warnings first, release them, and wait for a little while before creating a new major release to give Copier users a smoother transition experience.

Additionally looping in some contributors who have been active around the update algorithm: @tpluscode @tguillemot @barrywhart @thurse93


docs/updating.md
```diff
@@ -38,15 +38,11 @@ other Git ref you want.
...
-Copier to know what to do with a diff code hunk. In those cases, copier handles the
-conflict in one of two ways, controlled with the `--conflict` option:
-
--   `--conflict rej`: Creates a separate `.rej` file for each file with conflicts. These
-    files contain the unresolved diffs.
--   `--conflict inline` (default): Updates the file with conflict markers. This is quite
-    similar to the conflict markers created when a `git merge` command encounters a
-    conflict. For more information, see the "Checking Out Conflicts" section of the
-    [`git` documentation](https://git-scm.com/book/en/v2/Git-Tools-Advanced-Merging).
+Copier to know what to merge the changes from the evolved template into the evolved
```
yajo:
```suggestion
Copier to know how to merge the changes from the evolved template into the evolved
```

yajo (approved):
I'm pretty amazed with the new algorithm. I have not tested it yet, but with reduced low-level plumbing and half of the lines, all tests pass! At least, except the ones you're inherently breaking.

I suggest to add `BREAKING CHANGE:` trailers to the git commit message listing specific things that break, so they appear in the changelog once released.

Thank you! ❤️

sisp:
We might need to run pre-update migrations also on the fresh copy of the old template version – see https://github.com/copier-org/copier/discussions/2612#discussioncomment-16545129. But this needs some deep thinking to make sure it's _generally_ correct.

PR #2377, opened 2025-10-28
fix: use .gitignore for exludes

### LINKED ISSUE #2244: Issue with too many files in (ignored) .venv
### Describe the problem

I'm having an issue with a project that has a large virtual env, and the .venv folder lives within the project folder. `.venv` is ignored in both `copier.yaml` & `.gitignore`. I also tried to not ignore the `.venv` folder in the `copier.yaml` file and only rely on `.gitignore`, but that also didn't help.

The issue is that the generated `git apply` command becomes too long with all the files in the `.venv` directory being excluded separately (traceback below).

I've created an example project (needs [`uv`](https://docs.astral.sh/uv/) installed) to make it easy to reproduce the issue (although not sure if other OSs than my Fedora might have other limits with how long arguments to a command can be):

https://github.com/makkus/copier_issue

Note: I've installed copier using [pixi](https://pixi.sh/latest/), but couldn't select that below. Don't think it matters though.

### Template

https://github.com/makkus/copier_issue

### To Reproduce

```
copier copy --vcs-ref=HEAD gh:makkus/copier_issue test_template
cd test_template && git init && git add .* * && git commit -m 'chore: initial commit'
copier update --vcs-ref=HEAD  # works fine
uv run python main.py
copier update --vcs-ref=HEAD  # raises error
```

### Logs

```console
  File "/home/markus/.pixi/envs/copier/bin/copier", line 10, in <module>
    sys.exit(CopierApp.run())
             ~~~~~~~~~~~~~^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/cli/application.py", line 640, in run
                    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/cli/application.py", line 635, in run
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_cli.py", line 428, in main
    return _handle_exceptions(inner)
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_cli.py", line 71, in _handle_exceptions
    method()
    ~~~~~~^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_cli.py", line 418, in inner
    with self._worker(
         ~~~~~~~~~~~~^
        dst_path=destination_path,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        overwrite=True,
        ^^^^^^^^^^^^^^^
    ) as worker:
    ^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_main.py", line 271, in __exit__
    raise value
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_cli.py", line 426, in inner
    worker.run_update()
    ~~~~~~~~~~~~~~~~~^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_main.py", line 95, in _wrapper
    return func(*args, **kwargs)
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_main.py", line 1124, in run_update
    self._apply_update()
    ~~~~~~~~~~~~~~~~~~^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/copier/_main.py", line 1293, in _apply_update
    (apply_cmd << diff)(retcode=None)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/commands/base.py", line 115, in __call__
    return self.run(args, **kwargs)[1]
           ~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/commands/base.py", line 253, in run
    with self.bgrun(args, **kwargs) as p:
         ~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/contextlib.py", line 141, in __enter__
    return next(self.gen)
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/commands/base.py", line 209, in bgrun
    p = self.popen(args, **kwargs)
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/commands/base.py", line 563, in popen
    return self.cmd.popen(args, **kwargs)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/commands/base.py", line 334, in popen
    return self.cmd.popen(self.args + list(args), **kwargs)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/commands/base.py", line 361, in popen
    return self.cmd.popen(
           ~~~~~~~~~~~~~~^
        args,
        ^^^^^
    ...<2 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/machines/local.py", line 116, in popen
    return self.machine._popen(
           ~~~~~~~~~~~~~~~~~~~^
        self.executable,
        ^^^^^^^^^^^^^^^^
    ...<3 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/machines/local.py", line 298, in _popen
    proc = PlumbumLocalPopen(
        argv,
    ...<6 lines>...
        **kwargs,
    )  # bufsize = 4096
  File "/home/markus/.pixi/envs/copier/lib/python3.13/site-packages/plumbum/machines/local.py", line 30, in __init__
    self._proc = Popen(*args, **kwargs)  # pylint: disable=consider-using-with
                 ~~~~~^^^^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/subprocess.py", line 1039, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                        pass_fds, cwd, env,
                        ^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
                        gid, gids, uid, umask,
                        ^^^^^^^^^^^^^^^^^^^^^^
                        start_new_session, process_group)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markus/.pixi/envs/copier/lib/python3.13/subprocess.py", line 1972, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 7] Argument list too long: '/home/markus/.pixi/bin/git'
```

### Expected behavior

There would be no error, even with a `.venv` (or other ignored/excluded folder for that matter) that has tons of files in it.

_No response_

### Operating system

Linux

Fedora 42

### Copier version

9.8.0

### Python version

CPython 3.13.5

### Installation method

uvx+pypi

### Additional context

_No response_

tnijboer:
Fixes #2244.

### Problem

Running `copier update` fails with `OSError: [Errno 7] Argument list too long` on projects with extensive `.gitignore` files or `_skip_if_exists` lists.

The cause is that `copier` builds the `git apply` command by adding an individual `--exclude={pattern}` argument for every exclusion. This exceeds the OS command-line argument length limit.

### Solution

This patch refactors the exclusion handling for `git apply`:

1.  All exclusion patterns (from `.gitignore`, `all_skip_if_exists`, `answers_relpath`) are consolidated into a single list.
2.  This list is written to a temporary file.
3.  The `git apply` command is modified to use a single `--exclude-from=<temp_file>` argument, pointing to the new file.
4.  A `try...finally` block is used to guarantee the temporary file is cleaned up.

This change avoids the argument length limit and allows `copier update` to succeed on projects with many exclusions.

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1306,46 +1306,42 @@ class Worker:
             with local.cwd(subproject_top):
                 apply_cmd = git["apply", "--reject", "--exclude", self.answers_relpath]
                 ignored_files = git["status", "--ignored", "--porcelain"]()
-                # Parse .gitignore for directory-level patterns (robust)
-
-                def is_dir_pattern(pattern):
-                    pattern = pattern.strip()
-                    if not pattern or pattern.startswith("#"):
-                        return False
-                    # Matches foo/, foo/*, foo/**/, **/test/, **/test/*, etc.
-                    return (
-                        pattern.endswith("/")
-                        or pattern.endswith("/*")
-                        or pattern.endswith("/**/")
-                        or re.match(r".*\*\*/.*", pattern)
-                    )
-
                 gitignore_path = Path(subproject_top, ".gitignore")
                 dir_patterns = set()
                 if gitignore_path.exists():
                     with gitignore_path.open() as gitignore_file:
                         for line in gitignore_file:
                             line = line.strip()
-                            if is_dir_pattern(line):
-                                # Remove trailing slash or /* for matching
+                            if not line or line.startswith("#"):
+                                continue
+                            # Matches foo/, foo/*, foo/**/, **/test/, **/test/*, etc.
+                            if (
+                                line.endswith("/")
+                                or line.endswith("/*")
+                                or line.endswith("/**/")
+                                or re.match(r".*\*\*/.*", line)
+                            ):
                                 cleaned = re.sub(r"(\/\*|\/\*\*\/|\/$)", "", line)
                                 dir_patterns.add(cleaned)
-                # Build extra_exclude: only add files not covered by dir_patterns
+                # Single loop: process ignored files and build exclude lists
+                ignored_dirs = set()
                 extra_exclude = []
                 for filename in ignored_files.splitlines():
                     if filename.startswith("!! "):
                         fname = filename.split("!! ").pop()
-                        # Check if file is in a directory pattern
-                        if any(fname.startswith(p + "/") for p in dir_patterns):
-                            continue  # Skip, will be excluded by directory pattern
-                        extra_exclude.append(fname)
-                # Add directory patterns first
-                for dir_pattern in dir_patterns:
+                        matched_dir = False
+                        for p in dir_patterns:
+                            if fname == p or fname.startswith(p + "/"):
+                                if fname.endswith("/"):
+                                    ignored_dirs.add(fname.rstrip("/"))
+                                matched_dir = True
+                                break
+                        if not matched_dir:
+                            extra_exclude.append(fname)
+                for dir_pattern in ignored_dirs:
                     apply_cmd = apply_cmd["--exclude", dir_pattern + "/*"]
-                # Add skip_if_exists patterns
                 for skip_pattern in map(self._render_string, self.all_skip_if_exists):
                     apply_cmd = apply_cmd["--exclude", skip_pattern]
-                # Add remaining specific files
                 for skip_pattern in extra_exclude:
                     apply_cmd = apply_cmd["--exclude", skip_pattern]
                 (apply_cmd << diff)(retcode=None)


copier/_main.py
```diff
@@ -1312,11 +1312,26 @@ def _apply_update(self) -> None:  # noqa: C901
...
-                for skip_pattern in chain(
-                    map(self._render_string, self.all_skip_if_exists), extra_exclude
-                ):
-                    apply_cmd = apply_cmd["--exclude", skip_pattern]
-                (apply_cmd << diff)(retcode=None)
+                # Consolidate ALL patterns into one list
+                all_exclusions = [self.answers_relpath]
+                all_exclusions.extend(map(self._render_string, self.all_skip_if_exists))
+                all_exclusions.extend(extra_exclude)
+                import tempfile
```
sisp: I'd move this import to the top of the file and add `NamedTemporaryFile` to the `from tempfile import ...` import statement we already have there.

copier/_main.py
```diff
@@ -1312,11 +1312,26 @@ def _apply_update(self) -> None:  # noqa: C901
...
+
+                # Add the --exclude-from argument to the command
+                apply_cmd = apply_cmd["--exclude-from", exclude_file]
+
+                try:
+                    # Run the command, piping the diff to stdin
+                    (apply_cmd << diff)(retcode=None)
+                finally:
+                    # Ensure the temporary file is always deleted
+                    exclude_file.unlink()
```
sisp: Just an idea to avoid using a temporary file with disabled cleanup followed by manual cleanup:


```suggestion
                # Write all patterns to a temporary file
                with TemporaryDirectory(prefix=f"{__name__}.git_apply.") as tmpdir:
                    with (exclude_file := Path(tmpdir, "exclude")).open("w") as f:
                        f.writelines(f"{pattern}\n" for pattern in all_exclusions)

                    # Add the --exclude-from argument to the command
                    apply_cmd = apply_cmd["--exclude-from", exclude_file]

                    # Run the command, piping the diff to stdin
                    (apply_cmd << diff)(retcode=None)
```

WDYT?

sisp (commented):
Thanks for contributing this fix, @tnijboer! :pray:

It looks like the right approach to fixing the problem. I left two inline comments.

copier/_main.py
tnijboer: works for me, updated the code

copier/_main.py
tnijboer: not needed anymore with latest change.

tnijboer:
Yes, you're totally right. My bad, was confused with `rsync`. I updated the code to just check the `.gitignore` for folders, and if there are any files listed in `git status --ignored --porcelain` that are in an ignored folder just add the exclude for the folder instead so the list of excludes get way shorter.

tests/test_updatediff.py
```diff
@@ -2089,3 +2090,118 @@ def test_skip_if_exists_templated(tmp_path_factory: pytest.TempPathFactory) -> N
 
     run_update(str(dst), overwrite=True)
     assert (dst / "skip-if-exists.txt").read_text() == "baz"
+
+
+def test_exclude_with_gitignore(tmp_path_factory: pytest.TempPathFactory) -> None:
```
sisp, line 2095: This test passes even without the changes in the update algorithm, so it doesn't seem to cover the real scenario under test. 

This test looks way too complex for the scenario under test. Could you please scope it down to the scenario where too many exclude patterns cause the `OSError: [Errno 7] Argument list too long` error?

tests/test_updatediff.py
```diff
@@ -62,6 +62,7 @@ def test_updatediff(tmp_path_factory: pytest.TempPathFactory) -> None:
                 Thanks for your attention.
                 """
             ),
+            (repo / ".gitignore.jinja"): ".venv/\n**/.cache/\n",
```
sisp, line 65: I don't think this added file serves any purpose in this test case.


```suggestion
```

copier/_main.py
```diff
@@ -1305,16 +1306,43 @@ def _apply_update(self) -> None:  # noqa: C901
...
+                                    ignored_dirs.add(fname.rstrip("/"))
+                                matched_dir = True
+                                break
+                        if not matched_dir:
+                            extra_exclude.append(fname)
+                for dir_pattern in ignored_dirs:
+                    apply_cmd = apply_cmd["--exclude", dir_pattern + "/*"]
+                for skip_pattern in map(self._render_string, self.all_skip_if_exists):
+                    apply_cmd = apply_cmd["--exclude", skip_pattern]
+                for skip_pattern in extra_exclude:
```
sisp, line 1345: TBH, this looks quite complex for a problem, I'm a bit worried that we're missing some edge cases.

I haven't looked into it in detail, but I think there are two ways to exclude files in the update:

1. Exclude files when applying the patch via `git apply --exclude`, which has the limitation you've encountered.
1. Omit the files to be excluded from the patch.

Option 2 sounds potentially simpler and more robust. WDYT?

PR #2389, opened 2025-11-10
feat: Add postrender tasks phase

rcoup:
### Problem

A Copier template for Java applications requires every Java file to use `.jinja` extensions solely to template package names from `com.example.boilerplate` to `com.example.{{package}}`.  Ideally we'd have a way to keep most template files as plain `.java` with a fixed `com.example.boilerplate` package (`src/{test,main}/java/com/example/boilerplate/`), then execute a simple refactoring script at copy/update time to rename package directories (eg:`boilerplate/` → `myservice/`) and bulk-replace across files (eg: `com.example.boilerplate.` to `com.example.myservice.`). 

Copier's existing task system runs _after_ the diff calculation during updates, which breaks 3-way merge semantics. The diff compares `boilerplate` in the template vs `myservice` in the user's project, making changes undetectable.

### Solution

Solving this in a generic way (since directory-per-package isn't anything Java-specific), add a new `Phase.POSTRENDER` that executes _after_ template rendering but _before_ diff calculation during updates. This enables template authors to define tasks that transform rendered output without needing `.jinja` extensions on every file, while maintaining the 3-way merge semantics.

### Key changes

- Add `Phase.POSTRENDER` enum to execution phases
- Templates declare tasks in `copier.yml` using `_postrender_tasks` with support for conditional execution and custom working directories
- Execute postrender tasks in copy and update flows (on old-copy, new-copy, and destination)
- Add `_update_stage` variable to distinguish between previous, current, and new rendering contexts during updates
- Tests and documentation

Comments/suggestions most welcome!

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -246,6 +246,7 @@ class Worker:
 
     answers: AnswersMap = field(default_factory=AnswersMap, init=False)
     _cleanup_hooks: list[Callable[[], None]] = field(default_factory=list, init=False)
+    _update_stage: str = field(default="current", init=False)
 
     def __enter__(self) -> Worker:
         """Allow using worker as a context manager."""
@@ -363,6 +364,7 @@ class Worker:
         for i, task in enumerate(tasks):
             extra_context = {f"_{k}": v for k, v in task.extra_vars.items()}
             extra_context["_copier_operation"] = operation
+            extra_context["_update_stage"] = self._update_stage
 
             if not cast_to_bool(self._render_value(task.condition, extra_context)):
                 continue
@@ -1062,6 +1064,8 @@ class Worker:
                 # TODO Unify printing tools
                 print("")  # padding space
             if not self.skip_tasks:
+                with Phase.use(Phase.POSTRENDER):
+                    self._execute_tasks(self.template.postrender_tasks)
                 with Phase.use(Phase.TASKS):
                     self._execute_tasks(self.template.tasks)
         except Exception:
@@ -1179,6 +1183,7 @@ class Worker:
                 # https://github.com/orgs/copier-org/discussions/2345
                 exclude=[*self.template.exclude, *self.exclude],
             ) as old_worker:
+                old_worker._update_stage = "previous"
                 old_worker.run_copy()
             # Run pre-migration tasks
             with Phase.use(Phase.MIGRATE):
@@ -1236,6 +1241,7 @@ class Worker:
                 # TODO
                 quiet=True,
             ) as current_worker:
+                current_worker._update_stage = "current"
                 current_worker.run_copy()
                 self.answers = current_worker.answers
                 self.answers.external = self._external_data()
@@ -1254,6 +1260,7 @@ class Worker:
                 exclude=exclude_plus_removed,
                 vcs_ref=self.resolved_vcs_ref,
             ) as new_worker:
+                new_worker._update_stage = "new"
                 new_worker.run_copy()
             with local.cwd(new_copy):
                 self._git_initialize_repo()
--- a/copier/_template.py
+++ b/copier/_template.py
@@ -535,6 +535,32 @@ class Template:
         return tasks
 
     @cached_property
+    def postrender_tasks(self) -> Sequence[Task]:
+        """Get postrender tasks defined in the template.
+
+        These run after template rendering is complete but before regular
+        tasks execute. During updates, they run on both old_copy and new_copy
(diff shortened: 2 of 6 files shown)


navkast (approved):

sisp:
And why do you consider bulk rename/replace better (simpler, more manageable, ...?) than templating directory and package names? I'd argue that the latter is more robust / less error-prone and involves less "magic" than maintaining a bulk rename/replace script? I mean, I imagine such a script needs to walk the file tree, and the non-Jinja placeholders need to be unique to avoid false matches. :thinking:

sisp:
Thanks for elaborating on your scenario so clearly! :1st_place_medal: 

I understand your point that the Java syntax will be broken by the Jinja markup and that linters etc. won't work. I might not consider it as big a problem as you, but I acknowledge that this is probably a matter of opinion and your position seems valid, too.

I've finally found some time to think about this feature request and PR more thoroughly – sorry for the delay.

I think the current behavior of [tasks](https://copier.readthedocs.io/en/stable/configuring/#tasks) is (almost) what you need, as [tasks are executed right after template rendering](https://github.com/copier-org/copier/blob/3abf23606408bae501457b84b4056bad504fd388/copier/_main.py#L1065). Copier's current update algorithm [generates a fresh copy in a temporary location from the old template version using a regular `run_copy`](https://github.com/copier-org/copier/blob/3abf23606408bae501457b84b4056bad504fd388/copier/_main.py#L1181) call and performs two more copies from the new template version ([one in the real destination directory](https://github.com/copier-org/copier/blob/3abf23606408bae501457b84b4056bad504fd388/copier/_main.py#L1238) and [one in a temporary location](https://github.com/copier-org/copier/blob/3abf23606408bae501457b84b4056bad504fd388/copier/_main.py#L1256)), all of which execute the tasks after rendering. _But_ there is one problem, which I believe would also affect your suggested solution: The task execution after the [copy in the real destination directory]((https://github.com/copier-org/copier/blob/3abf23606408bae501457b84b4056bad504fd388/copier/_main.py#L1238)) occurs in the context of the real destination directory whose file tree is unpredictable, as the project may have diverged arbitrarily from the template. So it is entirely possible that your placeholder directory exists in the project, it would be (partially) overwritten by the internal `run_copy` call, and your post-rendering task would illegally rename it, thereby modifying the project in an unintended way. The kind of post-rendering tasks you need work _only_ on fresh copies from the old and new template versions _in temporary locations_ to avoid unintended side effects due to unpredictable project file trees. I have implemented a new update algorithm in #2376 which satisfies this requirement – except there's [post-update tasks execution](https://github.com/copier-org/copier/pull/2376/changes#diff-b55c5e654639cc2f01fc39698134da496f874d2e706c97c8eeba50681978abe4R1329) for backwards compatibility with making a [copy in the real destination directory](https://github.com/copier-org/copier/blob/3abf23606408bae501457b84b4056bad504fd388/copier/_main.py#L1238). That said, I don't quite like these tasks running in the real destination in the middle of the update algorithm, I don't think the purpose of these tasks is well-defined there. So, perhaps we can remove these task runs in the new update algorithm and advise template authors to use a post-migration task in case they intend to run a task after an update, which has clear semantics IMO. This will be a breaking change – but releasing the new update algorithm will be a breaking change anyway, as we'll remove support for merge conflicts via `.rej` files and possibly introduce slightly different 3-way merge behavior due to using the `git merge` command instead of a sequence of low-level Git plumbing commands.

If the new update algorithm with regular tasks offers a solution for your use case, it still means you're dependent on tasks which is an unsafe feature, so template users need to pass the [`--trust`](https://copier.readthedocs.io/en/stable/configuring/#unsafe) flag. I've been thinking about a more native approach to this kind of problem inspired by middlewares in web frameworks. Perhaps Copier could offer a way to hook into the generation/rendering process, similar to [`copier-template-extensions`](https://github.com/copier-org/copier-template-extensions) but using plain Jinja, allowing to perform rewrites like file/directory names, Jinja template post-processing, etc. This is an ad-hoc sketch of how I imagine this could look in `copier.yml`:

```yaml
package:
  type: str

_middleware: |
  {%- if _copier.middleware.event == 'create_path' -%}
  {%- if _copier.middleware.path.match('src/*/java/com/example/boilerplate') -%}
  {{ _copier.middlware.path.parent / package }}
  {%- endif -%}
  {%- endif -%}
```

We'd need to define the available events, context data, and expected "return" values of such a middleware.

WDYT?

PR #2393, opened 2025-11-11
refactor: drop support for Python 3.9

sisp:
I've dropped support for Python 3.9, as it reached its [EOL on October 31, 2025](https://devguide.python.org/versions/#unsupported-versions). This change comes with some cleanup, mostly automated through Ruff, to reduce technical debt, as we can use features that are available only for Python 3.10+ now.


sisp:
Thanks for your review, @pawamoy, much appreciated! :bow:

Issue #2426, opened 2025-12-14
Symlinks to outside the destination directory not updatable since v9.9.1


cbrown1234:
### Describe the problem

Following this change https://github.com/copier-org/copier/commit/fdbc0167cc22780b497e4db176feaf6f024757d6 `_preserve_symlinks: true` templates with symlinks that pointed outside the destination directory could not update.

This broke a template I use for subprojects of a monorepo that had symlinks referencing top level (parent dir) content. 

### Template

See test case in linked PR

### To Reproduce

See test case in linked PR

### Logs

```console

```

### Expected behavior

Templates should be able to create symlinks that point outside the destination directory, but not render content there

### Operating system

Linux

### Operating system distribution and version

Ubuntu 22.04

### Copier version

v9.9.1

### Python version

Any

### Installation method

pipx+pypi

PR #2427, opened 2025-12-14
fix: Re-enable updating of symlinks to outside subproject

cbrown1234:
Fix for https://github.com/copier-org/copier/issues/2426

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -715,7 +715,7 @@ class Worker:
     def _render_template(self) -> None:
         """Render the template in the subproject root."""
         follow_symlinks = not self.template.preserve_symlinks
-        cwd = Path.cwd()
+        dst_root = self.dst_path.resolve()
         for src in scantree(str(self.template_copy_root), follow_symlinks):
             src_abspath = Path(src.path)
             src_relpath = Path(src_abspath).relative_to(self.template.local_abspath)
@@ -723,7 +723,17 @@ class Worker:
                 Path(src_abspath).relative_to(self.template_copy_root)
             )
             for dst_relpath, ctx in dst_relpaths_ctxs:
-                if not cwd.joinpath(dst_relpath).resolve().is_relative_to(cwd):
+                dst_abspath = dst_root / dst_relpath
+                if dst_abspath.is_symlink() and self.template.preserve_symlinks:
+                    # If destination path is a symlink, it can safely point outside the
+                    # subproject dir, while still itself existing within the subproject.
+                    # (So long as nothing is templated into it (if it is a directory),
+                    # which would be caught by that path's own check.)
+                    # Therefore avoid resolving the symlink itself:
+                    dst_realpath = dst_abspath.parent.resolve() / dst_abspath.name
+                else:
+                    dst_realpath = dst_abspath.resolve()
+                if not dst_realpath.is_relative_to(dst_root):
                     raise ForbiddenPathError(path=dst_relpath)
                 if self.match_exclude(dst_relpath):
                     continue
--- a/tests/test_symlinks.py
+++ b/tests/test_symlinks.py
@@ -505,11 +505,11 @@ def test_symlinked_to_outside_destination_absolute(
         git("add", "-A")
         git("commit", "-m", "first commit on dst")
 
-        # While update appears like a no-op, this part of the test captures
-        # possible changes in behaviour when the symlinks actually exist
-        # (having been created by the initial copy) as was the case for
-        # https://github.com/copier-org/copier/issues/2426
-        run_update(dst, defaults=True, overwrite=True)
+    # While update appears like a no-op, this part of the test captures
+    # possible changes in behaviour when the symlinks actually exist
+    # (having been created by the initial copy) as was the case for
+    # https://github.com/copier-org/copier/issues/2426
+    run_update(dst, defaults=True, overwrite=True)
 
     assert (dst / "symlink_dir").is_symlink()
     assert (dst / "symlink_dir" / "outside.txt").read_text() == "outside"
@@ -559,11 +559,11 @@ def test_symlinked_to_outside_destination_relative(
         git("add", "-A")
         git("commit", "-m", "first commit on dst")
 
-        # While update appears like a no-op, this part of the test captures
-        # possible changes in behaviour when the symlinks actually exist
-        # (having been created by the initial copy) as was the case for
-        # https://github.com/copier-org/copier/issues/2426
-        run_update(dst, defaults=True, overwrite=True)
+    # While update appears like a no-op, this part of the test captures
+    # possible changes in behaviour when the symlinks actually exist
+    # (having been created by the initial copy) as was the case for
+    # https://github.com/copier-org/copier/issues/2426
+    run_update(dst, defaults=True, overwrite=True)
 
     assert (dst / "symlink_dir").is_symlink()
     assert (dst / "symlink_dir" / "outside.txt").read_text() == "outside"


tests/test_symlinks.py
```diff
@@ -460,3 +460,52 @@ def test_symlinked_dir_expanded(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+                # Changes here will be overwritten by Copier
+                {{ _copier_answers|to_nice_yaml }}
+            """,
+            src / "copier.yaml": """\
+            _preserve_symlinks: true
+            _subdirectory: template
+            """,
+            tmpl / "a_file.txt": "some content",
+            tmpl / "a_symlink.txt": other / "outside.txt",
+            tmpl / "symlink_dir": other,
```
sisp: Just an idea why the test might be failing on Windows (I'm on mobile right now, can't test it): Perhaps an absolute path to the symlink target doesn't work. Looking at other tests in this file, the targets are relative paths.

copier/_main.py
```diff
@@ -715,15 +715,23 @@ def match_skip(self) -> Callable[[Path], bool]:
...
-                if not cwd.joinpath(dst_relpath).resolve().is_relative_to(cwd):
+                path_to_check = dst_root.joinpath(dst_relpath)
+                if path_to_check.is_symlink():
+                    # If destination path is a symlink, it can safely point outside subproject dir,
+                    # so long as nothing is templated into it (which would be caught by its own check),
+                    # while still itself existing within the subproject.
+                    # Therefore check parent is within subproject to avoid resolving the symlink itself
+                    # Note: Typically occurs when updating preserve_symlinks = true templates
+                    path_to_check = path_to_check.parent
+                if not path_to_check.resolve().is_relative_to(dst_root):
```
sisp: I think the fix is correct, but I needed a moment to consider the implications of checking the parent directory of a symlink to avoid following the symlink via `<symlink>.resolve()`. I checked the `pathlib` docs, and it seems there really is no alternative to `Path.resolve()` that only normalizes a path without resolving a symlink. But all we really need is a purely lexical normalization of the symlink file path to _normalize_ path segments like `..`; [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath) offers exactly this. What do you think about simplifying the check by applying `os.path.normpath` on the full path and removing the parent directory "hack"?

cbrown1234: Yeah, I did review the pathlib docs too, and also couldn't find a good alternative, unfortunately. 

I had assumed that the behaviour of  `.resolve()` in resolving any intermediate symlinks on the way to the target file was desired/intentional, and it wasn't just the path normalisation that was needed. e.g. to check an intermediate symlink in the path didn't redirect you outside the destination directory. We'd lose this behaviour with [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath)

The check of the parent seems sound as `dst_root.is_relative_to(dst_root) == True` and the resolved parent dir + filename has to be the same file to my understanding

We could even potentially simply the logic and check the resolved parent for every case. It shouldn't matter if the file/dir is a symlink or not

tests/test_symlinks.py
cbrown1234: Nice spot, I've added a relative symlink path test that passes and xfailed on windows for this for now, although maybe it's another separate bug?

tests/test_symlinks.py
```diff
@@ -460,3 +461,109 @@ def test_symlinked_dir_expanded(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+
+
+def test_symlinked_to_outside_destination_relative(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    src, dst_parent = map(tmp_path_factory.mktemp, ("src", "dst_parent"))
+    dst = dst_parent / "dst"
+    dst.mkdir()
+    other = dst_parent / "other"
+    other.mkdir()
```
sisp: What do you think about some syntactic sugar? 🙂 


```suggestion
    (dst := dst_parent / "dst").mkdir()
    (other := dst_parent / "other").mkdir()
```

tests/test_symlinks.py
```diff
@@ -460,3 +461,109 @@ def test_symlinked_dir_expanded(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+                # Changes here will be overwritten by Copier
+                {{ _copier_answers|to_nice_yaml }}
+            """,
+            src / "copier.yaml": """\
+            _preserve_symlinks: true
+            _subdirectory: template
+            """,
+            tmpl / "a_file.txt": "some content",
+            tmpl / "a_symlink.txt": other_rel / "outside.txt",
+            tmpl / "symlink_dir": other_rel,
```
sisp: Same as above regarding removal of the `_subdirectory` setting:

```suggestion
    build_file_tree(
        {
            src / ".copier-answers.yml.jinja": """\
                # Changes here will be overwritten by Copier
                {{ _copier_answers|to_nice_yaml }}
            """,
            src / "copier.yaml": """\
            _preserve_symlinks: true
            """,
            src / "a_file.txt": "some content",
            src / "a_symlink.txt": other_rel / "outside.txt",
            src / "symlink_dir": other_rel,
```

WDYT?

tests/test_symlinks.py
```diff
@@ -460,3 +461,109 @@ def test_symlinked_dir_expanded(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+                # Changes here will be overwritten by Copier
+                {{ _copier_answers|to_nice_yaml }}
+            """,
+            src / "copier.yaml": """\
+            _preserve_symlinks: true
+            _subdirectory: template
+            """,
+            tmpl / "a_file.txt": "some content",
+            tmpl / "a_symlink.txt": other / "outside.txt",
+            tmpl / "symlink_dir": other,
```
sisp: I think I'd prefer removing the `_subdirectory` setting to avoid testing it here implicitly as wel (and to simplify the test a little):


```suggestion
    build_file_tree(
        {
            src / ".copier-answers.yml.jinja": """\
                # Changes here will be overwritten by Copier
                {{ _copier_answers|to_nice_yaml }}
            """,
            src / "copier.yaml": """\
            _preserve_symlinks: true
            """,
            src / "a_file.txt": "some content",
            src / "a_symlink.txt": other / "outside.txt",
            src / "symlink_dir": other,
```

WDYT?

tests/test_symlinks.py
sisp: I've tried debugging this a bit more but learned that Windows 11 requires "developer mode" enabled or admin privileges, none of which I have on my Windows VM. 😕 Perhaps the problem is that Windows – for some reason – decides to create a _junction_ instead of a true symlink in this case, so `.is_symlink()` returns `False`. Anyway, I think the `@pytest.mark.xfailed` decorator is sufficient here. Windows ... :shrug:

tests/test_symlinks.py
```diff
@@ -460,3 +461,109 @@ def test_symlinked_dir_expanded(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        git("init")
+        git("add", "-A")
+        git("commit", "-m", "first commit on dst")
+
+    run_update(dst, defaults=True, overwrite=True)
+
+    assert (dst / "symlink_dir").is_symlink()
+    assert (dst / "symlink_dir" / "outside.txt").read_text() == "outside"
+    assert (dst / "a_symlink.txt").is_symlink()
+    assert (dst / "a_symlink.txt").read_text() == "outside"
```
sisp, line 517: I wonder whether we need to test updating as well, especially since this update is a no-op. The test above that perform an update do actually test symlink-specific update behavior. Just trying to keep a clear test scope.

```suggestion
```

WDYT?

tests/test_symlinks.py
```diff
@@ -460,3 +461,109 @@ def test_symlinked_dir_expanded(tmp_path_factory: pytest.TempPathFactory) -> Non
...
```
sisp, line 571: Same as above regarding the removal of testing a no-op update:


```suggestion
```

WDYT?

tests/test_symlinks.py
cbrown1234: Even with those settings enabled on my PC I couldn't get a clean test run for the existing tests, so I didn't have confidence in debugging new ones

copier/_main.py
cbrown1234: Outdated discussion

tests/test_symlinks.py
cbrown1234: I'm mentally still living in the world of Python 3.7😂 sure, why not, I've updated it as suggested

tests/test_symlinks.py
cbrown1234: Updated as suggested

tests/test_symlinks.py
cbrown1234: Updated as suggested

tests/test_symlinks.py
cbrown1234: The updates were actually needed to demonstrate the original issue https://github.com/copier-org/copier/issues/2426, this was because of the side effect of the `.resolve()` behaviour being different when the symlink actually existing, having been created in the first copy.

This bug was sort of masked by the other bug of the directories being compared relative to the cwd, rather than the destination directory until this fix, which is why the run_update is unindented in the fix commit

Given it wasn't obvious why it was there, I've added a comment

tests/test_symlinks.py
cbrown1234: See reply in other thread

tests/test_symlinks.py
sisp: Thanks for clarifying this and adding a comment! :ok_hand:

PR #2432, opened 2025-12-17
Insure git hooks are ignored on checkout during conflicts

TheSuperiorStanislav:
Issues arise when you have post-checkout hooks, and you have conflicts, and also updated `.pre-commit-hooks.yaml` file

```console
plumbum.commands.processes.ProcessExecutionError: Unexpected exit code: 1
Command line: | /usr/bin/git checkout -- .github/settings.yaml
Stderr:       | [ERROR] Your pre-commit configuration is unstaged.
              | `git add .pre-commit-config.yaml` to fix this.
```

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1335,7 +1335,15 @@ class Worker:
                         # Remove ".rej" suffix
                         fname = fname[:-4]
                         # Undo possible non-rejected chunks
-                        git("checkout", "--", fname)
+                        git(
+                            # Ignore hooks to avoid errors from them or
+                            # issues when .pre-commit-config.yaml is changed
+                            "-c",
+                            "core.hooksPath=/dev/null",
+                            "checkout",
+                            "--",
+                            fname,
+                        )
                         # 3-way-merge the file directly
                         git(
                             "merge-file",
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1339,7 +1339,7 @@ class Worker:
                             # Ignore hooks to avoid errors from them or
                             # issues when .pre-commit-config.yaml is changed
                             "-c",
-                            "core.hooksPath=/dev/null",
+                            f"core.hooksPath={os.devnull}",
                             "checkout",
                             "--",
                             fname,
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -491,7 +491,7 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
 )
 @pytest.mark.impure
 def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) -> None:
-    """Ignore post-checkout hook when conflicts are encountered and .pre-commit-config.yaml is updated."""
+    """Ignore post-checkout hook when conflicts are encountered."""
     # Prepare source template v1
     src, dst1 = map(tmp_path_factory.mktemp, ("src", "dst1"))
     with local.cwd(src):
@@ -503,8 +503,8 @@ def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) ->
                     _templates_suffix: {SUFFIX_TMPL}
                     _tasks:
                         - git init
-                        - pre-commit install
-                        - pre-commit run -a || true
+                        - git config --local core.hooksPath hooks/
+                        - chmod +x hooks/post-checkout
                     """
                 ),
                 "[[ _copier_conf.answers_file ]].tmpl": (
@@ -513,29 +513,9 @@ def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) ->
                     """
                 ),
                 "test.txt": "This is a file",
-                ".pre-commit-config.yaml": (
+                "hooks/post-checkout": (
                     r"""
-                    default_install_hook_types: [
-                        pre-commit,
-                        pre-push,
-                        post-checkout,
-                        post-merge,
-                        post-rewrite,
-                    ]
-                    repos:
-                    -   repo: local
-                        hooks:
-                        -   id: echo
-                            name: echo V1
-                            entry: exit 0
-                            language: system
-                            pass_filenames: false
-                            types: [ file ]
-                            stages: [
-                                post-checkout,
-                                post-merge,
-                                post-rewrite,
-                            ]
+                    echo "Post-checkout hook executed"
                     """
                 ),
             }
@@ -561,44 +541,23 @@ def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) ->
         git("add", ".")
         git("commit", "-am", "feat: edit test.txt")
     # Evolve source template to v2
+    # No errors should be raised due to post-checkout hook
     with local.cwd(src):
         build_file_tree(
             {
-                ".pre-commit-config.yaml": (
+                "test.txt": "This is a edited file in v2",
+                "hooks/post-checkout": (
                     r"""
-                    default_install_hook_types: [
-                        pre-commit,
-                        pre-push,
-                        post-checkout,
-                        post-merge,
-                        post-rewrite,
-                    ]
-                    repos:
-                    -   repo: local
-                        hooks:
-                        -   id: echo
-                            name: echo V2
-                            entry: exit 0
-                            language: system
-                            pass_filenames: false
-                            types: [ file ]
-                            stages: [
-                                post-checkout,
-                                post-merge,
-                                post-rewrite,
-                            ]
+                    exit 1
                     """
                 ),
-                "test.txt": "This is a edited file in v2",
             }
         )
         git("init")
         git("add", ".")
-        git("commit", "-m", "feat: update .pre-commit-config.yaml")
+        git("commit", "-m", "feat: Update post-checkout")
         git("tag", "v2")
     # Update subproject to v2
-    # No errors should be raised due to post-checkout hook when
-    # encountering conflicts and .pre-commit-config.yaml is updated
     run_update(
         dst_path=dst1,
         defaults=True,
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import platform
+import stat
 from pathlib import Path
 from shutil import rmtree
 from textwrap import dedent
@@ -321,7 +322,7 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
                     _templates_suffix: {SUFFIX_TMPL}
                     _tasks:
                         - git init
-                        - pre-commit install
+                        - pre-commit install -t pre-commit -t commit-msg
                         - pre-commit run -a || true
                     what: grog
                     """
@@ -333,10 +334,6 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
                 ),
                 ".pre-commit-config.yaml": (
                     r"""
-                    default_install_hook_types: [
-                        pre-commit,
-                        commit-msg,
-                    ]
                     repos:
                     -   repo: https://github.com/pre-commit/mirrors-prettier
                         rev: v2.0.4
@@ -484,16 +481,11 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
         assert Path(f"{life}.rej").is_file()
 
 
-# Checkout test_commit_hooks_respected
-# FIXME Some generous Windows power user please fix this test!
-@pytest.mark.xfail(
-    condition=platform.system() == "Windows", reason="Git broken on Windows?"
-)
 @pytest.mark.impure
 def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) -> None:
     """Ignore post-checkout hook when conflicts are encountered."""
     # Prepare source template v1
-    src, dst1 = map(tmp_path_factory.mktemp, ("src", "dst1"))
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
     with local.cwd(src):
         build_file_tree(
             {
@@ -503,8 +495,6 @@ def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) ->
                     _templates_suffix: {SUFFIX_TMPL}
                     _tasks:
                         - git init
-                        - git config --local core.hooksPath hooks/
-                        - chmod +x hooks/post-checkout
                     """
                 ),
                 "[[ _copier_conf.answers_file ]].tmpl": (
@@ -513,11 +503,6 @@ def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) ->
                     """
                 ),
                 "test.txt": "This is a file",
-                "hooks/post-checkout": (
-                    r"""
-                    echo "Post-checkout hook executed"
-                    """
-                ),
             }
         )
         git("init")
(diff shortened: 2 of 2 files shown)


pawamoy:
Can you explain your setup in more details please? Are you using pre-commit, the tool? Why does it prevent a `git checkout -- file` on an file unrelated to pre-commit config? Would `git restore file` bypass this check?

copier/_main.py
```diff
@@ -1335,7 +1335,15 @@ def _apply_update(self) -> None:  # noqa: C901
                         # Remove ".rej" suffix
                         fname = fname[:-4]
                         # Undo possible non-rejected chunks
-                        git("checkout", "--", fname)
+                        git(
+                            # Ignore hooks to avoid errors from them or
+                            # issues when .pre-commit-config.yaml is changed
+                            "-c",
+                            "core.hooksPath=/dev/null",
```
pawamoy: You mention a Windows issue in tests, that might be the cause: there's no `/dev/null` on Windows.

TheSuperiorStanislav: No i copied message from test above, since it's similar

TheSuperiorStanislav: Git doesn't complain about unknown paths

image

pawamoy: I see, thanks!

pawamoy:
OK, thanks. Could you try to see if `git restore` triggers the same post-checkout hook? That's the modern version of `git checkout -- file`.

TheSuperiorStanislav:
> OK, thanks. Could you try to see if `git restore` triggers the same post-checkout hook? That's the modern version of `git checkout -- file`.

Yes it triggers it

sisp:
I wonder, is it generally valid to ignore post-checkout hooks? AFAIK, for example Git LFS' smudge filter is run after checkout to resolve the LFS pointers and fetch the actual blobs from the LFA server. Not 100% sure whether that's relevant in this case though.

copier/_main.py
sisp: Is [`os.devnull`](https://docs.python.org/3/library/os.html#os.devnull) perhaps more cross-platform compatible (or at least less hardcoded)?


```suggestion
                            f"core.hooksPath={os.devnull}",
```

tests/test_updatediff.py
sisp, line 1: I'd prefer a pure, minimal test that uses a small post-checkout hook script to test this scenario in isolation as much as possible rather than using pre-commit.

sisp (commented):
> I wonder if #2376 wouldn't fix this too.

@pawamoy I imagine a similar problem would occur because [there's also a `git checkout` call](https://github.com/copier-org/copier/pull/2376/changes#diff-b55c5e654639cc2f01fc39698134da496f874d2e706c97c8eeba50681978abe4R1319-R1325).

---

> I wonder, is it generally valid to ignore post-checkout hooks? AFAIK, for example Git LFS' smudge filter runs after checkout to resolve the LFS pointers and fetch the actual blobs from the LFS server. Not 100% sure whether that's relevant in this case though.

I've read up on Git LFS, it uses a [post-checkout hook](https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-post-checkout.adoc) for making locked files read-only in the working copy; this is probably irrelevant in our case. But I still wonder a bit about the general validity of ignoring all hooks here.

tests/test_updatediff.py
TheSuperiorStanislav: Simplified test to use simple post-checkout hook file

tests/test_updatediff.py
```diff
@@ -321,7 +321,7 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
                     _templates_suffix: {SUFFIX_TMPL}
                     _tasks:
                         - git init
-                        - pre-commit install -t pre-commit -t commit-msg
+                        - pre-commit install
```
sisp: Please revert this change, as it's unrelated to this PR.

tests/test_updatediff.py
```diff
@@ -333,6 +333,10 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
                 ),
                 ".pre-commit-config.yaml": (
                     r"""
+                    default_install_hook_types: [
+                        pre-commit,
+                        commit-msg,
+                    ]
```
sisp: Please revert this change, as it's unrelated to this PR.

tests/test_updatediff.py
```diff
@@ -480,6 +484,90 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+# Checkout test_commit_hooks_respected
+# FIXME Some generous Windows power user please fix this test!
+@pytest.mark.xfail(
+    condition=platform.system() == "Windows", reason="Git broken on Windows?"
+)
+@pytest.mark.impure
+def test_post_checkout_hook_ignored(tmp_path_factory: pytest.TempPathFactory) -> None:
+    """Ignore post-checkout hook when conflicts are encountered."""
+    # Prepare source template v1
+    src, dst1 = map(tmp_path_factory.mktemp, ("src", "dst1"))
```
sisp: For simplicity:


```suggestion
    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
```

tests/test_updatediff.py
```diff
@@ -480,6 +484,90 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
         assert Path(f"{life}.rej").is_file()
 
 
+# Checkout test_commit_hooks_respected
+# FIXME Some generous Windows power user please fix this test!
+@pytest.mark.xfail(
+    condition=platform.system() == "Windows", reason="Git broken on Windows?"
+)
```
sisp: Does this new test really fail on Windows? If not, please remove the decorator.

tests/test_updatediff.py
```diff
@@ -480,6 +484,90 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
         assert Path(f"{life}.rej").is_file()
 
 
+# Checkout test_commit_hooks_respected
```
sisp: A leftover from before? If so, let's remove it.


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +484,90 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        build_file_tree(
+            {
+                "copier.yml": (
+                    f"""
+                    _envops: {BRACKET_ENVOPS_JSON}
+                    _templates_suffix: {SUFFIX_TMPL}
+                    _tasks:
+                        - git init
+                        - git config --local core.hooksPath hooks/
+                        - chmod +x hooks/post-checkout
```
sisp: I'd prefer a manual hook setup instead of using Copier tasks, as I think it's (a) a little simpler, (b) doesn't depend on an unrelated Copier feature, and (c) doesn't depend on the `core.hooksPath` config. Instead, I'd set up the hook in the destination path after the commits have been made:

```python
run_copy(...)
with local.cwd(dst):
    ...
hook_file = dst / ".git" / "hooks" / "post-checkout"
hook_file.write_text("exit 1")
hook_file.chmod(hook_file.stat().st_mode | stat.S_IXUSR)
...
```

(You need to add `import stat` at the top.)

tests/test_updatediff.py
```diff
@@ -480,6 +484,90 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        # Introduce conflict
+        Path(f"{dst1}/test.txt").open(mode="w").write("This is a conflicting change")
+        git("add", ".")
+        git("commit", "-am", "feat: edit test.txt")
+    # Evolve source template to v2
+    # No errors should be raised due to post-checkout hook
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "test.txt": "This is a edited file in v2",
```
sisp:
```suggestion
                "test.txt": "This is an edited file in v2",
```

🤓

tests/test_updatediff.py
```diff
@@ -480,6 +484,90 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        defaults=True,
+        overwrite=False,
+        unsafe=True,
+    )
+    with local.cwd(dst1):
+        git("add", ".")
+        # Commit initial copy
+        git("commit", "-am", "feat: copied v1")
+        # Introduce conflict
+        Path(f"{dst1}/test.txt").open(mode="w").write("This is a conflicting change")
```
sisp:
```suggestion
        (dst / "test.txt").write_text("This is a conflicting change")
```

tests/test_updatediff.py
```diff
@@ -480,6 +484,90 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        git("add", ".")
+        git("commit", "-am", "feat: edit test.txt")
+    # Evolve source template to v2
+    # No errors should be raised due to post-checkout hook
```
sisp: If you'd like to keep this comment, it should probably be moved further down to right above the `run_update` function call.

sisp (changes requested):

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "copier.yml": (
+                    f"""
+                    _envops: {BRACKET_ENVOPS_JSON}
+                    _templates_suffix: {SUFFIX_TMPL}
+                    _tasks:
+                        - git init
```
sisp: I'd remove the task and just run `git init` in the destination directory manually to keep it simple.


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        git("add", ".")
+        git("commit", "-m", "feat: commit 1")
+        git("tag", "v1")
+    # Copy source template
+    run_copy(
+        src_path=str(src),
+        dst_path=dst,
+        defaults=True,
+        overwrite=False,
+        unsafe=True,
```
sisp: When you remove the task, there's no need for running with unsafe features enabled anymore:


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        git("init")
+        git("add", ".")
+        git("commit", "-m", "feat: commit 1")
+        git("tag", "v1")
+    # Copy source template
+    run_copy(
+        src_path=str(src),
+        dst_path=dst,
+        defaults=True,
+        overwrite=False,
```
sisp: This is the default setting anyway, so we can remove it:


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        )
+        git("init")
+        git("add", ".")
+        git("commit", "-m", "feat: commit 1")
+        git("tag", "v1")
+    # Copy source template
+    run_copy(
+        src_path=str(src),
+        dst_path=dst,
+        defaults=True,
```
sisp: There are no questions, so no need to use default answers:


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        overwrite=False,
+        unsafe=True,
+    )
+    with local.cwd(dst):
+        git("add", ".")
+        # Commit initial copy
+        git("commit", "-am", "feat: copied v1")
+        # Introduce conflict
+        Path(dst / "test.txt").open(mode="w").write("This is a conflicting change")
```
sisp: `(dst / "test.txt")` is already a path, no need to wrap it with `Path(...)`. However, the current working directory is already `dst`, so using the absolute path isn't necessary. We could simplify this to `Path("test.txt")`. And instead of chaining `.open(...).write(...)`, which leaves the file handle open, it's better to write the file content with `.write_text(...)`.


```suggestion
        Path("test.txt").write_text("This is a conflicting change")
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        git("add", ".")
+        git("commit", "-am", "feat: edit test.txt")
+        # Add post-checkout hook that fails
+        hook_file = dst / ".git" / "hooks" / "post-checkout"
```
sisp: Since the current working directory is `dst`, there's no need to use the absolute path:


```suggestion
        hook_file = Path(".git") / "hooks" / "post-checkout"
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+    # Copy source template
+    run_copy(
+        src_path=str(src),
+        dst_path=dst,
+        defaults=True,
+        overwrite=False,
+        unsafe=True,
+    )
+    with local.cwd(dst):
+        git("add", ".")
```
sisp, line 516: When you remove the task, we need to run `git init` here:


```suggestion
        git("init")
        git("add", ".")
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        hook_file.write_text("exit 1")
+        hook_file.chmod(hook_file.stat().st_mode | stat.S_IXUSR)
+    # Evolve source template to v2
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "test.txt": "This is an edited file in v2",
+            }
+        )
+        git("init")
```
sisp: The source location is already a Git repository, no need to re-initialize it:


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        git("add", ".")
+        git("commit", "-m", "feat: Update post-checkout")
+        git("tag", "v2")
+    # Update subproject to v2
+    # No errors should be raised due to post-checkout hook
+    run_update(
+        dst_path=dst,
+        defaults=True,
```
sisp: There are no questions, so no need to use default answers:


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        overwrite=True,
+        unsafe=True,
+        skip_tasks=True,
```
sisp: When you remove the task, there's no need for skipping tasks anymore:


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
...
+        conflict="inline",
```
sisp: This is the default setting anyway, so we can remove it:


```suggestion
```

tests/test_updatediff.py
```diff
@@ -480,6 +481,77 @@ def test_commit_hooks_respected(tmp_path_factory: pytest.TempPathFactory) -> Non
         assert Path(f"{life}.rej").is_file()
 
 
+@pytest.mark.impure
```
sisp: This test doesn't depend on external (network) resources anymore (like pre-commit repos), so we don't need to mark it as impure:


```suggestion
```

sisp (changes requested):

sisp (approved):
LGTM! :tada: Thanks for your contribution and patience with my recurring review suggestions. :bow:

PR #2456, opened 2026-01-13
fix: disallow symlink-based includes outside template root

sisp:
WDYT @cbrown1234?

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -718,6 +718,20 @@ class Worker:
         dst_root = self.dst_path.resolve()
         for src in scantree(str(self.template_copy_root), follow_symlinks):
             src_abspath = Path(src.path)
+            # If the source is a symlink, we are not preserving symlinks, and the
+            # symlink target is outside the template root, this means that we are
+            # copying a file/directory from outside the template, which is
+            # forbidden, so raise an error.
+            if (
+                src_abspath.is_symlink()
+                and not self.template.preserve_symlinks
+                and not (src_abspath.resolve()).is_relative_to(
+                    self.template.local_abspath
+                )
+            ):
+                raise ForbiddenPathError(
+                    path=src_abspath.relative_to(self.template_copy_root)
+                )
             src_relpath = Path(src_abspath).relative_to(self.template.local_abspath)
             dst_relpaths_ctxs = self._render_path(
                 Path(src_abspath).relative_to(self.template_copy_root)
--- a/tests/test_symlinks.py
+++ b/tests/test_symlinks.py
@@ -1,11 +1,13 @@
 import os
+import re
 from pathlib import Path
+from tempfile import gettempdir
 
 import pytest
 from plumbum import local
 
 from copier import run_copy, run_update
-from copier.errors import DirtyLocalWarning
+from copier.errors import DirtyLocalWarning, ForbiddenPathError
 
 from .helpers import build_file_tree, git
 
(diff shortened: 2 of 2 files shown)


copier/_main.py
```diff
@@ -718,6 +718,20 @@ def _render_template(self) -> None:
...
+            if (
+                src_abspath.is_symlink()
+                and not self.template.preserve_symlinks
+                and not (src_abspath.resolve()).is_relative_to(
+                    self.template.local_abspath
+                )
+            ):
+                raise ForbiddenPathError(
+                    path=src_abspath.relative_to(self.template_copy_root)
+                )
```
cbrown1234, line 734: Could the errors be clearer to users if there were a separate error for reading/src content from writing/dst content?

copier/_main.py
```diff
@@ -718,6 +718,20 @@ def _render_template(self) -> None:
         dst_root = self.dst_path.resolve()
         for src in scantree(str(self.template_copy_root), follow_symlinks):
             src_abspath = Path(src.path)
+            # If the source is a symlink, we are not preserving symlinks, and the
+            # symlink target is outside the template root, this means that we are
+            # copying a file/directory from outside the template, which is
+            # forbidden, so raise an error.
+            if (
+                src_abspath.is_symlink()
+                and not self.template.preserve_symlinks
```
cbrown1234, line 727: Do we need these conditions? Isn't any attempt to read from outside the src (template_copy_root) directory unsafe?

If we are preserving symlinks we just need to not resolve the path for the check below I think

copier/_main.py
```diff
@@ -718,6 +718,20 @@ def _render_template(self) -> None:
         dst_root = self.dst_path.resolve()
         for src in scantree(str(self.template_copy_root), follow_symlinks):
             src_abspath = Path(src.path)
+            # If the source is a symlink, we are not preserving symlinks, and the
+            # symlink target is outside the template root, this means that we are
+            # copying a file/directory from outside the template, which is
+            # forbidden, so raise an error.
+            if (
+                src_abspath.is_symlink()
```
cbrown1234, line 726: Because symlinks have already been followed above when we're not preserving symlinks, isn't the check here for symlinks going to be false unless the src directory is messed with while the loop is iterating? I don't think it's catching the standard case you intend to catch

cbrown1234 (commented):
~~I'm not 100% sure I'm right here, but this doesn't seem quite correct to me currently~~

copier/_main.py
cbrown1234: To clarify a bit: 
> If we are preserving symlinks we just need to not resolve the path for the check below I think

I think to be fully correct, we should check `src_abspath.parent.resolve() / src_abspath.name` when we are preserving symlinks

copier/_main.py
cbrown1234: Thinking about it again, I realised that `scantree` is following symlinks and not resolving, so this might not be relevant

copier/_main.py
```diff
@@ -718,6 +718,20 @@ def _render_template(self) -> None:
...
+                and not self.template.preserve_symlinks
+                and not (src_abspath.resolve()).is_relative_to(
+                    self.template.local_abspath
+                )
+            ):
```
cbrown1234, line 731: Finally found some time to properly check the suggestion I was making:
```suggestion
            if self.template.preserve_symlinks:
                 # Avoid resolving a potential symlink itself
                src_realpath = src_abspath.parent.resolve() / src_abspath.name
            else:
                src_realpath = src_abspath.resolve()

            if not src_realpath.is_relative_to(self.template.local_abspath):
```

copier/_main.py
sisp: Exactly, `scantree` follows directory symlinks when walking the source tree, it doesn't resolve any symlinks.

copier/_main.py
sisp: I _think_ the result is the same.

When we preserve symlinks, [`scantree` will _not_ follow symlinks](https://github.com/copier-org/copier/blob/f307b3889c81903f1e832f1cbee8413f74edaa3b/copier/_main.py#L717-L719), so the symlink will be copied and its target should not be constrained. This also means, a resolved source path can never resolve to a location outside the template root because its _unresolved_ path can never contain a directory symlink. In other words, `src_abspath.parent.resolve()` will always resolve to a path inside the template root, so `ForbiddenPathError` will never be raised.

When we don't preserve symlinks, then we only need to check whether a symlink is not inside the template root. [`scantree` yields a path before following a directory symlink](https://github.com/copier-org/copier/blob/f307b3889c81903f1e832f1cbee8413f74edaa3b/copier/_tools.py#L265-L266), so for both a directory or file symlink `(src_abspath.resolve()).is_relative_to(self.template.local_abspath)` will be `False` and the `ForbiddenPathError` will be raised before any content render operation is happening.

WDYT?

copier/_main.py
sisp: Might be, but I'm not sure whether this fine distinction is worth the effort at the moment. If you see a benefit, I'd like to defer this change to a follow-up PR.

copier/_main.py
sisp: I think this isn't needed – see https://github.com/copier-org/copier/pull/2456#discussion_r2704547332.

copier/_main.py
cbrown1234: I might have been overthinking this a little. I was thinking about attack vectors where someone could try to introduce symlinks into the src during the iteration of scantree, tricking the symlink following. However, the manner that scantree lazily yields the dir first means this seems unlikely. The write checks should also prevent writing into the src dir (assuming it's not also the destination dir).

So I agree with your logic, given: 
- that the paths we are iterating over are from the template copy / src (rather than the destination directory), because it is safe from modification by safe template actions
- the current implementation of scantree

I guess my motivation for reducing the number of conditions on the if statement is that it felt like it was reducing the possible surface area for attack, and making the checks less dependent on the specific implementation of the outer loop (scantree)

Your approach is probably also faster due to the lazy evaluation of the `and`s compared to my approach of requiring the path resolution.

cbrown1234:
LGTM, given the discussion

copier/_main.py
sisp: Thank you very much for the thorough review and thoughtful discussion about this delicate piece of code! :pray: It's important to get this right, your feedback gives me confidence about this. :+1:

PR #2458, opened 2026-01-14
docs: add filenames

Silvanoc:
Code blocks showing the content of files don't show the filename of the files being shown. This patch fixes it in `README.md` and `docs/creating.md`.

This is a screenshot of the current state:
image

And this is a screenshot of the result with my patch:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ To create a template:
 ```
 
 Content of the `copier.yml` file:
+
 ```yaml title="copier.yml"
 # questions
 project_name:
@@ -64,11 +65,13 @@ module_name:
 ```
 
 Content of the `{{project_name}}/{{module_name}}.py.jinja` file:
+
 ```python+jinja title="{{project_name}}/{{module_name}}.py.jinja"
 print("Hello from {{module_name}}!")
 ```
 
 Content of the `{{_copier_conf.answers_file}}.jinja` file:
+
 ```yaml+jinja title="{{_copier_conf.answers_file}}.jinja"
 # Changes here will be overwritten by Copier
 {{ _copier_answers|to_nice_yaml -}}


docs/creating.md
```diff
@@ -35,10 +36,12 @@ module_name:
     help: What is your Python module name?
 ```
```
sisp: This code blocks is currently rendered with a title row at the top:

Image

Same for most other code blocks in this file.

Do you think this isn't sufficient?

docs/creating.md
```diff
@@ -48,17 +51,20 @@ Generating a project from this template with `super_project` and `world` as answ
 the `project_name` and `module_name` questions respectively would create in the
 following directory and files:
 
+Content target directory:
```
sisp: I think the preceding paragraph

> Generating a project from this template with `super_project` and `world` as answers for the `project_name` and `module_name` questions respectively would create in the following directory and files:

clarifies that the code block shows a filesystem tree.

README.md
sisp, line 1: I agree that the code blocks should show filenames for better clarity. The [MkDocs site shows filenames](https://copier.readthedocs.io/en/stable/#quick-start) via the `title` attribute, but GFM doesn't support it. If you see the need for showing filenames on GitHub as well, I'm fine with a preceding paragraph in this file. But a minor request:

```diff
-Content file `copier.yml`:
+Content of the `copier.yml` file:
+
```

(similar for the other code blocks)

WDYT?

sisp (commented):
Thanks for improving our docs, @Silvanoc! :bow: I've left a few inline comments.

docs/creating.md
Silvanoc: That would suffice, but, as you can see in my screenshots, my browser doesn't render that title row 🤔

Silvanoc: Now with [this comment of your](https://github.com/copier-org/copier/pull/2458#discussion_r2689848520) I understand what's going on.

If the rendering is different and you expect people to read the docs, then I'd remove that documentation from the README.md and show a very prominent *The documentation is available in the [copier documentation site](https://copier.readthedocs.io)* or similar.

sisp: Are you viewing https://copier.readthedocs.io/en/stable/creating/?

sisp: Yes, the `docs/` content isn't intended to be viewed on GitHub because it contains non-standard Markdown syntax from Material for MkDocs (and pymdown-extensions).

Silvanoc: > Are you viewing https://copier.readthedocs.io/en/stable/creating/?

Now, yes. But not previously. In general, if I get the documentation directly in the README.md, then I don't visit the readthedocs site. Therefore my comment above. I wouldn't try to use the same markdown for both MkDocs and GFM.

Silvanoc: Now I think I understand you! You are asking to leave "creating" untouched and only README.md would need the changes, right?

README.md
Silvanoc: Done

docs/creating.md
Silvanoc: Excluding `docs/creating.md` from the changes. This document is expected to be viewed over https://copier.readthedocs.io/en/stable/creating/ and not over GitHub.

docs/creating.md
Silvanoc: Excluding `docs/creating.md` from the changes. This document is expected to be viewed over https://copier.readthedocs.io/en/stable/creating/ and not over GitHub.

sisp (approved):
LGTM! :tada: Ready to merge from my side, feel free to unset the draft state when you're ready, too. 🙂

PR #2462, opened 2026-01-19
Fix cast_to_bool to treat whitespace-only strings as False

developreiloyo:
### What does this change?

`cast_to_bool` was treating whitespace-only strings (e.g. `" "`, `"\n"`) as `True` due to Python’s default `bool(str)` behavior.

This PR adds an explicit check to treat empty and whitespace-only strings as `False`.

### What was done?

- Added a guard to return `False` for whitespace-only strings.
- Added a unit test covering this behavior.

### Why this approach?

The change is minimal and does not alter existing behavior for numbers, YAML booleans, or standard truthy values.

### DIFF
--- a/copier/_tools.py
+++ b/copier/_tools.py
@@ -127,19 +127,15 @@ def cast_to_bool(value: Any) -> bool:
     # Assume it's a number
     with suppress(TypeError, ValueError):
         return bool(float(value))
-    
-    # Treat whitespace-only string  as False
-    if isinstance(value,str) and not value.strip():
-        return False
-    
     # Assume it's a string
     with suppress(AttributeError):
-        lower = value.lower()
+        stripped = value.strip()
+        lower = stripped.lower()
         if lower in {"y", "yes", "t", "true", "on"}:
             return True
-        elif lower in {"n", "no", "f", "false", "off", "~", "null", "none"}:
+        elif lower in {"n", "no", "f", "false", "off", "~", "null", "none", ""}:
             return False
-    # Assume nothing
+    # Fallback
     return bool(value)
 
 
--- a/copier/_tools.py
+++ b/copier/_tools.py
@@ -127,6 +127,11 @@ def cast_to_bool(value: Any) -> bool:
     # Assume it's a number
     with suppress(TypeError, ValueError):
         return bool(float(value))
+    
+    # Treat whitespace-only string  as False
+    if isinstance(value,str) and not value.strip():
+        return False
+    
     # Assume it's a string
     with suppress(AttributeError):
         lower = value.lower()
--- a/copier/_tools.py
+++ b/copier/_tools.py
@@ -127,14 +127,19 @@ def cast_to_bool(value: Any) -> bool:
     # Assume it's a number
     with suppress(TypeError, ValueError):
         return bool(float(value))
+
+    # Handle whitespace-only strings first
+    if isinstance(value, str) and not value.strip():
+        return False
+
     # Assume it's a string
     with suppress(AttributeError):
-        stripped = value.strip()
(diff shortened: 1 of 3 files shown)


tests/test_cast_to_bool.py
sisp, line 1: Would you mind moving this test to `tests/test_tools.py`?

copier/_tools.py
```diff
@@ -127,6 +127,11 @@ def cast_to_bool(value: Any) -> bool:
     # Assume it's a number
     with suppress(TypeError, ValueError):
         return bool(float(value))
+    
+    # Treat whitespace-only string  as False
+    if isinstance(value,str) and not value.strip():
```
sisp:
```suggestion
    # Treat whitespace-only string as False
    if isinstance(value, str) and not value.strip():
```

copier/_tools.py
sisp, line 141: Would it make sense to avoid the special case handling for whitespace-only strings and extend the string handling instead?

```diff
 with suppress(AttributeError):
-    lower = value.lower()
+    lower = value.lower().strip()
     if lower in {"y", "yes", "t", "true", "on"}:
         return True
-    elif lower in {"n", "no", "f", "false", "off", "~", "null", "none"}:
+    elif lower in {"n", "no", "f", "false", "off", "~", "null", "none", ""}:
         return False
```

sisp (commented):
Thanks for your contribution around bool casting improvements, @developreiloyo! :bow: I've left a few inline comments.

tests/test_cast_to_bool.py
developreiloyo: Done! I've moved the test to tests/test_tools.py as suggested. Thanks for the feedback!

copier/_tools.py
developreiloyo: Thanks, @sisp! I included "" because without it, whitespace-only strings like "   " fall through to bool(value), which returns True. Since .strip() normalizes them to "", treating it as a falsy keyword feels like the cleanest fix.

sisp: Isn't this the same as

```python
lower = value.lower().strip()
if ...:
    ...
elif lower in {..., ""}:
    return False
```

but without a special case? `" "` would be captured by the `elif` branch.

developreiloyo: Thanks for the suggestion  I realized the redundant check wasn’t needed, so I updated the implementation to use:  
```
if isinstance(value, str) and not value.strip():
    return False
```

sisp: Ah, this is not what I meant. When we're at

https://github.com/copier-org/copier/blob/cd45f37a44e88f4e6da16af221dfb9b9a9f84b8a/copier/_tools.py#L130-L136

then the value is likely a string with methods like `.lower()` and also `.strip()`. So, I was thinking we only need to also strip leading and trailing whitespaces via `.strip()` and add the empty string to the set of values that shall cast to `False`. For this, I believe we don't need the special-case handling

```python
if isinstance(value, str) and not value.strip():
    return False
```

but instead can extend the string handling logic just a little bit.

copier/_tools.py
```diff
@@ -127,14 +127,20 @@ def cast_to_bool(value: Any) -> bool:
     # Assume it's a number
     with suppress(TypeError, ValueError):
         return bool(float(value))
+
+    # Handle whitespace-only strings first
+    if isinstance(value, str) and not value.strip():
+        return False
+
```
sisp, line 134: See https://github.com/copier-org/copier/pull/2462#discussion_r2755120117:

```suggestion
```

sisp (changes requested):

sisp:
Superseded by #2675.

PR #2463, opened 2026-01-20
feat: add `check-update` subcommand

### LINKED ISSUE #1020: Add a 'check' CLI command / API to check for new template version

natescherer:
Hello! This is my attempt to complete the work that was done in #1031. (Should close #1020, as well.)

# What's Changed

Here's what I've changed versus the previous PR, by type:

## Outstanding changes requested in discussions on old PR

- Subcommand updated to `check-update`
- Hard-coded exit code when an update is detected to `5`
	- `2` was discussed in the previous PR but I changed it based on https://github.com/copier-org/copier/issues/1328
- Tests added

## Necessary Updates

- Restructured to match updates made to project in the intervening years

## Other Changes

- Minor rewording
- Added `--check-update-output-as-json` flag for integrating with CI/scripts/etc
	- I'm not in love with this name but seeing as how I had to add it at the Worker level I figured I should make it pretty specific

# Questions/To-Do

I have a few questions for the maintainers:

- Are we sure that having the app exit with an error when there is an update is the right pattern?
	- I preserved it based on the work previously done, but if I was designing this from scratch, I wouldn't consider an update being available an error in the classic sense
	- My addition of `--check-update-output-as-json` provides a hook for scripts to detect if there is an update, from the discussions I saw before that seemed to be the overall intention of the exit code
- What level of documentation would you like for this subcommand?
	- Maybe a `docs/checking_for_updates.md`?

Thanks for making such an awesome tool! Please let me know if there are changes you'd like me to make.


copier/_cli.py
```diff
@@ -426,3 +438,52 @@ def inner() -> None:
...
+        If that file contains also `_commit` and `destination_path` is a git
+        repository, this command will do its best to determine whether a newer
+        version is available, applying PEP 440 to the template's history.
+        """
+    )
+
+    check_update_output_as_json = cli.Flag(
+        ["--check-update-output-as-json"],
+        help="Output a JSON object to stdout containing information about if an update is needed",
+    )
```
sisp: How about renaming this flag to `--output-format` with two allowed values `plain` and `json` similar to, e.g., Ruff's corresponding flag?

```
--output-format <OUTPUT_FORMAT>
    Output serialization format for violations. The default serialization
    format is "full" [env: RUFF_OUTPUT_FORMAT=] [possible values:
    concise, full, json, json-lines, junit, grouped, github, gitlab,
    pylint, rdjson, azure, sarif]
```

copier/_main.py
```diff
@@ -1450,6 +1457,55 @@ def _git_commit(self, message: str = "dumb commit") -> None:
...
+            raise UserMessageError(
+                "Cannot check: version from last update not detected."
+            )
+        if not self.template.version:
+            raise UserMessageError("Cannot check: version from template not detected.")
+
+        update_json = json.dumps(
+            {
+                "update_available": self.template.version
+                > self.subproject.template.version,
```
sisp: This might be a bit more legible with surrounding parentheses:


```suggestion
                "update_available": (
                    self.template.version > self.subproject.template.version
                ),
```

copier/_main.py
```diff
@@ -1450,6 +1457,55 @@ def _git_commit(self, message: str = "dumb commit") -> None:
...
+        )
+
+        if self.template.version > self.subproject.template.version:
+            if not self.quiet:
+                if self.check_update_output_as_json:
+                    # TODO Unify printing tools
+                    print(update_json, file=sys.stdout)
+                raise SubprojectOutdatedError(
+                    "Parent template update available. "
+                    f"\nCurrent version is {self.subproject.template.version}, "
```
sisp: - I think "template" suffices, _parent_ template sounds like the parent of a template in a hierarchy. Or perhaps even "New template version available"? WDYT?
- I'd move the line break to the first line for better legibility.


```suggestion
                    "New template version available.\n"
                    f"Current version is {self.subproject.template.version}, "
```

tests/test_check_update.py
sisp, line 1: Since this is a new test file for an entirely new feature, can we reduce the test cases code to a minimum?

- Avoid Git bundles.
- Avoid custom `_envops` settings.
- Minimize test file content (`README.txt.jinja`).
- Keep commit messages simple, e.g. `git("commit", "-m", "v1")`.

It might make sense to factor out the test template to a fixture, similar to, e.g., this:

https://github.com/copier-org/copier/blob/d3dd5808bc36e327c1a9dc097394b8ece861fddb/tests/test_answersfile_templating.py#L13-L30

sisp (changes requested):
Thanks for contributing this feature to Copier, @natescherer! :bow:

I've left a couple of inline comments.

tests/test_check_update.py
natescherer: Sure, I can simplify it. I just grabbed an existing test more or less at random, I agree it is a little over complex.

copier/_main.py
natescherer: Good point, I agree with the terminology change. I *have* been making a series of hierarchical templates myself and got my wires crossed I think.

copier/_cli.py
natescherer: Works for me, is an updated help message like this acceptable?

`Output format, either 'plain' (the default) or 'json'. This flag is only supported by the 'check-update' subcommand.`

sisp: For the CLI flag, the context is clear, so the second sentence isn't needed. But it makes sense to keep it for the `Worker` class attribute. One day, I'll try to split this class, so we won't have attributes that are only meaningful for some operations. :crossed_fingers:

sisp (commented):
Regarding the exit code: I think exiting with 0 when there's no update (i.e., the project is up-to-date) and with non-zero otherwise seems reasonable. At least [Cruft follows a similar approach](https://cruft.github.io/cruft/#checking-a-project). The specific non-zero exit code is another question. Cruft exits with 1 when there is an update, but 1 is already used for, e.g., precondition errors like a missing old template reference. Exit code 2 is used for indicating unsafe templates without explicit user trust, but this error can only occur for copy and update operations but never when checking for an update. I _think_ it's fine to use the same exit code for different errors by different subcommands.

> [...] but if I was designing this from scratch, I wouldn't consider an update being available an error in the classic sense

What exactly do you mean by "in the classic sense"?

---

Regarding documentation: How about adding a section on our "[Updating a project](https://copier.readthedocs.io/en/stable/updating/)" page, similar to [Cruft's corresponding section](https://cruft.github.io/cruft/#checking-a-project)?

natescherer:
Hi @sisp, I've completed updating the following:

- Your suggested changes
- Adding doc to Updating a Project
- Changing status code when an update is detected from 5 to 1

Let me know if there's anything else you'd like me to change!

copier/_main.py
```diff
@@ -1504,6 +1562,22 @@ def run_update(
     return worker
 
 
+def run_check_update(
```
sisp: I just checked Cruft's implementation of this feature, and there the Python function that performs the update check is called [`is_project_updated`](https://github.com/cruft/cruft/blob/33f6b722fc6fe4b5d26a351e487372e5e4375ab2/cruft/_commands/check.py#L30). In contrast, a function called `run_check_update` (looking especially at the `run_` prefix) seems more related to a (CLI) command and doesn't feel idiomatic as a Python API.

With this new insight, I wonder about the current design of the feature. Would it be sufficient to provide a function like `is_project_updated(...) -> bool` which only returns a boolean value without printing to stdout/stderr? Then, the CLI command could map the return value to an exit code and (unless `--quiet` is passed) print messages like `Project is up-to-date!` and ``Project is outdated! Run `copier update` to update the project``.

Or do you need detailed information about the old and new version for your use case?

I'm just trying to guard Copier's public API because once this is part of our public API it's hard to change.

copier/_main.py
```diff
@@ -1504,6 +1562,22 @@ def run_update(
...
+    dst_path: StrOrPath = ".",
+    data: AnyByStrDict | None = None,
+    **kwargs: Any,
```
sisp: Similar to #2496, could you use explicit parameters please?

copier/_main.py
```diff
@@ -1450,6 +1458,56 @@ def _git_commit(self, message: str = "dumb commit") -> None:
             "--no-verify",
         )
 
+    def run_check_update(self) -> None:
```
sisp: I'm (slowly) trying to refactor the codebase, eventually breaking down the generic `Worker` class into smaller, single-responsibility classes to better separate concerns and create a more maintainable and extensible design. Thinking about this extension of the `Worker` class some more, especially the `output_format` attribute is only useful when checking for updates. We can't change the architecture now, but perhaps it's cleaner to move the body of this method to the function below. This way, we avoid adding more weight to the `Worker` class, easing future refactoring. WDYT?

docs/updating.md
```diff
@@ -259,3 +259,74 @@ git clean -d -i     # remove untracked files and folders
...
+
+#### No Update Found
+
+If an update is found, `copier check-update` will exit with status code 1 and output
+similar to the following:
+
+```
+New template version available.
+Current version is 1.0.0, latest version is 2.0.0.
+```
```
sisp: I find these sections a bit verbose for something rather simple. How about extending the code block above just a little? For example:

````markdown
```console
$ copier check-update
Project is up-to-date!

$ copier check-update
New template version available.
Current version is 1.0.0, latest version is 2.0.0.
````

(Or whatever the final messages will be.)

File: tests/test_check_update.py
Comment (sisp) [inline tests/test_check_update.py:1]: I think the test setup can be simplified a bit. How about just a single template fixture like

```python
@pytest.fixture(scope="module")
def template_path(tmp_path_factory: pytest.TempPathFactory) -> str:
    ...
```

that creates template repository with 3 Git tags `v1`, `v2`, and `v2-alpha`? Then, a test case can generate copy using either `v1` or `v2` and with or without allowing prereleases to test the different scenarios.

sisp (changes requested):

copier/_main.py
natescherer: Sure, I can refactor this, but to supporting printing something like ```Current version is 1.0.0, latest version is 2.0.0.```, both versions need to be emitted as well. I'll refactor to `get_update_data` that emits a bool and two str's for now, unless I hear back from you before I finish. No problem if I need to refactor again after.

docs/updating.md
```diff
@@ -259,3 +259,53 @@ git clean -d -i     # remove untracked files and folders
...
+
+# Prerelease update available
+$ copier check-update --prereleases
+New template version available.
+Current version is 1.0.0, latest version is 2.0.0a0.
+```
+
+### Automated Checking
+
+To facilitate automated checking for updates, `copier check-update` provides json output
```
sisp:
```suggestion
To facilitate automated checking for updates, `copier check-update` provides JSON output
```

🤓

tests/test_check_update.py
```diff
@@ -0,0 +1,358 @@
...
+
+import pytest
+from plumbum import local
+
+from copier._cli import CopierApp
+
+from .helpers import (
+    build_file_tree,
+    git,
+)
```
sisp: Just some nitpicking 🤓:

```suggestion
from .helpers import build_file_tree, git
```

tests/test_check_update.py
```diff
@@ -0,0 +1,358 @@
...
+def template_path(tmp_path_factory: pytest.TempPathFactory) -> str:
+    src = tmp_path_factory.mktemp("src")
+
+    # Create template v1.0.0
+    build_file_tree(
+        {
+            src / "{{ _copier_conf.answers_file }}.jinja": """\
+                # Changes here will be overwritten by Copier
+                {{ _copier_answers|to_nice_yaml }}
+                """,
```
sisp: Just some nitpicking for consistency with (many) other similar occurrences (according to personal preference) 🤓:


```suggestion
            src / "{{ _copier_conf.answers_file }}.jinja": (
                """\
                # Changes here will be overwritten by Copier
                {{ _copier_answers|to_nice_yaml }}
                """
            ),
```

Same for the other similar occurrences.

copier/_cli.py
```diff
@@ -423,3 +436,84 @@ def inner() -> None:
...
+                if self.quiet:
+                    raise SubprojectOutdatedError()
+                else:
+                    if self.output_format == "json":
+                        # TODO Unify printing tools
+                        print(update_json, file=sys.stdout)
+                    raise SubprojectOutdatedError(
+                        "New template version available.\n"
+                        f"Current version is {current_version}, "
+                        f"latest version is {latest_version}."
```
sisp: I'd prefer managing the printed message without an exception because:

- Currently, custom exceptions are defined in the public `copier.errors` module, so a new `SubprojectOutdatedError` exception would become part of the public API, although we don't publicly expose a function for checking for updates.
- If we publicly exposed a function for checking for updates, I don't think we'd need to raise an exception, so we don't need to publicly expose the exception.
- We're already determining the result of this check in the CLI layer, so I think it's simpler and more consistent to simply print the error message and return the exit code instead of raising an exception that is handled elsewhere.

To avoid this exception, I think this `inner` function needs to return an exit code and the `_handle_exceptions` function needs to optionally handle a returned exit code. I imagine this will be unified in the future, but for now we need to deal with it and defer cleanup to later.

copier/_cli.py
```diff
@@ -423,3 +436,84 @@ def inner() -> None:
...
+
+    output_format = cli.SwitchAttr(
+        ["--output-format"],
+        cli.Set("plain", "json"),
+        default="plain",
+        help="Output format, either 'plain' (the default) or 'json'.",
+    )
+
+    def main(self, destination_path: cli.ExistingDirectory = ".") -> int:
+        """Call [get_update_data][copier.main.get_update_data].
```
sisp: This reference will be broken once we remove the deprecated modules and deprecated API compatibility code. I'd keep it simple, just summarize what the function is doing without code references.

sisp (changes requested):
Looks pretty good already, I've left a few more comments. 🫣

copier/_cli.py
natescherer: All things considered, is this a good opportunity to just drop returning different status codes based on update status? I never liked that and only kept it because it was part of the design in #1031. My preference here would just be to always have it exit zero barring some kind of actual error (UserMessageError, KeyboardInterrupt, etc)  being thrown.

If you think status code should be part of the design, however, I will refactor.

sisp: Thanks for being persistent about the status code topic. 😄 I've tried to find some comparable tool examples to align with existing patterns. Here is what I've found:

- `git diff` always exits with "0", but this behavior can be changed:

    ```console
    $ git diff --help
    ...

    --exit-code
        Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences. 

    --quiet
        Disable all output of the program. Implies --exit-code. [...]

    ...
    ```

- `ruff check` exits with non-zero on lint errors, but this behavior can be changed:

    ```console
    $ ruff check --help
    ...

    -e, --exit-zero                        Exit with status code "0", even upon detecting lint violations
        --exit-non-zero-on-fix             Exit with a non-zero status code if any files were modified via fix, even if no lint violations remain

    ...
    ```

Should we make the status code behavior configurable? For example:

- Inspired by `git diff`, `copier check-update` could always exit with "0" (because a message is printed for human or programmatic consumption) while `copier check-update --quiet` could exit with "2" (because "1" already indicates precondition errors like "no old template reference found") when the project is not up-to-date.

- Inspired by `ruff check`, `copier check-update` could exit with "2" when the project is not up-to-date while `copier check-update --exit-zero` could _always_ exit with "0".

WDYT?

natescherer: I think setting it up like `git diff` offers a real best of both worlds approach. If, like me, you don't want a non-zero exit code for anything that isn't a failure to execute, you get that from the default behavior. If you simply want a boolean yes or no on if there are updates, run it with `--quiet`. This also makes the `--quiet` flag useful in a way it wouldn't be if there was no nonverbal way of conveying update status.

I'll proceed to refactor along these lines!

copier/_cli.py
```diff
@@ -423,3 +435,83 @@ def inner() -> None:
...
+                            f"latest version is {latest_version}."
+                        )
+            elif not self.quiet:
+                if self.output_format == "json":
+                    # TODO Unify printing tools
+                    print(update_json, file=sys.stdout)
+                else:
+                    # TODO Unify printing tools
+                    print("Project is up-to-date!", file=sys.stdout)
+            return 0
```
sisp, line 515: - Just some cosmetics to reduce nested `if`/`else` a little.
- `print(..., file=sys.stdout)` → `print(...)` – it's the same, isn't it?
- I've added the `# TODO Unify printing tools` to all `print` statements for consistency.

```suggestion
            if update_available:
                if self.quiet:
                    return 2
                if self.output_format == "json":
                    # TODO Unify printing tools
                    print(update_json)
                else:
                    # TODO Unify printing tools
                    print(
                        "New template version available.\n"
                        f"Current version is {current_version}, "
                        f"latest version is {latest_version}."
                    )
            elif not self.quiet:
                if self.output_format == "json":
                    # TODO Unify printing tools
                    print(update_json)
                else:
                    # TODO Unify printing tools
                    print("Project is up-to-date!")
            return 0
```

WDYT?

copier/_main.py
```diff
@@ -1663,6 +1663,45 @@ def run_update(
     return worker
 
 
+def get_update_data(
+    dst_path: Path | str = ".",
+    use_prereleases: bool = False,
+) -> tuple[bool, str, str]:
```
sisp, line 1668:
```suggestion
def get_update_data(
    dst_path: Path | str = ".", use_prereleases: bool = False
) -> tuple[bool, str, str]:
```

🤓

copier/_main.py
```diff
@@ -1663,6 +1663,45 @@ def run_update(
...
+    Returns a tuple containing a bool for if the project has an update,
+    a string containing a project's current version, and a string containing
+    the latest version of the template available.
+
+    See [checking a project][checking-a-project].
+    """
+    with Worker(
+        dst_path=Path(dst_path),
+        use_prereleases=use_prereleases,
+    ) as worker:
```
sisp:
```suggestion
    with Worker(dst_path=Path(dst_path), use_prereleases=use_prereleases) as worker:
```

🤓

docs/updating.md
```diff
@@ -259,3 +259,62 @@ git clean -d -i     # remove untracked files and folders
...
+
+1. JSON output via the flag `--output-format json`..
+2. Exit code output via the flag `--quiet`
+
+Sample output is provided for different scenarios:
+
+```console
+# No update available
+$ copier check-update --output-format json
+{ "update_available": false, "current_version": "1.0.0", "latest_version": "1.0.0" }
```
sisp: I think the exact output is this:

```suggestion
{"update_available": false, "current_version": "1.0.0", "latest_version": "1.0.0"}
```

🤓

(Same for the others below.)

docs/updating.md
```diff
@@ -259,3 +259,62 @@ git clean -d -i     # remove untracked files and folders
...
+New template version available.
+Current version is 1.0.0, latest version is 2.0.0a0.
+```
+
+### Automated Checking
+
+To facilitate automated checking for updates, `copier check-update` provides two
+options:
+
+1. JSON output via the flag `--output-format json`..
```
sisp:
```suggestion
1. JSON output via the flag `--output-format json`
```

🤓

sisp (changes requested):
Nice! 🤩 Just some minor polishing, then we should be good to go. 🚀

copier/_cli.py
natescherer: All makes sense. The `, file=sys.stdout` was an artifact from when I was previously writing to stderr and stdout and I wanted it there for consistency. Unneeded now.

docs/updating.md
natescherer: Yep, I think this was a copypasta issue that got linted at some point, good catch.

tests/test_check_update.py
```diff
@@ -0,0 +1,547 @@
...
+            src / "copier.yml": (
+                """\
+                version: 1.0.0
+                """
+            ),
+            src / "README.md.jinja": (
+                """\
+                # Version {{ version }}
+                """
+            ),
```
sisp: We can simplify the template here (sorry for not noticing this earlier):


```suggestion
            src / "README.md.jinja": "# Version 1.0.0\n",
```

tests/test_check_update.py
```diff
@@ -0,0 +1,547 @@
...
+            src / "copier.yml": (
+                """\
+                version: 2.0.0
+                """
+            ),
+            src / "README.md.jinja": (
+                """\
+                # Version {{ version }}
+                """
+            ),
```
sisp: With the change above, we can simplify the template v2 here:

- The answers file is identical, we don't need to redefine it.
- Similar to the simplification in template v1.

```suggestion
           src / "README.md.jinja": "# Version 2.0.0\n",
```

tests/test_check_update.py
```diff
@@ -0,0 +1,547 @@
...
+            src / "copier.yml": (
+                """\
+                version: 3.0.0-alpha
+                """
+            ),
+            src / "README.md.jinja": (
+                """\
+                # Version {{ version }}
+                """
+            ),
```
sisp: With the change above, we can simplify the template v3 here:

- The answers file is identical, we don't need to redefine it.
- Similar to the simplification in template v1.

```suggestion
            src / "README.md.jinja": "# Version 3.0.0-alpha\n",
```

tests/test_check_update.py
```diff
@@ -0,0 +1,547 @@
...
+    dst = str(tmp_path_factory.mktemp("dst"))
+
+    CopierApp.run(
+        [
+            "copier",
+            "copy",
+            src,
+            dst,
+            "--quiet",
+            "--defaults",
```
sisp: With the changes above, we don't need `--defaults` anymore because we have no questions anymore.

```suggestion
```

There are few more occurrences like this below.

sisp (changes requested):
Just a few minor simplifications remaining ... 🫣

natescherer:
Those updates all make sense to me, and I've made all of them!

sisp (approved):
Fantastic work, @natescherer! 🥇 Thank you very much for this great contribution (and your patience with my change requests 🫣). :bow:

PR #2487, opened 2026-02-05
feat: add 'copier adopt' command for adopting templates in existing projects

### LINKED ISSUE #2486: Feature: `copier adopt` — first-class workflow for adopting a template in an existing project
### Actual Situation

When adopting a copier template in an **existing project** that wasn't originally created with that template, users are forced into a binary choice with no good middle ground:

**`copier copy --overwrite`** destroys project-specific content:
- `pyproject.toml` dependencies get wiped
  - `src/pkg/__init__.py` exports get replaced
  - Custom CLI code, test fixtures, etc. are overwritten
**`copier copy --skip` / `_skip_if_exists`** preserves project files but skips all template improvements:
- Project doesn't get updated ruff rules, poe tasks, coverage config
  - Project doesn't get updated pre-commit hooks
  - No way to "merge" template config sections with existing config
**`copier update`** is designed for projects already using the template. It uses 3-way merge, but requires a valid `_commit` pointing to a template version the project was created from. For adoption, there's no "old version" to diff against.

The current documented workaround (from #955 / PR #981) is to `copier copy --overwrite` and then use `git difftool` to manually cherry-pick changes back. This is tedious, error-prone, and doesn't scale when adopting a template across many existing repositories.

### Root cause

Copier's merge logic only works in `update` mode with a known baseline. For first-time adoption, there's no baseline to diff against, so it's binary: overwrite or skip. This gap means there is no practical path from "existing project" to "template-managed project" without significant manual effort.

### Related issues

- #955 — Requested this workflow; closed with docs-only fix that the author noted doesn't address the hard part
- #2005 — Requests conflict markers during `copier copy` on existing files (open)
- #1738 — Requests `--merge-files` for multi-template scenarios (open)

### Desired Situation

A first-class `copier adopt` command (or equivalent workflow) that lets an existing project adopt a template with intelligent handling of conflicts:

1. **Add new files** from the template that don't exist in the project (docs, CI configs, etc.) — no conflict
2. **Merge overlapping config files** (`pyproject.toml` tool sections, `.pre-commit-config.yaml`, `.gitignore`, etc.) — using conflict markers or a 3-way merge where the "base" is an empty file or a synthetic baseline
3. **Preserve project-specific content** (dependencies, custom code, package exports) — never silently overwrite
4. **Establish a baseline for future `copier update`** — after adoption, the project should be in the same state as if it had been created from the template, enabling the normal update workflow going forward

The result: a single command that takes a project from "not template-managed" to "template-managed and updatable" with explicit, reviewable conflict resolution rather than silent data loss.

### Proposed solution

A new `copier adopt` subcommand that works roughly like this:

```bash
copier adopt https://github.com/org/template.git ./my-existing-project
```

**MVP behavior:**

1. Render the template into a temporary directory (answering questions as normal)
2. For each rendered file:
   - If the file **doesn't exist** in the project → copy it in
   - If the file **exists and is identical** → skip (no-op)
   - If the file **exists and differs** → write the file with **git-style conflict markers** (treating the existing file as "ours" and the template output as "theirs"), or use a 3-way merge with an empty base
3. Write `.copier-answers.yml` with the current template version and answers
4. Commit (or leave uncommitted for user review)

After this, the project has a valid `.copier-answers.yml` and `_commit`, so future `copier update` calls work normally with proper 3-way merge.

**Key design consideration:** The synthetic baseline. Since there's no "old template version" for a project that was never created from the template, the adopt command could use an empty tree or the template's initial version as the merge base. This is similar to how `git merge --allow-unrelated-histories` handles merging two trees with no common ancestor.

**Alternative (lighter-weight):** Rather than a new subcommand, a `--adopt` flag on `copier copy` that triggers the conflict-marker behavior instead of the overwrite-or-skip binary. This would also write valid `.copier-answers.yml` for future updates.

ichoosetoaccept:
## Summary

This PR adds a new `copier adopt` subcommand that allows users to adopt a template for an existing project that was not originally created with Copier.

Closes #2486

## Motivation

Many users have existing projects they want to bring under Copier management. Currently, the only option is `copier copy --overwrite`, which destroys existing customizations. The `adopt` command provides a safer alternative by merging template files with existing files using conflict markers.

## What it does

Unlike `copier copy`, the `adopt` command:

- **Creates** new files from the template (files that don't exist in the project)
- **Skips** identical files (no changes needed)
- **Merges** differing files with git-style conflict markers (`<<<<<<< existing` / `>>>>>>> template`)
- **Writes** a valid `.copier-answers.yml` for future `copier update` calls
- **Skips tasks** by default since conflicts need resolution first (with a helpful message)
- **Handles binary files** by creating `.rej` files instead of corrupting them with conflict markers

## Usage

```bash
copier adopt <template_src> [destination_path]
```

### Options

- `--conflict {rej,inline}` - Behavior on conflict (default: `inline`)
  - `inline`: Add git-style conflict markers to the file
  - `rej`: Create a `.rej` file with the template version
- `--defaults` - Use default answers to questions

### Example workflow

```bash
# Adopt a template for an existing project
copier adopt gh:org/template ./my-project

# Review and resolve conflicts
git diff
# ... edit files to resolve conflicts ...

# Run template tasks manually
uv sync  # or whatever tasks the template defines

# Commit the adoption
git add -A
git commit -m "chore: adopt org/template"

# Future updates now work with proper 3-way merge!
copier update
```

## Implementation

The implementation:

1. Renders the template into a temporary directory using the existing `run_copy` machinery
2. Iterates through rendered files and compares with the destination
3. Uses `git merge-file` with an empty base for text files (creates conflict markers)
4. Creates `.rej` files for binary files (detected by null bytes in content)
5. Writes `.copier-answers.yml` directly (not from a template source)

## Tests

Added 13 comprehensive tests covering:

- Core functionality (create, skip, merge, answers file generation)
- CLI interface
- Both conflict modes (`inline` and `rej`)
- Binary file handling (uses `.rej` to avoid corruption)
- Exclude patterns
- Non-git destination projects
- Deep nested directory structures
- Templates with `_subdirectory`
- Pretend mode
- Enabling future updates

All existing tests continue to pass (215 total).

## Checklist

- [x] Code follows project style (ruff, editorconfig)
- [x] Tests added and passing
- [x] Pre-commit hooks pass
- [x] Conventional commit message

### DIFF
--- a/copier/__init__.py
+++ b/copier/__init__.py
@@ -21,6 +21,7 @@ except importlib.metadata.PackageNotFoundError:
 
 def __getattr__(name: str) -> Any:
     if not name.startswith("_") and name not in {
+        "run_adopt",
         "run_copy",
         "run_recopy",
         "run_update",
--- a/copier/_cli.py
+++ b/copier/_cli.py
@@ -352,6 +352,68 @@ class CopierRecopySubApp(_Subcommand):
         return _handle_exceptions(inner)
 
 
+@CopierApp.subcommand("adopt")
+class CopierAdoptSubApp(_Subcommand):
+    """The `copier adopt` subcommand.
+
+    Use this subcommand to adopt a template for an existing project that was not
+    originally created with Copier. This enables future `copier update` calls.
+    """
+
+    DESCRIPTION = "Adopt a template for an existing project."
+    DESCRIPTION_MORE = dedent(
+        """\
+        Use this command to bring an existing project under Copier management.
+        Unlike `copy`, this command will merge template files with existing files
+        using conflict markers, rather than overwriting them.
+
+        After adoption, the project will have a valid answers file, enabling
+        future `copier update` calls to work with proper 3-way merge.
+        """
+    )
+
+    conflict = cli.SwitchAttr(
+        ["-o", "--conflict"],
+        cli.Set("rej", "inline"),
+        default="inline",
+        help=(
+            "Behavior on conflict: Create .rej files, or add inline conflict markers."
+        ),
+    )
+    defaults = cli.Flag(
+        ["-l", "--defaults"],
+        help="Use default answers to questions, which might be null if not specified.",
+    )
+
+    def main(
+        self, template_src: str, destination_path: cli.ExistingDirectory = "."
+    ) -> int:
+        """Call [run_adopt][copier.main.Worker.run_adopt].
+
+        Params:
+            template_src:
+                Indicate where to get the template from.
+
+                This can be a git URL or a local path.
+
+            destination_path:
+                The existing project to adopt the template for.
+                If not specified, the currently working directory is used.
+        """
+
+        def inner() -> None:
+            with self._worker(
+                template_src,
+                destination_path,
+                conflict=self.conflict,
+                defaults=self.defaults,
+                overwrite=False,  # Never overwrite in adopt mode
+            ) as worker:
+                worker.run_adopt()
+
+        return _handle_exceptions(inner)
+
+
 @CopierApp.subcommand("update")
 class CopierUpdateSubApp(_Subcommand):
     """The `copier update` subcommand.
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -30,6 +30,7 @@ from typing import (
 )
 from unicodedata import normalize
 
+import yaml
 from jinja2.loaders import FileSystemLoader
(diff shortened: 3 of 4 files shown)


oliverpetersen:
May I ask why you have closed your PR @ichoosetoaccept  / @detailobsessed? I was really looking forward to being able to use `copier adopt`

PR #2490, opened 2026-02-06
fix: remove incorrect import deprecation warnings for `types.Phase` and `types.VcsRef` enums

### LINKED ISSUE #2485: Importing `Phase` enum raise a deprecation warning
### Describe the problem

#1948 introduced the `copier.types.Phase` enum exposed in a `_copier_phase` to fix [an issue in `copier-template-extensions`](https://github.com/copier-org/copier-template-extensions/issues/7). It allowed to scope template extensions processing to phases with exit conditions like:

```python
if context.get("_copier_phase") is not Phase.RENDER:
    return
```

#2069 deprecated direct access to `copier.types` making all templates using it raise a deprecation warning.

### Template

N/A

### To Reproduce

1. Create a template using `copier-template-extensions`
2. Add an exit condition on the phase like

```python
from copier_template_extensions import ContextHook

class ContextUpdater(ContextHook):
    def hook(self, context: dict[str, Any]) -> dict[str, Any] | None:
        if context.get("_copier_phase") is not Phase.RENDER:
            return
```
3.  Apply or update the template

### Logs

```console
DeprecationWarning: Importing `Phase` from `copier.types` is deprecated. This module member is intended for internal use only and will become inaccessible in the future. If you have any questions or concerns, please raise an issue at <https://github.com/copier-org/copier/issues>.
```

### Expected behavior

`copier.types.Phase` is considered part of the public API and does not raise a `DeprecationWarning`.
(Maybe it needs to be moved somewhere else, in a module dedicated to public types) 

_No response_

### Operating system

Linux

Archlinux

### Copier version

9.11.3

### Python version

3.13

### Installation method

uvx+pypi

### Additional context

_No response_

sisp:
I've removed the module-level deprecation warning for the `types` module, thereby restoring it as a public module, and excluded the `Phase` and `VcsRef` enums from raising import deprecation warnings. The `Phase` enum is useful for running a context hook using [copier-template-extensions](https://github.com/copier-org/copier-template-extensions) only in a subset of the rendering phases (see #2485). The `VcsRef` enum is necessary for running Copier programmatically with `vcs_ref=VcsRef.CURRENT`.

Fixes #2485.

### DIFF
--- a/copier/types.py
+++ b/copier/types.py
@@ -5,19 +5,13 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any
 
 from copier import _types
-from copier._deprecation import (
-    deprecate_member_as_internal,
-    deprecate_module_as_internal,
-)
+from copier._deprecation import deprecate_member_as_internal
 
 if TYPE_CHECKING:
     from copier._types import *  # noqa: F403
 
 
-deprecate_module_as_internal(__name__)
-
-
 def __getattr__(name: str) -> Any:
-    if not name.startswith("_"):
+    if not name.startswith("_") and name not in {"Phase", "VcsRef"}:
         deprecate_member_as_internal(name, __name__)
     return getattr(_types, name)


pawamoy (approved):

sisp:
@pawamoy I just realized that `VcsRef` is also being exported in `copier/__init__.py`:

https://github.com/copier-org/copier/blob/0fda385cbbba64b675b4653a9076fe56941d86ed/copier/__init__.py#L11

Should we deprecate this import path now that it's also exported in `copier/types.py`? WDYT? :thinking:

PR #2512, opened 2026-02-20
feat: re-export `Phase` enum at package level

sisp:
I've added a re-export of the `Phase` enum from `copier/_types.py` in `copier/__init__.py` and reverted 47afec32cf9843a6d7dadb04b89e8d929ea99d8b and 0fda385cbbba64b675b4653a9076fe56941d86ed.

After thinking about our module layout more deeply, re-reading [Griffe's recommendation about public APIs](https://mkdocstrings.github.io/griffe/guide/users/recommendations/public-apis/), and going over @pawamoy's and my discussion at https://github.com/copier-org/copier/pull/2490#issuecomment-3859991822 (and subsequent comments) again, I think having several public modules adds unnecessary complexity for this comparatively small project. For example, if we kept `copier`, `copier.errors`, and `copier.types` as public modules, where should `Settings` be exported? These layout decisions seem to be rather arbitrary and opinionated without adding value to the project, I don't think that's good use of our time.

With this PR, I'm proposing to follow the [single top-level re-export location](https://mkdocstrings.github.io/griffe/guide/users/recommendations/public-apis/#single-location-top-level-good) approach recommended by Griffe and preferred by @pawamoy.

### DIFF
--- a/copier/__init__.py
+++ b/copier/__init__.py
@@ -6,12 +6,12 @@ Docs: https://copier.readthedocs.io/
 import importlib.metadata
 from typing import TYPE_CHECKING, Any
 
-from . import _main, _types
-from ._deprecation import deprecate_member, deprecate_member_as_internal
+from . import _main
+from ._deprecation import deprecate_member_as_internal
+from ._types import VcsRef as VcsRef
 
 if TYPE_CHECKING:
     from ._main import *  # noqa: F403
-    from ._types import VcsRef  # noqa: F401
 
 try:
     __version__ = importlib.metadata.version(__name__)
@@ -20,10 +20,6 @@ except importlib.metadata.PackageNotFoundError:
 
 
 def __getattr__(name: str) -> Any:
-    if name == "VcsRef":
-        deprecate_member(name, __name__, f"{__name__}.types.{name}")
-        return getattr(_types, name)
-
     if not name.startswith("_") and name not in {
         "run_copy",
         "run_recopy",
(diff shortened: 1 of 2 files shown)


copier/__init__.py
```diff
@@ -6,12 +6,12 @@
 import importlib.metadata
 from typing import TYPE_CHECKING, Any
 
-from . import _main, _types
-from ._deprecation import deprecate_member, deprecate_member_as_internal
+from . import _main
+from ._deprecation import deprecate_member_as_internal
+from ._types import Phase as Phase, VcsRef as VcsRef
```
pawamoy: Redundant alias to make them public (by convention)? I think it's unnecessary since you already list them in `__all__`.

sisp: Good catch, thank you! :+1: I've removed the aliases.

pawamoy:
Just to be clear: are you planning to expose `Settings` and all errors from top-level only too? Or just `Settings`? Or is it already done?

sisp:
Good question. :smile:

For `Settings`, I'm planning to deprecate imports from `copier.settings` _and_ the current settings API altogether after our [discussion](https://github.com/copier-org/copier/pull/2497#issuecomment-3887255316) about internal-only yet publicly visible `Settings` methods, `pydantic.BaseModel` inheritance contributing to the public API, etc. I have a local branch ready with a revised settings API that – I believe :crossed_fingers: – offers a minimal public contract without leaking internals. I'll try to explain the details in the PR description.

Regarding errors, I haven't made up my mind about it. But I guess it would be consistent to also re-export all errors in the top-level location, too, and deprecate the `copier.errors` module.

pawamoy (approved):

PR #2513, opened 2026-02-20
feat: add new settings API with minimal surface

sisp:
I've deprecated the public `settings` module and its public symbols because it turned out that [the current settings API leaks internal-only methods and public methods from Pydantic's `BaseModel` class](https://github.com/copier-org/copier/pull/2497#issuecomment-3887255316), effectively adding part of Pydantic's public API to ours.

I've given the Pydantic situation deeper thought, this is my current conclusion: In a well-architected Python library, Pydantic should live strictly at the input boundary for parsing and validating untrusted data (e.g., config files) and never define the public domain model. Public APIs should expose plain Python types (e.g., primitive types, dataclasses, `dict`/`TypedDict`, etc.) that represent already-validated domain objects, while Pydantic models and validators remain private implementation details. Validation errors from Pydantic should be caught and mapped to custom exceptions, since exception types are also part of the public API contract. This keeps the public surface minimal and stable, prevents framework types and exceptions from leaking, and allows validation mechanisms to be replaced without breaking the library's API contract.

When thinking about Pydantic's place in a proper architecture, I also wondered about where validation rules should be implemented because Pydantic offers many runtime validation types and field/model validators for additional checks; this is my current conclusion (it might deserve a longer explanation for better clarity): The input boundary layer (Pydantic models) is responsible for validating and normalizing _how data is provided_, i.e., parsing raw input, enforcing structural constraints, handling mutually exclusive or deprecated fields, and converting untrusted representations into typed values. The domain layer is responsible for enforcing _invariants that must always hold for a valid domain object_, regardless of how it was constructed. In other words, the boundary ensures that data is well-formed and acceptable as input, while the domain guarantees that any constructed object is logically valid and cannot exist in an inconsistent state. From my perspective, this reasoning gives dataclasses an edge over typed dicts because domain invariants can be enforced via checks implemented in the `__post_init__` method while typed dicts are only dicts at runtime which cannot enforce invariants at construction time.

With these insights, I've created a new settings API:

- A dataclass `Settings` without any methods.
- A `load_settings` function.

These symbols are re-exported at the top-level location now.

I've tried hard to retain backward compatibility at all levels including raised exceptions. The new settings API, on the other hand, raises only a custom exception `SettingsError` instead of Pydantic or PyYAML exceptions. I've also retained the tests for the deprecated settings API in the renamed `tests/test_settings_deprecated.py` file to ensure continued functionality, and added modified/extended tests for the new settings API.

WDYT, @pawamoy?

### DIFF
--- a/copier/_settings.py
+++ b/copier/_settings.py
@@ -1,0 +1,78 @@
+"""User settings models and helper functions."""
+
+from __future__ import annotations
+
+import os
+import warnings
+from os.path import expanduser
+from pathlib import Path
+from typing import Any
+
+import yaml
+from platformdirs import user_config_path
+from pydantic import BaseModel, Field
+
+from ._tools import OS
+from .errors import MissingSettingsWarning
+
+ENV_VAR = "COPIER_SETTINGS_PATH"
+
+
+class Settings(BaseModel):
+    """User settings model."""
+
+    defaults: dict[str, Any] = Field(
+        default_factory=dict, description="Default values for questions"
+    )
+    trust: set[str] = Field(
+        default_factory=set, description="List of trusted repositories or prefixes"
+    )
+
+    @staticmethod
+    def _default_settings_path() -> Path:
+        return user_config_path("copier", appauthor=False) / "settings.yml"
+
+    @classmethod
+    def from_file(cls, settings_path: Path | None = None) -> Settings:
+        """Load settings from a file."""
+        env_path = os.getenv(ENV_VAR)
+        if settings_path is None:
+            if env_path:
+                settings_path = Path(env_path)
+            else:
+                settings_path = cls._default_settings_path()
+
+                # NOTE: Remove after a sufficiently long deprecation period.
+                if OS == "windows":
+                    old_settings_path = user_config_path("copier") / "settings.yml"
+                    if old_settings_path.is_file():
+                        warnings.warn(
+                            f"Settings path {old_settings_path} is deprecated. "
+                            f"Please migrate to {settings_path}.",
+                            DeprecationWarning,
+                            stacklevel=2,
+                        )
+                        settings_path = old_settings_path
+        if settings_path.is_file():
+            data = yaml.safe_load(settings_path.read_bytes())
+            return cls.model_validate(data)
+        elif env_path:
+            warnings.warn(
+                f"Settings file not found at {env_path}", MissingSettingsWarning
+            )
+        return cls()
+
+    def is_trusted(self, repository: str) -> bool:
+        """Check if a repository is trusted."""
+        return any(
+            repository.startswith(self.normalize(trusted))
+            if trusted.endswith("/")
+            else repository == self.normalize(trusted)
+            for trusted in self.trust
+        )
+
+    def normalize(self, url: str) -> str:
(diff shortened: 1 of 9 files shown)


copier/_settings.py
```diff
@@ -0,0 +1,153 @@
...
+from typing import Any
+
+import yaml
+from platformdirs import user_config_path
+from pydantic import BaseModel, Field, ValidationError
+
+from ._tools import OS
+from .errors import MissingSettingsWarning, SettingsError
+
+__all__ = ["is_trusted_repository", "load_settings", "Settings"]
```
pawamoy: It's not necessary to declare `__all__` in private modules. Also, `is_trusted_repository` doesn't seem to be exported at the top-level: should it?

pawamoy (approved):

copier/_settings.py
sisp: > It's not necessary to declare `__all__` in private modules.

Ah, good point. :+1: I've removed it.

> Also, `is_trusted_repository` doesn't seem to be exported at the top-level: should it?

Ah, yes. Let me explain and then let's see what we should do about it. :smile: The current implementation basically doesn't touch the `Worker` class, leaving the `settings: SettingsModel` attribute (except for the renamed class) untouched. This is less invasive than also extending it to `settings: Settings | SettingsModel` like I've done for `run_{copy,recopy,update}`. Now, as long as `SettingsModel` is used in `Worker`, `is_trusted_repository` isn't actually used internally, and I see no use for it by external users (please feel free to disagree), that's why it isn't exported at the top-level. Once we remove the deprecated API, `Worker` will have `settings: Settings` and we'll use `is_trusted_repository(settings.trust, ...)`.

That said, do you think this is an acceptable situation now? And should we accept `settings: Settings | SettingsModel` in the `Worker` class? :thinking:

sisp: ~~Ah, wait, regarding `is_trusted_repository`: We could use it even with `settings: SettingsModel`. Then, `SettingsModel.is_trusted` would remain only for backward compatibility.~~

Okay, not correct because a `set[str]` value can't be passed to a `frozenset[str]` function argument. :shrug:

Perhaps we can loosen the type to `Iterable[str]`?

sisp: I've loosened types to `Iterable[str]` (in `is_trusted_repository`) and `Sequence[str]` (in `Settings`). I think there's no need to require a frozen set. Now, `is_trusted_repository` is used internally.

pawamoy: > Now, as long as SettingsModel is used in Worker, is_trusted_repository isn't actually used internally, and I see no use for it by external users (please feel free to disagree), that's why it isn't exported at the top-level. Once we remove the deprecated API, Worker will have settings: Settings and we'll use is_trusted_repository(settings.trust, ...)

OK the utility is just there for a future refactor. No problem with that :slightly_smiling_face: 

> And should we accept settings: Settings | SettingsModel in the Worker class? 🤔

The worker class is private right? If you feel like it would make your life easier, go for it :shrug: If it's the contrary, then don't :smile: 

> types

Loosening input types is always a good idea I think :+1: (contrary to output types)

sisp: Thanks for your feedback! :bow:

PR #2541, opened 2026-03-10
fix(updating): run tasks in Git repository with subproject's Git config

### LINKED ISSUE #2478: Tasks with copier update fail if they need a Git repo
### Describe the problem

I have a task that relies on the project tree being a Git repository. i.e. I'm using `gh` (Github CLI) to perform some automation after a template has been rendered.

Everything works fine when using `copier copy`, but when doing a `copier update` things fail. Doing a cursory check it looks like copier does not create a Git repo under `/tmp/copier._main.old_copy.0si_f3td`, which causes tools that expect the source tree to be such to fail.

### Template

`copier.yaml`:

```yaml
_tasks:
  - |-
        gh api --method PATCH repos/{owner}/{repo}/properties/values -f 'properties[][property_name]=something' -f 'properties[][value]=true'
```

### To Reproduce

1. Run `copier copy`
2. Then run `copier update`

### Logs

```console

```

### Expected behavior

Tasks should run the same, whether for `copy` or `update` 

_No response_

### Operating system

Linux

Archlinux

### Copier version

9.11.0

### Python version

CPython 3.13

### Installation method

distro package

### Additional context

_No response_

sisp:
I've fixed the update algorithm to Git-initialize the temporary locations wherein the fresh copies of old and new template versions are generated such that tasks are run inside Git repositories. Also, these temporary Git repositories now reuse the subproject's Git config, so settings like Git remotes are accessible by tasks and custom Jinja extensions.

I think it's a valid assumption that temporary locations are Git repositories because the update algorithm requires Git to be installed and the subproject to be a Git repository, too. Hence, it may come as a surprise to template authors when tasks or Jinja extensions are run in the temporary locations, managed by the update algorithm, that aren't Git repositories. After all, having temporary locations is an implementation detail of the update algorithm, I think we should avoid leaking internals to template authors as much as possible.

Just to be sure I'm not missing anything, would the @copier-org/maintainers mind double-checking this idea?

Closes #2478.

### DIFF
--- a/tests/test_vcs.py
+++ b/tests/test_vcs.py
@@ -183,7 +183,11 @@ def test_update_using_local_source_path_with_tilde(tmp_path: Path) -> None:
 
     # generate project and assert correct path in answers
     worker = run_copy(
-        src_path=user_src_path, dst_path=tmp_path, defaults=True, unsafe=True
+        src_path=user_src_path,
+        dst_path=tmp_path,
+        defaults=True,
+        unsafe=True,
+        skip_tasks=True,
     )
     assert worker.answers.combined["_src_path"] == user_src_path
 
@@ -198,6 +202,7 @@ def test_update_using_local_source_path_with_tilde(tmp_path: Path) -> None:
         overwrite=True,
         answers_file=".copier-answers.autopretty.yml",
         unsafe=True,
+        skip_tasks=True,
     )
     assert worker.answers.combined["_src_path"] == user_src_path
 
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1363,6 +1363,15 @@ class Worker:
             ).strip()
         )
         subproject_subdir = self.subproject.local_abspath.relative_to(subproject_top)
+        subproject_git_config_path = Path(
+            git(
+                "-C",
+                self.subproject.local_abspath,
+                "rev-parse",
+                "--absolute-git-dir",
+            ).strip(),
+            "config",
+        )
 
         with (
             TemporaryDirectory(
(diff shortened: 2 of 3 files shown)


tests/test_vcs.py
```diff
@@ -177,7 +177,11 @@ def test_update_using_local_source_path_with_tilde(tmp_path: Path) -> None:
 
     # generate project and assert correct path in answers
     worker = run_copy(
-        src_path=user_src_path, dst_path=tmp_path, defaults=True, unsafe=True
+        src_path=user_src_path,
+        dst_path=tmp_path,
+        defaults=True,
+        unsafe=True,
+        skip_tasks=True,
```
sisp, line 190: I decided to disable tasks here (and also below) because the generated project of this archived template uses a [pre-commit hook from https://gitlab.com/pycqa/flake8](https://github.com/copier-org/autopretty/blob/9e825f4c361e8af11454f4b2a3c38568ae9f36d5/src/.pre-commit-config.yaml.jinja#L159) which doesn't exist anymore. When trying to clone a non-existing Git repository on gitlab.com, Git prompts for credentials which blocks the test case. The [pre-commit hooks are installed via a task](https://github.com/copier-org/autopretty/blob/9e825f4c361e8af11454f4b2a3c38568ae9f36d5/copier.yml#L22) and thus executed on `git commit` run internally by the update algorithm. Before the changes of this PR, there was no problem because the temporary locations managed by the update algorithm were no Git repositories at the time of project generation, so the pre-commit hooks were not installed by the task.

Gara-Dorta:
I don't mean to create noise here but is this PR abandoned? I'm also hit by the same issue as #2478 in [python-package-template#39](https://gitlab.ewi.tudelft.nl/reit/python-package-template/-/work_items/39)

sisp:
Thanks for the reminder, @Gara-Dorta! :+1: No, it isn't abandoned, just never got merged because I was hoping for a sanity check. But I believe this should be a safe change after re-thinking it again. I just resolved a merge conflict, will merge once CI passes.

Issue #2421, opened 2025-12-09
Enable conditionally pre-selected multiselect choices that cannot be unchecked


SoheilSalmani:
### Actual Situation

I would like to create a Copier template for a React application in which the user can select any number of ready-made components from a single `multiselect` question. Before displaying that selection, I also want to ask whether they prefer to use a preset (e.g. `ecommerce`, `trading`, `blog`). If a preset is chosen, some components should be automatically selected and the template should generate the pages and configuration relevant to that preset.

Example components:

- `Cart` (required for `ecommerce` preset)
- `BarChart` (required for `trading` preset)
- `HeroSection`, (required for `blog` preset)
- `Carousel`
- etc.

Each preset relies on some of these components to render correctly.  If the user removes those components, the generated preset pages would break (missing imports, routes, or sections).

What I’d like:

- A single `multiselect` question with all available components.
- Some components can be toggled freely.
- Some components are **required** and must stay selected when a given preset is chosen, because the preset pages depend on them.
 - Required components cannot be unchecked in the interactive prompt.

Currently, as far as I understand:

- The `validator` for choices is used to **enable or disable** a choice.
- When a choice is disabled, it is also **unchecked**.

That means I can’t use the validator to express “this choice is required and must remain selected”:
  - If I disable it, it gets unchecked.
  - If I leave it enabled, the user can still uncheck it.

### Desired Situation

The ability to conditionally pre-select choices in a multiselect and prevent the user from unchecking them.

### Proposed solution

I think the simplest way to implement this, without introducing any new properties in the configuration, would be to conditionally set the `default` property based on the selected preset using Jinja, and then disable the choice using the `validator` property. For this to work, the validator should only disable the choice and a **disabled choice should not be automatically unchecked.**

Minimal example with a `ecommerce` preset with a required `Cart` component:

```yaml
preset:
  type: str
  choices: [ecommerce, none]
  default: none

components:
  type: yaml
  multiselect: true
  choices:
    Cart:
      value: cart
      validator: "{% if preset == 'ecommerce' %}Required by the preset{% endif %}"
  default: |
    {%- if preset == 'ecommerce' %}
    - cart
    {%- endif %}
```

However, at the moment, setting a choice as `default` and then disabling it with a `validator` results in an error.

PR #2556, opened 2026-03-17
fix(user_data): support default=false in multi-select choice syntax

BillionClaw:
Allow setting default=false in the extended choice syntax for multi-select questions. Previously, the default property in choice definitions was ignored.

With this change, users can specify:

choices:
  - value: feature_a
    default: false
  - value: feature_b
    default: true

The choice-level default is respected when no question-level default is set. When a question-level default list is provided, it overrides the choice-level defaults for backward compatibility.

Related to #2421

### DIFF
--- a/copier/_user_data.py
+++ b/copier/_user_data.py
@@ -352,6 +352,7 @@ class Question:
             # Extract the extended syntax for dict-like (dict-style or
             # tuple-style) choices if applicable
             disabled = ""
+            checked = None  # Use None to indicate "not explicitly set"
             if isinstance(choice, (tuple, list)) and isinstance(value, dict):
                 if "value" not in value:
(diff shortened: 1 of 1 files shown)


bobo-xxx:
Just checking in — is this ready for review?

onthebed:
I'll get the CLA signed — will follow up once it's done.

onthebed:
Not quite yet — I still need to address the ambiguity between question-level and choice-level defaults. I’m going to update the PR to handle that cleanly before marking it ready for review.

onthebed:
Thanks for the feedback. I don't see a narrower change that would resolve the ambiguity cleanly without broadening the default/validator model, so I'm going to leave this PR open rather than try to force a redesign into it.

PR #2561, opened 2026-03-19
chore: use `rumdl` as Markdown formatter and linter

sisp:
I've introduced [rumdl](https://rumdl.dev/) as our new Markdown formatter and linter and fixed raised errors accordingly. At the same time, I've configured Prettier to ignore Markdown files.

rumdl brings built-in Python/MkDocs-flavored Markdown support which is a major improvement over Prettier. Also, we've been stuck with an old Prettier version, as more recent versions break formatting compliance with the Python/MkDocs flavor. After delegating Markdown formatting to rumdl, we should finally be able to update to a recent Prettier version.

### DIFF
--- a/mkdocs_hooks.py
+++ b/mkdocs_hooks.py
@@ -1,0 +1,22 @@
+"""MkDocs plugin hooks."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from mkdocs.config.defaults import MkDocsConfig
+    from mkdocs.structure.files import Files
+    from mkdocs.structure.pages import Page
+
+
+def on_page_markdown(
+    markdown: str,
+    page: Page,  # noqa: ARG001
+    config: MkDocsConfig,  # noqa: ARG001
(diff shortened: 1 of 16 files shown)


pyproject.toml
```diff
@@ -162,6 +163,10 @@ warn_return_any = false
 module = ["tests.test_cli", "tests.test_prompt"]
 disable_error_code = ["no-untyped-def"]
 
+[[tool.mypy.overrides]]
+module = "mkdocs_hooks"
+disable_error_code = ["import-not-found"]
+
```
sisp, line 169: I've added this mypy override to ignore `import-not-found` errors in the hook module to avoid installing the `docs` dependency group type-checking.

mkdocs_hooks.py
sisp, line 1: This is a little hack to avoid a broken link from `CONTRIBUTING.md` to `CHANGELOG.md` in the MkDocs site because we lowercase the filenames there for prettier URL paths, and MkDocs' link validator is case-sensitive.

sisp (commented):
Just a few remarks from myself about potentially non-obvious changes.

PR #2586, opened 2026-04-07
fix(vcs): support git worktrees in `is_git_repo_root`

rcoup:
`is_git_repo_root()` assumed `.git` is always a directory, but in [git worktrees](https://git-scm.com/docs/git-worktree) [`.git` is a file](https://git-scm.com/docs/gitrepository-layout). This caused `get_repo()` to return `None` for worktree paths, omitting `_commit` from `.copier-answers.yml` and breaking subsequent `copier update`.

Replace the `local.cwd(.git)` approach with `git -C <path> rev-parse --show-toplevel`, which works for both regular repos and worktrees.

### DIFF
--- a/copier/_vcs.py
+++ b/copier/_vcs.py
@@ -55,7 +55,9 @@ def is_git_repo_root(path: StrOrPath) -> bool:
     """Indicate if a given path is a git repo root directory."""
     try:
         return (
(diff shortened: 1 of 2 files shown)


sisp (approved):
LGTM! 🎉 Thanks for this contribution, @rcoup! 🙇

Issue #2482, opened 2026-02-02
Conditional exclude fails on update for new question


mschoettle:
### Describe the problem

I added a new question to our template and added a conditional (templated) `_exclude` to exclude a file if the question is answered with no.

When calling `copier update` I get the following error:

```console
Traceback (most recent call last):
  File "/home/<user>/.cache/uv/archive-v0/LPweG3ZFVAPurgjpe8gG_/lib/python3.13/site-packages/copier/_main.py", line 1088, in run_copy
    self._render_template()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/<user>/.cache/uv/archive-v0/LPweG3ZFVAPurgjpe8gG_/lib/python3.13/site-packages/copier/_main.py", line 757, in _render_template
    if self.match_exclude(dst_relpath):
       ^^^^^^^^^^^^^^^^^^
  File "/home/<user>/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/functools.py", line 1042, in __get__
    val = self.func(instance)
  File "/home/<user>/.cache/uv/archive-v0/LPweG3ZFVAPurgjpe8gG_/lib/python3.13/site-packages/copier/_main.py", line 710, in match_exclude
    return self._path_matcher(
           ~~~~~~~~~~~~~~~~~~^
        self._render_string(exclusion, extra_context=extra_context)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        for exclusion in self.all_exclusions
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/<user>/.cache/uv/archive-v0/LPweG3ZFVAPurgjpe8gG_/lib/python3.13/site-packages/copier/_main.py", line 447, in _path_matcher
    spec = PathSpec.from_lines(_pathspec_pattern, normalized_patterns)
  File "/home/<user>/.cache/uv/archive-v0/LPweG3ZFVAPurgjpe8gG_/lib/python3.13/site-packages/pathspec/pathspec.py", line 260, in from_lines
    patterns = [pattern_factory(line) for line in lines if line]
                                                  ^^^^^
  File "/home/<user>/.cache/uv/archive-v0/LPweG3ZFVAPurgjpe8gG_/lib/python3.13/site-packages/copier/_main.py", line 446, in <genexpr>
[...]

PR #2587, opened 2026-04-08
feat: add support for enabling StrictUndefined in Jinja environment

mschoettle:
This makes it possible to fail fast when there are undefined variables in templates. Also, it avoids having to define a custom extension to enable StrictUndefined.

Also adds a test to check the module not found error to increase coverage.

Closes #1522

PR #2590, opened 2026-04-08
feat!: emit warning that StrictUndefined will be enabled by default in the future

mschoettle:
BREAKING CHANGES: Existing templates will see a FutureWarning emitted.

Follow-up to #2587 as a second phase towards enabling StrictUndefined by default.


mschoettle:
@sisp Before I put more effort into this. Is there any interest in having this change (even if it is not merged right away)?

sisp:
@mschoettle I think eventually we'll want to use `StrictUndefined` by default because it's the safer behavior, but #2482 is holding me back from even announcing the future change at this point. I'd leave it up you whether you'd like to update this PR and keep it open until we've resolved #2482 or close it in favor of an issue that tracks it with a reference to resolving #2482 as a precondition.

mschoettle:
@sisp I brought it up to date and will leave it open. Is there anything I can help to advance #2482?

sisp:
Thanks for your efforts and offer to help advance #2482, @mschoettle! :pray: The way I see it, #2482 is best addressed after a major refactoring that decouples prompting from generation better. I've been working towards it with as much time as I can dedicate, hopefully it will be ready soon-ish. :crossed_fingers: I anticipate that implementing this fix and several other enhancements will become much easier then. I realize this is a somewhat unsatisfying, somewhat vague reply, but I'm convinced the wait will be worth it. :muscle: :slightly_smiling_face:

PR #2591, opened 2026-04-09
fix(vcs): fix Git tag lookup for relative local template paths

sisp:
Copier uses `git ls-remote` to look up available tags in a template repository. With relative paths containing parent traversal segments (e.g. `../../my-template`), `git ls-remote` can behave unexpectedly, working correctly in some cases but not others. Resolve local file paths to their absolute form before passing them to `git ls-remote` to ensure consistent behavior regardless of the working directory.

Fixes #2589.

### DIFF
--- a/copier/_vcs.py
+++ b/copier/_vcs.py
@@ -7,6 +7,7 @@ import sys
 from contextlib import suppress
 from pathlib import Path
 from tempfile import TemporaryDirectory, mkdtemp
+from urllib.parse import urlparse
 from warnings import warn
 
 from packaging import version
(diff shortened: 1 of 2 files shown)


copier/_vcs.py
```diff
@@ -143,6 +144,15 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
     """
+    # For local Git repos, `git ls-remote` requires an absolute path to work correctly,
+    # it behaves unexpectedly with some relative paths, especially with parent path
+    # traversal.
+    #
+    # See:
+    # - https://github.com/copier-org/copier/issues/2589
+    # - https://stackoverflow.com/q/59981939
+    if urlparse(url).scheme in {"", "file"}:
+        url = Path(url).resolve().as_posix()
```
sisp, line 155: For future reference: This is hacky and shouldn't stay forever. I think a better solution would be to distinguish between remote URLs and paths via different types, possibly using [yarl](https://github.com/aio-libs/yarl)'s `URL` class and `pathlib.Path`. The `get_repo` function could return `URL | Path` instead of `str`, so we don't need to infer it here. But this means changing the signature of `get_repo` which I'll avoid until the public API surface has been cleaned up.

PR #2605, opened 2026-04-10
fix: propagate executable-bit changes on `copier update`

### LINKED ISSUE #2604: copier update does not propagate executable-bit changes from template to destination
## Summary

`copier update` does not reliably propagate executable-bit (file mode) changes from a template to its destinations. Two distinct failure modes:

1. **Mode-only template change** (mode bit gained or lost, no content change): `copier update` is a complete no-op for the file. Neither the working tree nor the git index ever sees the new mode. The only sign that anything happened is `_commit:` being bumped in `.copier-answers.yml`.
2. **Mode + content template change on Windows / `core.fileMode=false` filesystems**: copier writes the new content and `chmod`s the file on disk, but `chmod` is meaningless on Windows for the executable bit, and the destination's git index is never updated. The mode change is silently lost on push.

On POSIX with `core.fileMode=true` (the default on Linux/macOS), the second case is masked by the user's normal post-update `git add -A` (which records the on-disk mode into the index). But the first case fails on every platform because copier doesn't even reach the `chmod` call: `_render_allowed` short-circuits as "identical" before `_render_file` runs.

`copier copy` of a fresh destination is unaffected (every file is brand new, so `_render_allowed` always returns `True`, the on-disk `chmod` always fires, and the user's first `git add -A` records the mode on POSIX).

This is the same shape as #361, half-fixed in 2021 by PR #404. That fix added the on-disk `chmod` after `write_bytes`, which makes `copier copy` work and partially helps `copier update` — but it never extended `_render_allowed` to consider mode bits, and it never updates the destination's git index, so the no-op case and the Windows case are still broken today.

## Versions tested

| Version | Mode-only update | Mode promotion (gain `+x`) on Windows |
| --- | --- | --- |
| `copier==9.14.2` (latest) | ❌ silent no-op | ❌ bit lost on push |
| `copier==9.12.0` | ❌ silent no-op | ❌ bit lost on push |

POSIX cases (Linux, macOS, `core.fileMode=true`) verified directly with the reproducers below. Windows / `core.fileMode=false` behaviour traced through copier's source rather than reproduced on a Windows host — happy to also reproduce on Windows if maintainers want that.

Reproduction host: macOS 26.4 (Darwin), Python 3.11.14, git 2.50.1 (Apple Git-155), APFS.

Both directions reproduce: gaining `+x` and losing `+x`.

## Reproducer 1 — mode-only update (silent no-op)

Self-contained, copy-paste runnable:

```bash
set -e
rm -rf /tmp/copier-bug && mkdir /tmp/copier-bug && cd /tmp/copier-bug

# 1. Template at v1.0.0 — launcher.sh is NOT executable
mkdir tpl && cd tpl
git init -q -b main
echo '_min_copier_version: "9.10.0"' > copier.yml
printf '#!/bin/sh\necho hi\n' > launcher.sh
cat > '{{ _copier_conf.answers_file }}.jinja' <<'EOF'
{{ _copier_answers|to_nice_yaml -}}
EOF
git add -A
git -c user.email=t@t -c user.name=t commit -qm init
git tag v1.0.0
cd ..

# 2. copier copy at v1.0.0 → dst is also 644
uvx copier@latest copy /tmp/copier-bug/tpl ./dst --defaults --trust >/dev/null
( cd dst && git init -q -b main && git add -A \
    && git -c user.email=t@t -c user.name=t commit -qm init )
echo "== dst at v1.0.0 =="
ls -la dst/launcher.sh
( cd dst && git ls-files -s launcher.sh )

# 3. Template v1.1.0 — gains +x but content unchanged
cd tpl
git update-index --chmod=+x launcher.sh
chmod +x launcher.sh
git -c user.email=t@t -c user.name=t commit -qm "make executable"
git tag v1.1.0
cd ..

# 4. copier update
cd dst
uvx copier@latest update --defaults --trust
echo "== dst after update =="
ls -la launcher.sh           # observed: -rw-r--r--   (still 644, unchanged)
git ls-files -s launcher.sh  # observed: 100644 ...   (still 644, unchanged)
git status --short
```

Expected: `copier update` should propagate the new executable bit so a subsequent `git add -A && git commit` records `100755` for the file. Observed: copier doesn't touch the file or its index entry at all. The mode change is silently dropped on every platform.

## Reproducer 2 — mode + content update on a `core.fileMode=false` filesystem (Windows behaviour)

Same template setup, but the v1.1.0 commit changes both mode and content, and the destination has `core.fileMode=false` (the Windows default):

```bash
set -e
rm -rf /tmp/copier-bug-fm && mkdir /tmp/copier-bug-fm && cd /tmp/copier-bug-fm

# Template — same as reproducer 1 setup
mkdir tpl && cd tpl
git init -q -b main
echo '_min_copier_version: "9.10.0"' > copier.yml
printf '#!/bin/sh\necho hi\n' > launcher.sh
cat > '{{ _copier_conf.answers_file }}.jinja' <<'EOF'
{{ _copier_answers|to_nice_yaml -}}
EOF
git add -A
git -c user.email=t@t -c user.name=t commit -qm init
git tag v1.0.0
cd ..

uvx copier@latest copy /tmp/copier-bug-fm/tpl ./dst --defaults --trust >/dev/null
cd dst
git init -q -b main
git config core.fileMode false   # simulate Windows behaviour
git add -A
git -c user.email=t@t -c user.name=t commit -qm init
echo "== dst at v1.0.0, core.fileMode=false =="
git ls-files -s launcher.sh   # 100644
cd ..

cd tpl
git update-index --chmod=+x launcher.sh
chmod +x launcher.sh
echo "# updated" >> launcher.sh
git add -A
git -c user.email=t@t -c user.name=t commit -qm "make executable and update"
git tag v1.1.0
cd ..

cd dst
uvx copier@latest update --defaults --trust >/dev/null
git add -A
echo "== dst after update + git add (the standard post-update flow) =="
git ls-files -s launcher.sh   # observed: 100644 ...   (mode change LOST)
```

Expected: `git ls-files -s launcher.sh` should show `100755` after `git add -A`, matching the template at v1.1.0. Observed: index stays at `100644`. With `core.fileMode=false`, `git add` does not look at on-disk mode bits, and copier never tells the index about the mode change directly. On Windows the same scenario produces the same result for the same reasons.

## Where in the source

In `copier/_main.py` at v9.14.2:

- [`_render_allowed`, lines 507–563](https://github.com/copier-org/copier/blob/v9.14.2/copier/_main.py#L507-L563) compares `previous_content == expected_contents`. File mode is not part of the equality check, so when only the mode changes the function returns `False` ("identical") and `_render_file` short-circuits without ever stat'ing or chmod'ing the destination. **This is the no-op bug (Reproducer 1).**
- [`_render_file`, lines 786–848](https://github.com/copier-org/copier/blob/v9.14.2/copier/_main.py#L786-L848) calls `_render_allowed` at line 829 and returns early on `False`. When it does proceed, lines 837–840 write the new content and `chmod` the destination on disk to match `src_mode`. There is no corresponding `git update-index --chmod=…` call, so when `core.fileMode=false` (Windows default) the index keeps the old mode. **This is the Windows bug (Reproducer 2).**
- The merge-conflict path at [lines 1459–1489](https://github.com/copier-org/copier/blob/v9.14.2/copier/_main.py#L1459-L1489) does call `git update-index --index-info`, but it reuses the *current* `perms` parsed from `git ls-files --stage`, so even in the conflict path the mode bit is never updated. (Edge case; not the primary bug, but worth noting as the same root cause: copier never tells git about new mode bits.)

Historical context: PR #404 (commit `4d51d6b`, 2021) added the on-disk `chmod` to fix #361. That fix is what makes `copier copy` work — but it was never extended to (a) detect mode-only changes, or (b) propagate the mode to the index. So #361 was only half-fixed.

## Impact

Templates that ship anything mode-sensitive break silently on `copier update`:

- Shell launchers and `.command` files (macOS double-click launchers)
- CI/CD scripts that get invoked directly
- Git hooks installed into `.git/hooks` (or `.githooks/` and similar)
- Test fixtures that exercise execute permissions
- Anything using `chmod a-w` or directory mode bits

The no-op variant breaks every platform; the index variant breaks Windows users specifically. Both are silent: the file appears to be present and `copier update` reports success.

## Workaround

We hit this in a project template that ships `.command` launchers. After every `copier copy` / `copier update`, we maintain a list file `.config/executable-files.txt` and re-apply the bit ourselves with both `chmod +x` (on-disk) and `git update-index --chmod=+x` (cross-platform, regardless of `core.fileMode`):

```python
def sync_executable_bits(scope: Path, paths: list[str]) -> None:
    for rel in paths:
        target = scope / rel
        if target.is_file():
            target.chmod(target.stat().st_mode | 0o111)
    subprocess.run(
        ["git", "-C", str(scope), "update-index", "--chmod=+x", *paths],
        check=True,
    )
```

It's idempotent and works around both failure modes on every platform. Mentioning here so other users hitting this can find it while waiting for an upstream fix.

## Suggested fix (PR follow-up)

Two changes in `copier/_main.py`:

1. **Extend `_render_allowed` to consider mode bits.** Thread an `expected_mode` parameter from `_render_file` into `_render_allowed`, and treat the file as "needs rendering" when the destination's current mode differs from the template's, even if content is identical. This fixes Reproducer 1 directly: a mode-only template change will now go through the full write/chmod path.
2. **Have `_render_file` propagate the mode to the git index when the destination is git-tracked.** After the existing on-disk `chmod` (lines 838–848), if the destination subproject is a git repo (`self.subproject.local_abspath` has a `.git` dir or is inside one), shell out to `git update-index --add --cacheinfo <mode>,<sha>,<path>` (or `git update-index --chmod=+x|-x <path>` if the file is already tracked). This fixes Reproducer 2 and the Windows case directly. It does add a new git interaction from the write path, which copier doesn't currently have outside of the merge-conflict block; happy to discuss whether maintainers prefer that or a different approach (e.g., a warning + a `--fix-mode-bits` flag).

The merge-conflict path mode propagation (lines 1459–1489) is the same root cause but a rare edge case; can be a follow-up issue.

I'll open a PR with both changes plus regression tests covering each reproducer above. Will reference this issue.

## Ask

- Is this a bug, or intended behaviour?
- For the PR: are maintainers OK with copier shelling out to `git update-index` from `_render_file` when the destination is git-tracked? It's a small extension of copier's existing git interactions (clone, conflict-recording) but I want to flag the design point before sending the diff. If not, the fallback is "subfix 1 only + a runtime warning on Windows".
- If this is intended: please document the asymmetry with `copier copy` prominently in the update docs, and emit a warning at update time when a template's mode bits would otherwise be silently dropped.

willemkokke:
## Summary

Fixes #2604.

`copier update` did not reliably propagate template executable-bit (file mode) changes to destinations. Three distinct failure modes:

1. **Mode-only template change** (no content change): `_render_allowed` classified the file as "identical" and `_render_file` short-circuited before reaching the `chmod` call — a silent no-op on every platform.
2. **Mode + content change on `core.fileMode=false`** (Windows default): copier wrote the new content and `chmod`d the file on disk, but the destination's git index was never updated. Since git ignores on-disk mode bits when `core.fileMode=false`, the executable bit was silently lost on the user's next commit.
3. **Template authored on Linux/macOS, consumed on Windows**: copier read the template's intended mode from `Path.stat().st_mode`, which on Windows never carries UNIX exec bits. Templates committed with `100755` on Linux were indistinguishable from `100644` when rendered on Windows, so the executable bit was lost end-to-end.

### Changes

**`copier/_main.py`**:

- `_render_allowed`: new `expected_mode` parameter. Compares only the executable bits (`0o111`) against the existing destination file so that a mode-only template change is no longer misclassified as "identical".
- `_render_file`: now consults `self.template.git_index_modes` for exec-bit determination, merging the git-tracked exec bits with the filesystem's non-exec flags (fall back to pure `stat()` for non-git templates and untracked files). This makes the fix work even on filesystems that don't represent exec bits on disk.
- `_sync_git_index_executable_bit` (new): runs only when `core.fileMode=false`. Uses `git update-index --cacheinfo <mode>,<sha>,<path>` (not `--chmod`), which rewrites *only* the mode on the existing blob SHA. `--chmod` looks simpler but has a hidden side effect of re-staging the current working-tree content as the new blob, which would stomp the downstream-edited blob that `_apply_update` needs to reconstruct merge conflicts. When `core.fileMode=true` the method is a no-op — git picks up the on-disk `chmod` naturally as an unstaged modification, matching copier's normal "leave changes unstaged for user review" behavior. All git failures are silently swallowed so the render path cannot be broken by unrelated git problems.

**`copier/_template.py`**:

- New `Template.git_index_modes` cached property: reads file modes directly from the template clone's git index via a single `git ls-files --stage` call. Returns an empty mapping for non-git templates, silently falling back to `stat()` callers.

**`tests/test_updatediff.py`** — new regression tests:

| Test | What it covers |
| --- | --- |
| `test_update_propagates_executable_bit_addition[True\|False]` | Mode-only promotion (gain `+x`, no content change), parametrized over `core.fileMode`. The `True` case asserts the change is unstaged after update, guarding against auto-staging regression; the `False` case asserts the index records `100755`. |
| `test_update_propagates_executable_bit_removal[True\|False]` | Symmetric: template drops `+x`, destination index/working tree must lose the exec bit. |
| `test_update_preserves_exec_bit_authored_on_unix` | Windows-only end-to-end test. Sets up the template using only `git update-index --chmod` (no `Path.chmod`) and verifies the exec bit survives all the way into the destination's git index on `core.fileMode=false` — exercising the `Template.git_index_modes` code path that's load-bearing on Windows. |
| `test_update_with_exec_bit_change_and_merge_conflict[True\|False]` | Deliberately creates a merge conflict on a file that gains `+x` between template versions, parametrized over `core.fileMode`. Asserts the conflict is properly registered (`UU`, conflict markers, stages 1/2/3 present). Locks in the ordering guarantee that `_sync_git_index_executable_bit` runs in `_render_file` *before* the conflict registration in `_apply_update`, and that `--cacheinfo` does not corrupt the conflict stages. |

## Test plan

- [x] All new regression tests pass on macOS
- [x] `test_update_preserves_exec_bit_authored_on_unix` passes on a real Windows machine (first run — previously only reasoned about)
- [x] Full copier test suite on macOS: 1063 passed, 5 skipped, 8 xfailed, 1 xpassed
- [x] Full copier test suite on Windows: all exec-bit-related tests green. 5 pre-existing symlink failures in `tests/test_symlinks.py` and `tests/test_dirty_local.py` reproduce identically on `master` on the same machine (local environment issues, unrelated to this PR)
- [x] ruff check, ruff format, mypy all clean
- [x] Rebased onto latest master and re-verified

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -512,6 +512,7 @@ class Worker:
         is_dir: bool = False,
         is_symlink: bool = False,
         expected_contents: bytes | Path = b"",
+        expected_mode: int | None = None,
     ) -> bool:
         """Determine if a file or directory can be rendered.
 
@@ -525,6 +526,12 @@ class Worker:
             expected_contents:
                 Used to compare existing file contents with them. Allows to know if
                 rendering is needed.
+            expected_mode:
+                Used to compare the existing file's mode bits with the
+                template's, so a mode-only change between template versions
+                is not misclassified as "identical". Only the executable bits
+                (``0o111``) are compared, since those are the only mode bits
+                tracked by git.
         """
         assert not dst_relpath.is_absolute()
         assert not expected_contents or not is_dir, "Dirs cannot have expected content"
@@ -551,8 +558,18 @@ class Worker:
                 raise
         except IsADirectoryError:
             assert is_dir
+        mode_matches = True
+        if expected_mode is not None and not previous_is_symlink and not is_dir:
+            try:
+                dst_exec_bits = dst_abspath.stat().st_mode & 0o111
+            except (FileNotFoundError, PermissionError):
+                dst_exec_bits = None
+            else:
+                mode_matches = dst_exec_bits == (expected_mode & 0o111)
         if is_dir or (
-            previous_content == expected_contents and previous_is_symlink == is_symlink
+            previous_content == expected_contents
+            and previous_is_symlink == is_symlink
+            and mode_matches
         ):
             printf(
                 "identical",
@@ -835,8 +852,30 @@ class Worker:
         else:
             new_content = src_abspath.read_bytes()
         dst_abspath = self.subproject.local_abspath / dst_relpath
-        src_mode = src_abspath.stat().st_mode
-        if not self._render_allowed(dst_relpath, expected_contents=new_content):
+        # Prefer the template's git-index mode over ``stat().st_mode`` so
+        # that executable bits committed by the template author are
+        # honored even on filesystems that don't represent them on disk
+        # (notably Windows). ``stat().st_mode`` is used for
+        # non-executable-bit flags (user/group/world read/write, etc.)
+        # and as a fallback when the file isn't tracked in the
+        # template's git index — e.g. local directory templates without
+        # a git repo, or untracked files.
+        stat_mode = src_abspath.stat().st_mode
+        git_mode = self.template.git_index_modes.get(
+            PurePosixPath(src_relpath.as_posix())
+        )
+        if git_mode is None:
+            src_mode = stat_mode
+        else:
+            # Merge the git-tracked executable bits with the filesystem's
+            # non-exec bits so we don't lose read/write flags that matter
+            # for the destination chmod.
+            src_mode = (stat_mode & ~0o111) | (git_mode & 0o111)
+        if not self._render_allowed(
+            dst_relpath,
+            expected_contents=new_content,
+            expected_mode=src_mode,
+        ):
             return
         if not self.pretend:
             dst_abspath.parent.mkdir(parents=True, exist_ok=True)
@@ -856,6 +895,93 @@ class Worker:
                         f"{stat.filemode(dst_mode)} to {stat.filemode(src_mode)}",
                         stacklevel=2,
                     )
+            self._sync_git_index_executable_bit(dst_relpath, src_mode)
+
+    def _sync_git_index_executable_bit(self, dst_relpath: Path, src_mode: int) -> None:
+        """Propagate executable-bit changes to the destination's git index.
+
+        Only needed when ``core.fileMode`` is ``false`` (the Windows default
+        and a common opt-out elsewhere): in that case git ignores on-disk
+        mode bits, so the ``chmod`` performed by :meth:`_render_file` is
+        invisible to git, and the executable bit would be silently lost on
+        the user's next commit. To keep the destination's index in sync
+        with the template, we explicitly rewrite the entry's mode via
+        ``git update-index --cacheinfo``.
+
+        When ``core.fileMode`` is ``true`` (or unset, the unix default),
+        this method is a no-op: git already picks up the on-disk ``chmod``
+        as an *unstaged* modification, which matches copier's normal
+        behavior of leaving rendered changes unstaged for user review.
+
+        Also a no-op when the destination is not in a git repository, when
+        the file is not yet tracked (the user's eventual ``git add`` will
+        record the on-disk mode where the platform allows it), or when git
+        is unavailable. All git failures are swallowed so that copying or
+        updating cannot be broken by an unrelated git problem.
+
+        .. note::
+
+            ``git update-index --chmod=±x`` cannot be used here even
+            though it looks simpler: it has a side effect of also
+            re-staging the current working-tree content as the blob (it
+            implicitly refreshes the index entry). During ``copier
+            update`` the working tree has already been overwritten with
+            the new template content at the time this method runs, so
+            ``--chmod`` would stomp the downstream-edited blob that the
+            update flow needs to reconstruct merge conflicts.
+            ``--cacheinfo`` rewrites the mode on the *existing* blob
+            SHA only, leaving the rest of the entry (and therefore the
+            conflict-reconstruction flow) untouched.
+        """
+        subproject_root = self.subproject.local_abspath
+        git = get_git(context_dir=subproject_root)
+        try:
+            # ``--type=bool`` normalizes truthy/falsy spellings to
+            # ``true``/``false``.  Exits 1 if the key is unset.
+            file_mode_setting = git(
+                "config", "--type=bool", "--get", "core.fileMode"
+            ).strip()
+        except ProcessExecutionError:
+            # Not a git repo, or ``core.fileMode`` is unset — unix default
+            # is ``true``, so plain ``chmod`` is enough; nothing to do.
+            return
+        if file_mode_setting != "false":
+            # git will pick up the plain on-disk ``chmod`` from
+            # :meth:`_render_file`; no index manipulation needed.
+            return
+        try:
+            # TODO: simplify with ``--format %(objectmode)`` once the
+            # minimum git version is raised to 2.38+ (--format cannot be
+            # combined with --stage; see git-ls-files(1)).
+            result = git("ls-files", "--stage", "--", str(dst_relpath)).strip()
+        except ProcessExecutionError:
+            # git can't read the index — fall back to a silent no-op.
+            return
+        if not result:
+            # File is not tracked yet; nothing to update.
+            return
+        # Format: "<mode> <sha> <stage>\t<path>"
+        meta = result.split("\t", 1)[0].split()
+        current_index_mode = int(meta[0], 8)
+        current_index_sha = meta[1]
+        desired_executable = bool(src_mode & 0o111)
+        current_executable = bool(current_index_mode & 0o111)
+        if desired_executable == current_executable:
+            return
+        new_mode = "100755" if desired_executable else "100644"
+        try:
+            # ``--cacheinfo`` rewrites the entry's mode on the *existing*
+            # blob SHA. Unlike ``--chmod``, it does NOT re-read the
+            # working tree or restage its content.
+            git(
+                "update-index",
+                "--cacheinfo",
+                f"{new_mode},{current_index_sha},{dst_relpath}",
+            )
+        except (OSError, ProcessExecutionError):
+            # git not installed, or some other unrelated git failure
+            # — silently fall back so we never break the render path.
+            pass
 
     def _render_symlink(self, src_relpath: Path, dst_relpath: Path) -> None:
         """Render one symlink.
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -510,6 +510,7 @@ class Worker:
         is_dir: bool = False,
         is_symlink: bool = False,
         expected_contents: bytes | Path = b"",
+        expected_mode: int | None = None,
     ) -> bool:
         """Determine if a file or directory can be rendered.
 
@@ -523,6 +524,12 @@ class Worker:
             expected_contents:
                 Used to compare existing file contents with them. Allows to know if
                 rendering is needed.
+            expected_mode:
+                Used to compare the existing file's mode bits with the
+                template's, so a mode-only change between template versions
+                is not misclassified as "identical". Only the executable bits
+                (``0o111``) are compared, since those are the only mode bits
+                tracked by git.
         """
         assert not dst_relpath.is_absolute()
         assert not expected_contents or not is_dir, "Dirs cannot have expected content"
@@ -549,8 +556,18 @@ class Worker:
                 raise
         except IsADirectoryError:
             assert is_dir
+        mode_matches = True
+        if expected_mode is not None and not previous_is_symlink and not is_dir:
+            try:
+                dst_exec_bits = dst_abspath.stat().st_mode & 0o111
+            except (FileNotFoundError, PermissionError):
+                dst_exec_bits = None
+            else:
+                mode_matches = dst_exec_bits == (expected_mode & 0o111)
         if is_dir or (
-            previous_content == expected_contents and previous_is_symlink == is_symlink
+            previous_content == expected_contents
+            and previous_is_symlink == is_symlink
+            and mode_matches
         ):
             printf(
                 "identical",
@@ -825,8 +842,30 @@ class Worker:
         else:
             new_content = src_abspath.read_bytes()
         dst_abspath = self.subproject.local_abspath / dst_relpath
-        src_mode = src_abspath.stat().st_mode
-        if not self._render_allowed(dst_relpath, expected_contents=new_content):
+        # Prefer the template's git-index mode over ``stat().st_mode`` so
+        # that executable bits committed by the template author are
+        # honored even on filesystems that don't represent them on disk
+        # (notably Windows). ``stat().st_mode`` is used for
+        # non-executable-bit flags (user/group/world read/write, etc.)
+        # and as a fallback when the file isn't tracked in the
+        # template's git index — e.g. local directory templates without
+        # a git repo, or untracked files.
+        stat_mode = src_abspath.stat().st_mode
+        git_mode = self.template.git_index_modes.get(
+            PurePosixPath(src_relpath.as_posix())
+        )
+        if git_mode is None:
+            src_mode = stat_mode
+        else:
+            # Merge the git-tracked executable bits with the filesystem's
+            # non-exec bits so we don't lose read/write flags that matter
+            # for the destination chmod.
+            src_mode = (stat_mode & ~0o111) | (git_mode & 0o111)
+        if not self._render_allowed(
+            dst_relpath,
+            expected_contents=new_content,
+            expected_mode=src_mode,
+        ):
             return
         if not self.pretend:
             dst_abspath.parent.mkdir(parents=True, exist_ok=True)
@@ -846,6 +885,93 @@ class Worker:
                         f"{stat.filemode(dst_mode)} to {stat.filemode(src_mode)}",
                         stacklevel=2,
                     )
+            self._sync_git_index_executable_bit(dst_relpath, src_mode)
+
+    def _sync_git_index_executable_bit(self, dst_relpath: Path, src_mode: int) -> None:
+        """Propagate executable-bit changes to the destination's git index.
+
+        Only needed when ``core.fileMode`` is ``false`` (the Windows default
+        and a common opt-out elsewhere): in that case git ignores on-disk
+        mode bits, so the ``chmod`` performed by :meth:`_render_file` is
+        invisible to git, and the executable bit would be silently lost on
+        the user's next commit. To keep the destination's index in sync
+        with the template, we explicitly rewrite the entry's mode via
+        ``git update-index --cacheinfo``.
+
+        When ``core.fileMode`` is ``true`` (or unset, the unix default),
+        this method is a no-op: git already picks up the on-disk ``chmod``
+        as an *unstaged* modification, which matches copier's normal
+        behavior of leaving rendered changes unstaged for user review.
+
+        Also a no-op when the destination is not in a git repository, when
+        the file is not yet tracked (the user's eventual ``git add`` will
+        record the on-disk mode where the platform allows it), or when git
+        is unavailable. All git failures are swallowed so that copying or
+        updating cannot be broken by an unrelated git problem.
+
+        .. note::
+
+            ``git update-index --chmod=±x`` cannot be used here even
+            though it looks simpler: it has a side effect of also
+            re-staging the current working-tree content as the blob (it
+            implicitly refreshes the index entry). During ``copier
+            update`` the working tree has already been overwritten with
+            the new template content at the time this method runs, so
+            ``--chmod`` would stomp the downstream-edited blob that the
+            update flow needs to reconstruct merge conflicts.
+            ``--cacheinfo`` rewrites the mode on the *existing* blob
+            SHA only, leaving the rest of the entry (and therefore the
+            conflict-reconstruction flow) untouched.
+        """
+        subproject_root = self.subproject.local_abspath
+        git = get_git(context_dir=subproject_root)
+        try:
+            # ``--type=bool`` normalizes truthy/falsy spellings to
+            # ``true``/``false``.  Exits 1 if the key is unset.
+            file_mode_setting = git(
+                "config", "--type=bool", "--get", "core.fileMode"
+            ).strip()
+        except ProcessExecutionError:
+            # Not a git repo, or ``core.fileMode`` is unset — unix default
+            # is ``true``, so plain ``chmod`` is enough; nothing to do.
+            return
+        if file_mode_setting != "false":
+            # git will pick up the plain on-disk ``chmod`` from
+            # :meth:`_render_file`; no index manipulation needed.
+            return
+        try:
+            # TODO: simplify with ``--format %(objectmode)`` once the
+            # minimum git version is raised to 2.38+ (--format cannot be
+            # combined with --stage; see git-ls-files(1)).
+            result = git("ls-files", "--stage", "--", str(dst_relpath)).strip()
+        except ProcessExecutionError:
+            # git can't read the index — fall back to a silent no-op.
+            return
+        if not result:
+            # File is not tracked yet; nothing to update.
+            return
+        # Format: "<mode> <sha> <stage>\t<path>"
+        meta = result.split("\t", 1)[0].split()
+        current_index_mode = int(meta[0], 8)
+        current_index_sha = meta[1]
+        desired_executable = bool(src_mode & 0o111)
+        current_executable = bool(current_index_mode & 0o111)
+        if desired_executable == current_executable:
+            return
+        new_mode = "100755" if desired_executable else "100644"
+        try:
+            # ``--cacheinfo`` rewrites the entry's mode on the *existing*
+            # blob SHA. Unlike ``--chmod``, it does NOT re-read the
+            # working tree or restage its content.
+            git(
+                "update-index",
+                "--cacheinfo",
+                f"{new_mode},{current_index_sha},{dst_relpath}",
+            )
+        except (OSError, ProcessExecutionError):
+            # git not installed, or some other unrelated git failure
+            # — silently fall back so we never break the render path.
+            pass
 
     def _render_symlink(self, src_relpath: Path, dst_relpath: Path) -> None:
         """Render one symlink.
--- a/copier/_template.py
+++ b/copier/_template.py
@@ -19,6 +19,7 @@ import packaging.version
 import yaml
 from funcy import lflatten
 from packaging.version import Version, parse
+from plumbum.commands.processes import ProcessExecutionError
 from plumbum.machines import local
 from pydantic.dataclasses import dataclass
 
@@ -578,6 +579,46 @@ class Template:
         with suppress(OSError):
             result = result.resolve()
         return result
+
+    @cached_property
+    def git_index_modes(self) -> Mapping[PurePosixPath, int]:
+        """Read file modes from the template's git index.
+
+        This returns the mode bits for every tracked file as recorded in
+        git's index, keyed by POSIX path relative to
+        :attr:`local_abspath`. Git always records executable-bit
+        information in the index (as mode ``100755``/``100644``), even
+        on filesystems where the bit can't be represented on disk — most
+        notably Windows, where ``os.stat().st_mode`` never reports
+        ``S_IXUSR``/``S_IXGRP``/``S_IXOTH`` for regular files.
+
+        Callers that want to know the template's *intended* file mode
+        (as committed by the template author) should consult this
+        mapping before falling back to ``Path.stat().st_mode``.
+
+        Returns an empty mapping when the template is not a git
+        checkout, when git is unavailable, or when git fails for any
+        other reason — callers must be ready to fall back.
+        """
+        if self.vcs != "git":
+            return {}
+        try:
+            git = get_git(context_dir=self.local_abspath)
+            # TODO: simplify with ``--format`` once the minimum git
+            # version is raised to 2.38+ (--format cannot be combined
+            # with --stage; see git-ls-files(1)).
+            raw = git("ls-files", "--stage").strip()
+        except (OSError, ProcessExecutionError):
+            return {}
+        modes: dict[PurePosixPath, int] = {}
+        for line in raw.splitlines():
+            # Format: "<mode> <sha> <stage>\t<path>"
+            if "\t" not in line:
+                continue
+            meta, path = line.split("\t", 1)
+            mode_str = meta.split(" ", 1)[0]
+            modes[PurePosixPath(path)] = int(mode_str, 8)
+        return modes
 
     @cached_property
     def url_expanded(self) -> str:
--- a/copier/_template.py
+++ b/copier/_template.py
@@ -19,6 +19,7 @@ import packaging.version
 import yaml
 from funcy import lflatten
 from packaging.version import Version, parse
+from plumbum.commands.processes import ProcessExecutionError
 from plumbum.machines import local
 from pydantic.dataclasses import dataclass
 
@@ -578,6 +579,46 @@ class Template:
         with suppress(OSError):
             result = result.resolve()
         return result
+
+    @cached_property
+    def git_index_modes(self) -> Mapping[PurePosixPath, int]:
+        """Read file modes from the template's git index.
+
+        This returns the mode bits for every tracked file as recorded in
+        git's index, keyed by POSIX path relative to
+        :attr:`local_abspath`. Git always records executable-bit
+        information in the index (as mode ``100755``/``100644``), even
+        on filesystems where the bit can't be represented on disk — most
+        notably Windows, where ``os.stat().st_mode`` never reports
+        ``S_IXUSR``/``S_IXGRP``/``S_IXOTH`` for regular files.
+
+        Callers that want to know the template's *intended* file mode
+        (as committed by the template author) should consult this
+        mapping before falling back to ``Path.stat().st_mode``.
+
+        Returns an empty mapping when the template is not a git
+        checkout, when git is unavailable, or when git fails for any
+        other reason — callers must be ready to fall back.
+        """
+        if self.vcs != "git":
+            return {}
+        try:
+            git = get_git(context_dir=self.local_abspath)
+            # TODO: simplify with ``--format`` once the minimum git
+            # version is raised to 2.38+ (--format cannot be combined
+            # with --stage; see git-ls-files(1)).
+            raw = git("ls-files", "--stage").strip()
+        except (OSError, ProcessExecutionError):
+            return {}
+        modes: dict[PurePosixPath, int] = {}
+        for line in raw.splitlines():
+            # Format: "<mode> <sha> <stage>\t<path>"
+            meta, _, path = line.partition("\t")
+            if not path:
+                continue
+            mode_str, _, _ = meta.partition(" ")
+            modes[PurePosixPath(path)] = int(mode_str, 8)
+        return modes
 
     @cached_property
     def url_expanded(self) -> str:
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -9,6 +9,7 @@ from typing import Literal
 
 import pexpect
 import pytest
+from inline_snapshot import snapshot
 from plumbum import local
 
 from copier._cli import CopierApp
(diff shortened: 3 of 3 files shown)


copier/_main.py
```diff
@@ -549,8 +556,18 @@ def _render_allowed(
                 raise
         except IsADirectoryError:
             assert is_dir
+        mode_matches = True
+        if expected_mode is not None and not previous_is_symlink and not is_dir:
+            try:
+                dst_exec_bits = dst_abspath.stat().st_mode & 0o111
+            except (FileNotFoundError, PermissionError):
+                dst_exec_bits = None
+            if dst_exec_bits is not None:
```
sisp: How about a minor simplification here?

```suggestion
            else:
```

copier/_main.py
```diff
@@ -846,6 +867,49 @@ def _render_file(  # noqa: C901
...
+            try:
+                ls = git("ls-files", "--stage", "--", str(dst_relpath)).strip()
+            except ProcessExecutionError:
+                # Not a git repo, or git can't read the index — fall back
+                # to a silent no-op.
+                return
+            if not ls:
+                # File is not tracked yet; nothing to update.
+                return
+            current_index_mode = int(ls.split()[0], 8)
```
sisp: We can simplify and improve robustness of the file mode extraction by adding `--format %(objectmode)` to the `git ls-files` command to print only the file mode:

```suggestion
                result = git("ls-files", "--stage", "--format", "%(objectmode)", "--", str(dst_relpath)).strip()
            except ProcessExecutionError:
                # Not a git repo, or git can't read the index — fall back
                # to a silent no-op.
                return
            if not result:
                # File is not tracked yet; nothing to update.
                return
            current_index_mode = int(result, 8)
```

copier/_main.py
```diff
@@ -846,6 +867,49 @@ def _render_file(  # noqa: C901
...
+            desired_executable = bool(src_mode & 0o111)
+            current_executable = bool(current_index_mode & 0o111)
+            if desired_executable == current_executable:
+                return
+            flag = "+x" if desired_executable else "-x"
+            git("update-index", f"--chmod={flag}", "--", str(dst_relpath))
```
sisp: I think it's better style to surround this call with `try`/`except` rather than surrounding the entire method body.

Essentially:


```suggestion
        try:
            git("update-index", f"--chmod={flag}", "--", str(dst_relpath))
```

And remove the topmost `try` and dedent the lines above here accordingly.

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+    reason="Windows does not have UNIX-like permissions",
+)
+def test_update_propagates_executable_bit_with_no_content_change(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    """A template that gains ``+x`` between versions (without changing the
+    file's content) must propagate that bit to existing destinations on
+    ``copier update``. Without the fix, ``_render_allowed`` short-circuits
+    on identical content and ``_render_file`` never reaches the ``chmod``
+    or the index update.
```
sisp: Don't mention some fix for which the context is missing here – it only exists as part of this PR but is unclear when just reading the docstring. Also, the docstring should describe the scenario under test and not explain implementation details (which may change by the way) – IMHO.

```suggestion
    ``copier update``.
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+        )
+        git_save(tag="v1")
+
+    run_copy(str(src), dst, defaults=True, overwrite=True)
+    with local.cwd(dst):
+        git_init("init")
+    assert (dst / "launcher.sh").stat().st_mode & 0o111 == 0
+    with local.cwd(dst):
+        index_entry = git("ls-files", "--stage", "--", "launcher.sh").strip()
+    assert index_entry.split()[0] == "100644"
```
sisp: Same as above – we can simplify mode extraction here:

```suggestion
        mode = git("ls-files", "--stage", "--format", "%(objectmode)", "--", "launcher.sh").strip()
    assert mode == "100644"
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+
+    # Template v2: launcher.sh gains +x with no content change.
+    with local.cwd(src):
+        (src / "launcher.sh").chmod(
+            (src / "launcher.sh").stat().st_mode
```
sisp: The current working directory is already `src`, so `(src / ...)` is redundant:

```suggestion
        Path("launcher.sh").chmod(
            Path("launcher.sh").stat().st_mode
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+    run_update(str(dst), defaults=True, overwrite=True)
+
+    # On disk, the bit must be set after the update.
+    assert (dst / "launcher.sh").stat().st_mode & 0o111 != 0
+    # And copier must have propagated the bit to the destination's git
+    # index, so the change is durable across a checkout / on
+    # ``core.fileMode=false`` filesystems.
+    with local.cwd(dst):
+        index_entry = git("ls-files", "--stage", "--", "launcher.sh").strip()
+    assert index_entry.split()[0] == "100755"
```
sisp: Same as above:


```suggestion
        mode = git("ls-files", "--stage", "--format", "%(objectmode)", "--", "launcher.sh").strip()
    assert mode == "100755"
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+        git_save(tag="v1")
+
+    run_copy(str(src), dst, defaults=True, overwrite=True)
+    with local.cwd(dst):
+        git("init")
+        git("config", "core.fileMode", "false")
+        git("add", "-A")
+        git("commit", "-m", "init")
+        index_entry = git("ls-files", "--stage", "--", "launcher.sh").strip()
+    assert index_entry.split()[0] == "100644"
```
sisp: Same as above:


```suggestion
        mode = git("ls-files", "--stage", "--format", "%(objectmode)", "--", "launcher.sh").strip()
    assert mode == "100644"
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+
+    run_update(str(dst), defaults=True, overwrite=True)
+
+    # The destination's git index must already record 100755, even before
+    # the user runs ``git add``. With ``core.fileMode=false``, ``git add``
+    # would otherwise leave the index at 100644 because git refuses to
+    # look at the on-disk mode bit.
+    with local.cwd(dst):
+        index_entry = git("ls-files", "--stage", "--", "launcher.sh").strip()
+    assert index_entry.split()[0] == "100755"
```
sisp: Same as above:

```suggestion
        mode = git("ls-files", "--stage", "--format", "%(objectmode)", "--", "launcher.sh").strip()
    assert mode == "100755"
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+    # Template v2: gains +x AND changes content.
+    with local.cwd(src):
+        (src / "launcher.sh").chmod(
+            (src / "launcher.sh").stat().st_mode
+            | stat.S_IXUSR
+            | stat.S_IXGRP
+            | stat.S_IXOTH
+        )
+        git("update-index", "--chmod=+x", "launcher.sh")
+        (src / "launcher.sh").write_text("#!/bin/sh\necho hi\n# updated\n")
```
sisp: Why is it important for this test to update the file content? It seems the main purpose is to test with `core.fileMode=false`.

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+@pytest.mark.skipif(
+    condition=platform.system() == "Windows",
+    reason="Windows does not have UNIX-like permissions",
+)
+def test_update_propagates_executable_bit_removal(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    """The symmetric direction of the previous tests: when a template
+    drops the executable bit between versions, the destination's index
+    must lose ``100755`` too.
```
sisp: Please rephrase the docstring to avoid referring to "previous tests" because this reference might not be stable.

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+            | stat.S_IXGRP
+            | stat.S_IXOTH
+        )
+        git_save(tag="v1")
+
+    run_copy(str(src), dst, defaults=True, overwrite=True)
+    with local.cwd(dst):
+        git_init("init")
+        index_entry = git("ls-files", "--stage", "--", "launcher.sh").strip()
+    assert index_entry.split()[0] == "100755"
```
sisp: Same as above:


```suggestion
        mode = git("ls-files", "--stage", "--format", "%(objectmode)", "--", "launcher.sh").strip()
    assert mode == "100755"
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+        git("update-index", "--chmod=-x", "launcher.sh")
+        git("commit", "-am", "drop executable")
+        git("tag", "v2")
+
+    run_update(str(dst), defaults=True, overwrite=True)
+
+    assert (dst / "launcher.sh").stat().st_mode & 0o111 == 0
+    with local.cwd(dst):
+        index_entry = git("ls-files", "--stage", "--", "launcher.sh").strip()
+    assert index_entry.split()[0] == "100644"
```
sisp: Same as above:

```suggestion
        mode = git("ls-files", "--stage", "--format", "%(objectmode)", "--", "launcher.sh").strip()
    assert mode == "100644"
```

tests/test_updatediff.py
sisp, line 1: Does it make sense to run each test with `core.fileMode=true` and `core.fileMode=false` to make sure that the executable bit update is working under both conditions?

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_nice_yaml }}",
+                "launcher.sh": "#!/bin/sh\necho hi\n",
+            }
+        )
+        (src / "launcher.sh").chmod(
+            (src / "launcher.sh").stat().st_mode
```
sisp: Same as above:


```suggestion
        Path("launcher.sh").chmod(
            Path("launcher.sh").stat().st_mode
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,169 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+        git("config", "core.fileMode", "false")
+        git("add", "-A")
+        git("commit", "-m", "init")
+        index_entry = git("ls-files", "--stage", "--", "launcher.sh").strip()
+    assert index_entry.split()[0] == "100644"
+
+    # Template v2: gains +x AND changes content.
+    with local.cwd(src):
+        (src / "launcher.sh").chmod(
+            (src / "launcher.sh").stat().st_mode
```
sisp: Same as above:

```suggestion
        Path("launcher.sh").chmod(
            Path("launcher.sh").stat().st_mode
```

sisp (changes requested):
Thanks for discovering this bug and contributing a fix, @willemkokke! :bow: I've left a couple of remarks.

willemkokke:
All very valid, thanks!

Will update my PR now

copier/_main.py
willemkokke: the ls-file --format argument is introduced in Git 2.38 (Oct 20, 2022)(https://github.com/git/git/blob/master/Documentation/RelNotes/2.38.0.adoc)

Copier advertises backwards compatibility to Git 2.27 (Jun 2, 2020) (https://copier.readthedocs.io/en/stable/#installation)

Using --format would break this backwards compatibility guarantee. 

Do you want to raise the minimum git version or use the current backwards compatible version?

sisp: Oh, thanks for catching this! 👌 In this case, let's retain compatibility to Git 2.27 and keep extracting the mode manually.

willemkokke:
Thanks for the thorough review, @sisp! I've addressed all your feedback. Here's a summary of the changes:

**`copier/_main.py`**
- `_render_allowed`: simplified `if dst_exec_bits is not None:` → `else:`
- `_sync_git_index_executable_bit`: narrowed the outer `try`/`except` to wrap only the `git update-index` call instead of the entire method body

**`tests/test_updatediff.py`**
- Parametrized over `core.fileMode` (`True`/`False`) instead of separate test functions — collapsed 3 tests into 2
- Removed implementation-detail references from docstrings ("the fix", `_render_allowed`, "previous tests")
- Replaced `(src / "launcher.sh")` with `Path("launcher.sh")` where already inside `local.cwd(src)`
- Dropped the content change from the `core.fileMode=false` scenario (no longer needed since `_render_allowed` now detects mode-only changes)

**Intentionally kept `--stage` + `.split()[0]`** instead of switching to `--format %(objectmode)`: the `--format` option for `git ls-files` requires Git 2.38+ ([git-ls-files(1)](https://git-scm.com/docs/git-ls-files#Documentation/git-ls-files.txt---formatltformatgt)), while copier documents Git 2.27+ as the minimum. I've added TODO comments in both the production code and tests to simplify this once the minimum git version is raised. Happy to switch if bumping the minimum to 2.38+ is acceptable.

sisp:
Thanks for updating the PR, @willemkokke! :+1:

I noticed that running `git update-index --chmod=±x <file>` stages the file, whereas changing a file's mode with `chmod +x <file>` with `core.fileMode=true` lets Git detect the change _without_ staging it. According to my research, there seems to be no way to avoid staging the file with `git update-index --chmod=±x <file>`, and a subsequent `git restore --staged <file>` undoes the file mode change. Do you agree? I also wonder if there are any side effects when the same file has a merge conflict on update. Would you mind checking this (perhaps a simple test case might help)?

willemkokke:
Thanks for raising this, @sisp! You're right on both counts — I did some investigation and here's what I found:

## 1. Auto-staging with `git update-index --chmod`

Confirmed: there's no way to update the index mode without also staging the change. The git plumbing simply doesn't distinguish between "change this field in the index" and "stage this field". So `update-index --chmod` always stages.

However, this side effect is **only an issue when `core.fileMode=true`**. In that case, the existing on-disk `chmod` in [`_render_file`](https://github.com/copier-org/copier/blob/fix/preserve-mode-bits-on-update/copier/_main.py#L861) is already picked up by git as an *unstaged* modification — exactly matching copier's normal "leave things unstaged for user review" behavior. So the `update-index` call is redundant *and* introduces the auto-staging inconsistency.

When `core.fileMode=false`, `update-index --chmod` is the only way to record the mode in the index, and pre-staging is not a behavioral regression because git would have ignored plain `chmod` anyway.

**Fix**: skip `update-index --chmod` entirely when `core.fileMode` is not `false`. I'll push this shortly.

## 2. Merge conflict interaction

This is where it gets interesting. I reproduced the scenario in isolation and confirmed your suspicion was warranted — and worse than you might have thought. Running `git update-index --chmod=±x` on a file in stages 1/2/3 doesn't just stage the mode change: it **silently collapses all stages into stage 0**, using the current working-tree content (including conflict markers) as the blob. `git status` changes from `UU` to `MM` — the conflict indicator disappears. A subsequent `git commit` would commit conflict markers with no warning.

**However**, after tracing the update flow carefully, this cannot actually occur during `copier update`:

- [`run_update`](https://github.com/copier-org/copier/blob/fix/preserve-mode-bits-on-update/copier/_main.py#L1229) refuses to run on a dirty working tree via `is_dirty()`, which uses `git status --porcelain`. Unmerged entries show up in porcelain output, so pre-existing merge conflicts are blocked with *"Destination repository is dirty; cannot continue."*
- In `_apply_update`, conflicts created *during* the update are registered *after* `_render_file` runs. The sequence is:
    1. [L1353](https://github.com/copier-org/copier/blob/fix/preserve-mode-bits-on-update/copier/_main.py#L1353) `current_worker.run_copy()` → `_render_file` → `_sync_git_index_executable_bit`. **Index has only stage 0 here.**
    2. [L1457](https://github.com/copier-org/copier/blob/fix/preserve-mode-bits-on-update/copier/_main.py#L1457) `git apply --reject` creates `.rej` files.
    3. [L1487](https://github.com/copier-org/copier/blob/fix/preserve-mode-bits-on-update/copier/_main.py#L1487) `git merge-file` writes inline conflict markers.
    4. [L1553](https://github.com/copier-org/copier/blob/fix/preserve-mode-bits-on-update/copier/_main.py#L1553) `git update-index --index-info` registers stages 1/2/3.
- The conflict registration at [L1529](https://github.com/copier-org/copier/blob/fix/preserve-mode-bits-on-update/copier/_main.py#L1529) reads `ls-files --stage` for the current perms when constructing stages 1/2/3, so the exec bit we set in stage 0 propagates correctly through the conflict flow.

In other words, copier's clean-index guarantee + the order of operations in `_apply_update` means our `update-index --chmod` always runs on a conflict-free index. The catastrophic scenario I reproduced in isolation can't happen here.

## Plan

1. Skip `update-index --chmod` when `core.fileMode` is not `false`.
2. Update the existing parametrized tests: the `file_mode=True` case should assert the change is **unstaged** after update (guards against regression into auto-staging).
3. Add a new regression test that deliberately creates a merge conflict on a file gaining `+x` (parametrized over `core.fileMode`), asserting that the conflict is properly registered (`UU`, stages 1/2/3 present, exec bit propagated through) — to lock in the timing guarantee against future refactors.

Does this sound reasonable? Will push once you confirm the direction.

sisp:
Sounds reasonable, let's go ahead as you suggested.

willemkokke:
Hey @sisp — update on this PR. The second round of feedback prompted a deeper investigation that uncovered a few things worth flagging, and the fix is now broader than originally planned. TL;DR: cross-platform support is in, conflict corruption is guarded against, and I verified on Windows.

## What changed from the [previous plan I outlined](https://github.com/copier-org/copier/pull/2605#issuecomment-4236596571)

### 1. `git update-index --chmod` has a hidden side effect — switched to `--cacheinfo`

While writing the merge-conflict regression test, the `file_mode=False` case failed in a way that didn't match my earlier reasoning. Digging in, I found that **`git update-index --chmod=±x` does not just change the mode** — it implicitly re-reads the working tree and re-stages the current file content as the new blob. Minimal repro:

```
$ echo "original" > file && git add file && git commit -qm init
$ echo "modified" > file
$ git ls-files --stage file
100644 4b48deed3a433909bfd6b6ab3d4b91348b6af464 0	file
$ git update-index --chmod=+x file
$ git ls-files --stage file
100755 2e0996000b7e9019eabcad29391bf0f5c7702f0b 0	file   ← new blob!
```

During `copier update`, `_render_file` writes the new template content to disk **before** `_sync_git_index_executable_bit` runs. `--chmod` was therefore stomping the downstream-edited blob that `_apply_update`'s conflict-reconstruction flow ([L1487-L1556](https://github.com/copier-org/copier/blob/master/copier/_main.py#L1487-L1556)) needs to derive stages 1/2/3. The merge conflict silently disappeared — exactly the kind of data-corruption scenario you'd originally flagged, just via a different mechanism than I'd predicted.

**Fix**: switched to `git update-index --cacheinfo <mode>,<sha>,<path>`, which rewrites *only* the mode on the existing blob SHA. Left a detailed code comment explaining why we don't use `--chmod`, because this is the kind of non-obvious trap a future refactor could regress into. Verified comma-form `--cacheinfo` has been available since Git ~1.9/2014, so it's well under the 2.27 minimum.

### 2. Only touch the index when `core.fileMode=false` (auto-staging side effect)

As planned in the previous comment. When `core.fileMode=true`, `_sync_git_index_executable_bit` is a no-op — git picks up the on-disk `chmod` naturally as an unstaged modification, matching copier's normal "leave rendered changes unstaged for user review" behavior.

### 3. Scope expanded to fix Windows end-to-end (new)

Looking at the proposed "update existing tests to run on both `core.fileMode=true/false`" suggestion, I realized copier had a deeper, unaddressed issue on Windows that our fix didn't touch:

1. Template authored on Linux/macOS commits `launcher.sh` with index mode `100755`.
2. Windows user clones → on-disk file has no exec bit (Windows filesystems don't represent them). Git checkout with `core.fileMode=false` (Windows default) preserves this.
3. In `_render_file`, `src_mode = src_abspath.stat().st_mode` — on Windows this returns a mode without `S_IXUSR`/`GRP`/`OTH`, because Python's `os.stat` doesn't derive exec bits from filesystem state for regular files.
4. So `src_mode & 0o111 == 0`, and copier thinks the template file is non-executable. The exec bit is lost end-to-end, regardless of `core.fileMode` on the destination.

**Fix**: new `Template.git_index_modes` cached property reads the template's intended file modes directly from the template clone's git index via a single `git ls-files --stage` call. `_render_file` now consults this for exec-bit determination, merging the git-tracked exec bits with the filesystem's non-exec flags, and falls back to pure `stat()` for non-git templates and untracked files.

This means `copier` now correctly propagates template exec bits across all platform combinations: Linux↔Linux, Linux→Windows, macOS↔macOS, etc.

### 4. Windows-only regression test (new)

`test_update_preserves_exec_bit_authored_on_unix` — deliberately sets up a template using only `git update-index --chmod` (no `Path.chmod`, which would be a no-op on Windows anyway) and verifies the exec bit survives end-to-end into the destination's git index on `core.fileMode=false`. This test is Windows-only because on Linux/macOS `git checkout` restores the exec bit on disk during the template clone, so `stat().st_mode` independently arrives at the right answer — the test would pass on Linux even if the `git_index_modes` code path were broken, giving a false positive for the Windows scenario it's meant to cover. Keeping it Windows-only keeps the intent clear.

### 5. Merge conflict regression test (new)

`test_update_with_exec_bit_change_and_merge_conflict` — parametrized over `core.fileMode=[True, False]`. Deliberately creates a merge conflict on a file that also gains `+x` between template versions, then asserts the conflict is properly registered (`UU` in porcelain, conflict markers in the working tree). This locks in the ordering guarantee discussed above, and specifically catches the failure mode that the `--chmod`→`--cacheinfo` fix addresses.

## Verification

- **macOS**: `uv run poe test` → 1063 passed, 5 skipped, 8 xfailed, 1 xpassed. Zero failures. All new tests green. Rebased onto latest master and re-ran: still green.
- **Windows**: all new exec-bit tests green, including `test_update_preserves_exec_bit_authored_on_unix` on a real Windows machine for the first time (previously only reasoned about). The full suite on this Windows machine also reported 5 pre-existing symlink-related failures in `tests/test_symlinks.py` and `tests/test_dirty_local.py`. These look like local-environment issues (symlink emulation / developer mode / autocrlf) and they reproduce identically on `master` on the same machine, so they're unrelated to this PR.

## Summary of commit

Single amended commit with:

- `copier/_main.py`:
  - `_render_allowed` gains an `expected_mode` parameter so mode-only changes are no longer classified as "identical"
  - `_render_file` consults `self.template.git_index_modes` for exec-bit determination
  - New `_sync_git_index_executable_bit` helper, only active on `core.fileMode=false`, uses `--cacheinfo`
- `copier/_template.py`:
  - New `Template.git_index_modes` cached property
- `tests/test_updatediff.py`:
  - 4 parametrized exec-bit tests (addition/removal × `core.fileMode=true/false`), with the `true` case asserting the change is unstaged after update as an auto-staging regression guard
  - Windows-only end-to-end test
  - Merge-conflict regression test

copier/_main.py
```diff
@@ -846,6 +885,93 @@ def _render_file(  # noqa: C901
...
+            return
+        if file_mode_setting != "false":
+            # git will pick up the plain on-disk ``chmod`` from
+            # :meth:`_render_file`; no index manipulation needed.
+            return
+        try:
+            # TODO: simplify with ``--format %(objectmode)`` once the
+            # minimum git version is raised to 2.38+ (--format cannot be
+            # combined with --stage; see git-ls-files(1)).
+            result = git("ls-files", "--stage", "--", str(dst_relpath)).strip()
```
sisp, line 956: What about untracked files that were introduced by the project update? Should their file mode be synced, too? Correct me if I'm wrong, on Windows (or with `core.fileMode=false`) their executable bits would be lost in the project's Git index, wouldn't they?

That said, `git update-index` doesn't work on untracked files. I wonder whether we could run `git add --intent-to-add <untracked file>` on untracked files – perhaps even in general to indicate that a new file introduced by the update algorithm is different from a random untracked file and is intended to be tracked? Then, `git update-index` does work. If you agree, I think this small enhancement of the update algorithm should be made in a dedicated PR (also to have a dedicated changelog entry about a new feature). WDYT?

copier/_template.py
```diff
@@ -579,6 +580,46 @@ def local_abspath(self) -> Path:
...
+            # TODO: simplify with ``--format`` once the minimum git
+            # version is raised to 2.38+ (--format cannot be combined
+            # with --stage; see git-ls-files(1)).
+            raw = git("ls-files", "--stage").strip()
+        except (OSError, ProcessExecutionError):
+            return {}
+        modes: dict[PurePosixPath, int] = {}
+        for line in raw.splitlines():
+            # Format: "<mode> <sha> <stage>\t<path>"
+            meta, _, path = line.partition("\t")
```
sisp: Why not just use

```suggestion
            meta, path = line.split("\t", 1)
```

similar to `Worker._sync_git_index_executable_bit`?

copier/_template.py
```diff
@@ -579,6 +580,46 @@ def local_abspath(self) -> Path:
...
+            if not path:
+                continue
+            mode_str, _, _ = meta.partition(" ")
```
sisp: Perhaps this is a little simpler / more intuitive?

```suggestion
            mode_str = meta.split(" ", 1)[0]
```

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,313 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+        )
+        git_save(tag="v1")
+
+    run_copy(str(src), dst, defaults=True, overwrite=True)
+    with local.cwd(dst):
+        git("init")
+        git("config", "core.fileMode", str(file_mode).lower())
+        git("add", "-A")
+        git("commit", "-m", "init")
+    assert (dst / "launcher.sh").stat().st_mode & 0o111 == 0
```
sisp: I think this assertion is unrelated to the Git repository, I'd move it right below the `run_copy` call. Then, the next `with local.cwd(dst):` line can also be removed because the Git file mode retrieval can come right after `git("commit", ...)`.

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,313 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+
+    # Template v2: launcher.sh gains +x with no content change.
+    with local.cwd(src):
+        Path("launcher.sh").chmod(
+            Path("launcher.sh").stat().st_mode
+            | stat.S_IXUSR
+            | stat.S_IXGRP
+            | stat.S_IXOTH
+        )
+        git("update-index", "--chmod=+x", "launcher.sh")
```
sisp: I think we don't need this index update because this test runs only on Unix. Right?

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,313 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+        mode = git("ls-files", "--stage", "--", "launcher.sh").strip().split()[0]
+    assert mode == "100755"
+
+    # Template v2: drop +x (no content change).
+    with local.cwd(src):
+        Path("launcher.sh").chmod(
+            Path("launcher.sh").stat().st_mode
+            & ~(stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+        )
+        git("update-index", "--chmod=-x", "launcher.sh")
```
sisp: I think we don't need this index update because this test runs only on Unix. Right?

tests/test_updatediff.py
```diff
@@ -2225,3 +2225,313 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+    # --chmod`` were called after stages 1/2/3 were registered, it
+    # would collapse them into stage 0 and the conflict would silently
+    # disappear.
+    assert "<<<<<<< before updating" in (dst / filename).read_text()
+    with local.cwd(dst):
+        lines = git("status", "--porcelain=v1").strip().splitlines()
+        assert any(
+            line.startswith("UU") and normalize_git_path(line[3:]) == filename
+            for line in lines
+        ), f"expected UU conflict for {filename}, got {lines!r}"
```
sisp: Just thinking whether the test would be easier to verify with simple `git status` and `git diff` snapshot assertions

```suggestion
    assert git("-C", dst, "status") == snapshot()
    assert git("-C", dst, "diff") == snapshot()
```

with `from inline_snapshot import snapshot` and the the snapshots added via:

```shell
uv run pytest -n0 --inline-snapshot=review ...
```

WDYT?

sisp (changes requested):
Looks pretty solid, thanks for updating the PR and addressing the problem of updating the index while there are merge conflicts. I've left a few rather minor comments.

willemkokke:
Thanks @sisp! Addressed the minor stuff and deferring the untracked-files case:

**`copier/_template.py`**
- Switched `line.partition("\t")` → `line.split("\t", 1)` and `meta.partition(" ")` → `meta.split(" ", 1)[0]` (consistency with the existing `_sync_git_index_executable_bit` idiom)

**`tests/test_updatediff.py`**
- Moved the on-disk exec-bit assertion above the `local.cwd(dst)` block in `test_update_propagates_executable_bit_addition` and folded the two `local.cwd(dst)` blocks into one
- Removed the redundant `git update-index --chmod` calls in the three template-side test setups — on Unix with `core.fileMode=true` the subsequent `git commit -am` already picks up the `Path.chmod` as a mode change. Kept the one in the removal test's destination setup since it's load-bearing for `core.fileMode=false` (on-disk chmod is invisible to git there).
- Adopted your `inline_snapshot` suggestion for the merge-conflict regression test. Scoped `status`/`diff` to the conflicted filename so unstable tmp paths in the `_src_path` of `.copier-answers.yml` don't leak into the snapshot. Both `core.fileMode` parametrizations produced identical output for the scoped diff, so a single snapshot covers both.

**Untracked files + `core.fileMode=false`**

Agree with your assessment — it's a real gap and `git add --intent-to-add <file>` is a reasonable approach (the intent-to-add marker would be an accurate signal to users that these files aren't random untracked garbage but deliberate template output). Agree this should land in a separate PR so it gets its own changelog entry as a feature. I've opened #2630 to track.

copier/_main.py
willemkokke: Excellent idea. I've verified this works locally. Will follow up with a PR once this has landed, I don't want to deal with the mental overhead of having a PR based on another PR.

tests/test_updatediff.py
```diff
@@ -2225,3 +2226,328 @@ def test_render_copier_conf_as_json_without_circular_reference(
...
+    # Status + diff for the conflicted file. If ``_sync_git_index_executable_bit``
+    # ever regresses into stomping conflict stages, these snapshots will
+    # diverge immediately. Scoped to ``filename`` to avoid unstable
+    # tmp-path strings in the ``.copier-answers.yml`` diff. Both
+    # ``core.fileMode`` parametrizations produce the same output here,
+    # so a single snapshot suffices.
+    with local.cwd(dst):
+        assert git("status", "--porcelain=v1", "--", filename) == snapshot(
+            "UU launcher.sh\n"
+        )
```
sisp, line 2539: Sorry if I was a bit unclear about what I meant here. I was thinking about asserting the snapshot of literally `git status` (without `--porcelain=v1` and not for a specific file) because this is a natural user view that (at least for me) is easy to grasp. WDYT?

willemkokke:
Sorry for the mis-read — switched to raw `git status` (no `--porcelain`, no path scoping). The snapshot is now a natural user view:

```
On branch main
Unmerged paths:
  (use "git restore --staged <file>..." to unstage)
  (use "git add <file>..." to mark resolution)
	both modified:   launcher.sh

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   .copier-answers.yml

no changes added to commit (use "git add" and/or "git commit -a")
```

Verified that `pytest-gitconfig` pins `init.defaultBranch=main`, so `On branch main` is stable across test environments.

Kept `git diff` scoped to the conflicted filename (`git diff -- launcher.sh`) — the unscoped diff includes `.copier-answers.yml`'s `_src_path`, which is a per-run tmp path, and would make the snapshot unstable. Happy to revisit if you'd prefer a different approach there.

sisp:
Mhhh, it looks like `editorconfig-checker` disagrees with the `git diff` and `git status` snapshots' indentation. :see_no_evil: It seems there's no option to disable it around those snapshots. I guess we'll need to revert those assertions, possibly reconsider the checker or how to combine it with snapshots in a follow-up. Sorry for my bad request and the extra effort now. :face_with_peeking_eye:

willemkokke:
I’ll try this once I get home: https://github.com/editorconfig-checker/editorconfig-checker#excluding-blocks

willemkokke:
Fix pushed. Installed `editorconfig-checker` 3.6.1 locally to reproduce — confirmed the same 5 violations CI was seeing (git's output uses 2-space hints, tabs on unmerged paths, 1-space diff context prefix — all conflicting with `.editorconfig`). Wrapped the two snapshots in `# editorconfig-checker-disable` / `# editorconfig-checker-enable` comments, which the tool [supports](https://github.com/editorconfig-checker/editorconfig-checker#excluding-blocks) since v3.1.0. `editorconfig-checker tests/test_updatediff.py` now exits 0, `uv run pre-commit run --all-files` passes editorconfig-checker, merge-conflict tests still pass against the committed snapshots.

sisp (approved):
LGTM! 🎉 Thanks again for discovering and fixing thie bug, @willemkokke! 🙏

PR #2625, opened 2026-04-20
test: add test for additional Jinja2 extensions error

mschoettle:
Pulled out of #2587 since it is unrelated.

Noticed during working on that PR that there is missing coverage for the error when an additional extension can not be found.


sisp (approved):
Perfect, thanks, @mschoettle! 👌

PR #2640, opened 2026-04-26
feat(updating): propagate exec bit on new template files when core.fileMode=false

### LINKED ISSUE #2630: `copier update` loses executable bit on NEW files introduced by the template when `core.fileMode=false`
Follow-up to #2605.

#2605 fixes executable-bit propagation for files that already exist in the destination. But files *introduced* by the template for the first time on a `copier update` run are still affected when `core.fileMode=false`:

1. Template v2 adds a new `script.sh` with `100755` in the index.
2. Destination has `core.fileMode=false` (Windows default, and a common opt-out elsewhere).
3. copier writes the file, `chmod`s it on disk (invisible to git with `fileMode=false`), then calls `_sync_git_index_executable_bit` — which short-circuits because the file is untracked.
4. The user's next `git add script.sh` records `100644` because git ignores on-disk mode with `fileMode=false`. Exec bit is lost.

Suggested approach, from @sisp's review on #2605: run `git add --intent-to-add <file>` for new files introduced by the template. This registers an index entry (without a blob), after which `git update-index --cacheinfo` can set the correct mode. The intent-to-add marker is also a nice semantic signal that these files aren't random untracked content but deliberate template output.

This is a separate PR / changelog entry because it expands copier's update algorithm with a new feature, rather than fixing a regression on existing behavior.

willemkokke:
## Summary

Closes #2630. Follow-up to #2605.

#2605 fixed executable-bit propagation for files that *already exist* in the destination. Files *introduced* by the template for the first time on `copier update` were still affected when `core.fileMode=false`:

1. Template v2 adds a new `script.sh` with `100755` in the index.
2. Destination has `core.fileMode=false` (Windows default).
3. copier writes the file, `chmod`s it on disk (invisible to git with `fileMode=false`), then calls `_sync_git_index_executable_bit` — which short-circuited because the file is untracked.
4. The user's next `git add script.sh` records `100644` because git ignores on-disk mode with `fileMode=false`. Exec bit lost.

### Changes

**`copier/_main.py` — `_sync_git_index_executable_bit`**:

For untracked files that should be executable, run `git add --intent-to-add <path>` to register an empty index entry, then fall through to the existing `--cacheinfo` path to set the mode to `100755`. The user's eventual `git add` replaces the empty blob with the file's actual content while preserving the mode bit. The intent-to-add marker (`A` in `git status`) is also a more accurate signal than `??` for files that are deliberate template output.

- Skipped when the file does not need an executable bit (the user's `git add` would record `100644` either way).
- Skipped when `core.fileMode=true` (git picks up the on-disk chmod naturally — pre-staging would be inconsistent with copier's "leave changes unstaged for user review" behavior).

**`tests/test_updatediff.py`** — 3 new tests:

| Test | What it covers |
| --- | --- |
| `test_update_introduces_new_executable_file[True\|False]` | New +x file in template v2: index records `100755` on `fileMode=false`; file stays untracked on `fileMode=true` |
| `test_update_introduces_new_non_executable_file[True\|False]` | New non-exec file is NOT pre-staged in either fileMode case |
| `test_update_introduces_new_executable_file_then_user_commits` | End-to-end on `core.fileMode=false`: copier update → `git add` → `git commit` → committed tree records `100755` and the file's actual content. The template setup uses `git update-index --chmod=+x` (rather than `Path.chmod`) so the test runs identically on every platform — Linux, macOS, *and* Windows. |

## Test plan

- [x] All new tests pass on macOS and windows
- [x] Full test suite still green on macOS (1073 passed)
- [x] ruff check, ruff format, mypy, editorconfig-checker all clean

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -895,7 +895,49 @@ class Worker:
                         f"{stat.filemode(dst_mode)} to {stat.filemode(src_mode)}",
                         stacklevel=2,
                     )
+            self._register_in_git_index(dst_relpath)
             self._sync_git_index_executable_bit(dst_relpath, src_mode)
+
+    def _register_in_git_index(self, dst_relpath: Path) -> None:
+        """Register a freshly-rendered file in the destination's git index.
+
+        Runs ``git add --intent-to-add`` on the file when the destination
+        is a git repository and the file is not yet tracked, regardless
+        of the executable bit or ``core.fileMode``. This makes the file
+        appear in ``git status`` as ``A`` (added with intent) rather
+        than ``??`` (untracked), accurately reflecting that the file is
+        deliberate template output rather than random user-created
+        content. It also gives :meth:`_sync_git_index_executable_bit`
+        an existing index entry whose mode it can rewrite via
+        ``update-index --cacheinfo`` on filesystems that don't preserve
+        executable bits.
+
+        No-op when the destination is not a git repository, when the
+        file is already tracked, or when git is unavailable. All git
+        failures are swallowed so copying or updating cannot be broken
+        by an unrelated git problem — most notably, ``git add
+        --intent-to-add`` fails on gitignored files, and that's the
+        user's explicit signal not to track them.
+        """
+        subproject_root = self.subproject.local_abspath
+        git = get_git(context_dir=subproject_root)
+        try:
+            # TODO: simplify with ``--format %(objectmode)`` once the
+            # minimum git version is raised to 2.38+ (--format cannot be
+            # combined with --stage; see git-ls-files(1)).
+            result = git("ls-files", "--stage", "--", str(dst_relpath)).strip()
+        except ProcessExecutionError:
+            # Not a git repo, or git can't read the index.
+            return
+        if result:
+            # Already tracked; nothing to do.
+            return
+        try:
+            git("add", "--intent-to-add", "--", str(dst_relpath))
+        except (OSError, ProcessExecutionError):
+            # Most likely the file is gitignored (or git is unavailable);
+            # silently fall back so we never break the render path.
+            pass
 
     def _sync_git_index_executable_bit(self, dst_relpath: Path, src_mode: int) -> None:
         """Propagate executable-bit changes to the destination's git index.
@@ -914,9 +956,10 @@ class Worker:
         behavior of leaving rendered changes unstaged for user review.
 
         Also a no-op when the destination is not in a git repository, when
-        the file is not yet tracked (the user's eventual ``git add`` will
-        record the on-disk mode where the platform allows it), or when git
-        is unavailable. All git failures are swallowed so that copying or
+        the file is not in the index (a previous call to
+        :meth:`_register_in_git_index` would normally have intent-to-add'd
+        it, but a gitignored file may not have an entry), or when git is
+        unavailable. All git failures are swallowed so that copying or
         updating cannot be broken by an unrelated git problem.
 
         .. note::
@@ -958,7 +1001,7 @@ class Worker:
             # git can't read the index — fall back to a silent no-op.
             return
         if not result:
-            # File is not tracked yet; nothing to update.
+            # File isn't in the index (e.g. gitignored). Nothing to update.
             return
         # Format: "<mode> <sha> <stage>\t<path>"
         meta = result.split("\t", 1)[0].split()
--- a/docs/updating.md
+++ b/docs/updating.md
@@ -97,6 +97,28 @@ repos:
     you can add both hooks to your `pre-commit-config.yaml` file, making sure that no
     unresolved merge conflicts are committed.
 
+## How Copier registers new files in your Git index
+
+When `copier update` adds files that didn't exist in your project before — say, the
+template gained a new helper script in its latest version — Copier marks them in your
+Git index with
+[`git add --intent-to-add`](https://git-scm.com/docs/git-add#Documentation/git-add.txt---intent-to-add).
+They show up in `git status` as `A` (added with intent) instead of `??` (untracked),
+signalling that they're deliberate template output and not something you happened to
+drop in the directory yourself.
+
+You still need to run `git add <file>` to actually stage the file's content
+for committing.
+
+There's a second reason: on filesystems that don't preserve the executable bit
+on disk, most notably Windows (where `core.fileMode` is `false` by default),
+having an existing index entry lets Copier record the file's mode as `100755`
+directly, so the executable bit isn't silently dropped on your next `git add`.
+
+If you change your mind, [aborting an update](#aborting-an-update) still works
+the same way: `git reset` removes the intent-to-add markers, then `git clean`
+picks up the new files like any other untracked content.
+
 ## Never change the answers file manually
 
 !!! important
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -2571,3 +2571,246 @@ index f163f4b,d20125d..0000000
 ++>>>>>>> after updating
 """)
         # editorconfig-checker-enable
+
+
+@pytest.mark.skipif(
+    condition=platform.system() == "Windows",
+    reason="Windows does not have UNIX-like permissions",
+)
+@pytest.mark.parametrize("file_mode", [True, False])
+def test_update_introduces_new_executable_file(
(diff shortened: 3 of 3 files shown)


sisp:
Correct me if I'm wrong, but doesn't this PR only mark files added by the update algorithm as intended to be added if `core.fileMode=false`. My original intent was to mark all files added by an update as intended to be added, regardless of the file mode and changes in the executable bit.

willemkokke:
Ah I misunderstood then, I thought we wanted to deviate as little as possible from existing behaviour. Happy to adjust, brb!

willemkokke:
I've adjusted the PR to always intent-to-add new files:

**`copier/_main.py`** — split into two helpers:

- New `_register_in_git_index`: runs `git add --intent-to-add` on any newly-rendered untracked file, regardless of `core.fileMode` or executable bit. Silently no-ops if the file is gitignored, the destination isn't a git repo, or git is unavailable.
- `_sync_git_index_executable_bit`: simplified — it now just rewrites the index entry's mode via `--cacheinfo` when `core.fileMode=false`, relying on the entry that `_register_in_git_index` left behind. The intent-to-add fallback was removed.

**Tests** — updated to match:

- `test_update_introduces_new_executable_file[True\|False]` now asserts both parametrizations land at `100755` in the index. The two cases exercise different paths to get there: with `fileMode=true` the intent-to-add call reads the on-disk chmod directly; with `fileMode=false` Copier additionally rewrites the entry's mode (the on-disk chmod is invisible there).
- `test_update_introduces_new_non_executable_file[True\|False]` now asserts the file is intent-to-add'd at `100644` (previously asserted untracked).
- New `test_update_introduces_new_file_can_be_aborted` regression-tests the [aborting an update](https://copier.readthedocs.io/en/latest/updating/#aborting-an-update) flow: `git reset` drops the intent-to-add markers so `git clean` picks the files up like any other untracked content. There were no existing tests for the abort flow at all, which seemed worth fixing while we're here.

**`docs/updating.md`** — new section "How Copier registers new files in your Git index" explaining the user-visible behavior change (`A` instead of `??` in `git status`), why it matters on Windows/`core.fileMode=false`, and that the abort flow still works the same way.

Verified locally: 1077 passed, 5 skipped, 8 xfailed, 1 xpassed on macOS. ruff/mypy/editorconfig-checker all clean.

copier/_main.py
```diff
@@ -895,8 +895,50 @@ def _render_file(  # noqa: C901
...
+        --intent-to-add`` fails on gitignored files, and that's the
+        user's explicit signal not to track them.
+        """
+        subproject_root = self.subproject.local_abspath
+        git = get_git(context_dir=subproject_root)
+        try:
+            # TODO: simplify with ``--format %(objectmode)`` once the
+            # minimum git version is raised to 2.38+ (--format cannot be
+            # combined with --stage; see git-ls-files(1)).
+            result = git("ls-files", "--stage", "--", str(dst_relpath)).strip()
```
sisp, line 928: I think this check is not sufficient because it also passes for files with merge conflicts which shouldn't be marked with "intent-to-add" because it changes the state. Perhaps we can check the status with `git status --porcelain=v1 <dst_relpath>` and only proceed for files without merge conflicts.

sisp (commented):
In addition to my inline comment, I've been giving this PR some more thought. Once #2376 lands, the update algorithm will use `git merge` instead of a complex sequence of low-level `git` commands. `git merge` automatically tracks new files (added in the source branch), so marking new files with "intent-to-add" would be inconsistent with the future (and arguably more correct) behavior. This means, auto-staging will naturally occur even when `core.fileMode=true` (we currently assert files to remain unstaged)

https://github.com/copier-org/copier/blob/c4310ac6b553bb673090f5268116bfd85f78669d/tests/test_updatediff.py#L2293-L2300

so the future behavior will be fully consistent for both `core.fileMode=true` and `core.fileMode=false`.

However, the `Worker._render_file` method is currently not aware of the operation mode, so staging (with or without "intent-to-add") applies to both initial copying and updating. I wonder what the best behavior in anticipation of #2376 is. 🤔

And another note that I had missed when reviewing #2605: We need to support `copier copy` even when `git` is not installed. I've restored this regression with #2651.

copier/_main.py
willemkokke: I think the existing check already excludes conflicted files, so I'd argue the porcelain check is redundant — but happy to add it as belt-and-suspenders if you prefer.

Conflicted files are tracked: they have stages 1, 2, and 3 in the index. `git ls-files --stage <path>` returns three lines for them (one per stage), so `result` is non-empty and the early return at line 932 triggers. The cases are:

- **Tracked, clean**: `100644 abc 0\tfile` (1 line) → return
- **Tracked, conflicted**: stages 1/2/3 (3 lines) → return
- **Untracked**: empty → proceed to intent-to-add

The function is called from `_render_file` for every rendered file, but only untracked files reach the `git add --intent-to-add` line.

As a backstop, `git add --intent-to-add` itself refuses to run on conflicted files (`fatal: cannot add the path '<file>' because of a merge conflict`), and our `try/except (OSError, ProcessExecutionError)` would swallow that.

That said: under the current update flow, I don't think `_register_in_git_index` can even be *reached* with a conflicted file. Conflicts get registered at L1487-L1556 of `_apply_update`, which runs *after* `current_worker.run_copy()` (and thus after `_render_file` for every file). So at the time our function runs, the index has only stage 0 entries.

Let me know if there's a scenario I'm missing — happy to swap in `git status --porcelain=v1` if you'd prefer.

willemkokke:
#2376 looks great — leveraging `git merge` directly is much cleaner than the low-level `git ls-files` / `update-index` plumbing we're adding here.

I had a look and ported the exec-bit tests from #2605 and this PR onto `refactor/new-update-algorithm` to see what holds up. **Most of them pass cleanly** — the `git merge`-based flow handles exec-bit propagation natively (correct mode preserved through git's trees regardless of `core.fileMode`), and new files get auto-staged with real content. That's strictly better than the intent-to-add empty-blob marker we're adding here.

Concrete results running the tests on `refactor/new-update-algorithm`:

- `executable_bit_addition[False]` and `executable_bit_removal[False]` (the core `core.fileMode=false` propagation cases): ✅ pass
- All four `introduces_new_*` tests: ✅ pass
- `introduces_new_executable_file_then_user_commits` (end-to-end): ✅ pass
- `introduces_new_file_can_be_aborted` (the abort flow): ✅ pass
- `executable_bit_addition[True]` / `executable_bit_removal[True]` fail only because they assert "remain unstaged" — no longer true under #2376's auto-staging. Behavior is correct, assertion is stale.
- The merge-conflict tests fail purely on the `run_update(conflict=...)` API removal.

So you're right: the `git_index_modes` / `_sync_git_index_executable_bit` / `_register_in_git_index` machinery from #2605 + this PR becomes redundant under #2376. Once #2376 lands, all of it can be ripped out.

Happy to open a follow-up PR against `refactor/new-update-algorithm` that ports the exec-bit propagation and new-file behavior tests (with the stale assertions updated for the auto-staging semantics, and the conflict-test API calls adjusted) — most of the coverage carries over with minor changes.

One observation while I was poking around: `refactor/new-update-algorithm` is currently ~78 commits behind master (last touched 2026-03-16) and would have a non-trivial rebase against #2605 and a few other shipped fixes (#2651, the `git clone` cleanup, the PathSpec backslash fix, etc.). Is there anything blocking it, or anything I can help with to get it landed? I'm happy to take on rebase work, conflict resolution, or specific items if you have a list of what's left.

sisp:
It turns out that the test suite (specifically `tests/test_updatediff.py::test_update_propagates_executable_bit_removal[False]`) of #2541 is failing because the internal fresh copies of the old and new template versions are Git repos there and

https://github.com/copier-org/copier/blob/b80196e5e075eaa1810323b1dc5836f9698d418b/copier/_main.py#L959

returns an empty string because none of the files are staged in those empty repos. This leads to incorrect permission updates:

https://github.com/copier-org/copier/blob/b80196e5e075eaa1810323b1dc5836f9698d418b/copier/_main.py#L1621-L1646

Staging (not just as intent-to-add) all rendered files would fix the failing test. It would also imitate a regular Git 3-way merge better as we discussed before. What do you think about staging all rendered files (if the destination is a Git repo)? This would also propagate the exec bit on new template files as well as on newly generated projects whose destination is already a Git repo.

PR #2662, opened 2026-05-10
fix: better error on invalid choice

### LINKED ISSUE #2126: "ValueError: Invalid Choice" with neither property name nor value
### Describe the problem

During `copier update` I get `ValueError: Invalid Choice` but there is no mention of the property name or the invalid value.
This is caused an earlier update in copier.yaml making difficult to debug (i.e. I don't know which change is causing this.)
Issue may be related to multi select choice, I will update if I figure it out.

### Template

Sorry, cannot provide the template.

### To Reproduce

_No response_

### Logs

```console

```

### Expected behavior

The exception message includes property name, valid choices and bad choice causing the error.

_No response_

### Operating system

Linux

Fedora 40

### Copier version

9.7.1

### Python version

3.12.9

### Installation method

uvx+pypi

### Additional context

Install is `uv venv ...; uv pip ...`

obestwalter:
Add var_name and valid choices to the error message, so the user does not have to guess.

fixes #2126

### DIFF
--- a/copier/_user_data.py
+++ b/copier/_user_data.py
@@ -520,15 +520,18 @@ class Question:
         if not choices:
             return ans
         choice_error = ""
+        valid = []
         for choice in choices:
-            if ans == self.cast_answer(choice.value):
+            cast_ans = self.cast_answer(choice.value)
+            if not choice.disabled:
+                valid.append(cast_ans)
+            if ans == cast_ans:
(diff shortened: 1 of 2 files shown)


sisp (approved):
Fantastic! 🎉 Thank you, @obestwalter! :pray:

PR #2668, opened 2026-05-14
fix(update): don't skip gitignored .rej files for inline-conflict resolution

rcoup:
When a downstream project gitignores `*.rej`, `copier update --conflict=inline` silently skips the inline-merge step for affected files. The `.rej` files survive in the working tree, the original target files don't get the inline conflict markers, and nothing is reported as conflicted.

## Fix

`Worker._apply_update` ([_main.py:1532](https://github.com/copier-org/copier/blob/master/copier/_main.py#L1532)) discovers `.rej` files via `git status --porcelain`, which respects `.gitignore` and omits ignored paths.

The same function at the line above already uses `git status --ignored --porcelain` to enumerate ignored files for `--exclude` patterns to `git apply`. The fix brings the two invocations into alignment:

- Add `--ignored` to the `git status --porcelain` call.
- Accept both `?? ` (untracked) and `!! ` (ignored) prefixes.

The existing `endswith(".rej")` filter scopes processing to rejection witnesses regardless of tracked/untracked/ignored status, so behaviour for non-rej ignored files is unchanged.

## Tests

Added a `gitignore_rej` parameter to `test_conflicted_files_are_marked_unmerged` which adds a destination `.gitignore` containing `*.rej`. Asserts the same post-conditions: inline markers in the file, no `.rej` left over, file reported as unmerged in the index.

Without the fix, the new `gitignore_rej=True` parametrizations fail (no markers, `.rej` survives). With the fix, all parametrizations pass.

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1529,12 +1529,16 @@ class Worker:
                     conflicted = []
                     old_path = Path(old_copy)
                     new_path = Path(new_copy)
-                    status = git("status", "--porcelain").strip().splitlines()
+                    # `--ignored` so we still find .rej files when the
+                    # destination has a `*.rej` ignore rule.
+                    status = (
+                        git("status", "--porcelain", "--ignored").strip().splitlines()
+                    )
                     for line in status:
                         # Filter merge rejections (part 1/2)
-                        if not line.startswith("?? "):
+                        if not line.startswith(("?? ", "!! ")):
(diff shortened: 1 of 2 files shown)


sisp (approved):
LGTM! 🎉  Thanks for discovering this subtle bug and fixing it! 🏆

PR #2679, opened 2026-05-19
fix(copy): preserve answers file subdirectories

### LINKED ISSUE #1257: Answers file setting / CLI switch behavior inconsistent with docs
### Describe the problem

The [answers file setting / CLI switch](https://copier.readthedocs.io/en/v8.1.0/configuring/#answers_file) behavior is inconsistent with what the docs suggest. The docs state that this setting is a path relative to the destination path, but the subdirectory in a path to the answers file is ignored by the rendering and only the file _name_ is used.

### Template

```
📁 src/
├── 📄 {{ _copier_conf.answers_file }}.jinja
└── 📄 copier.yml
```

`copier.yml` content:

```yaml
_answers_file: subdir/.custom-answers.yml
```

### To Reproduce

```console
$ copier copy src/ dst
No git tags found in template; using HEAD as ref

Copying from template version 0.0.0.post1.dev0+a4f2b63
    create  .
    create  .custom-answers.yml
```

### Logs

```console
n/a
```

### Expected behavior

IIRC, changing not only the file _name_ but also its _location_ via the answers file setting / CLI switch used to work some time ago, but I think #927 made it impossible. Should it be possible to override the _location_  of the answers file or just its _name_ when generating a new project?

_No response_

### Operating system

Linux

Ubuntu 20.04

### Copier version

8.1.0

### Python version

CPython 3.9

### Installation method

pipx+pypi

### Additional context

_No response_

gcomneno:
Fixes #1257.

When `_answers_file` was configured with a relative subdirectory, for example:

    _answers_file: subdir/.custom-answers.yml

Copier flattened the rendered answers file path and created:

    .custom-answers.yml

instead of:

    subdir/.custom-answers.yml

This change preserves the configured relative path while still avoiding duplicated path prefixes when the answers file template is already located inside that same directory.

It also keeps dynamic file path rendering working by clearing the yield context after path-part expansion.

Added regression coverage to ensure the answers file is created in the configured subdirectory.

Tests run:

- `pytest -q tests/test_answersfile.py -k "answers_file_keeps_configured_subdirectory" -rs`
- `pytest -q tests/test_cli.py -k "good_cli_run_dot_config" -rs`
- `pytest -q tests/test_dynamic_file_structures.py -k "file_loop" -rs`
- `LC_ALL=C pytest -q tests/test_answersfile.py tests/test_answersfile_templating.py tests/test_subdirectory.py tests/test_copy.py tests/test_updatediff.py tests/test_cli.py tests/test_dynamic_file_structures.py -rs`

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1040,19 +1040,22 @@ class Worker:
             dst_abspath = self.subproject.local_abspath / dst_relpath
             dst_abspath.mkdir(parents=True, exist_ok=True)
 
-    def _adjust_rendered_part(self, rendered_part: str) -> str:
-        """Adjust the rendered part if necessary.
+    def _adjust_rendered_part(
+        self, rendered_parts: tuple[str, ...], rendered_part: str
+    ) -> str:
+        """Avoid duplicating already rendered path prefixes.
 
-        If `{{ _copier_conf.answers_file }}` becomes the full path,
-        restore part to be just the end leaf.
-
-        Args:
-            rendered_part:
-                The rendered part of the path to adjust.
-
+        If a path part renders to the full answers file path while the current
+        rendered path already contains its parent directory, keep only the leaf
+        name. Otherwise, preserve the rendered part as-is.
         """
-        if str(self.answers_relpath) == rendered_part:
-            return Path(rendered_part).name
+        rendered_path = Path(rendered_part)
+        current_path = Path(*rendered_parts) if rendered_parts else Path()
+        if (
+            rendered_path == self.answers_relpath
+            and rendered_path.parent == current_path
+        ):
+            return rendered_path.name
         return rendered_part
 
     def _render_parts(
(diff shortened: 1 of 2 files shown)


sisp:
Good catch! :+1: This is in part related to #2186, just mentioning it because it might be touching related code blocks.

gcomneno:
Thanks for the pointer, @sisp.

I agree this touches the same answers-file path handling area as #2186. I tried to keep this PR narrow: preserve the configured relative `_answers_file` path, while still keeping the existing non-root answers-file-template behavior covered by the `good_cli_run_dot_config` tests.

I also added the yield-context cleanup after noticing the first CI run broke dynamic file path rendering, and covered that with `test_file_loop`.

Happy to adjust if you think this should be aligned differently with the direction from #2186.

sisp:
I've just merged #2186, it turns out the bug fixed by this PR was also fixed by d76195a649eee1daff477c9a8f7312ce10cef300 and is covered by the added test case `test_answersfile_with_custom_path`. I've also confirmed this by running the test case you've added here on the new tip of the `master` branch.

Thanks for your initiative and effort nevertheless, much appreciated. :pray:

gcomneno:
Thanks for checking and for confirming the test case against the new `master`, @sisp.

That makes perfect sense if #2186 now covers the same behavior. I’m glad the investigation was at least useful to validate the issue.

Thanks again for the clear explanation.

PR #2683, opened 2026-05-21
fix(update): preserve executable bit for new files

### LINKED ISSUE #2630: `copier update` loses executable bit on NEW files introduced by the template when `core.fileMode=false`

gcomneno:
Fixes #2630.

When a template update introduces a new executable file in a destination repository with `core.fileMode=false`, Git ignores the on-disk executable bit. As a result, the user's later `git add` records the new file as `100644` instead of `100755`.

This change handles new executable files in the same executable-bit sync path used for existing files:

- if the destination has `core.fileMode=false`;
- and the rendered file is executable;
- and the file is not tracked yet;

Copier first creates an empty intent-to-add index entry, then updates only that entry's mode via `git update-index --cacheinfo`. This prepares the executable bit without staging the rendered file contents.

Added regression coverage for a template update that introduces a new executable file, verifying that:

- the file is executable on disk after update;
- the intermediate index entry has mode `100755`;
- the rendered file contents are not staged;
- a later user `git add` records the file as `100755`.

Also added coverage for new non-executable files with `core.fileMode=false`, ensuring Copier leaves them untracked and does not create an unnecessary intent-to-add index entry.

Tests run:

- `pytest -q tests/test_updatediff.py -k "new_files_with_filemode_false or new_non_executable_files" -rs`
- `LC_ALL=C pytest -q tests/test_updatediff.py -k "executable_bit or exec_bit or new_non_executable_files" -rs`
- `LC_ALL=C pytest -q tests/test_updatediff.py -rs`
- `LC_ALL=C pytest -q tests/test_updatediff.py tests/test_copy.py -rs`
- `ruff format --check copier/_main.py tests/test_updatediff.py`
- `ruff check copier/_main.py tests/test_updatediff.py`

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -915,11 +915,12 @@ class Worker:
         as an *unstaged* modification, which matches copier's normal
         behavior of leaving rendered changes unstaged for user review.
 
-        Also a no-op when the destination is not in a git repository, when
-        the file is not yet tracked (the user's eventual ``git add`` will
-        record the on-disk mode where the platform allows it), or when git
-        is unavailable. All git failures are swallowed so that copying or
-        updating cannot be broken by an unrelated git problem.
+        Also a no-op when the destination is not in a git repository, or
+        when git is unavailable. New executable files are first registered
+        with ``git add --intent-to-add`` so an empty index entry exists; then
+        only that entry's executable mode is prepared, without staging the
+        rendered file contents. All git failures are swallowed so that copying
+        or updating cannot be broken by an unrelated git problem.
 
         .. note::
 
@@ -951,6 +952,7 @@ class Worker:
             # git will pick up the plain on-disk ``chmod`` from
             # :meth:`_render_file`; no index manipulation needed.
             return
+        desired_executable = bool(src_mode & 0o111)
         try:
             # TODO: simplify with ``--format %(objectmode)`` once the
             # minimum git version is raised to 2.38+ (--format cannot be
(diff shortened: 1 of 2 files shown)


sisp:
Isn't this the same as #2640? :thinking: (I haven't made up my mind about the best path forward yet because #2376 will resolve the same problem more fundamentally.)

gcomneno:
Thanks for pointing that out, @sisp.

Yes, after looking at #2640, this is essentially the same direction: create an intent-to-add entry for new executable files with `core.fileMode=false`, then preserve the executable bit through the later user `git add`.

I started from #2630 and missed the existing PR, sorry for the duplicate work. #2640 is older and more complete, so I’m happy to close this PR to avoid adding noise.

If useful, feel free to reuse anything from the additional non-executable-file regression test here, but otherwise I’ll consider this superseded by #2640 / the larger #2376 direction.

willemkokke:
Haha, just checking in on my PR (#2640) and I got really confused for a minute ;) Great minds think alike and all that! 

#2376 is definitely the better solution

gcomneno:
Thanks for checking and confirming! Since #2376 is the better solution, I'm closing this PR to avoid duplicate work.

PR #2686, opened 2026-05-22
fix(copy): report template path for Jinja errors

### LINKED ISSUE #2267: Jinja error without clear filename
### Describe the problem

I run copier and get a stacktrace without filename. That super hard to debug. 

### Template

just an empty dir (you can add files, but this is minimal): 

```
mkdir -p 'my-template/some-{{{{ jinja_error }}}}'

```

### To Reproduce

```
mkdir -p 'my-template/some-{{{{ jinja_error }}}}'

# note we have no copier file, so by default it assumes "{{" escape pattern

copier copy my-template .
```

### Logs

```console
copier copy template generated --trust

  File "/opt/homebrew/bin/copier", line 8, in <module>
    sys.exit(CopierApp.run())
             ^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/plumbum/cli/application.py", line 640, in run
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/plumbum/cli/application.py", line 635, in run
              ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_cli.py", line 284, in main
    return _handle_exceptions(inner)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_cli.py", line 71, in _handle_exceptions
    method()
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_cli.py", line 275, in inner
    with self._worker(
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_main.py", line 271, in __exit__
    raise value
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_cli.py", line 282, in inner
    worker.run_copy()
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_main.py", line 95, in _wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_main.py", line 1039, in run_copy
    self._render_template()
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_main.py", line 717, in _render_template
    for dst_relpath, ctx in dst_relpaths_ctxs:
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_main.py", line 931, in _render_path
    yield from self._render_parts(relpath.parts, is_template=is_template)
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_main.py", line 885, in _render_parts
    rendered_part = self._render_string(part, extra_context=extra_context)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/copier/_main.py", line 945, in _render_string
    tpl = self.jinja_env.from_string(string)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/jinja2/environment.py", line 1111, in from_string
    return cls.from_code(self, self.compile(source), gs, None)
                               ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/jinja2/environment.py", line 771, in compile
    self.handle_exception(source=source_hint)
  File "/opt/homebrew/lib/python3.11/site-packages/jinja2/environment.py", line 942, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<unknown>", line 1, in template
jinja2.exceptions.TemplateSyntaxError: expected token ':', got '}'
```

### Expected behavior

I would get a clear filename/directory name that cannot be templated. The stack trace is ugly already, but `File "<unknown>", line 1, in template` is hard. 

_No response_

### Operating system

macOS

MacOS

### Copier version

copier 9.9.0

### Python version

python3.13

### Installation method

uvx+pypi

### Additional context

_No response_

gcomneno:
Fixes #2267.

When a template path contains invalid Jinja syntax, Jinja raises a `TemplateSyntaxError` with `<unknown>` as the filename. This makes it difficult to identify which template path failed, especially in larger templates.

This change wraps Jinja errors raised while rendering template path parts and re-raises them as `UserMessageError` with the original template path included in the message.

Added regression coverage for an invalid Jinja expression in a rendered path.

Tests run:

- `pytest -q tests/test_copy.py -k "invalid_jinja_in_rendered_path" -rs`
- `pytest -q tests/test_copy.py -k "render_path or invalid_jinja_in_rendered_path" -rs`
- `LC_ALL=C pytest -q tests/test_copy.py -rs`
- `pytest -q tests/test_dynamic_file_structures.py -rs`
- `pytest -q tests/test_jinja2_undefined.py -rs`
- `pytest -q tests/test_context.py -k "copier_operation or exclude or skip" -rs`
- `ruff format --check copier/_main.py tests/test_copy.py`
- `ruff check copier/_main.py tests/test_copy.py`

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -30,6 +30,7 @@ from typing import (
 )
 from unicodedata import normalize
 
+from jinja2.exceptions import TemplateError
 from jinja2.loaders import FileSystemLoader
 from jinja2.sandbox import SandboxedEnvironment
 from jinja2.utils import import_string
@@ -1047,6 +1048,7 @@ class Worker:
         rendered_parts: tuple[str, ...] | None = None,
         extra_context: AnyByStrDict | None = None,
         is_template: bool = False,
+        source_parts: tuple[str, ...] = (),
     ) -> Iterable[tuple[Path, AnyByStrDict | None]]:
         """Render a set of parts into path and context pairs.
 
@@ -1070,20 +1072,32 @@ class Worker:
 
         part = parts[0]
         parts = parts[1:]
+        source_parts = source_parts + (part,)
+        source_path = Path(*source_parts)
 
         if not extra_context:
             extra_context = {}
 
         # If the `part` has a yield tag, `self.jinja_env` will be set with the yield
         # name and iterable
-        rendered_part = self._render_string(part, extra_context=extra_context)
+        try:
+            rendered_part = self._render_string(part, extra_context=extra_context)
+        except TemplateError as error:
+            raise UserMessageError(
+                f"Error rendering template path {source_path}: {error}"
+            ) from error
 
         ctx = get_yield_context(self.jinja_env)
         yield_name = ctx.yield_name
         if yield_name:
             for value in ctx.yield_iterable or ():
                 new_context = {**extra_context, yield_name: value}
-                rendered_part = self._render_string(part, extra_context=new_context)
+                try:
+                    rendered_part = self._render_string(part, extra_context=new_context)
+                except TemplateError as error:
+                    raise UserMessageError(
+                        f"Error rendering template path {source_path}: {error}"
+                    ) from error
                 if str(self.answers_relpath) == rendered_part:
                     if rendered_parts:
                         deprecate_answers_file_template_path()
@@ -1095,7 +1109,11 @@ class Worker:
                     continue
 
                 yield from self._render_parts(
-                    parts, rendered_parts + (rendered_part,), new_context, is_template
+                    parts,
+                    rendered_parts + (rendered_part,),
+                    new_context,
+                    is_template,
+                    source_parts,
                 )
 
             return
@@ -1111,7 +1129,11 @@ class Worker:
             return
 
         yield from self._render_parts(
-            parts, rendered_parts + (rendered_part,), extra_context, is_template
+            parts,
+            rendered_parts + (rendered_part,),
+            extra_context,
+            is_template,
+            source_parts,
         )
 
     def _render_path(self, relpath: Path) -> Iterable[tuple[Path, AnyByStrDict | None]]:
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -22,7 +22,7 @@ from plumbum import CommandNotFound, local
 import copier
 from copier import run_copy
 from copier._types import AnyByStrDict
-from copier.errors import ForbiddenPathError
+from copier.errors import ForbiddenPathError, UserMessageError
 
 from .helpers import (
     BRACKET_ENVOPS,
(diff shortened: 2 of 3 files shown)


copier/_main.py
```diff
@@ -1041,9 +1042,24 @@ def _render_folder(self, dst_relpath: Path) -> None:
...
+            return self._render_string(part, extra_context=extra_context)
+        except TemplateError as error:
+            raise UserMessageError(
+                f"Error rendering template path {source_relpath}: {error}"
+            ) from error
+
     def _render_parts(  # noqa: C901
         self,
         parts: tuple[str, ...],
+        source_relpath: Path,
```
sisp: Can we avoid this argument by constructing the path from `parts` again when reporting the render error?

sisp (commented):
Thanks for contributing a fix to this long-standing problem, @gcomneno! :bow:

I've left a minor inline question/comment.

In addition, I think errors raised while attempting to render template _file content_ don't include the file path either. This might be worth looking into as well (also for error consistency) – perhaps in a separate PR.

gcomneno:
Thanks for the review!

I updated the PR to avoid passing the extra `source_relpath` argument through `_render_parts`. The `TemplateError` wrapping now happens in `_render_path()`, where the original `relpath` is already available.

I agree that adding similar path context for errors while rendering file contents would be useful for consistency. I kept that out of this PR and can look into it separately.

copier/_main.py
```diff
@@ -1095,7 +1096,10 @@ def _render_parts(  # noqa: C901
                     continue
 
                 yield from self._render_parts(
-                    parts, rendered_parts + (rendered_part,), new_context, is_template
+                    parts,
+                    rendered_parts + (rendered_part,),
+                    new_context,
+                    is_template,
```
sisp, line 1115: This is just diff noise, there's no actual change except the trailing comma which leads to the reformatting. Could you please revert this snippet?

copier/_main.py
```diff
@@ -1111,7 +1115,10 @@ def _render_parts(  # noqa: C901
             return
 
         yield from self._render_parts(
-            parts, rendered_parts + (rendered_part,), extra_context, is_template
+            parts,
+            rendered_parts + (rendered_part,),
+            extra_context,
+            is_template,
```
sisp, line 1135: This is just diff noise, there's no actual change except the trailing comma which leads to the reformatting. Could you please revert this snippet?

copier/_main.py
```diff
@@ -1131,7 +1139,12 @@ def _render_path(self, relpath: Path) -> Iterable[tuple[Path, AnyByStrDict | Non
         if self.template.templates_suffix and is_template:
             relpath = relpath.with_suffix("")
 
-        yield from self._render_parts(relpath.parts, is_template=is_template)
+        try:
+            yield from self._render_parts(relpath.parts, is_template=is_template)
+        except TemplateError as error:
+            raise UserMessageError(
+                f"Error rendering template path {source_relpath}: {error}"
+            ) from error
```
sisp: I think catching the `TemplateError` exception here is the wrong place because the error message for an invalid path part in the middle shows the full path, not the path up to the invalid part. Instead, I think we should raise in the `Worker._render_parts` method. Could you please add a test case to cover this scenario, too?

sisp (changes requested):
Thanks for the update. I've left a few new remarks/requests.

sisp:
Hmm, I just realized that `parts` passed to `Worker._render_parts` is the tuple of remaining path parts, not the tuple of already rendered parts. In this case, we may need an additional argument holding the current source path parts for accurate error reporting. 🤔 Sorry for the inaccurate review earlier.

gcomneno:
Thanks for the follow-up review!

I updated the implementation to catch `TemplateError` inside `_render_parts()` again, while carrying the original source path parts through the recursion. This lets the error report the path up to the failing rendered part instead of the whole source path.

I also added a regression test for a nested path where an intermediate path part has invalid Jinja syntax, checking that the reported path includes the failing part but not the remaining child path.

gcomneno:
I added one more regression test covering the `yield` path-part branch as well: the first render succeeds, then the yielded render raises a `TemplateError`, which is wrapped with the failing template path context.

copier/_main.py
```diff
@@ -1070,20 +1072,32 @@ def _render_parts(  # noqa: C901
 
         part = parts[0]
         parts = parts[1:]
+        source_part_path = Path(*source_parts, part)
+        source_parts = source_parts + (part,)
```
sisp: Just some nitpicking 🫣:

```suggestion
        source_parts = source_parts + (part,)
        source_path = Path(*source_parts)
```

sisp (changes requested):
Just some a polishing request, otherwise great work and ready for merge. 👌

gcomneno:
Thanks! Applied the suggested polishing change.

sisp (approved):
LGTM! :tada: Thanks again for contributing a fix for this problem! :bow:

PR #2687, opened 2026-05-22
docs: clarify validator helper filters

### LINKED ISSUE #1529: Documentation is unclear to available validator methods
### Describe the problem

In this documentation example, a function(?) called `regex_search` is used:

```
 validator: >-
        {% if not (project_name | regex_search('^[a-z][a-z0-9\-]+$')) %}
        project_name must start with a letter, followed one or more letters, digits or dashes all lowercase.
        {% endif %}
```

There is no further mention of this in the documentation or any additional validator methods that are discussed.  It is also not part of the jinja2 docs.

### Template

NA

### To Reproduce

_No response_

### Logs

_No response_

### Expected behavior

Please include available validation commands in the documentation.

_No response_

### Operating system

Linux

NA

### Copier version

9.1.1

### Python version

NA

### Installation method

pipx+pypi

### Additional context

I am trying to find if it is possible to validate input based on existing file names or bash command outputs (for instance trying to validate that a response to a question is unique based on `grep` output in bash... regex seems to exist with `regex_search` but no further validation methods are documented).

gcomneno:
Fixes #1529.

Clarifies that question validators are rendered as Jinja templates and can use:

- built-in Jinja filters, tests, and functions;
- Copier helpers such as `pathjoin`;
- filters from `jinja2-ansible-filters`, including `regex_search`.

Also clarifies that `regex_search` is a Jinja filter, not a dedicated validation method, and points users to `_jinja_extensions` for project-specific validation helpers.

Checks run:

- `uv run rumdl check docs/configuring.md`

### DIFF
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -281,6 +281,18 @@ Supported keys:
     will be rendered with the combined answers as variables; it should render _nothing_
     if the value is valid, and an error message to show to the user otherwise.
 
+    !!! example "Validate a slug with `regex_search`"
+
+        ```yaml title="copier.yml"
+        project_slug:
+            type: str
+            help: Project slug
+            validator: >-
+                {% if not (project_slug | regex_search('^[a-z][a-z0-9-]+$')) %}
+                Use lowercase letters, digits, and dashes; start with a letter.
+                {% endif %}
+        ```
+
 - **when**: Condition that, if `false`, skips the question.
 
     If it is a boolean, it is used directly. Setting it to `false` is useful for


docs/configuring.md
```diff
@@ -277,9 +277,21 @@ Supported keys:
...
+    !!! info "Available validator helpers"
+
+        Validators are rendered as Jinja templates. They can use built-in Jinja
+        filters, tests, and functions; Copier helpers such as `pathjoin`; and
+        filters from [`jinja2-ansible-filters`][jinja2-ansible-filters],
+        including `regex_search`.
+
+        `regex_search` is a Jinja filter, not a dedicated validation method. For
+        project-specific validation helpers, configure custom
+        [Jinja extensions](#jinja_extensions).
```
sisp: TBH, I'm not sure about the added value of these documentation changes.

- It's already stated above that the validator field is a Jinja template.
- I imagine it's expected that the same Jinja filters, tests, and functions are available here; we don't suggest otherwise.
- The `pathjoin` helper is an uncommon function to use in a validator, it might confuse readers why this one is mentioned here.
- Why is it necessary to clarify that, e.g., `regex_search` is a Jinja filter and not a dedicated validation method? It isn't mentioned or confusingly suggested to be a validation method.

If you feel there's an opportunity to improve the validator documentation, how about adding an example admonition with a common (perhaps `regex_search`-based) validator?

gcomneno:
Thanks for the feedback! I replaced the generic helper note with a focused validator example using `regex_search`.

docs/configuring.md
```diff
@@ -277,9 +277,22 @@ Supported keys:
 - **multiline**: When set to `true`, it allows multiline input. This is especially
     useful when `type` is `json` or `yaml`.
 
-- **validator**: Jinja template with which to validate the user input. This template
-    will be rendered with the combined answers as variables; it should render _nothing_
-    if the value is valid, and an error message to show to the user otherwise.
+- **validator**: Jinja template with which to validate the user input. This
+    template will be rendered with the combined answers as variables; it should
+    render _nothing_ if the value is valid, and an error message to show to the
+    user otherwise.
```
sisp: Would you mind reverting this snippet to avoid diff noise? It's only a formatting change and unrelated to this PR.

sisp (changes requested):
Just a minor pedantic diff noise avoidance request, otherwise ready to merge. 🎉

gcomneno:
Thanks! Reverted the unrelated formatting-only change.

sisp (approved):
LGTM! :tada: Thanks! :bow:

PR #2688, opened 2026-05-22
docs: warn against moving template tags

### LINKED ISSUE #1097: document that Copier isn't compatible with moving tags
Copier isn't compatible with moving tags. Even less if those tags are valid PEP440 version identifiers. 

If you have moving refs, try using branches for that. Or you could use tags that aren't recognized as version strings.

_Originally posted by @yajo in https://github.com/copier-org/copier/issues/987#issuecomment-1505391701_

gcomneno:
Fixes #1097.

Adds a warning to the template versioning documentation that template version tags should be stable and should not be moved after projects have been generated from them, especially when those tags are valid PEP 440 versions.

For moving development references, the documentation now recommends using branches or passing the desired ref explicitly with `--vcs-ref`.

Checks run:

- `uv run rumdl check docs/generating.md`

### DIFF
--- a/docs/generating.md
+++ b/docs/generating.md
@@ -45,6 +45,13 @@ By default, Copier will copy from the last release found in template Git tags, s
 [PEP 440](https://peps.python.org/pep-0440/), regardless of whether the template is from
 a URL or a local clone of a Git repository.
 
+!!! warning
+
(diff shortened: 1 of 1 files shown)


sisp (approved):
LGTM: :tada: Thanks! :bow:

PR #2691, opened 2026-05-24
fix(copy): honor configured skip_tasks in unsafe check

### LINKED ISSUE #2690: Copier throws error about tasks being unsafe even when disabled
### Describe the problem

When running `copier copy` with a template that includes tasks but has tasks disabled by default, it still throws an error about the template using the unsafe feature tasks. I would think that by disabling tasks, I would be able to use the template without needing to pass the `--unsafe` flag.

The reason I would like to use it this way is that I need to run this task manually when setting up existing Git repositories with this template, and then have Renovate make MRs when there are new updates to the upstream template, **but I don't want Renovate to be able to run the tasks**, as this allows for arbitrary code execution in pipelines. However, I want the functionality of the task to still be there when running it manually for the first time.

I would think an update could be made to the [`_check_unsafe` function](https://github.com/copier-org/copier/blob/659f4e85c47bb74f5ae51e0d34b22cdbfa488f81/copier/_main.py#L292) to be aware of whether or not `skip_tasks` has been set to `true`.

Seems vaguely related to #2046, but not exactly. Also possibly related to #2087 

### Template

Not going to paste the full template, but I basically have:

```
_skip_tasks: true

_tasks:
- some tasks
```

### To Reproduce

1. Use a template similar to the one provided
2. Run `copier copy` with that template
3. It will not run due to it using unsafe features

### Logs

```console

```

### Expected behavior

I would expect to still be able to use the rest of the features of `copier copy` without being forced to pass the `--unsafe` flag for a feature that is not being used.

_No response_

### Operating system

Windows

Windows 11

### Copier version

9.15.1

### Python version

3.14.3

### Installation method

pip+pypi

### Additional context

_No response_

gcomneno:
Fixes #2690.

When a template defines `_skip_tasks: true` together with `_tasks`, Copier currently refuses to run without `--unsafe`, even though those tasks are configured not to run.

The unsafe check already skips task-related unsafe warnings when `skip_tasks=True` is passed through the CLI/API. However, it did not take the template-configured `_skip_tasks` value into account, so these two equivalent ways of disabling tasks behaved differently:

- `copier copy --skip-tasks ...`
- `_skip_tasks: true` in `copier.yml`

This change introduces a single effective skip-tasks value that combines both sources:

- the CLI/API `skip_tasks` option;
- the template configuration value from `_skip_tasks`.

That effective value is then used consistently for:

- the unsafe-feature check;
- `_copier_conf.skip_tasks` in the render context;
- deciding whether template tasks should actually be executed.

This does not make task execution trusted. It only avoids requiring `--unsafe` when tasks are configured not to run. Other unsafe features, such as Jinja extensions and migrations, are still checked as before.

The configured value is parsed with Copier's existing boolean coercion helper, so values such as `_skip_tasks: "false"` do not accidentally bypass the unsafe check.

Added regression coverage for:

- `_skip_tasks: true` with `_tasks`, which should not require `--unsafe` and should not execute tasks;
- `_skip_tasks: "false"` with `_tasks`, which should still require `--unsafe`.

Tests run:

- `pytest -q tests/test_unsafe.py -k "configured_skip_tasks" -rs`
- `pytest -q tests/test_unsafe.py tests/test_tasks.py -rs`
- `pytest -q tests/test_config.py tests/test_context.py tests/test_unsafe.py tests/test_tasks.py -rs`
- `ruff format --check copier/_main.py tests/test_unsafe.py`
- `ruff check copier/_main.py tests/test_unsafe.py`

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -296,7 +296,7 @@ class Worker:
         features: set[str] = set()
         if self.template.jinja_extensions:
             features.add("jinja_extensions")
-        if self.template.tasks and not self.skip_tasks:
+        if self.template.tasks and not self.effective_skip_tasks:
             features.add("tasks")
         if mode == "update" and self.subproject.template:
             if self.subproject.template.jinja_extensions:
@@ -449,7 +449,7 @@ class Worker:
                 "context_lines": lambda: self.context_lines,
                 "unsafe": lambda: self.unsafe,
                 "skip_answered": lambda: self.skip_answered,
-                "skip_tasks": lambda: self.skip_tasks,
+                "skip_tasks": lambda: self.effective_skip_tasks,
                 "sep": lambda: os.sep,
                 "os": lambda: OS,
             }
@@ -687,6 +687,13 @@ class Worker:
     def all_skip_if_exists(self) -> Sequence[str]:
         """Combine default and template skip-if-exists patterns."""
         return tuple(chain(self.skip_if_exists, self.template.skip_if_exists))
+
+    @cached_property
+    def effective_skip_tasks(self) -> bool:
+        """Check whether template tasks should be skipped."""
+        return self.skip_tasks or cast_to_bool(
+            self.template.config_data.get("skip_tasks", False)
+        )
 
     @cached_property
     def jinja_env(self) -> SandboxedEnvironment:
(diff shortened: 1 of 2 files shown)


sisp:
Thanks for intending to fix #2690, but it's a false positive bug report – see https://github.com/copier-org/copier/issues/2690#issuecomment-4529456350.

PR #2705, opened 2026-05-30
feat: add --ask option to `update`/`copy`/`recopy`

### LINKED ISSUE #2680: Add option to modify specific answer when `update`ing
### Actual Situation

One of our template questions is a listing of all the internal dependencies of a project (they are excluded from min-age verifications and the like)

```yml
internal_dependencies:
  help: "Dependencies local to the org included in the project"
  default: []
```

when adding/removing an internal dependency to a project, a user has two options:
* Either they call `copier update`, and press `enter` a dozen times to update this one answer inline (the prompt is already filled in, so adding new entries to the list is easy).
* Or they call `copier update -A --data internal_dependencies=[...]` where `...` is the shorthand yaml syntax for the list (not as it appears in `.copier-answers.yml`), with the user's modification

Both are rather unwieldy

### Desired Situation

It would be easiest for to be able to skip all questions when `update`ing except for the few that the user knows ahead of time that they know they want to change.

### Proposed solution

introduce a new switch to `copier copy` named `--answer`, that the user may use to specify which questions they want to answer. acceptable values:
* `--answer=foo` answer only the question named `foo`
* `--answer=foo-*` answer only questions starting with `foo-*` (this is useful if the user wants to modify only some sub-section of the template)
* `--answer=a,b`, filters can be combined with a comma, in this case, only the questions `a` and `b` will b answered

In addition, `--answer` can include categories starting with a dollar sign, for example
* `--answer=$SHOWN` answer all non-hidden questions in the template (this can be considered the default value)
* `--answer=$UNANSWERED` answer only questions that haven't been previously answered (`-A` can be an alias for this)
* `--answer=` answer no questions (`--defaults` can be an alias for this)

When this argument is supplied to `copier copy`, it skips all questions that don't match any of the filters provided. Users can then call `copier update --answer=internal_dependencies` and they will only be prompted with the filled-in current value of `internal_dependencies`

bentheiii:
closes #2680


bentheiii:
Hi @sisp, this is the PR that adds the `--ask` arg we discussed. I couldn't get the formatting to work (I'm using windows), but the tests pass

bentheiii:
Hi @sisp, what can be done do move this PR forward?

sisp:
I'd like to conclude the [design discussion](https://github.com/copier-org/copier/issues/2680#issuecomment-4649267435) with @lhupfeldt first to try to make sure UX will be good.

copier/_ask_pattern.py
sisp, line 1: Even though this is an internal module, I think we should avoid creating a new one just for this class. Once the public API surface has been fully clarified after the next major release, we can move things around without risking breaking changes. Until then, how about just adding the `AskPattern` class to `copier/_main.py` for lack of a better place?

copier/_main.py
```diff
@@ -633,24 +646,36 @@ def _ask(self) -> None:  # noqa: C901
...
+            else:
+                implicitly_skipped = False
+
+            if implicitly_skipped:
+                # skip the question unless the user explicitly asked for it with `--ask`
+                for ask_pattern in asks:
+                    if ask_pattern.matches(var_name):
+                        break
+                else:
+                    continue
```
sisp: Perhaps it's easier and results in less change to move the ask pattern matching up:

```diff
-            # Skip a question when the user already answered it.
-            if self.skip_answered and var_name in self.answers.last:
-                 continue
+            for ask_pattern in asks:
+                if ask_pattern.matches(var_name):
+                    break
+            else:
+                # Skip a question when the user already answered it.
+                if self.skip_answered and var_name in self.answers.last:
+                     continue
```

copier/_ask_pattern.py
```diff
@@ -0,0 +1,20 @@
...
+
+class AskPattern:
+    """A string pattern for questions for usage in `--ask` arguments."""
+    
+    def __init__(self, pattern: str) -> None:
+        self.orginal = pattern
+        escaped = re.escape(pattern)
+        self.pattern = re.compile(
+            escaped.replace(r"\*", ".*")
+        )
```
sisp: I'm not sure about regex matching, I think the CLI ergonomics is suboptimal and I anticipate glob-style patterns to be sufficient for this use case. How about using [`fnmatch.fnmatchcase`](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatchcase) instead of partially escaped regex patterns matched with `re.fullmatch`?

docs/configuring.md
```diff
@@ -808,6 +808,17 @@ Don't forget to read [the docs about the answers file](#the-copier-answersyml-fi
     _answers_file: .my-custom-answers.yml
     ```
 
+### `ask`
+
+- Format: `List[str]`
+- CLI flags: `-k`, `--ask`
```
sisp: I'm not sure about the `-k` short flag. How about starting with just `--ask`?

tests/test_prompt.py
sisp, line 1: We need to test the different scenarios extensively for all operations (`copy`/`recopy`/`update`), also with `--defaults`, also interplay with `--data`, also no-op cases, and also cases where a warning is expected. I think tailored tests will be better than reusing the `MARIO_TREE` test template.

copier/_main.py
```diff
@@ -659,6 +684,12 @@ def _ask(self) -> None:  # noqa: C901
 
         # Reload external data, which may depend on answers
         self.answers.external = self._external_data()
+        for ask_pattern in asks:
+            if not ask_pattern.was_matched:
+                warnings.warn(
+                    f"Asked pattern {ask_pattern.orginal!r} did not match any question that would be skipped."
+                    " Did you specify it in --data?"
```
sisp: We can check whether an answer was provided via `--data` via `self.answers.init`.

sisp (changes requested):
Thanks for contributing the `--ask` flag, @bentheiii. :bow: I've left some inline comments.

copier/_ask_pattern.py
bentheiii: Done, TIL about the fnmatch module

copier/_main.py
bentheiii: Done, though note that this may cause some warnings to not fire, since patterns are now checked regardless of when they would be skipped. For example: if a template has questions `a, b`, and the user has answered `a`, then we now differ in `copier update -A --ask b`
 
* in the old implementation, this would show a warning (since the `--ask b` is redundant)
* in the new implementation, this would not show a warning (we could argue that this makes sense too, a user might want to ensure that specific questions are asked regardless of any other flags, even if it would be asked anyway)

Your call

bentheiii:
Thanks for the review @sisp, I've pushed a commit addressing the comments. Note the reply concerning moving the ask matching up (I ultimately think it's for the better)

tests/test_prompt.py
```diff
@@ -110,12 +110,20 @@
         ("None", ("--data=your_name=None",)),
     ],
 )
+@pytest.mark.parametrize(
+    "ask",
+    (
+        ("your*",),
+        (),
+    )
+)
```
sisp: Let's not extend existing tests but rather create minimal tailored tests for `--ask` behavior.

docs/configuring.md
```diff
@@ -808,6 +808,17 @@ Don't forget to read [the docs about the answers file](#the-copier-answersyml-fi
     _answers_file: .my-custom-answers.yml
     ```
 
+### `ask`
+
+- Format: `List[str]`
+- CLI flags: `--ask`
+
+Ask the questions specified, even if they would be skipped by [defaults](#defaults) or [skip-answered](#skip_answered). Names may include fnmatch-style wildcards
```
sisp:
```suggestion
Ask the matched questions, even if they would be skipped by [defaults](#defaults) or [skip-answered](#skip_answered). Names may be glob-style patterns.
```

copier/_main.py
```diff
@@ -228,6 +229,9 @@ class Worker:
 
         skip_tasks:
             When `True`, skip template tasks execution.
+
+        ask:
+            List of question names to ask, even if they would be skipped by `skip_answered` or `defaults`. Can also fnmatch-style include wildcards.
```
sisp:
```suggestion
            List of question names to ask, even if they would be skipped by `skip_answered` or `defaults`. Supports glob-style patterns.
```

copier/_main.py
```diff
@@ -290,6 +295,30 @@ def _cleanup(self) -> None:
...
+        if in_init:
+            questions = ", ".join(repr(k) for k in in_init)
+            warnings.warn(
+                f"Asked pattern {ask.pattern!r} did not match any question in the template except for {questions},"
+                " which were already provided in --data not asked."
+            )
+        else:
+            warnings.warn(
+                f"Asked pattern {ask.pattern!r} did not match any question in the template."
+            )
```
sisp: I wonder whether we should even warn about unused `--ask` patterns. I think there's little harm in an unmatched pattern because it's just a no-op.

And would it be more intuitive if `--ask` took precedence over `--data` in the sense that `--ask` always forces prompting while `--data` overrides any default or recorded answers? For example, `--data X=v --ask X` would force prompting of `X` which is pre-filled with `v`. This would avoid cumbersome logic for raising warnings which even feels late (IMO) when raised _after_ completing the questionnaire – which is the only way because of conditional questions using templated `when` fields. WDYT? If you agree, this has implications on the logic in `Worker._ask`.

copier/_main.py
```diff
@@ -1253,6 +1293,7 @@ def run_copy(self) -> None:
 
         self._check_unsafe("copy")
         self._print_message(self.template.message_before_copy)
+        self._check_unneccessary_ask()
```
sisp: I think this check should run before printing the "before copy" message.


```suggestion
        self._check_unneccessary_ask()
        self._print_message(self.template.message_before_copy)
```

sisp (changes requested):

copier/_main.py
bentheiii: I agree that the warnings are superfluous (it might even be part of someone's workflow to ensure a question is asked no matter what)
Under the same reasoning, I also removed the `_check_unnecessary_ask` warning

bentheiii:
Thanks @sisp, I changed the behavior so that `--ask` has precedence over `--data` (while preserving `--data`'s answer as the initial prompt), I also removed all warnings over no-op `--ask`

tests/test_prompt.py
```diff
@@ -127,7 +127,12 @@ def test_copy_default_advertised(
     with local.cwd(dst):
         # Copy the v1 template
         tui = spawn(
-            COPIER_PATH + ("copy", str(src), ".", "--vcs-ref=v1") + args,
+            COPIER_PATH + (
+                "copy",
+                str(src),
+                ".",
+                "--vcs-ref=v1",
+            ) + args,
```
sisp: Would you mind reverting this formatting-only change? 🤓

copier/_main.py
```diff
@@ -228,6 +229,9 @@ class Worker:
 
         skip_tasks:
             When `True`, skip template tasks execution.
+
+        ask:
+            List of question names to ask, even if they would be skipped by other options. Supports glob-style patterns.
```
sisp: Ruff complains about the line being too long.


```suggestion
            List of question names to ask, even if they would be skipped by other options.
            Supports glob-style patterns.
```

copier/_cli.py
```diff
@@ -239,6 +239,14 @@ class CopierCopySubApp(_Subcommand):
         ["-w", "--overwrite"],
         help="Overwrite files that already exist, without asking.",
     )
+    ask = cli.SwitchAttr(
+        ["--ask"],
+        str,
+        list=True,
+        help=(
+            "Ask the questions matching the given glob-pattern, even if they would be skipped by other options"
```
sisp: Let's wrap this long string:

```suggestion
            "Ask the questions matching the given glob-pattern, even if they would be "
            "skipped by other options"
```

Same for the other occurrences below.

copier/_main.py
```diff
@@ -622,35 +627,39 @@ def _ask(self) -> None:  # noqa: C901
...
-            # Skip a question when the user already answered it.
-            if self.skip_answered and var_name in self.answers.last:
-                continue
 
-            # Display TUI and ask user interactively only without --defaults
-            try:
+            for ask_pattern in self.ask:
+                if fnmatchcase(var_name, ask_pattern):
+                    break
+            else:
```
sisp: I find this a bit hard to reason about (it might just be me). How about this?

```suggestion
            if not any(fnmatchcase(var_name, ask_pattern) for ask_pattern in self.ask):
```

tests/test_prompt.py
```diff
@@ -1202,3 +1209,367 @@ def test_interactive_session_required_for_overwrite_prompt(
     assert (
         b"Interactive session required: Consider using `--overwrite`" in process.stderr
     )
+
+
+ANIMAL_TREE: Mapping[StrOrPath, str | bytes] = {
+    "copier.yml": (
+        f"""\
+        _templates_suffix: {SUFFIX_TMPL}
+        _envops: {BRACKET_ENVOPS_JSON}
```
sisp: I know that some tests use these explicit settings. For new tests, I'd prefer omitting them to use default settings.

```suggestion
```

 This means, we need to use curly braces for template syntax below and the `.jinja` suffix for template files.

sisp (changes requested):
Just some final polishing requests, I think. And the CLI snapshot assertions need to be updated after the reformatting commit.

copier/_cli.py
```diff
@@ -318,6 +328,14 @@ class CopierRecopySubApp(_Subcommand):
         default=False,
         help="Skip questions that have already been answered",
     )
+    ask = cli.SwitchAttr(
+        ["--ask"],
+        str,
+        list=True,
+        help=(
+            "Ask the questions matching the given fnamtch-pattern, even if they would be skipped by other options"
```
sisp: This one still needs line-wrapping.

```suggestion
            "Ask the questions matching the given fnamtch-pattern, even if they would be "
            "skipped by other options"
```

sisp (changes requested):
One last thing, otherwise ready for merge from my side. 🎉

sisp (approved):
Thanks for this fantastic contribution, @bentheiii! 🙇 I really like the feature, can't wait to use it. 🎉

PR #2717, opened 2026-06-08
fix: resolve local template paths to absolute before storing in answers file

### LINKED ISSUE #335: Shouldn't `_src_path` be stored as absolute path?
When creating a project using a local path, shouldn't it be stored as absolute path in the `answers_file`?

If created using `copier . /path/to/destination`, it's saved as `_src_path: .`.

One can do it "auto-manually" using `copier "$PWD" /path/to/destination`.

Oscar-SR:
This fix resolves local template paths to absolute before storing them in the answers file, so they remain valid regardless of the CWD when running `copier update`.

Remote URLs (gh:, https://, git@) are unaffected since they don't resolve to local paths.

Fixes #335

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -364,12 +364,20 @@ class Worker:
         answers: AnyByStrDict = {}
         commit = self.template.commit
         src = self.template.url
-        # For local template paths, store as absolute so the path remains
-        # valid if the CWD changes between copy and update operations
+        # Resolve template path to a relative one when possible
         if src:
             src_path = Path(src)
-            if src_path.exists():
-                src = str(src_path.resolve())
+            # Only resolve local relative paths that exist on disk
+            if not src_path.is_absolute() and src_path.exists():
+                src_resolved = src_path.resolve()
+                dst_resolved = self.subproject.local_abspath.resolve()
+                try:
+                    src = os.path.relpath(
+                        str(src_resolved), str(dst_resolved)
+                    )
+                except ValueError:
+                    # Fallback for cross-drive paths on Windows
+                    src = str(src_resolved)
         for key, value in (("_commit", commit), ("_src_path", src)):
             if value is not None:
                 answers[key] = value
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -364,6 +364,12 @@ class Worker:
         answers: AnyByStrDict = {}
         commit = self.template.commit
         src = self.template.url
+        # For local template paths, store as absolute so the path remains
+        # valid if the CWD changes between copy and update operations
+        if src:
+            src_path = Path(src)
+            if src_path.exists():
+                src = str(src_path.resolve())
         for key, value in (("_commit", commit), ("_src_path", src)):
             if value is not None:
                 answers[key] = value
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -364,20 +364,15 @@ class Worker:
         answers: AnyByStrDict = {}
         commit = self.template.commit
         src = self.template.url
-        # Resolve template path to a relative one when possible
-        if src:
-            src_path = Path(src)
-            # Only resolve local relative paths that exist on disk
-            if not src_path.is_absolute() and src_path.exists():
-                src_resolved = src_path.resolve()
-                dst_resolved = self.subproject.local_abspath.resolve()
-                try:
-                    src = os.path.relpath(
-                        str(src_resolved), str(dst_resolved)
-                    )
-                except ValueError:
-                    # Fallback for cross-drive paths on Windows
-                    src = str(src_resolved)
+        # Resolve local relative template paths
+        src_path = Path(src)
+        if not src_path.is_absolute() and src_path.exists():
+            src_resolved = src_path.resolve()
+            dst_resolved = self.subproject.local_abspath.resolve()
+            if src_resolved.is_relative_to(dst_resolved):
+                src = str(src_resolved.relative_to(dst_resolved))
+            else:
+                src = str(src_resolved)
         for key, value in (("_commit", commit), ("_src_path", src)):
             if value is not None:
                 answers[key] = value
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -364,15 +364,21 @@ class Worker:
         answers: AnyByStrDict = {}
         commit = self.template.commit
         src = self.template.url
-        # Resolve local relative template paths
-        src_path = Path(src)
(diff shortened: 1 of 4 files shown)


sisp:
I think this change would break the use case described in #2589. One can always provide an absolute path to `copier copy` to record an absolute path in `.copier-answers.yml`. Always converting to an absolute path seems like convenience for your use case at the cost of (likely) breaking some (admittedly rare but known) other use cases.

copier/_main.py
```diff
@@ -364,6 +364,12 @@ def _answers_to_remember(self) -> Mapping[str, Any]:
         answers: AnyByStrDict = {}
         commit = self.template.commit
         src = self.template.url
+        # For local template paths, store as absolute so the path remains
```
lhupfeldt: This would break configurations with template and subprojects in the same repo.
It cannot be assumed that the repo will always be checked out in the same location.

lhupfeldt (changes requested):

copier/_main.py
lhupfeldt: It would make sense if copier resolved the `relative` path from the subproject  to the template if a relative path was specified. Then it would be possible to update.

copier/_main.py
```diff
@@ -364,6 +364,20 @@ def _answers_to_remember(self) -> Mapping[str, Any]:
         answers: AnyByStrDict = {}
         commit = self.template.commit
         src = self.template.url
+        # Resolve template path to a relative one when possible
+        if src:
```
lhupfeldt: In which scenario is self.template.url falsy?

copier/_main.py
```diff
@@ -364,6 +364,20 @@ def _answers_to_remember(self) -> Mapping[str, Any]:
...
+            src_path = Path(src)
+            # Only resolve local relative paths that exist on disk
+            if not src_path.is_absolute() and src_path.exists():
+                src_resolved = src_path.resolve()
+                dst_resolved = self.subproject.local_abspath.resolve()
+                try:
+                    src = os.path.relpath(
```
lhupfeldt: Is it necessary to min pathlib and os.path?

lhupfeldt (changes requested):
There should be test cases.

Note that I'm not a maintainer.

copier/_main.py
sisp: We can use [`Path.is_relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to) instead.

sisp (commented):
Strictly speaking, this is a breaking change because a template path relative to the subproject may break moving the project directory to a different location. But it might be a valuable change to be included in the next major release.

copier/_main.py
sisp: It should always be a non-empty string.

Oscar-SR:
Thanks for the feedback! I have refactored the code using `pathlib`, removed the `self.template.url` validation and added the missing test cases. Let me know if everything looks good now.

sisp:
With the current implementation, relative paths including `..` are not computed, so the conversion of an absolute path to a relative path is limited to a template path within the project root, which is quite an edge case. I'm not sure I see the value in the current implementation because you can navigate to the project root and run `copier copy` with a relative path to the local template directory. And updating a project with `copier update` currently only works when the current working directory is the project root which may not be intuitive if your current working directory was different when running `copier copy`. The only potential enhancement I can imagine is recording the relative path to a local template converted to the path relative to the project root when Copier internally changes the working directory to the project root on `copier update`, so you can run `copier update path/to/project` no matter the current working directory.

sisp:
Re-reading https://github.com/copier-org/copier/issues/335#issuecomment-3917217569 again, I think converting a local template path that is relative to the current working directory to a path that is relative to the project root is strictly better:

1. `copier update` works regardless of the current working directory.
1. Moving the local template or project root independently invalidates the recorded relative template path in both cases, so there's no degradation.

For this to be useful, path traversal (involving `..`) must be supported. `pathlib.Path.relative_to(other, walk_up=True)` would work, but `walk_up` was added in Python 3.12 while Copier supports Python 3.10+. This means, we'll need to use `os.path.relpath` until Copier drops support for Python 3.11 presumably after October 31, 2027.

There might be a way to make this change non-breaking: I think we could look up a local template according to the new relative path (relative to the project root) and fall back to looking it up according to the old relative path (relative to the current working directory). In both cases, the new relative path would be recorded in the updated answers file on `copier update`. This way, existing projects are automatically migrated to the new path. We would deprecate support for the old relative path and remove the fallback logic in the next major release (or perhaps after a sufficiently long period).

WDYT?

lhupfeldt:
Does the existing relative path work?
Let's say you run `copier copy mytemplate mysubproject` which would store `mytemplate` as the path.
Then you cd into `mysubproject` and run update, but now the stored path is wrong.

sisp:
Do you mean with the changes of this PR or with the state of `master`?

With my suggestion, the template path in `.copier-answers.yml` would be `../mytemplate` because that's the path relative to `mysubproject`.

lhupfeldt:
> Do you mean with the changes of this PR or with the state of `master`?
> 
I meant `master`. 
I mean, is there any scenario currently working which would necessitate backwards compatibility?
I have no problem with our cd'ing hack breaking if this is implemented.

sisp:
Hmmm, I think you're right. Currently, `copier update` with a recorded relative template path only works when executed in the same working directory as `copier copy` before. This still works when the relative template path is relative to the project root. Hence, no breaking change. :bulb: But we still need the fallback to the current template path resolution.

copier/_main.py
```diff
@@ -364,15 +364,21 @@ def _answers_to_remember(self) -> Mapping[str, Any]:
...
-                src = str(src_resolved)
+        # Check if the original path was relative
+        src_path = Path(self.src_path) if self.src_path else None
+        if src_path is None:
+            last_src = self.subproject.last_answers.get("_src_path")
+            src_path = Path(last_src) if last_src else None
+        was_relative = src_path is not None and not src_path.is_absolute()
+
+        # If original was relative and it is not a remote Git repo, save as relative to the subproject root
+        if was_relative and src and not src.startswith(("http://", "https://", "git@", "git+", "gh:", "gl:", "bb:")):
```
lhupfeldt: I don't think the list of prefixes for "remote paths" belongs here. There must be other places (when doing the copy) where it is also determined if the path is remote.

PR #2720, opened 2026-06-11
feat: cache remote git templates as local mirrors with worktrees

### LINKED ISSUE #450: Cache git templates locally
> the fetch that copier does for this case is too slow.

_Originally posted by @bburgin in https://github.com/copier-org/copier/discussions/449#discussioncomment-1371169_

And it's true. Copier should cache downloaded templates, so next fetches are faster.

DrKat0m:
Remote templates were re-cloned into a throwaway temp directory on every run.
Following @yajo's suggestion, this caches a persistent `git --mirror` clone per template under the user cache dir and checks out temporary worktrees from it.

- First use: `git clone --mirror` into the cache (created atomically via staging + rename), then `git worktree add` a temporary checkout.
- Later uses: `git remote update --prune` (no full re-download) + a fresh worktree.
- Local templates keep the original behavior (including dirty-change handling).
- Cache dir via `platformdirs`, overridable with `COPIER_CACHE_DIR`. No new dependency.
- Robustness: explicit `--git-dir` (works with `safe.bareRepository=explicit`), stale-worktree pruning, read-only-safe cleanup, corrupt-mirror recovery.
- Existing VCS tests unchanged; 5 new network free tests added (20 passing under pytest-xdist).

Closes #450

### DIFF
--- a/copier/_vcs.py
+++ b/copier/_vcs.py
@@ -2,18 +2,23 @@
 
 from __future__ import annotations
 
+import os
 import re
 import sys
 from contextlib import suppress
+from hashlib import sha256
 from pathlib import Path
+from shutil import rmtree
 from tempfile import TemporaryDirectory, mkdtemp
 from warnings import warn
 
 from packaging import version
 from packaging.version import InvalidVersion, Version
+from platformdirs import user_cache_dir
 from plumbum import TF, CommandNotFound, ProcessExecutionError, colors, local
 from plumbum.commands.base import BaseCommand
 
+from ._tools import handle_remove_readonly
 from ._types import OptBool, OptStrOrPath, StrOrPath
 from .errors import DirtyLocalWarning, ShallowCloneWarning
 
@@ -21,6 +26,9 @@ GIT_USER_NAME = "Copier"
 GIT_USER_EMAIL = "copier@copier"
 
 CLONE_PREFIX = f"{__name__}.clone."
+
+# Environment variable to override the on-disk location of the git mirror cache.
+CACHE_DIR_ENV_VAR = "COPIER_CACHE_DIR"
 
 
 class _PathStr(str):
@@ -190,8 +198,135 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
         return "HEAD"
 
 
+def get_cache_dir() -> Path:
+    """Get the directory where cached git mirrors are stored.
+
+    Defaults to a per-user cache directory (via `platformdirs`), but can be
+    overridden with the `COPIER_CACHE_DIR` environment variable.
+    """
+    override = os.environ.get(CACHE_DIR_ENV_VAR)
+    if override:
+        return Path(override)
+    return Path(user_cache_dir("copier")) / "git"
+
+
+def _mirror_path(url: str) -> Path:
+    """Compute the deterministic on-disk location of the mirror for `url`."""
+    digest = sha256(url.encode("utf-8")).hexdigest()
+    return get_cache_dir() / f"{digest}.git"
+
+
+def _is_remote(url: str) -> bool:
+    """Tell whether `url` points to a remote repository worth caching.
+
+    Local repositories (existing directories or bundle files, as well as
+    paths marked with [_PathStr][]) keep their original clone behavior, which
+    preserves features like including dirty changes.
+    """
+    if isinstance(url, _PathStr):
+        return False
+    try:
+        if Path(url).exists():
+            return False
+    except OSError:
+        pass
+    return True
+
+
+def _force_rmtree(path: Path) -> None:
+    """Remove `path`, coping with git's read-only files (e.g. on Windows)."""
+    if sys.version_info >= (3, 12):
+        rmtree(path, onexc=handle_remove_readonly)
+    else:
+        rmtree(path, onerror=handle_remove_readonly)
+
+
+def _is_valid_mirror(mirror: Path) -> bool:
+    """Tell whether `mirror` is a usable bare git repository.
+
+    Guards against partial or corrupt cache entries (e.g. left behind by an
+    interrupted clone or a failed deletion) that would otherwise be mistaken
+    for a healthy mirror.
+    """
+    if not (mirror / "objects").is_dir():
+        return False
+    try:
+        out = get_git()(
+            "--git-dir", str(mirror), "rev-parse", "--is-bare-repository"
+        ).strip()
+    except (OSError, ProcessExecutionError):
+        return False
+    return out == "true"
+
+
+def get_or_create_mirror(url: str) -> Path:
+    """Get a cached `--mirror` clone of `url`, creating or refreshing it.
+
+    On first use the remote is mirror-cloned into the cache. On subsequent
+    uses the existing mirror is refreshed via `git remote update` and stale
+    worktree registrations are pruned, so no full re-download is needed. A
+    corrupt or partial cache entry is discarded and re-created.
+    """
+    git = get_git()
+    mirror = _mirror_path(url)
+    if _is_valid_mirror(mirror):
+        # Refreshing the existing mirror. Use `--git-dir` explicitly when
+        # operating on the bare repository so this works even when Git is
+        # configured with `safe.bareRepository=explicit`.
+        git("--git-dir", str(mirror), "remote", "update", "--prune")
+        # Dropping registrations for worktrees whose directories were already
+        # removed by the cleanup step (from `Template._cleanup`).
+        git("--git-dir", str(mirror), "worktree", "prune")
+    else:
+        # Discarding any corrupt/partial remnant before recreating.
+        if mirror.exists():
+            _force_rmtree(mirror)
+
+        '''Creating the mirror atomically: clone into a staging directory, then
+        move it into place. If a concurrent Copier process created the
+        mirror first, discard ours and reuse theirs. This avoids colliding
+        on a half-written mirror without requiring full inter-process locks.
+        A fresh `--mirror` clone already has every ref, so no fetch is needed.'''
+
+        mirror.parent.mkdir(parents=True, exist_ok=True)
+        staging = Path(mkdtemp(prefix=CLONE_PREFIX, dir=mirror.parent))
+        try:
+            staging_repo = staging / "repo.git"
+            git("clone", "--mirror", url, str(staging_repo))
+            try:
+                staging_repo.rename(mirror)
+            except OSError:
+                if not _is_valid_mirror(mirror):
+                    raise
+        finally:
+            rmtree(staging, ignore_errors=True)
+    return mirror
+
+
+def _clone_via_cache(ref: str, location: str, mirror: Path) -> str:
+    """Create a temporary worktree of `mirror` at `ref` in `location`.
+    
+    `git worktree add` refuses a path that already exists, so remove the
+    pre-allocated (empty) placeholder directory and let git recreate it."""
+    git = get_git()
+    placeholder = Path(location)
+    if placeholder.exists():
+        placeholder.rmdir()
+    git(
+        "--git-dir", str(mirror),
+        "worktree", "add", "--detach", "--force", location, ref,
+    )
+    with local.cwd(location):
+        git("submodule", "update", "--checkout", "--init", "--recursive", "--force")
+    return location
+
+
 def clone(url: str, ref: str = "HEAD", location: str | None = None) -> str:
     """Clone repo into some temporary destination.
+
+    Remote repositories are cached as a local git mirror and checked out into
+    a temporary worktree, so repeated use of the same template avoids a full
+    re-download.
 
     Includes dirty changes for local templates by copying into a temp
     directory and applying a wip commit there.
@@ -210,6 +345,11 @@ def clone(url: str, ref: str = "HEAD", location: str | None = None) -> str:
     git_version = get_git_version()
     if location is None:
         location = mkdtemp(prefix=CLONE_PREFIX)
+    # Remote templates: use a cached mirror + temporary worktree.
+    if _is_remote(url):
+        mirror = get_or_create_mirror(url)
+        return _clone_via_cache(ref, location, mirror)
+    # Local templates: keep the original clone behavior.
     _clone = git["clone", "--no-checkout", url, location]
     # Faster clones if possible
     if git_version >= Version("2.27"):
--- a/copier/_vcs.py
+++ b/copier/_vcs.py
@@ -10,6 +10,7 @@ from hashlib import sha256
 from pathlib import Path
 from shutil import rmtree
 from tempfile import TemporaryDirectory, mkdtemp
+from urllib.parse import urlsplit, urlunsplit
 from warnings import warn
 
 from packaging import version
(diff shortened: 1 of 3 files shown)


copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
+def get_cache_dir() -> Path:
+    """Get the directory where cached git mirrors are stored.
+
+    Defaults to a per-user cache directory (via `platformdirs`), but can be
+    overridden with the `COPIER_CACHE_DIR` environment variable.
+    """
+    override = os.environ.get(CACHE_DIR_ENV_VAR)
+    if override:
+        return Path(override)
+    return Path(user_cache_dir("copier")) / "git"
```
sisp: We need to change the `user_cache_dir` function call to

```suggestion
    return Path(user_cache_dir("copier"), appauthor=False) / "git"
```

else the Windows path will be

```
C:\Users\<user>\AppData\Local\copier\copier\Cache
```

instead of:

```
C:\Users\<user>\AppData\Local\copier\Cache
```

copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
+
+
+def _mirror_path(url: str) -> Path:
```
sisp: How about a minor rename to `_get_mirror_path`? The word "mirror" in `_mirror_path` could be interpreted as a verb in imperative form.


```suggestion
def _get_mirror_path(url: str) -> Path:
```

copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
         return "HEAD"
 
 
+def get_cache_dir() -> Path:
```
sisp: Should we make this function internal?

```suggestion
def _get_cache_dir() -> Path:
```

(I know the codebase isn't very consistent in this regard yet, but we can try being more consistent for new code.)

copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
+    """
+    override = os.environ.get(CACHE_DIR_ENV_VAR)
+    if override:
+        return Path(override)
+    return Path(user_cache_dir("copier")) / "git"
+
+
+def _mirror_path(url: str) -> Path:
+    """Compute the deterministic on-disk location of the mirror for `url`."""
+    digest = sha256(url.encode("utf-8")).hexdigest()
```
sisp: What if the URL contains embedded credentials? For example:

```text
https://x-access-token:github_pat_xxxxxxx@github.com/org/repo.git
```

I think we should omit such credentials from the cache key.

copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
+    try:
+        out = get_git()(
+            "--git-dir", str(mirror), "rev-parse", "--is-bare-repository"
+        ).strip()
+    except (OSError, ProcessExecutionError):
+        return False
+    return out == "true"
+
+
+def get_or_create_mirror(url: str) -> Path:
```
sisp: Should we make this function internal?

```suggestion
def _get_or_create_mirror(url: str) -> Path:
```

(I know the codebase isn't very consistent in this regard yet, but we can try being more consistent for new code.)

copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
+    else:
+        # Discarding any corrupt/partial remnant before recreating.
+        if mirror.exists():
+            _force_rmtree(mirror)
+
+        '''Creating the mirror atomically: clone into a staging directory, then
+        move it into place. If a concurrent Copier process created the
+        mirror first, discard ours and reuse theirs. This avoids colliding
+        on a half-written mirror without requiring full inter-process locks.
+        A fresh `--mirror` clone already has every ref, so no fetch is needed.'''
```
sisp: I'd prefer making this a `#`-based comment block.

copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
+        try:
+            staging_repo = staging / "repo.git"
+            git("clone", "--mirror", url, str(staging_repo))
+            try:
+                staging_repo.rename(mirror)
+            except OSError:
+                if not _is_valid_mirror(mirror):
+                    raise
+        finally:
+            rmtree(staging, ignore_errors=True)
```
sisp: Probably only a stylistic difference: Would it be a bit more idiomatic to use [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) as a context manager?


```suggestion
        with TemporaryDirectory(
            prefix=CLONE_PREFIX, dir=mirror.parent, ignore_cleanup_errors=True
        ) as staging:
            staging_repo = Path(staging) / "repo.git"
            git("clone", "--mirror", url, str(staging_repo))
            try:
                staging_repo.rename(mirror)
            except OSError:
                if not _is_valid_mirror(mirror):
                    raise
```

copier/_vcs.py
```diff
@@ -190,9 +198,136 @@ def get_latest_tag(url: str, use_prereleases: OptBool = False) -> str:
...
+    return mirror
+
+
+def _clone_via_cache(ref: str, location: str, mirror: Path) -> str:
+    """Create a temporary worktree of `mirror` at `ref` in `location`.
+    
+    `git worktree add` refuses a path that already exists, so remove the
+    pre-allocated (empty) placeholder directory and let git recreate it."""
```
sisp:
```suggestion
    pre-allocated (empty) placeholder directory and let git recreate it.
    """
```

tests/test_vcs.py
```diff
@@ -95,6 +98,116 @@ def test_local_clone() -> None:
     shutil.rmtree(local_tmp, ignore_errors=True)
 
 
+def _make_remote_repo(path: Path) -> str:
```
sisp: It might be more idiomatic to make this a function-scoped pytest fixture rather than a helper function.

tests/test_vcs.py
```diff
@@ -95,6 +98,116 @@ def test_local_clone() -> None:
...
+    sentinel = mirror / "copier-cache-sentinel"
+    sentinel.write_text("kept")
+
+    dst2 = clone(url)
+    assert sentinel.exists(), "mirror was re-created instead of reused"
+    assert Path(dst2, "README.md").exists()
+    assert dst2 != dst1  # a fresh worktree per use
+
+    shutil.rmtree(dst1, ignore_errors=True)
+    shutil.rmtree(dst2, ignore_errors=True)
```
sisp: These functions are not executed if there's an assertion error above. Can we make sure the clone locations are always cleaned up, e.g. using [`contextlib.ExitStack`](https://docs.python.org/dev/library/contextlib.html#contextlib.ExitStack)? There are a few more such occurrences below.

sisp (changes requested):
Thanks for contributing this fantastic feature and enhancement of Copier, @DrKat0m! :bow:

I've left a couple of inline comments. In addition, I wonder whether it would be a good idea to add a CLI flag that forces ignoring the cache, e.g. `--no-cache`, to provide an escape hatch in case the cache is corrupted or otherwise broken. WDYT?

DrKat0m:
Thanks for the detailed review, @sisp! I've pushed a commit addressing all the inline comments.

I agree that a --no-cache CLI flag would be useful. One thing to note is that the cache already auto-rebuilds corrupt or partial mirrors via _is_valid_mirror, so for the "cache is broken" case, a copier cache clear (--clear-cache) flag may be the more useful escape hatch. In that sense, --no-cache could mainly serve as a debugging or bypass option.

tests/test_vcs.py
```diff
@@ -95,6 +99,127 @@ def test_local_clone() -> None:
...
+    base = "https://github.com/org/repo.git"
+    with_token = "https://x-access-token:github_pat_secret@github.com/org/repo.git"
+    assert _get_mirror_path(with_token) == _get_mirror_path(base)
+
+
+def test_remote_clone_creates_and_reuses_mirror(
+    remote_repo: str, clone_cleanup: Callable[[str], str]
+) -> None:
+    # First use creates the mirror and checks out a worktree.
+    dst1 = clone_cleanup(clone(remote_repo))
```
sisp: I hadn't thought about this in the first review iteration: The `clone` function accepts a `location` argument. We could use a temporary directory, e.g. managed by `tmp_path_factory` fixture or using `with tempfile.TemporaryDirectory() as tmpdir: ...` as the clone location instead of letting `clone` create a temporary directory itself which isn't garbage-collected. Then, we could remove the `clone_cleanup` fixture again.

tests/test_vcs.py
```diff
@@ -95,6 +99,127 @@ def test_local_clone() -> None:
...
+def remote_repo(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> str:
+    """A small local git repo exposed via a ``file://`` URL (treated as remote).
+
+    A ``file://`` URL is treated as remote by ``clone()``, so this exercises
+    the mirror-cache code path without requiring network access. The mirror
+    cache is redirected to a temp dir to keep the test isolated.
+    """
+    monkeypatch.setenv("COPIER_CACHE_DIR", str(tmp_path / "cache"))
+    path = tmp_path / "remote"
+    path.mkdir(parents=True, exist_ok=True)
```
sisp: I think it's better to use `tmp_path_factory` to create a temporary directory because `tmp_path` is the same path for all fixtures in the same function scope, i.e. if we use `tmp_path` elsewhere in the test function that uses `remote_repo`, it would be the same path.


```suggestion
def remote_repo(tmp_path_factory: pytest.TempPathFactory, monkeypatch: pytest.MonkeyPatch) -> str:
    """A small local git repo exposed via a ``file://`` URL (treated as remote).

    A ``file://`` URL is treated as remote by ``clone()``, so this exercises
    the mirror-cache code path without requiring network access. The mirror
    cache is redirected to a temp dir to keep the test isolated.
    """
    monkeypatch.setenv("COPIER_CACHE_DIR", str(tmp_path_factory.mktemp("cache")))
    path = tmp_path_factory.mktemp("remote_repo")
```

sisp (commented):
Thanks for pushing an update. I've left a few more really minor inline comments. This looks fantastic! 🥇

I agree with your argument regarding `--no-cache` and `--clear-cache`. I really like how cache validity is ensured and how it is invalidated/recreated otherwise. How about keeping the scope of this PR limited to the new cache feature (as it is now) and considering CLI extensions in a follow-up? Since you brought up the `--clear-cache` flag idea, perhaps it might even make sense to add a new subcommand `cache` with a few actions like `clear` to clear the cache and `dir` to print the cache path, slightly inspired by `uv cache ...`.

sisp:
One more thought: I think we should use a temporary directory as the cache dir during test suite runs to avoid polluting the real cache with tons of test templates. For example, we could add a (session?) fixture with `autouse=True` that monkeypatches `COPIER_CACHE_DIR` to refer to a temporary directory managed via the `tmp_path_factory` fixture.

tests/test_vcs.py
```diff
@@ -95,6 +98,120 @@ def test_local_clone() -> None:
...
+    """A small local git repo exposed via a ``file://`` URL (treated as remote).
+
+    A ``file://`` URL is treated as remote by ``clone()``, so this exercises
+    the mirror-cache code path without requiring network access. The mirror
+    cache is redirected to a temp dir to keep the test isolated.
+
+    Uses ``tmp_path_factory`` rather than ``tmp_path`` so the repo and cache
+    don't share a path with a test's own ``tmp_path``.
+    """
+    monkeypatch.setenv("COPIER_CACHE_DIR", str(tmp_path_factory.mktemp("cache")))
```
sisp: I think we can remove this environment variable monkeypatching because the `copier_cache_dir` fixture already takes care of it.

tests/conftest.py
```diff
@@ -95,6 +95,19 @@ def default_gitconfig(default_gitconfig: GitConfig) -> GitConfig:
...
+
+    Avoids polluting the user's real Copier cache with the templates cloned
+    during the test suite.
+    """
+    cache_dir = tmp_path_factory.mktemp("copier_cache")
+    with pytest.MonkeyPatch.context() as monkeypatch:
+        monkeypatch.setenv("COPIER_CACHE_DIR", str(cache_dir))
+        yield
+
+
```
sisp: Can we move this fixture elsewhere in this file, e.g. below the next fixture, so it isn't between the Git-related fixtures? Just cosmetics ... 🤓

tests/conftest.py
```diff
@@ -95,6 +95,19 @@ def default_gitconfig(default_gitconfig: GitConfig) -> GitConfig:
     return default_gitconfig
 
 
+@pytest.fixture(scope="session", autouse=True)
+def copier_cache_dir(tmp_path_factory: pytest.TempPathFactory) -> Iterator[None]:
```
sisp: Should we rename this fixture to simply `cache_dir` for better consistency with, e.g., `settings_path`? Or do you think there's a risk of a name collision because `cache_dir` is quite generic and might be used elsewhere? Since this fixture doesn't yield a value and is auto-used, we could also underscore-prefix it to avoid name collisions, e.g. `_cache_dir`.

sisp (changes requested):
Just some final polishing. 🤓

tests/conftest.py
DrKat0m: Yes, I like _cache_dir as the fixture name. Since it's an autouse fixture and not used directly by tests, the underscore prefix seems appropriate.

tests/test_vcs.py
```diff
@@ -95,6 +98,118 @@ def test_local_clone() -> None:
...
 
 
+@pytest.fixture
+def remote_repo(tmp_path_factory: pytest.TempPathFactory) -> str:
+    """A small local git repo exposed via a ``file://`` URL (treated as remote).
+
+    A ``file://`` URL is treated as remote by ``clone()``, so this exercises
+    the mirror-cache code path without requiring network access. The mirror
+    cache itself is redirected to a temp dir by the session-wide ``_cache_dir``
+    fixture in ``conftest.py``.
```
sisp: With the autouse fixture in `conftest.py`, this comment feels odd because it describes test setup outside this fixture.


```suggestion
    the mirror-cache code path without requiring network access.
```

tests/test_vcs.py
```diff
@@ -95,6 +98,118 @@ def test_local_clone() -> None:
...
+
+    Uses ``tmp_path_factory`` rather than ``tmp_path`` so the repo doesn't share
+    a path with a test's own ``tmp_path``.
```
sisp: This comment is clearly a result of review feedback, it feels odd here.


```suggestion
```

sisp (changes requested):
Sorry, some minor docstring polishing still. Other than that, ready to merge if CI passes.

sisp (approved):
Great work, @DrKat0m! 🏆 Thanks for contributing this super valuable enhancement! :bow:

Issue #1162, opened 2023-05-24
gitignored files removed after update


hparfr:
### Describe the problem

When a file from the template is in gitignore, this file is created with `copier copy` but then removed by `copier update`

I don't expect this file to be removed after an update.

### Template

```
.env
copier.yml
{{ _copier_conf.answers_file }}.jinja
```

### To Reproduce

```sh
$ git init && echo ".env" >> .gitignore && git add .gitignore && git commit -m "first"
$ copier copy file:///pathtothetemplate/.git .
Copying from template version 0.0.0.post10.dev0+a4f9293
 identical  .
    create  .copier-answers.yml
    create  .env

# answer to the questions
$ git add .copier-answers.yml && git commit -m "+answers"
$ ls .env
.env
$ copier update
# answer to the questions
Copying from template version 0.0.0.post10.dev0+a4f9293
 identical  .
 identical  .copier-answers.yml
 identical  .env

$ ls .env
error: file not found
```

### Expected behavior

1) consistency

after `copier copy && git add . && git commit -m"" `
`copier update` shoud not change a thing

2) feature

I expect to manage gitignored files with copier

### Operating system

Linux

### Operating system distribution and version

Ubuntu

### Copier version

copier 7.2.0

### Python version

Python 3.11.2

### Installation method

pip+pypi

### Additional context

Why I gitignore some files ?

I have a setup with two templates, one with answers shared with the team (project structure), and another with answsers not shared (db credentials, user related config...).
I make use of https://copier.readthedocs.io/en/stable/configuring/#applying-multiple-templates-to-the-same-subproject

hparfr:
This bug is still present in copier 8.3.0

PR #2730, opened 2026-06-18
fix(updating): don't exclude template-generated gitignored files from apply

### LINKED ISSUE #2729: gitignored files created by copy are deleted on update
### Describe the problem

This appears to be a reversion of the supposedly fixed  #1162.

### Template

Create a simple `copier.yml`:

```
_subdirectory: template

_message_after_copy: |
  Next steps:
    1. Add devcontainer.env to your .gitignore
    2. Fill in devcontainer.env at your project root with your credentials:
         GH_TOKEN=           # GitHub personal access token
         GIT_AUTHOR_NAME=    # Your name
         GIT_AUTHOR_EMAIL=   # Your email
         GIT_COMMITTER_NAME=
         GIT_COMMITTER_EMAIL=
    3. Open the project in VS Code and select "Reopen in Container"

_message_after_update: |
  Devcontainer updated. Your local changes have been merged with the template.
  devcontainer.env was not modified.
```

Add `template/devcontainer.env.jinja`.

### To Reproduce

Run `copier copy`  in a project that has a `.gitignore` with `*.env`

`devcontainer.env` is successfully created. 

Run `copier update`. 

`devcontainer.env` is silently deleted. 

### Logs

```console

```

### Expected behavior

The gitignored file should not be deleted. 

_No response_

### Operating system

macOS

15.4.1

### Copier version

9.15.2

### Python version

3.11

### Installation method

uvx+pypi

### Additional context

Copier installed via `uv tool install`

binggao1230:
Template-generated files matching .gitignore patterns were silently dropped during `copier update` because `git status --ignored` returned them and `--exclude` was unconditionally added for each to `git apply`.

Check if an ignored file exists in the fresh template copy before excluding it — if it does, it's template-managed and should be updated.

Fixes #2729, regression of #1162.

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1412,13 +1412,13 @@ class Worker:
                     "HEAD",
                     subproject_head,
                 ).splitlines()
-                exclude_plus_removed = list(
-                    set(self.exclude).union(
-                        f"/{escape_git_path(path)}"
-                        for path in map(normalize_git_path, files_removed)
-                        if not self.match_skip(Path(path))
-                    )
-                )
+                removed_paths = []
+                for path in map(normalize_git_path, files_removed):
+                    if (subproject_top / path).exists():
+                        continue
+                    if not self.match_skip(Path(path)):
+                        removed_paths.append(f"/{escape_git_path(path)}")
+                exclude_plus_removed = list(set(self.exclude).union(removed_paths))
             # Clear last answers cache to load possible answers migration, if
             # skip_answered flag is not set
             if self.skip_answered is False:
@@ -1543,9 +1543,7 @@ class Worker:
                         # Don't exclude template-generated files that happen to
                         # be gitignored — they should still be updated.
                         # Fixes #2729, regression of #1162.
-                        if (
-                            Path(new_copy) / normalize_git_path(filepath)
-                        ).exists():
+                        if (Path(new_copy) / normalize_git_path(filepath)).exists():
                             continue
                         apply_cmd = apply_cmd["--exclude", filepath]
                 (apply_cmd << diff)(retcode=None)
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -1539,7 +1539,15 @@ class Worker:
                 # adds `--exclude file1 --exclude file2` to `git apply` command
                 for filename in ignored_files.splitlines():
                     if filename.startswith("!! "):
-                        apply_cmd = apply_cmd["--exclude", filename[3:]]
+                        filepath = filename[3:]
+                        # Don't exclude template-generated files that happen to
+                        # be gitignored — they should still be updated.
+                        # Fixes #2729, regression of #1162.
+                        if (
+                            Path(new_copy) / normalize_git_path(filepath)
+                        ).exists():
+                            continue
+                        apply_cmd = apply_cmd["--exclude", filepath]
                 (apply_cmd << diff)(retcode=None)
                 if self.conflict == "inline":
                     conflicted = []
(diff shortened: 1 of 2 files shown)


sisp:
Thanks for contributing a fix for #2729, @gaoflow! :bow:

We'll need a test case to guard against future regression, probably a test function like

```python
def test_update_ignores_gitignored_files(
) -> None:
    build_file_tree(
        {
            src / "{{ _copier_conf.answers_file }}.jinja": (
                "{{ _copier_answers|to_yaml }}"
            ),
            src / ".env.jinja": "VERSION=v1",
        }
    )
    with local.cwd(src):
        git("init")
        git("add", ".")
        git("commit", "-m1")
        git("tag", "v1")

    build_file_tree({src / ".env.jinja": "VERSION=v2"})
    with local.cwd(src):
        git("add", ".")
        git("commit", "-m2")
        git("tag", "v2")

        git("commit", "-m1")

    run_copy(str(src), dst, vcs_ref="v1")
    assert (dst / ".env").is_file()
    assert (dst / ".env").read_text() == "VERSION=v1"
    with local.cwd(dst):
        git("add", ".")
        assert ".env" not in git("ls-files", "--", ".env").splitlines()
        git("commit", "-m2")

    run_update(dst, overwrite=True)
    assert (dst / ".env").is_file()
    assert (dst / ".env").read_text() == "VERSION=v2"
```

could be added to `tests/test_updatediff.py`. Does your fix pass this test?

binggao1230:
Thanks for the concrete test case. I added a regression test for the `.env.jinja` -> `.env` case, and it did fail against my previous patch: the ignored generated file was absent from the destination Git tree, so it was being treated as intentionally removed and then deleted during the old/new copy comparison.

I pushed an additional fix so removed-path exclusions now skip paths that still exist in the destination worktree. That keeps untracked/gitignored generated files eligible for update while still preserving the existing behavior for files that were actually removed.

Verified locally:

- `uv run pytest -o addopts="" tests/test_updatediff.py`
- `uv run ruff check copier/_main.py tests/test_updatediff.py`
- `uv run ruff format --check copier/_main.py tests/test_updatediff.py`

copier/_main.py
```diff
@@ -1412,13 +1412,13 @@ def _apply_update(self) -> None:  # noqa: C901
...
-                        if not self.match_skip(Path(path))
-                    )
-                )
+                removed_paths = []
+                for path in map(normalize_git_path, files_removed):
+                    if (subproject_top / path).exists():
+                        continue
+                    if not self.match_skip(Path(path)):
+                        removed_paths.append(f"/{escape_git_path(path)}")
+                exclude_plus_removed = list(set(self.exclude).union(removed_paths))
```
sisp: I think we can simplify this (and thus reduce the diff):


```suggestion
                exclude_plus_removed = list(
                    set(self.exclude).union(
                        f"/{escape_git_path(path)}"
                        for path in map(normalize_git_path, files_removed)
                        if not (subproject_top / path).exists()
                        and not self.match_skip(Path(path))
                    )
                )
```

sisp (changes requested):
Thanks for the update, looks great! 👌 Just a minor stylistic request.

binggao1230:
Addressed the stylistic request by simplifying the removed-path exclusion expression.

Verified locally:
- `uv run pytest -o addopts="" tests/test_updatediff.py`
- `uv run ruff check copier/_main.py tests/test_updatediff.py`
- `uv run ruff format --check copier/_main.py tests/test_updatediff.py`
- `git diff --check upstream/master...HEAD`

sisp (approved):
Thanks for contributing the fix for #2729, @gaoflow! :bow: LGTM! :tada:

Issue #1446, opened 2023-12-03
Workflow for updating from a local clone of a template


oprypin:
### Actual Situation

I have a few projects using the template `gh:oprypin/py-project-template`,
and I also have a local clone of the template `~/py-project-template`.

Currently Copier expects the following workflow from maintainers of a public template:

1. Edit something in the local clone
2. **Commit the template and publicly push it**
3. In a project using this template: `copier update -A -r HEAD`
4. Look at the new changes in the real project

I am not fully satisfied with this workflow because at first I want to *try* the template - whether it produces valid syntax and whether the destination project still works OK with the changes. And only then publish the changes.

### Desired Situation

I want the following workflow to be possible:

1. Edit something in the local clone
2. In a project using this template: `copier update -A -r HEAD`
3. Look at the new changes in the real project
4. If it looks good, **commit the template and publicly push it**

But Copier doesn't work that way, so instead this is what I have to do:

1. Edit something in the local clone
2. In a project using this template:
    1. `sed -i 's@gh:oprypin/py-project-template@~/py-project-template@' copier-answers.yml`
    2. `git commit -a` (because dirty state is not allowed)
    3. `copier update -A -r HEAD`
    4. `sed -i 's@~/py-project-template@gh:oprypin/py-project-template@' copier-answers.yml`
3. Look at the new changes in the real project
4. If it looks good, commit the template and publicly push it

### Proposed solution

I think my suggested workflow would be nice for many people and I wish it could just work. I want to publicly refer to my template so others can use it, but I want to use the local clone whenever I personally work with it.

One simple way is to support passing a **different path of a template as a flag**. Notably, I would like this to also **not write the overridden path of the template back into the answers file**, otherwise the last revert step will still be necessary.

Or perhaps even better, a mapping of such overrides could be a **permanent global config**: I want to tell copier that whenever it sees `gh:oprypin/py-project-template`, it should look at `~/py-project-template` instead.

In the meantime I hardcoded this override into my fork of Copier https://github.com/copier-org/copier/commit/ec09a7eefc6a48a348568c2cd5d358aac749ff7c and this works well for me.

pawamoy:
> otherwise the last revert step will still be necessary.

Let say Copier has a flag to specify a local template path when updating.
If it's only for testing your latest template changes, shouldn't you:

- update your project from the local template
- check that everything is OK
- **revert everything**
- commit and push the template
- do a real update from the remote template?

[...]

PR #2747, opened 2026-06-29
Fix issue live preview

### LINKED ISSUE #1451: Add template preview
### Actual Situation

During developing [my pretty advanced template](https://github.com/bswck/skeleton), when I want to check how my update will render in the downstream projects, I usually:
- commit my change,
- push it (up to some point I did push to a `dev` branch, but it became exhausting),
- open a special [downstream project](https://github.com/bswck/downstream) (lucky if I have skeleton and downstream open in one workspace, but it's still pretty hard to work with),
- `poe bump` (wrapper around `copier update`)
- see what needs fix,
- open back the skeleton project,
- commit my fix,
- push it...

I know I can update from a template in a directory that is locally in my filesystem, but it would conflict with my `poe bump` command that not only is configured to pull changes from the skeleton repository on GitHub, but also automatically makes a commit.

And this is a pretty messy workflow and it definitely shouldn't be like that.

### Desired Situation

I can easily inspect how my copier template renders as I am developing it, ideally with hot-reloading.

### Proposed solution

A browser-based or IDE-based GUI for previewing a template that is made in a temporary directory as the developer makes changes with hot-reloading (I suggest [watchfiles](https://github.com/samuelcolvin/watchfiles#readme)).
My goal would be getting a DX like when writing a React application.

karenemily:
## What does this PR do?

Adds a new `copier preview` command that allows template developers to preview
how their template renders locally with hot-reloading. When a developer runs
`copier preview <template_src> <destination_path>`, Copier renders the template
immediately and then watches the template directory for any file changes,
automatically re-rendering whenever a change is saved.

## Why was this PR needed?

Currently, template developers have no way to preview how their template renders
without going through a slow manual cycle of committing changes, pushing to
GitHub, switching to a downstream project, and running `copier update`. This
makes template development extremely tedious and slow, especially for small
iterative changes.

This PR implements the feature requested in issue #1451 by adding a
`run_preview` function that wraps the existing `_render_template` logic in a
`watchfiles` file watcher loop, and exposes it as a `copier preview` CLI
subcommand.

## What are the relevant issue numbers?

Closes #1451

## Does this PR meet the acceptance criteria?

- [x] Tests added for new/changed behavior
- [x] All tests passing (2 pre-existing Windows-only failures unrelated to
      this change)
- [x] Follows project style guide
- [x] No breaking changes introduced
- [ ] Documentation updated (if applicable)

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -3,7 +3,7 @@
 **Contribution Number:** 1
 **Student:** Karen Emily Muhwezi
 **Issue:** https://github.com/copier-org/copier/issues/1451
-**Status:** Phase II In Progress
+**Status:** Phase III In Progress
 
 ---
 
@@ -55,7 +55,7 @@ and running multiple commands every single time a change is made.
 ### Affected Components
 
 - Copier's CLI layer — a new `preview` or `watch` command needs to be added
-- Core template rendering pipeline in `copier/main.py` — the existing render 
+- Core template rendering pipeline in `copier/_main.py` — the existing render 
   logic will be reused
 - A new file watching component using the `watchfiles` library to detect 
   changes and trigger re-renders
@@ -80,9 +80,13 @@ x Failed to build copier @ file:///C:/Users/Administrator/codepath-copier
 |-> error sending request for url (https://files.pythonhosted.org/...)
 
 `-> operation timed out
-This appears to be a network connectivity issue on the current machine rather
-than a project configuration problem. Will retry on a stable network
-connection.
+
+Resolved by retrying on a stable network connection. Confirmed working
+environment by running `python -m uv run copier --version` which returned
+`copier 0.1.dev2247+g735b2ed0e`.
+
+Note: `uv` is not on the system PATH on Windows, so all commands must be
+run as `python -m uv` instead of `uv` directly.
 
 ### Steps to Reproduce
 
@@ -96,7 +100,9 @@ connection.
 
 - **Commit showing reproduction:** [Link to commit in your fork]
 - **Screenshots/logs:** Network timeout error documented above
-- **My findings:** [What you discovered during reproduction]
+- **My findings:** The issue is a missing feature — there is no existing 
+  preview or watch command in Copier's CLI. Confirmed by searching the 
+  codebase for any existing preview functionality and finding none.
 
 ---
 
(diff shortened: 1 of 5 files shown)


karenemily:
Hi @sisp — this is my first contribution to Copier. I've implemented the live preview feature requested in #1451 by adding a  copier preview command with hot-reloading support using `watchfiles`. Would appreciate a review when you have time!

CONTRIBUTION.md
sisp, line 1: I don't understand this file, it doesn't belong in this repo.

sisp (commented):
I don't think the current implementation is a correct solution to the problem.

- It is insufficient to run `Worker._render_template` on source file changes – this ignores changes in the questionnaire, which likely affects the render context, and doesn't (re)run tasks, which may be essential to the render output.
- It doesn't remove files that were previously rendered but are no longer rendered after a template change – files are only added or replaced.

I _imagine_ a (more) correct solution might involve making a full copy (using `run_copy`) in a temporary destination, computing the filetree diff between the temporary copy and the user-specified destination, and syncing the latter to the former. But this is just the first idea that has come to my mind right now. I imagine the devil is in the details. For example, for good UX you'll likely want to ask only questions that were changed in `copier.yml`, but a template might not render an answers file which is what `copier update --skip-answered` uses.

---

> Currently, template developers have no way to preview how their template renders
> without going through a slow manual cycle of committing changes, pushing to
> GitHub, switching to a downstream project, and running `copier update`. This
> makes template development extremely tedious and slow, especially for small
> iterative changes.

I think serious template authors need a test suite to test their templates, this manual cycle is not sustainable for non-trivial templates. FWIW, there's a [trick that lets you avoid pushing to a Git platform](https://github.com/copier-org/copier/issues/1446#issuecomment-1837543138) when testing a local change on an existing Copier-managed project.

karenemily:
Thank you for the detailed feedback @sisp! I understand the issues with the current approach. Before I attempt a revised implementation, I have a few questions to make sure I'm on the right track:

1. For the temporary directory approach, when `run_copy()` runs into the temp directory on each change, how should we handle the questionnaire? Should we capture the answers from the first run and pass them via the `data` parameter on subsequent runs to avoid re-prompting the user each time a file changes?

2. For syncing the temp directory to the destination, does Copier already have utilities for computing file diffs between two directories, or should we use Python's built-in `filecmp` module?

3. For deleted files, how do we distinguish between files that were rendered by Copier and files the user may have created themselves in the destination folder, so we only delete the former?

4. For tasks, should tasks re-run on every file change, or only on the initial render?

I want to make sure I implement this correctly before writing more code. Thanks for your patience with a first-time contributor!

sisp:
Sorry for the long delay to reply. These are important design and implementation questions that likely need some deep thinking. I'm afraid I don't have the capacity to do this work at the moment plus I don't consider this a high-priority feature because I believe serious template authors need to have automated tests for their templates anyway, so TDD seems like the superior approach that makes live previews mostly obsolete. Even before the above points you've raised, there are UX questions like how should Copier behave when questions are changed in arbitrary ways when those also affect downstream questions? Should only the changed questions be prompted? And should downstream questions, e.g. those with templated default values, be re-computed even when they are not computed values (i.e., not `when: false`)? A live preview update feels similar to a `copier update` run, but it seems much more dynamic. For example, structural changes like changes in the questionnaire are typically addressed by migration tasks on `copier update`, but I don't think this works for live previews because they are too dynamic.

I'd be happy to review and discuss a sound proposal, but I don't have the capacity to do this work myself at the moment. Thanks for your understanding.

PR #2754, opened 2026-07-05
fix: use destination path as working directory while rendering

### LINKED ISSUE #1708: Current working directory during `copier {copy,recopy,update}` should be destination path
### Describe the problem

While reimplementing `jinja-ansible-filters` (see #1398) – specifically testing the `fileglob` filter –, I noticed that the current working directory while running rendering template files through `copier {copy,recopy,update}` is the directory from which the command is executed, but I think it should be the destination path; otherwise, the `fileglob` filter (and other custom filesystem-based Jinja filters) wouldn't be particularly useful. WDYT?

### Template

[example.zip](https://github.com/user-attachments/files/16348922/copier-cwd-fileglob.zip)

The filesystem layout is as follows:

```
.
├── cwd.txt
├── dst
│   ├── a.txt
│   └── b.txt
└── src
    └── fileglob.txt.jinja
```

The example assumes that `src/` contains a template from which a *sub*project will be generated into `dst/`, which already contains files (e.g., from another Copier template).

### To Reproduce

1. Extract the content of the reproducible example in the ZIP archive above.
2. Change your current working directory to the root directory where you extracted archive (i.e. where the file `cwd.txt` is located).
3. Run `copier copy src/ dst/`.
4. Note the content of the generated file `dst/fileglob.txt`:

    ```console
    $ cat dst/fileglob.txt
    ['cwd.txt']
    ```

    As you can see, the file `cwd.txt` is listed as a result of `{{ '*.txt' | fileglob }}`.

### Logs

_No response_

### Expected behavior

I'd expect the following behavior:

```console
$ copier copy src/ dst/
$ cat dst/fileglob.txt
['a'.txt', 'b.txt']
```

I.e., filesystem-related Jinja filters should operate relative to the destination path.

That said, filesystem-related Jinja filters operating on the destination path may lead to unexpected behavior because Copier will generate files in arbitrary order, so the output of the `fileglob` filter may not be consistent across multiple runs.

_No response_

### Operating system

Linux

Ubuntu 22.04

### Copier version

9.3.1

### Python version

CPython 3.12

### Installation method

pipx+pypi

### Additional context

_No response_

nikzart:
Fixes #1708.

## What

While rendering a subproject through `copier {copy,recopy,update}`, the working directory was whatever directory Copier was invoked from, so filesystem-related Jinja filters and extensions (e.g. the `fileglob` filter from the always-loaded `jinja2_ansible_filters.AnsibleCoreFiltersExtension`) operated relative to the invocation directory. With this PR, rendering happens with the destination path as the working directory, so the reproducible example from #1708 produces the expected output:

```console
$ copier copy src/ dst/
$ cat dst/fileglob.txt
['a.txt', 'b.txt']
```

## Design decisions

Since #1708 ended with "WDYT?", here are the choices made, explicitly:

- **Scope**: the working-directory change wraps the whole render loop in `_render_template` (both path-name and content rendering), so it applies to `copy`, `recopy`, and `update` — all three renders of the update flow go through `run_copy`, so the temporary old/new copies render with the working directory at their own temporary destination, and the render into the real destination runs with the working directory there. Tasks and migrations already run with the destination as working directory, so this aligns rendering with the rest of the codebase.
- **Destination creation**: the destination folder is now created eagerly at the start of rendering, because it must exist to change into it. This matches `run_copy`'s documented behavior ("If `dst_path` was missing, it will be created"); the only observable difference is that a template which renders zero files now still creates the (empty) destination folder.
- **Pretend mode**: never creates the destination. If the destination already exists, rendering happens with the working directory there, like a real run; if it is missing, the working directory is left unchanged, since creating it would violate `--pretend`.
- **External data**: `_external_data` now resolves against the absolute destination path instead of the possibly-relative `dst_path`, because external data may be loaded lazily *during* rendering, when the working directory is the destination and a relative `dst_path` would no longer resolve correctly. No behavior change otherwise.
- **Non-determinism caveat**: as noted in the issue, files are generated in no particular order, so destination-relative filesystem operations may or may not observe files generated earlier in the same run. This is called out in a new docs note under `jinja_extensions`.

## Tests

- `copy`: working directory is the destination while rendering, with both absolute and relative `dst_path`, and the invocation directory is restored afterwards (a decoy file next to the invocation directory guards against regressions).
- `update`: working directory is the real destination while rendering into it.
- `pretend`: a missing destination is not created.
- external data loading lazily during rendering with a relative `dst_path`.

Full test suite passes locally (1143 passed), as do `mypy`, `ruff check`, `ruff format`, and `codespell`.

---

@sopw686 asked about working on this back in June — apologies if there's any overlap; happy to defer to your PR if you have one in progress.

### DIFF
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -596,7 +596,7 @@ def test_update_render_cwd_is_destination(
         git_init("v1")
         git("tag", "v1")
     run_copy(str(src), dst, defaults=True, overwrite=True)
-    build_file_tree({(dst / "a.txt"): "", (invocation / "decoy.txt"): ""})
+    build_file_tree({(invocation / "decoy.txt"): ""})
     with local.cwd(dst):
         git_init("copied")
     with local.cwd(src):
@@ -608,7 +608,13 @@ def test_update_render_cwd_is_destination(
         cwd_before = Path.cwd()
         run_update(dst, defaults=True, overwrite=True)
         assert Path.cwd() == cwd_before
-    assert (dst / "fileglob.txt").read_text() == "['a.txt', 'version.txt']"
+    # NOTE: Deliberately assert only files rendered by the template, not files
+    # added by the user in the destination. Whether the latter are visible to
+    # filesystem filters during an update depends on whether the update
+    # algorithm renders into the user-specified destination, which is an
+    # implementation detail that may change.
+    # https://github.com/copier-org/copier/pull/2754#discussion_r3529453935
+    assert (dst / "fileglob.txt").read_text() == "['version.txt']"
 
 
 @pytest.mark.parametrize("subproject_path", [Path(), Path("subproject")])
--- a/tests/test_updatediff.py
+++ b/tests/test_updatediff.py
@@ -578,6 +578,39 @@ def test_update_from_tagged_to_head(tmp_path_factory: pytest.TempPathFactory) ->
     assert load_answersfile_data(dst)["_commit"] == f"v1-1-g{sha}"
 
 
+def test_update_render_cwd_is_destination(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> None:
+    """While updating, the working directory during rendering is the destination.
+
+    https://github.com/copier-org/copier/issues/1708
+    """
+    src, dst, invocation = map(tmp_path_factory.mktemp, ("src", "dst", "invocation"))
+    with local.cwd(src):
+        build_file_tree(
+            {
+                "{{ _copier_conf.answers_file }}.jinja": "{{ _copier_answers|to_nice_yaml }}",
+                "version.txt": "1",
+            }
+        )
+        git_init("v1")
+        git("tag", "v1")
+    run_copy(str(src), dst, defaults=True, overwrite=True)
+    build_file_tree({(dst / "a.txt"): "", (invocation / "decoy.txt"): ""})
+    with local.cwd(dst):
+        git_init("copied")
+    with local.cwd(src):
+        build_file_tree({"fileglob.txt.jinja": "{{ '*.txt' | fileglob | sort }}"})
+        git("add", ".")
+        git("commit", "-m2")
+        git("tag", "v2")
+    with local.cwd(invocation):
+        cwd_before = Path.cwd()
+        run_update(dst, defaults=True, overwrite=True)
+        assert Path.cwd() == cwd_before
+    assert (dst / "fileglob.txt").read_text() == "['a.txt', 'version.txt']"
+
+
 @pytest.mark.parametrize("subproject_path", [Path(), Path("subproject")])
 @pytest.mark.parametrize("skip_pattern", ["skip_me", "/skip_me"])
 def test_skip_update(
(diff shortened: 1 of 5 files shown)


tests/test_updatediff.py
```diff
@@ -578,6 +578,39 @@ def test_update_from_tagged_to_head(tmp_path_factory: pytest.TempPathFactory) ->
...
+    with local.cwd(src):
+        build_file_tree({"fileglob.txt.jinja": "{{ '*.txt' | fileglob | sort }}"})
+        git("add", ".")
+        git("commit", "-m2")
+        git("tag", "v2")
+    with local.cwd(invocation):
+        cwd_before = Path.cwd()
+        run_update(dst, defaults=True, overwrite=True)
+        assert Path.cwd() == cwd_before
+    assert (dst / "fileglob.txt").read_text() == "['a.txt', 'version.txt']"
```
sisp: I wonder whether `a.txt` should be listed by the `fileglob` filter during an update, especially considering #2376 which only replays a fresh copy based on the old template, renders a fresh copy based on the new template, and merges the latter into `HEAD` given the former, which is the common ancestor (i.e., a "simple" 3-way merge). The new update algorithm implementation in #2376 never copies into the user-specified destination [unlike the current update algorithm implementation](https://github.com/copier-org/copier/blob/454ec4244132bce478e60c4707ee418312ca8922/copier/_main.py#L1437-L1448) – and I _think_ that's the only reason why `a.txt` is listed by `fileglob` here. When we switch to a final version of #2376, the behavior asserted by this PR might change.

Perhaps the more important question is even: Should the CWD be the user-specified destination during an update, or should it be the destination where files are being rendered? The latter seems more consistent and predictable and less susceptible to unexpected side-effects to me at first glance. Also, updating currently relies on replaying a fresh copy based on the old template, but from my perspective [this is conceptually flawed](https://github.com/copier-org/copier/issues/1170) – restoring a cached old copy would be correct IMO – although it's a mostly sufficient approximation of the correct approach. If we ever find/implement a good way cache and restore that cached old copy, setting the CWD to the destination where files are being rendered when replaying the fresh copy is more likely to yield the same result because the replayed copy is not affected by the content of the user-specified destination. We already have this [side-effect problem with loading external data from the user-specified destination](https://github.com/copier-org/copier/pull/2187).

 WDYT?

sisp (commented):
Thanks for taking the time to tackle #1708. It looks pretty solid, but I have a conceptual question.

tests/test_updatediff.py
nikzart: Good point — and I think we're aligned: the chdir lives inside `_render_template`, so each render's CWD is its *own* `dst_path` — the old-copy replay renders with CWD at the old-copy temp dir, the new-copy render at the new-copy temp dir, never the user-specified destination as such. The only reason `a.txt` showed up in this test is that the current update algorithm's "normal update in final destination" step renders into the real destination, so for that one render the two notions coincide. Under #2376, no render would target the user-specified destination, so `fileglob` would never see user files — which I agree is the more consistent and predictable semantics (and composes better with restoring a cached old copy per #1170).

So the `a.txt` assertion was pinning an artifact of the current algorithm rather than intended behavior. I've pushed a commit that drops the user-added file and asserts `['version.txt']` instead, which asserts only the invariant both algorithms share: renders see the destination being rendered, never the invocation directory (the decoy file still guards the original bug from #1708). It should survive the switch to #2376 unchanged, and I left a note in the test explaining why user-added files are deliberately not asserted.

PR #2762, opened 2026-07-15
feat(update): preserve developer-owned regions across updates (#2184)

### LINKED ISSUE #2184: Add support for initial content
### Actual Situation

Some of our template files just contain dummy content or comments like ToDo markers.

They get replaced by the developers, right after starting the project from the copier template.

To ease their work, the comments and ToDo markers show where and how some things should be done.

As soon as the project gets updated from the copier template, the actual implementation is suggested to be replaced by the dummy content.

### Desired Situation

It would be great, if there is some special `initial content` marker to tell copier changes in this section should be ignored and not beeing updated from the template.

### Proposed solution

Maybe a special marker of included jinja extension could help.

DrKat0m:
Templates often ship files with dummy content or TODO markers that the developer replaces right after generation. Previously `copier update` treated those regions like any other template-owned content, so a change to the placeholder in the template could clobber the developer's work or raise a needless merge conflict.

Template authors can now wrap a developer-owned region in literal marker comments that survive rendering:

    # copier:preserve:start <optional-id>
    ...developer-owned content...
    # copier:preserve:end <optional-id>

During an update the content between the markers is captured from the current project and injected into every intermediate render (old render, new render, and the freshly rendered destination) so all three merge sides agree on the region. The merge therefore keeps the developer's content conflict-free, while the marker lines and surrounding content keep updating from the template as usual.

- Added `copier/_preserve.py` with the marker parsing, capture and restore primitives.
- Wired capture/restore into `Worker._apply_update`, scoping the scan to template-managed files discovered in the old render.
- Documented the feature in `docs/updating.md`.
- Added unit and integration tests in `tests/test_preserve.py`.

Closes #2184


DrKat0m:
Hi @sisp, whenever you have a chance, could you please take a look at this PR? I'd really appreciate any feedback on the approach (especially the choice to keep the region conflict free by aligning all three merge sides rather than post processing conflict output) and on the marker syntax. Happy to iterate on anything. Thank you!

DrKat0m:
Hi @sisp, just a gentle follow up whenever you have some time. I'd really appreciate it if you could take a look at this PR and share any feedback you have. In particular, I'd love to know your thoughts on the overall approach and whether the marker syntax feels appropriate. Thank you!

sisp:
Just a quick thought on this (because I'm on vacation 🌴 ): If the preserve markers must remain in the rendered files (as you document), then they must be valid comments of the respective file format. This means, a hardcoded `#`-comment won't work for files like `.js`, `.ts`, `.go`, `.c`, etc. because this isn't how comments work in those languages. Like, e.g., [Prettier's error suppression comments](https://prettier.io/docs/ignore), we might need a comprehensive set of marker variants depending on this file format / programming language. Unlike Prettier, Copier supports an open set of file formats – because Copier renders _any_ text file no matter the underlying format. So I'm not sure yet how to tackle this best.

sisp:
Ideally, the markers aren't part of a rendered file, then we could use Jinja syntax and be language-agnostic. Also, use of Copier wouldn't leak into rendered files.

Perhaps we could omit developer-owned sections from old and new copies that are created by the update algorithm? This way, the 3-way merge should ignore changes in the template in those sections. #2376 might make this even easier.

We could try this quickly by exporting `_copier_operation` also in the file render context and conditionally render a developer-owned block like this:

```jinja
{%- if _copier_operation != 'update' %}
...
{%- endif %}
```

If this works, I'd prefer a custom tag like

```jinja
{%- ignore %}
...
{%- endignore %}
```

for better DX.

DrKat0m:
Thanks for the feedback - enjoy the vacation! 🌴 You're right that hardcoded comment markers can't work across an open set of formats, so I reworked this the way you suggested.

Dropped the hardcoded markers in favor of a Jinja tag -

{% ignore -%}
def greeting() -> str:
    return "dummy"   # developer replaces this after generation
{%- endignore %}

It's a small extension compiling to {% if _copier_operation != 'update' %}…{% endif %} - rendered on copier copy, omitted from the internal renders during copier update. Being pure Jinja, nothing leaks into rendered files and it's fully language agnostic.

However, since the region is omitted from both merge sides, the developer's content re-applies as a diff hunk, clean when there's stable context around the block, but if the template edits a line directly adjacent to it, it degrades to a normal merge conflict.

DrKat0m:
Hi @sisp, just a gentle follow up whenever you have some time. I'd really appreciate it if you could take a look at this PR and share any feedback you have.

copier/_main.py
```diff
@@ -852,7 +859,7 @@ def _render_file(  # noqa: C901
                 new_content = src_abspath.read_bytes()
             else:
                 new_content = tpl.render(
-                    **self._render_context(), **(extra_context or {})
+                    {**self._render_context(), **(extra_context or {})}
```
sisp: This is diff noise, not strictly needed to implement the new feature.


```suggestion
                    **self._render_context(), **(extra_context or {})
```

copier/_main.py
```diff
@@ -1175,7 +1182,7 @@ def _render_string(
                 Additional variables to use for rendering the template.
         """
         tpl = self.jinja_env.from_string(string)
-        return tpl.render(**self._render_context(), **(extra_context or {}))
+        return tpl.render({**self._render_context(), **(extra_context or {})})
```
sisp: This is diff noise, not strictly needed to implement the new feature.

```suggestion
        return tpl.render(**self._render_context(), **(extra_context or {}))
```

docs/updating.md
```diff
@@ -255,6 +255,85 @@ git clean -d -i     # remove untracked files and folders
...
+your `greeting()` is kept even if the template changed the placeholder, while `HEADER`
+and `FOOTER` still receive template updates.
+
+!!! tip "Nothing leaks into the rendered file"
+
+    Unlike a comment-based marker, the `{% ignore %}` tag is pure Jinja — it is stripped
+    during rendering, so no Copier-specific syntax ends up in your generated project. It
+    is therefore **language-agnostic**: it works in `.js`, `.go`, `.c`, `.rs` … or any
+    other text file, since it never has to be a valid comment in the target language.
+
```
sisp: I think this is pretty obvious, let's try keeping the docs concise. I presume AI was in a chatty mood. 😄


```suggestion
```

docs/updating.md
```diff
@@ -255,6 +255,85 @@ git clean -d -i     # remove untracked files and folders
...
+The tag relies on the `_copier_operation` render context variable, which is `"copy"`
+during generation and `"update"` during an update. If you prefer to be explicit, or need
+a condition the tag can't express, you can write the equivalent directly:
+
+```jinja
+{% if _copier_operation != "update" %}
+...developer-owned content...
+{% endif %}
+```
```
sisp: I'm not sure we should include this snippet, I don't quite see when anybody would want to use the explicit condition, so including it might suggest this is a reasonably common scenario.

```suggestion
```

docs/updating.md
```diff
@@ -255,6 +255,85 @@ git clean -d -i     # remove untracked files and folders
...
+
+    The `-%}` and `{%-` [whitespace-control][whitespace] markers strip the tag's own
+    lines so the output stays clean. Keep at least one unchanging line — such as the
+    blank lines above — directly around the block, and avoid editing the lines
+    immediately adjacent to it in later template versions. The update re-applies your
+    region as a diff against the surrounding lines; if the template changes a line right
+    next to the block, the merge falls back to a normal Copier
+    [conflict](#recover-from-a-broken-update) that you resolve by hand (your content is
+    never silently lost). Embedding the block among stable code, rather than at the very
+    top or bottom of a tiny file, gives the merge the anchors it needs.
```
sisp: I'm not sure we should phrase this as a recommendation because users may not be able to avoid touching adjacent lines that are part of the hunk context. It might be better to inform users about the behavior, so they aren't surprised when merge conflicts involving the ignored block occur.

tests/test_ignore.py
```diff
@@ -0,0 +1,178 @@
...
+    """Initial generation renders the block; no marker/tag survives."""
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    _commit_template(src, _V1, "v1")
+
+    run_copy(str(src), dst, defaults=True, overwrite=True, vcs_ref="v1")
+
+    rendered = (dst / "app.py").read_text(encoding="utf-8")
+    assert 'return "dummy v1"' in rendered  # scaffolding is present after copy
+    assert "ignore" not in rendered  # no Copier syntax leaks into the file
+    assert "{%" not in rendered and "%}" not in rendered
```
sisp: I'd omit this test because the (updated version of the) tests below also covers this implicitly.

tests/test_ignore.py
```diff
@@ -0,0 +1,178 @@
...
+def test_ignore_omits_body_on_update(env: SandboxedEnvironment) -> None:
+    template = env.from_string("a\n{% ignore %}b{% endignore %}\nc")
+    assert template.render({"_copier_operation": "update"}) == "a\n\nc"
+
+
+def test_ignore_trim_markers_clean_output(env: SandboxedEnvironment) -> None:
+    """The idiomatic ``-%}``/``{%-`` markers strip the tag lines entirely."""
+    template = env.from_string("a\n{% ignore -%}\nb\n{%- endignore %}\nc")
+    assert template.render({"_copier_operation": "copy"}) == "a\nb\nc"
+    assert template.render({"_copier_operation": "update"}) == "a\n\nc"
```
sisp: I'm not sure we need these tests. In the end, what really matters is the behavior involving `run_copy` and `run_update`.

tests/test_ignore.py
```diff
@@ -0,0 +1,178 @@
...
+
+def another_stable():
+    return "keep"
+"""
+
+_V2 = (
+    _V1.replace('CONFIG = "v1"', 'CONFIG = "v2"')
+    .replace('return "dummy v1"', 'return "dummy v2"')
+    .replace('return "keep"', 'return "kept in v2"')
+)
```
sisp: I think it would be more idiomatic to create a fixture that creates the test template with two commits/tags rather than having a helper function and constants. For brevity and convenience, you can use the [`git_save`](https://github.com/copier-org/copier/blob/69af2685efa4ab71455fd082ca539c344403e4c6/tests/helpers.py#L165-L170) helper function instead of the `git init/add/commit/tag` command sequence.

tests/test_ignore.py
```diff
@@ -0,0 +1,178 @@
...
+    _commit_template(src, _V2, "v2")
+    run_update(dst, defaults=True, overwrite=True, vcs_ref="v2")
+
+    result = app.read_text(encoding="utf-8")
+    assert "<<<<<<<" not in result  # clean merge
+    assert 'return "REAL IMPLEMENTATION"' in result  # developer content kept
+    assert 'return "dummy v2"' not in result  # template placeholder ignored
+    assert 'CONFIG = "v2"' in result  # surrounding code updated
+    assert 'return "kept in v2"' in result
+    assert "ignore" not in result  # still no leaked syntax
```
sisp: I think it would be simpler and more explicit to assert the full file content. You can use [`inline-snapshot`](https://15r10nk.github.io/inline-snapshot/latest/) to ease snapshot testing:

```suggestion
    assert result == snapshot()
```

For this, add `from inline_snapshot import snapshot` at the top. When you run `uv run pytest -n0 --inline-snapshot=review ...`, you'll be prompted to review the snapshot content and it'll be inserted for you if you agree.

Same applies to the other test below.

tests/test_ignore.py
```diff
@@ -0,0 +1,178 @@
...
+    """Developer content in an ``{% ignore %}`` block survives an update.
+
+    The template changes the surrounding code *and* the placeholder inside the
+    block; only the surroundings update while the developer's implementation is
+    kept, with no merge conflict.
+    """
+    src, dst = map(tmp_path_factory.mktemp, ("src", "dst"))
+    _commit_template(src, _V1, "v1")
+
+    run_copy(str(src), dst, defaults=True, overwrite=True, vcs_ref="v1")
```
sisp: Perhaps we can add an assertion that checks the content of `app.py` here as well. See my comment about snapshot testing below.

sisp (changes requested):
Thanks for updating the PR, @DrKat0m! :bow: I've left a few comments and suggestions.

tests/test_ignore.py
sisp, line 1: I think the tests look much better now. But would you mind adding a case where merge conflicts occur?

sisp (changes requested):
Looks much better, thanks! I've left an inline request regarding testing.

Full disclosure: I'm still a bit uncertain about this feature because we already know its limitations regarding changes in close proximity that cause merge conflicts. There's also no structural mechanism that prevents using the tag in a single line like

```jinja
This is {% ignore %}some developer-owned content{% endignore %}` that can be edited.
```

which, I believe, would always cause a merge conflict for any change outside the `{% ignore %}` block. I've been giving this more thought but haven't been able to think of a better approach so far. If we decide to merge this feature, I'm leaning towards making it a "preview" feature (inspired by [Ruff's preview mode](https://docs.astral.sh/ruff/preview/)) such that it requires opt-in by a template author (via a setting in `copier.yml`) and is not subject to stability guarantees. For this, we'd need to add the necessary preview mode infrastructure first.

PR #2799, opened 2026-08-10
docs: Fix broken Star History chart

Mubelotix:
This new provider doesn't require API tokens

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,11 +213,11 @@ And thanks to all financial supporters and folks that give us a shiny star! ⭐
 
 <!-- rumdl-capture -->
 <!-- rumdl-disable MD033 -->
-<a href="https://star-history.com/#copier-org/copier&Date">
-  <picture>
-    <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/svg?repos=copier-org/copier&type=Date&theme=dark" />
-    <source media="(prefers-color-scheme: light)" srcset="https://api.star-history.com/svg?repos=copier-org/copier&type=Date" />
-    <img alt="Star History Chart" src="https://api.star-history.com/svg?repos=copier-org/copier&type=Date" />
-  </picture>
+<a href="https://star-history.dera.page/#copier-org/copier&type=date&legend=top-left">
+ <picture>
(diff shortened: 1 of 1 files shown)


sisp:
Thanks for contributing a fix for the GitHub star history chart, @Mubelotix! :bow:

I have two questions:

1. Is this service hosted by you?

1. How does this service overcome the new GitHub requirements on accessing the data that feed into the chart generation?

    > GitHub is restricting its **stargazers API** — the endpoint that returns who starred a repository and when. As [announced on June 30, 2026](https://github.blog/changelog/2026-06-30-upcoming-access-restrictions-to-public-api-endpoints-and-ui-views/), access to this data is being **limited to a repository's own admins and collaborators**.
    >
    > &mdash; https://www.star-history.com/blog/github-stargazer-api-restriction/

Mubelotix:
Yes it is hosted by me, I happened to have the data available on my disk because I was building [another project](https://github.com/Mubelotix/SimRepo) that required it. Since github restricted the API, I just have to keep the counts up-to-date by fetching them from time to time

sisp:
And how do you intend to keep the counts up-to-date? By scraping the website? Because the stargazers API seems to require write permissions to the repo.

Mubelotix:
The history is the hard part and it's solved already. To keep up-to-date I can just query the repository metadata, it returns the stargazers count and it always will

sisp:
This means, you need to retrieve the stargazers count on an interval, right? ... which you can do for a list of repos but, it isn't a general solution that works for _all_ repos on github.com, is it? TBH, it doesn't sound like a _proper_ solution to the problem. But I also admit that I'm not eager to give a PAT with write permissions to star-history.com, that's also the main reason why I haven't fixed the chart yet.

Mubelotix:
> which you can do for a list of repos but, it isn't a general solution that works for all repos on github.com, is it

You are right. To be honest if I don't expect to ever be able to support repositories with just 50 stars, I won't ever be able to collect this much

>  I also admit that I'm not eager to give a PAT with write permissions to star-history.com

This is what motivated me to fork, such pattern is just a really bad idea and Github should be ashamed to have caused this

sisp:
GitHub has resolved the problem that broke the star history chart. See https://www.star-history.com/blog/new-github-star-history-api for more information.

PR #2801, opened 2026-08-11
feat: allow an exclude entry to render to multiple newline-separated patterns

sjortiz:
## What

Split each rendered `_exclude` entry on newlines, so a **single Jinja block can conditionally emit a whole list of patterns** instead of repeating the condition on every entry.

```yaml
skip_ci: bool
_exclude:
  - |
    {% if skip_ci %}
    .github/workflows/ci.yml
    .github/workflows/deploy.yml
    docs/ci/
    {% endif %}
```

When `skip_ci` is false the block renders empty and nothing extra is excluded.

## Why

`_exclude` patterns are templated per-item, so conditionally excluding a *group* of files today requires repeating the guard on every line:

```yaml
_exclude:
  - "{% if skip_ci %}.github/workflows/ci.yml{% endif %}"
  - "{% if skip_ci %}.github/workflows/deploy.yml{% endif %}"
  - "{% if skip_ci %}docs/ci/{% endif %}"
```

Authoring a multi-line block instead **silently does nothing**: the rendered value reaches `PathSpec.from_lines` as a single string containing newlines, i.e. one malformed pattern that matches no path. This change makes the intuitive form work.

## How

In `Worker.match_exclude`, each rendered exclusion is split on newlines and blank lines are dropped before building the `PathSpec`. This mirrors gitignore-style semantics (one pattern per line, blanks ignored).

- **Backward-compatible:** a single-line entry splits to itself; existing templates are unaffected (all 13 pre-existing exclude tests still pass).
- **No-op safety:** a false conditional renders to whitespace → dropped → excludes nothing.

## Tests

Added a parametrized `test_config_exclude_multiline_conditional` (true/false) asserting a one-block conditional excludes the whole list when the guard is true and nothing when false. Full `tests/test_exclude.py` passes (15 passed, 1 pre-existing xfail); `ruff check` + `ruff format` clean.

## Docs

Documented the multi-line behavior with an example under the `exclude` setting in `docs/configuring.md`.

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -769,10 +769,20 @@ class Worker:
         # Note: This method is a cached property, it needs to be regenerated
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
-        return self._path_matcher(
-            self._render_string(exclusion, extra_context=extra_context)
-            for exclusion in self.all_exclusions
-        )
+
+        # A single exclusion entry may render to multiple newline-separated
+        # patterns. This lets one Jinja block conditionally emit a whole list of
+        # patterns (e.g. `{% if flag %}a\nb\nc{% endif %}`) instead of repeating
+        # the condition on every entry. Blank lines (including the empty render of
+        # a false conditional) are dropped, so a no-op renders to nothing.
+        def _rendered_patterns() -> Iterable[str]:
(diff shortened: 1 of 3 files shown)


sjortiz:
Closing here — moving this to our fork for internal review first.

PR #2804, opened 2026-08-11
feat: Adds support for multiline exclusion under a single flag

### LINKED ISSUE #2803: support group file exclusion under a single condition
### Actual Situation

we have a single copier template for multiple types of deployments (thus multiple github actions pipelines), which lead us to a copier.yaml file that repeats the same condition per line on multiple lines, making it harder to update track. 

### Desired Situation

I want the copier.yaml exclude section to support group exclusions

### Proposed solution

So I'm proposing we support doing group exclusions:
```
- |
  {% if flag %}
  a.txt
  b.txt
  {% endif %}
- file-c
```
instead of

```
 - "{% if app_pipeline_disabled %}.github/workflows/deploy-type-b.yml{% endif %}"
 - "{% if app_pipeline_disabled %}.github/workflows/deploy-type-b.yml{% endif %}"
 - "{% if app_pipeline_disabled %}.github/workflows/security-check.yml{% endif %}"
....
```

sjortiz:
# description
This pr introduces the capability to do multiline exclusion under the same flag, removing the need to do a condition per line, thus resulting in cleaner config files

# approach
This one was a quick solution, after the logic is rendered it causes the result to be a pattern that doesn't match anything as 
`.github/workflows/ci.yml \n.github/workflows/deploy.yml\n docs/ci/` this will never match, so I then split by line jump

fixes #2803

### DIFF
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -769,18 +769,12 @@ class Worker:
         # Note: This method is a cached property, it needs to be regenerated
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
-
-        # A single exclusion entry may render to multiple newline-separated patterns.
-        # Blank lines are dropped, so an empty conditional renders to nothing.
-        def _rendered_patterns() -> Iterable[str]:
-            for exclusion in self.all_exclusions:
-                rendered = self._render_string(exclusion, extra_context=extra_context)
-                for line in rendered.splitlines():
-                    stripped = line.strip()
-                    if stripped:
-                        yield stripped
-
-        return self._path_matcher(_rendered_patterns())
+        return self._path_matcher(
+            chain.from_iterable(
+                self._render_string(exclusion, extra_context=extra_context).splitlines()
+                for exclusion in self.all_exclusions
+            )
+        )
 
     @cached_property
     def match_skip(self) -> Callable[[Path], bool]:
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -769,10 +769,20 @@ class Worker:
         # Note: This method is a cached property, it needs to be regenerated
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
-        return self._path_matcher(
-            self._render_string(exclusion, extra_context=extra_context)
-            for exclusion in self.all_exclusions
-        )
+
+        # A single exclusion entry may render to multiple newline-separated
+        # patterns. This lets one Jinja block conditionally emit a whole list of
+        # patterns (e.g. `{% if flag %}a\nb\nc{% endif %}`) instead of repeating
+        # the condition on every entry. Blank lines (including the empty render of
+        # a false conditional) are dropped, so a no-op renders to nothing.
+        def _rendered_patterns() -> Iterable[str]:
+            for exclusion in self.all_exclusions:
+                rendered = self._render_string(exclusion, extra_context=extra_context)
+                for line in rendered.splitlines():
+                    if line.strip():
+                        yield line
+
+        return self._path_matcher(_rendered_patterns())
 
     @cached_property
     def match_skip(self) -> Callable[[Path], bool]:
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import os
 import platform
+import re
 import stat
 import subprocess
 import sys
@@ -769,9 +770,12 @@ class Worker:
         # Note: This method is a cached property, it needs to be regenerated
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
+        regex_linebreak = re.compile(r"\r\n|\r|\n")
         return self._path_matcher(
             chain.from_iterable(
-                self._render_string(exclusion, extra_context=extra_context).splitlines()
+                regex_linebreak.split(
+                    self._render_string(exclusion, extra_context=extra_context),
+                )
                 for exclusion in self.all_exclusions
             )
         )
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -770,11 +770,15 @@ class Worker:
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
 
-        # A single exclusion entry may render to multiple newline-separated
-        # patterns. This lets one Jinja block conditionally emit a whole list of
-        # patterns (e.g. `{% if flag %}a\nb\nc{% endif %}`) instead of repeating
-        # the condition on every entry. Blank lines (including the empty render of
-        # a false conditional) are dropped, so a no-op renders to nothing.
+        # A single exclusion entry can render multiple newline-separated patterns.
+        # This lets one Jinja block conditionally emit a whole list of patterns e.g.:
+        # {% if flag %}
+        #   - a
+        #   - b
+        #   - c
+        # {% endif %}`) instead of repeating {% if flag %} on every line
+        # Blank lines (including the empty render of a false conditional) are dropped
+        # so a no-op renders to nothing.
         def _rendered_patterns() -> Iterable[str]:
             for exclusion in self.all_exclusions:
                 rendered = self._render_string(exclusion, extra_context=extra_context)
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -770,21 +770,15 @@ class Worker:
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
 
-        # A single exclusion entry can render multiple newline-separated patterns.
-        # This lets one Jinja block conditionally emit a whole list of patterns e.g.:
-        # {% if flag %}
-        #   - a
-        #   - b
-        #   - c
-        # {% endif %}`) instead of repeating {% if flag %} on every line
-        # Blank lines (including the empty render of a false conditional) are dropped
-        # so a no-op renders to nothing.
+        # A single exclusion entry may render to multiple newline-separated patterns.
+        # Blank lines are dropped, so an empty conditional renders to nothing.
         def _rendered_patterns() -> Iterable[str]:
             for exclusion in self.all_exclusions:
                 rendered = self._render_string(exclusion, extra_context=extra_context)
                 for line in rendered.splitlines():
-                    if line.strip():
-                        yield line
+                    stripped = line.strip()
+                    if stripped:
+                        yield stripped
 
         return self._path_matcher(_rendered_patterns())
 
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1121,7 +1121,9 @@ Each pattern can be templated using Jinja.
     entry individually:
 
     ```yaml
-    skip_ci: bool
+    skip_ci:
+        type: bool
+
     _exclude:
         - |
           {% if skip_ci %}
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -1112,10 +1112,15 @@ Each pattern can be templated using Jinja.
     The difference with [skip_if_exists](#skip_if_exists) is that it will never be
     rendered during an update, no matter if it exists or not.
 
-A single pattern may render to **multiple newline-separated patterns**. Each rendered
-line is treated as its own pattern (blank lines, including the empty render of a false
-conditional, are ignored). This lets one Jinja block conditionally emit a whole list of
-patterns instead of repeating the same condition on every entry.
+ A single exclusion entry can render multiple newline-separated patterns.
+ This lets one Jinja block conditionally emit a whole list of patterns e.g.:
+{% if flag %}
+    - a
+    - b
+    - c
+{% endif %}`) instead of repeating {% if flag %} on every line
+Blank lines (including the empty render of a false conditional) are dropped
+so a no-op renders to nothing.
 
 !!! example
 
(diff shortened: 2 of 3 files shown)


copier/_main.py
```diff
@@ -769,10 +769,18 @@ def match_exclude(self) -> Callable[[Path], bool]:
...
+
+        # A single exclusion entry may render to multiple newline-separated patterns.
+        # Blank lines are dropped, so an empty conditional renders to nothing.
+        def _rendered_patterns() -> Iterable[str]:
+            for exclusion in self.all_exclusions:
+                rendered = self._render_string(exclusion, extra_context=extra_context)
+                for line in rendered.splitlines():
+                    stripped = line.strip()
+                    if stripped:
+                        yield stripped
```
sjortiz: I wanted to do a walrus here but wasn't sure if it'd broke retro-compatibility or require systems to upgrade. I'm too new to this project to make such decisions.

copier/_main.py
```diff
@@ -769,10 +769,18 @@ def match_exclude(self) -> Callable[[Path], bool]:
...
+
+        return self._path_matcher(_rendered_patterns())
```
sisp: I think we can simplify this (requiring `from itertools import chain` in the import block):

```suggestion
        return self._path_matcher(
            chain.from_iterable(
                self._render_string(exclusion, extra_context=extra_context).splitlines()
                for exclusion in self.all_exclusions
            )
        )
```

`Worker._path_matcher` already accepts an iterable of strings (patterns), and valid patterns include empty lines and Gitignore comments, so there's no need to preprocess/filter the lines. See also the [PathSpec tutorial](https://github.com/cpburnz/python-pathspec/tree/master#tutorial) where lines are parsed individually:

```pycon
>>> from pathspec.patterns.gitignore.basic import GitIgnoreBasicPattern
>>> patterns = map(GitIgnoreBasicPattern, spec_text.splitlines())
>>> spec = PathSpec(patterns)
```

docs/configuring.md
```diff
@@ -1112,6 +1112,29 @@ Each pattern can be templated using Jinja.
     The difference with [skip_if_exists](#skip_if_exists) is that it will never be
     rendered during an update, no matter if it exists or not.
 
+    A single exclusion entry may render to multiple newline-separated patterns,
+    letting one Jinja block conditionally emit a whole list instead of repeating
+    the condition on every entry. Blank lines are dropped, so an empty conditional
+    renders to nothing.
```
sisp: I'd document the Jinja conditional as an example use case where a multiline exclusion entry comes in handy – as you've already done below. And how about describing the behavior to be equivalent to parsing a complete gitignore file instead of highlighting specific behavioral aspects?

```suggestion
    equivalent to a complete gitignore file including comments and blank lines.
```

In fact, we _could_ change the `_exclude` setting's value to be a (multiline) string and not a list of strings due to the change of this PR because there's no difference between multiple list items and one list item with multiple lines anymore. But I'd prefer keeping it as-is for backward compatibility and because plain exclude entries are probably more intuitive to write as individual list items than as lines in a multiline string. Just sharing my observation.

WDYT?

docs/configuring.md
```diff
@@ -1112,6 +1112,29 @@ Each pattern can be templated using Jinja.
...
+
+!!! example
+
+    Exclude a group of files with a single conditional block instead of guarding each
+    entry individually:
+
+    ```yaml
+    skip_ci: bool
```
sisp: This means `skip_ci` is a string-type question whose default value is `"bool"`.


```suggestion
    skip_ci:
        type: bool

```

docs/configuring.md
```diff
@@ -1112,6 +1112,29 @@ Each pattern can be templated using Jinja.
...
+    _exclude:
+        - |
+          {% if skip_ci %}
+          .github/workflows/ci.yml
+          .github/workflows/deploy.yml
+          docs/ci/
+          {% endif %}
```
sisp: - I think editorconfig-checker will require 4-space indentation for the YAML multiline string content.
- We should add a leading `/` to each path to anchor the paths at the subproject root, else any subpath matching these patterns will be excluded.


```suggestion
        - |
            {% if skip_ci %}
            /.github/workflows/ci.yml
            /.github/workflows/deploy.yml
            /docs/ci/
            {% endif %}
```

docs/configuring.md
```diff
@@ -1112,6 +1112,29 @@ Each pattern can be templated using Jinja.
...
+    ```
+
+    When `skip_ci` is false the block renders empty and nothing extra is excluded.
```
sisp:
```suggestion
    When `skip_ci` is `false` the block renders empty and nothing extra is excluded.
```

tests/test_exclude.py
```diff
@@ -54,6 +54,52 @@ def test_config_include(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+            src / "copier.yml": (
+                "skip: {type: bool}\n"
+                "_exclude:\n"
+                '  - "copier.yml"\n'
+                "  - |\n"
+                "    {% if skip %}\n"
+                "    a.txt\n"
+                "    b.txt\n"
+                "    sub/c.txt\n"
+                "    {% endif %}\n"
```
sisp: I think this is more readable, and we should add a leading `/` to each pattern to be precise (although it makes no difference in this test case)

```suggestion
                """\
                skip:
                    type: bool

                _exclude:
                    - copier.yml
                    - |
                        {% if skip %}
                        /a.txt
                        /b.txt
                        /sub/c.txt
                        {% endif %}
                """
```

tests/test_exclude.py
```diff
@@ -54,6 +54,52 @@ def test_config_include(tmp_path_factory: pytest.TempPathFactory) -> None:
...
+    if skip:
+        # When skip=True, the conditional block renders patterns, so files are excluded
+        assert not (dst / "a.txt").exists()
+        assert not (dst / "b.txt").exists()
+        assert not (dst / "sub" / "c.txt").exists()
+    else:
+        # When skip=False, the conditional block renders empty, so files are copied
+        assert (dst / "a.txt").exists()
+        assert (dst / "b.txt").exists()
+        assert (dst / "sub" / "c.txt").exists()
```
sisp: How about simplifying this?

```suggestion
    assert (dst / "keep.txt").exists()
    # Files are excluded when skip=True
    assert (dst / "a.txt").exists() is not skip
    assert (dst / "b.txt").exists() is not skip
    assert (dst / "sub" / "c.txt").exists() is not skip
```

sisp (changes requested):
Thanks for contributing this useful enhancement of the `_exclude` setting, @sjortiz! :bow: It has been on my mind for a while, but I've never actually acted on the idea. I thought it had been discussed before, but I can't seem to find any issue or comment about it – perhaps I'm hallucinating 😄. In any case, this is great!

I've left a few comments, mostly about simplifying things a bit. Looking forward to having this PR merged soon! 🤩

docs/configuring.md
sjortiz: agree on the backwards compatibility. I also prefer the list of strings to keep the visuals consistent

docs/configuring.md
sjortiz: my bad

tests/test_exclude.py
sjortiz: elegant

sjortiz:
I was surprised the test started failing after the last suggestions, but here is the address

sisp (commented):
Thanks for attempting a fix for the failing test case!

I've given this some deeper thought and learned a few things along the way.

1. The [gitignore spec](https://git-scm.com/docs/gitignore) seems to [not support patterns that include literal newline characters](https://software.codidact.com/posts/290259). So it seems like it should have not been possible to exclude a user-deleted file containing a `\n` character from an update by using its name as a synthesized exclude pattern in the internal `run_copy` call of the update algorithm. _But_ we don't use `git` to match exclude patterns, we use [PathSpec](https://github.com/cpburnz/python-pathspec), which _does_ support gitignore patterns with newline characters:

    ```pycon
    >>> from pathspec import PathSpec
    >>> filename = "\a\f\n\t\vcontrol\a\f\n\t\vcharacters\v\t\n\f\a"
    >>> spec = PathSpec.from_lines("gitignore", [f"/{filename}"])
    >>> spec.match_file(filename)
    True
    ```

1. Supporting multiline strings as exclude patterns means that each line is treated as a gitignore pattern, so lines must be split. We reuse the `exclude` setting to exclude user-deleted files from updates by synthesizing gitignore patterns for those files from the names of deleted files. Since POSIX filesystems support `\n` characters in filenames, synthesized patterns may include `\n` characters, so splitting a multiline string of exclude patterns into lines isn't safe.

1. The test case includes additional escape sequences like `\f` or `\v`. I just realized that [`str.splitlines` splits on a superset of universal newlines](https://docs.python.org/3/library/stdtypes.html#str.splitlines) including some of those in the test case. So, applying `str.splitlines` to the synthesized exclude patterns doesn't only split on `\n` but on a few more. That said, gitignore lines are separated only by universal newlines (according to my research), so we should limit ourselves to those characters, too.

1. The `NUL` character is forbidden in filenames of all major operating systems.

Given all of the above, I suggest to:

1. Revert to the state of a68b9af55d26778d976970950d7497fe5324ec28.
1. Substitute all occurrences of `\n` by `\0` in all synthesized patterns in `exclude_plus_removed`, and reverse-substitute in `Worker._path_matcher`. This way, `\n` characters in synthesized patterns survive the line split.
1. Replace `str.splitlines` by `re.split(r"\r\n|\r|\n", ...)` in `Worker.match_exclude` to split a multiline exclude string only on universal newline characters.

This is the diff of the changes in my local copy for which all tests are passing on my Linux environment:

```diff
diff --git a/copier/_main.py b/copier/_main.py
index d633529..e05f19c 100644
--- a/copier/_main.py
+++ b/copier/_main.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import os
 import platform
+import re
 import stat
 import subprocess
 import sys
@@ -479,7 +480,9 @@ class Worker:
     def _path_matcher(self, patterns: Iterable[str]) -> Callable[[Path], bool]:
         """Produce a function that matches against specified patterns."""
         # TODO Is normalization really needed?
-        normalized_patterns = (normalize("NFD", pattern) for pattern in patterns)
+        normalized_patterns = (
+            normalize("NFD", pattern.replace("\0", "\n")) for pattern in patterns
+        )
         spec = PathSpec.from_lines(_pathspec_pattern, normalized_patterns)
         return spec.match_file
 
@@ -769,9 +772,12 @@ class Worker:
         # Note: This method is a cached property, it needs to be regenerated
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
+        regex_linebreak = re.compile(r"\r\n|\r|\n")
         return self._path_matcher(
             chain.from_iterable(
-                self._render_string(exclusion, extra_context=extra_context).splitlines()
+                regex_linebreak.split(
+                    self._render_string(exclusion, extra_context=extra_context),
+                )
                 for exclusion in self.all_exclusions
             )
         )
@@ -1430,7 +1436,7 @@ class Worker:
                 ).splitlines()
                 exclude_plus_removed = list(
                     set(self.exclude).union(
-                        f"/{escape_git_path(path)}"
+                        f"/{escape_git_path(path)}".replace("\n", "\0")
                         for path in map(normalize_git_path, files_removed)
                         if not (subproject_top / path).exists()
                         and not self.match_skip(Path(path))
```

WDYT?

sjortiz:
@sisp I tried the exclude path before tbh, and based on the amount of characters to escape (according to the tests, and posix options) it looked rather cumbersome. I'd prefer keeping the two type of exclusion separated.  But your solution work as well and keeps it more consistent. 

so I'd leave that final decision to you.
@sisp

sisp:
I've refactored the update algorithm in #2815 a bit such that user-deleted files remain deleted without converting them to exclude patterns, which is causing the trouble here when the name of a deleted file contains the `\n` character. I believe that reverting to a68b9af55d26778d976970950d7497fe5324ec28 and rebasing onto `master` should let

```
tests/test_updatediff.py::test_update_deleted_path[\x07\x0c\n\t\x0bcontrol\x07\x0c\n\t\x0bcharacters\x0b\t\n\x0c\x07]
```

pass now, i.e. restoring to the state of https://github.com/copier-org/copier/pull/2804/changes/BASE..a68b9af55d26778d976970950d7497fe5324ec28. This would keep things simple and correct.

sjortiz:
let me restore and force pushe with lease

sisp:
Nice, the pipeline is passing now. :tada:

Just a final request: I missed that a68b9af55d26778d976970950d7497fe5324ec28 still uses `str.splitlines` to split the patterns of a multiline exclude string. Although the tests are passing now, I think it's better to split lines only on universal newlines using the regex from above:

```diff
         # Note: This method is a cached property, it needs to be regenerated
         # when reusing an instance in different contexts.
         extra_context = {"_copier_operation": _operation.get()}
+        regex_linebreak = re.compile(r"\r\n|\r|\n")
         return self._path_matcher(
             chain.from_iterable(
-                self._render_string(exclusion, extra_context=extra_context).splitlines()
+                regex_linebreak.split(
+                    self._render_string(exclusion, extra_context=extra_context),
+                )
                 for exclusion in self.all_exclusions
             )
         )
```

sisp (approved):
LGTM!: tada: Thanks! :bow:

PR #2824, opened 2026-08-22
fix: don't let cleanup replace the exception being handled

shawn-synvaren-AI:
Symptom: ran copier against a template that contained an error (a template where the only file contained {{ not_a_question.value }}) and got a FileNotFoundError that named the destination directory.
Cause: rmtree in the run_copy exception handler raises inside the except block and replaces the exception being handled. The original exception survives in __context__. This only happens when the render fails before the first file is written, as if any file is written then the destination exists and cleanup succeeds.

Fix: 2 branches, 1st handles the destination is missing - passes. 2nd warns on any other OSError so that it isn't lost. Original exception then raised outside of the if.
Additional warning type CleanupWarning added to errors.py, subclassing UserWarning and CopierWarning as the other warnings in that file do

Test: New test in tests/test_cleanup.py, fails against master with FileNotFoundError, passes after the fix, full suite green.

Notes: This is the only unguarded cleanup that I could find in the codebase. _vcs.py and _template.py already have guards. This is bringing _main.py into line with these.
There is a simpler 1-line alternative I considered, using ignore_errors=True, but it silences genuine failures so was discarded.
A case could be made to take the handle_remove_readonly helper approach and build/extend the helper to handle this case as well, however _main.py does not import that helper today, using a bare shutil.rmtree, however I went with the simpler fix. Let me know if you would like me to run a separate PR to fold that in (1 helper used everywhere and the helper updated to resolve this error as well)

### DIFF
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -23,6 +23,7 @@ else:
 
 
 __all__ = [
+    "CleanupWarning",
     "ConfigFileError",
     "CopierAnswersInterrupt",
     "CopierError",
--- a/copier/errors.py
+++ b/copier/errors.py
@@ -231,6 +231,10 @@ class MissingFileWarning(UserWarning, CopierWarning):
     """I still couldn't find what I'm looking for."""
 
 
+class CleanupWarning(UserWarning, CopierWarning):
+    """Failed to clean up the destination after an error."""
+
+
 class InteractiveSessionError(UserMessageError):
     """An interactive session is required to run this program."""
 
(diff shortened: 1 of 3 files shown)


copier/errors.py
```diff
@@ -231,6 +231,10 @@ class MissingFileWarning(UserWarning, CopierWarning):
     """I still couldn't find what I'm looking for."""
 
 
+class CleanupWarning(UserWarning, CopierWarning):
```
sisp, line 235: Could you add this warning to the `__all__` list at the top?

sisp (changes requested):
Thanks for fixing/improving the error handling around project cleanup, @shawn-synvaren-AI! :bow: Just a minor request from my side.

copier/errors.py
shawn-synvaren-AI: All done, anything else?

sisp (approved):
LGTM! :tada: Thanks! :bow:

